PR #293, opened 2023-09-25
[backend] Add migration to avoid deletion of injects when removing or…

### LINKED ISSUE #272: Delete cascade from user to injects should never happen
## Description

When deleting a user in the platform, all injects created by the user are deleted. We need a "set null" behavior on the creator....

guillaumejparis:
…iginal owner (#272)

### DIFF
--- a/openex-api/src/main/java/io/openex/migration/V2_60__Delete_fk_users_injects.java
+++ b/openex-api/src/main/java/io/openex/migration/V2_60__Delete_fk_users_injects.java
@@ -1,0 +1,26 @@
+package io.openex.migration;
+
+import org.flywaydb.core.api.migration.BaseJavaMigration;
+import org.flywaydb.core.api.migration.Context;
+import org.springframework.stereotype.Component;
+
+import java.sql.Connection;
+import java.sql.Statement;
+
+@Component
+public class V2_60__Delete_fk_users_injects extends BaseJavaMigration {
+
+    @Override
+    public void migrate(Context context) throws Exception {
+        Connection connection = context.getConnection();
+        Statement select = connection.createStatement();
+        select.execute("""
+              ALTER TABLE injects
+              DROP CONSTRAINT fk_a60839b2e20fc097,
+              ADD CONSTRAINT fk_injects_user_id
+                  foreign key (inject_user)
+                  references users(user_id)
+                  on delete set null;
+        """);
+    }
+}


openex-api/src/main/java/io/openex/migration/V3_3__Delete_fk_users_injects.java
```diff
@@ -0,0 +1,26 @@
...
+
+import org.flywaydb.core.api.migration.BaseJavaMigration;
+import org.flywaydb.core.api.migration.Context;
+import org.springframework.stereotype.Component;
+
+import java.sql.Connection;
+import java.sql.Statement;
+
+@Component
+public class V3_3__Delete_fk_users_injects extends BaseJavaMigration {
```
RomuDeuxfois, line 11: Why the number V3_3 ?

openex-api/src/main/java/io/openex/migration/V3_3__Delete_fk_users_injects.java
```diff
@@ -0,0 +1,26 @@
...
+
+    @Override
+    public void migrate(Context context) throws Exception {
+        Connection connection = context.getConnection();
+        Statement select = connection.createStatement();
+        select.execute("""
+              ALTER TABLE injects
+              DROP CONSTRAINT fk_a60839b2e20fc097,
+              ADD CONSTRAINT fk_user_id
```
RomuDeuxfois, line 20: Maybe better to have this name : fk_injects_user_id

**Foreign key** on **injects** table on **user_id** property
What do you think ?

openex-api/src/main/java/io/openex/migration/V3_3__Delete_fk_users_injects.java
guillaumejparis: At first i thought that it is linked to the version we release but seems not. I will fix it

openex-api/src/main/java/io/openex/migration/V3_3__Delete_fk_users_injects.java
guillaumejparis, line 20:
```suggestion
              ADD CONSTRAINT fk_injects_user_id
```

guillaumejparis: https://github.com/OpenEx-Platform/openex/pull/293/files#r1335785119

openex-api/src/main/java/io/openex/migration/V3_3__Delete_fk_users_injects.java
RomuDeuxfois: Could be a good nomenclature !

guillaumejparis: Yes but if one time we cannot release a migration on the version we thought it could lead to other problems

RomuDeuxfois:
Tests :
- existing base
- new base

openex-api/src/main/java/io/openex/migration/V3_3__Delete_fk_users_injects.java
RomuDeuxfois: https://www.red-gate.com/blog/database-devops/flyway-naming-patterns-matter

I not sure to understand which can of problem do you have in mind

RomuDeuxfois:
Test 👌

RomuDeuxfois (approved):

helene-nguyen (approved):

helene-nguyen:
Tested locally, it works fine :)

PR #356, opened 2023-10-19
[backend] Add a ProtectUser in place of a User in the execution user …

guillaumejparis:
…context and add pgpkey in the user execution context (#320)

### DIFF
--- a/openex-api/src/main/java/io/openex/helper/InjectHelper.java
+++ b/openex-api/src/main/java/io/openex/helper/InjectHelper.java
@@ -13,6 +13,8 @@ import io.openex.execution.ExecutionContext;
 import io.openex.service.ContractService;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
+import reactor.util.function.Tuple2;
+import reactor.util.function.Tuples;
 
 import javax.annotation.Resource;
 import javax.transaction.Transactional;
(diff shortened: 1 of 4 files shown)


guillaumejparis:
https://github.com/OpenEx-Platform/injectors/pull/28

openex-api/src/main/java/io/openex/helper/InjectHelper.java
```diff
@@ -30,92 +32,93 @@
...
+              .map(user -> Tuples.of(user, audience.getName())));
     }
+    throw new UnsupportedOperationException("Unsupported type of Injection");
+  }
 
-    private List<ExecutionContext> usersFromInjection(Injection injection) {
-        return getUsersFromInjection(injection)
-                .collect(groupingBy(ExecutionContext::getUser)).entrySet().stream()
-                .map(entry -> new ExecutionContext(openExConfig, entry.getKey(), injection,
-                        entry.getValue().stream().flatMap(ua -> ua.getAudiences().stream()).toList()))
```
guillaumejparis, line 83: We had a problem here with the ExecutionContext::getUser that was a ProtectUser and when we gave it to the ExecutionContext we had a type problem as ExecutionContext  want a User
So we remove the ExectutionContext creation in getUsersFromInjection and just pass a tuple of user & audience

openex-api/src/main/java/io/openex/injects/email/service/EmailPgp.java
```diff
@@ -23,57 +23,57 @@
 @Component
 public class EmailPgp {
 
-    public PGPPublicKey getUserPgpKey(User user) throws IOException {
```
guillaumejparis, line 26: Bingo ! It seems that we had a problem on the sending an email with a pgp key too

openex-injectors
guillaumejparis, line 1: https://github.com/OpenEx-Platform/injectors/pull/28

RomuDeuxfois (approved):

PR #386, opened 2023-11-03
[devex] Redirect manually to / after logout when in dev env  (#385)

guillaumejparis:


openex-front/src/admin/components/nav/TopBar.js
```diff
@@ -89,6 +89,10 @@ class TopBar extends Component {
   handleLogout() {
     this.handleClose();
     this.props.logout();
+    if (process.env.NODE_ENV === 'development') {
+      // api logout success handler redirect to / or in dev it is localhost:8080 and not localhost:3000
```
guillaumejparis, line 93:
```suggestion
      // api logout success handler redirect to / but in dev it is localhost:8080 and not localhost:3000
```

RomuDeuxfois:
Works fine, but I don't find any mention of NODE_ENV=production for Docker image..
It's maybe missing ?

We need to be sure about this point before merging I think

guillaumejparis:
> Works fine, but I don't find any mention of NODE_ENV=production for Docker image.. It's maybe missing ?
> 
> We need to be sure about this point before merging I think

We use it in the store.js file so I suppose it works but we can double check just to be sure yea !

RomuDeuxfois:
Put the logic in backend

guillaumejparis:
to do in backend

PR #390, opened 2023-11-07
[devex] add generate ts types from api

guillaumejparis:


openex-front/package.json
```diff
@@ -89,6 +89,7 @@
     "jest": "27.5.1",
     "prettier": "2.7.1",
     "redux-logger": "3.0.6",
+    "swagger-typescript-api": "^13.0.3",
```
Kedae, line 92: We use exact version

openex-front/package.json
```diff
@@ -99,7 +100,8 @@
     "build": "node builder/prod/prod.js",
     "lint": "cross-env DEBUG=eslint:cli-engine TIMING=1 eslint --max-warnings 0 --cache -c .eslintrc.json src",
     "control": "yarn audit --groups dependencies --summary",
-    "test": "jest --coverage"
+    "test": "jest --coverage",
+    "generate-types-from-api": "swagger-typescript-api -p http://localhost:8080/api-docs -o src/utils -n index.d.ts"
```
Kedae, line 104: Weird to enforce a localhost, wouldn't it be better to have it relative ?

guillaumejparis: Do you think we will generate the types on another instance than local ?

Kedae: May be I'm missing something, but for OpenCTI we generate types with drone when building. Won't you be missing something if we stick with localhost ? 
Especially when the package support relative path :
` npx swagger-typescript-api -p ./swagger.json -o ./src -n myApi.ts`

guillaumejparis: Ah ok I see what you mean.
I am not sure if we want to automatically generate types at first however we can have it relative to be future proof

guillaumejparis: _⛔   error while fetching data from URL "./swagger.json" TypeError: Only absolute URLs are supported_

:'(

openex-front/package.json
```diff
@@ -99,7 +100,8 @@
     "build": "node builder/prod/prod.js",
     "lint": "cross-env DEBUG=eslint:cli-engine TIMING=1 eslint --max-warnings 0 --cache -c .eslintrc.json src",
     "control": "yarn audit --groups dependencies --summary",
-    "test": "jest --coverage"
+    "test": "jest --coverage",
+    "generate-types-from-api": "swagger-typescript-api -p /swagger.json -o src/utils -n index.d.ts"
```
RomuDeuxfois, line 104:
```suggestion
    "generate-types-from-api": "swagger-typescript-api --no-client -p /swagger.json -o src/utils -n index.d.ts"
```

RomuDeuxfois (approved):

Issue #195, opened 2022-10-02
Exercise templating with global variables and variables by organization


SamuelHassine:
## Problem to Solve
Currently, a set of built-in variables are defined and can be used to customize the content of stimuli. But these built-in variables cannot cover all the specific needs of our users. And these variables cannot be used to configure stimuli

## Proposed Solution
A user (with enough rights) should be able to define new variables in the context of an exercice and be able to easily use it on any input field when creating stimuli or configuring exercice's content.
Having the same feature at Organizations level will allow to custom stimuli content by Organization's context.

PR #397, opened 2023-11-17
Exercise templating with dynamic variables (#195)

guillaumejparis:


RomuDeuxfois (approved):

openex-api/src/test/java/io/openex/service/VariableServiceTest.java
```diff
@@ -0,0 +1,93 @@
...
+    variable.setValue(value);
+
+    // -- EXECUTE --
+    Variable variableUpdated = this.variableService.updateVariable(variable);
+    assertNotNull(variable);
+    assertEquals(value, variableUpdated.getValue());
+  }
+
+  @DisplayName("Delete variable")
+  @Test
```
aHenryJard, line 82: If you don't want to verify the exception message content, you could also use something like:

```suggestion
  @Test(expected = NoSuchElementException.class)
```

aHenryJard: ah yes sorry, and assertThrow it's even better.

PR #611, opened 2024-02-21
Ensure uniqueness of email field when creating a user account

RomuDeuxfois:
### Related issues

* https://github.com/OpenEx-Platform/openex/issues/576

### DIFF
--- a/openex-api/src/test/java/io/openex/rest/UserApiTest.java
+++ b/openex-api/src/test/java/io/openex/rest/UserApiTest.java
@@ -4,6 +4,7 @@ import io.openex.IntegrationTest;
 import io.openex.database.model.User;
 import io.openex.database.repository.UserRepository;
 import io.openex.rest.user.form.login.LoginUserInput;
+import io.openex.rest.user.form.user.CreateUserInput;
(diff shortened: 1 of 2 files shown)


RomuDeuxfois:
Add tests to validate this behavior

RomuDeuxfois:
@savacano28  Ready for a second review

openex-api/src/main/java/io/openex/service/ChallengeService.java
```diff
@@ -23,51 +25,65 @@
...
-                        return Stream.empty();
-                    }
-                })
-                .distinct().toList();
-        return fromIterable(challengeRepository.findAllById(challenges)).stream()
-                .map(this::enrichChallengeWithExercises).toList();
-    }
+  private Stream<Challenge> resolveChallenges(@NotNull final List<Inject> injects) {
+    List<String> challenges = injects.stream()
+        .filter(inject -> inject.getContract().equals(CHALLENGE_PUBLISH))
```
savacano28, line 76: Is better "Challenege" on the left side of equals because inject.getContract can be null? is it is null, the program should be running?

guillaumejparis (approved):

PR #632, opened 2024-03-01
[backend/frontend] Issue/566 pagination : Retrieving contracts using pagination, filtering, and sorting parameters #607

savacano28:
### Proposed changes

* Endpoint for retrieving a paginated list of contracts using parameters such as: page, sizePage, and list attributes like sort condition, filtering on type, label, and text search.

### Related issues

* 607

### Checklist

### Further comments


openbas-api/src/main/java/io/openbas/contract/ContractApi.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.Base64;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static io.openbas.contract.ContractService.TYPE;
+
+@RequiredArgsConstructor
+@RestController
+@Slf4j
```
RomuDeuxfois, line 31: We use the @Log annotation for the logging system in order to use Java's native logging library. It is sufficient for our use at the moment.

openbas-api/src/main/java/io/openbas/contract/ContractApi.java
```diff
@@ -0,0 +1,106 @@
...
+@PreAuthorize("hasRole('ROLE_ADMIN')")
```
RomuDeuxfois, line 32: Could be great to have a new SecurityExpression for this, what do you think ?

openbas-api/src/main/java/io/openbas/contract/ContractApi.java
```diff
@@ -0,0 +1,106 @@
...
+                                    @ExtensionProperty(name = "httpMethod", value = "POST")
+                            }
+                    )
+            }
+    )
+    @ApiResponses(value = {
+            @ApiResponse(responseCode = "200", description = "Page of contracts"),
+            @ApiResponse(responseCode = "400", description = "Bad parameters")
+    })
+    //TODO ContractDTO
```
RomuDeuxfois, line 74: TODO ?

openbas-api/src/main/java/io/openbas/contract/ContractApi.java
```diff
@@ -0,0 +1,106 @@
...
+                                                 @RequestParam(defaultValue = "0") @Min(0) int page,
+                                                 @RequestParam(defaultValue = "10") @Max(20) int size,
+                                                 @RequestParam(required = false) List<String> sort) {
+
+        Sort sortFromQuery = convertToSort(sort);
+        Pageable pageable = PageRequest.of(page, size, sortFromQuery);
+        return contractService.searchContracts(contractSearchInput, pageable);
+    }
+
+    private Sort convertToSort(List<String> sortFields) {
```
RomuDeuxfois, line 85: sortFields is only two fileds ? First one is property and second one is direction ?
I think it's better if we have a sort object in ContractSearchInput with a defined java class to limit error
sort {
String property
String direction
}

openbas-api/src/test/java/io/openbas/contract/ContratApiTest.java
```diff
@@ -0,0 +1,237 @@
...
+
+    @BeforeEach
+    public void before() {
+        Mockito.when(contractService.getContracts()).thenReturn(ContractFixture.getContracts());
+        Mockito.when(contractService.searchContracts(any(), any())).thenCallRealMethod();
+    }
+
+    @Nested
+    @WithMockAdminUser
+    @DisplayName("Fecthing contracts")
```
RomuDeuxfois, line 39: Little typo ;)

openbas-api/src/test/java/io/openbas/utils/mockUser/WithMockAdminUser.java
```diff
@@ -0,0 +1,14 @@
...
+import org.springframework.security.test.context.support.WithSecurityContext;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+import static io.openbas.utils.mockUser.WithMockAdminUserSecurityContextFactory.MOCK_USER_ADMIN_EMAIL;
+
+@Retention(RetentionPolicy.RUNTIME)
+@WithSecurityContext(factory = WithMockAdminUserSecurityContextFactory.class)
+public @interface WithMockAdminUser {
```
RomuDeuxfois, line 12: @WithMockUser(roles = {"ADMIN"}) is not enough ?

openbas-front/tests_e2e/model/contracts/contract-api.ts
```diff
@@ -0,0 +1,360 @@
+import { Page } from '@playwright/test';
+
+class ContractApiMock {
+  constructor(private page: Page) {
+  }
+
+  async mockContracts() {
```
RomuDeuxfois, line 7: I don't undestand why we need to mock the back-end in e2e tests

openbas-api/src/main/java/io/openbas/contract/ContractApi.java
savacano28: sortFields is an array of string as: ['type:ASC', 'label:DESC', 'property:dir'].

openbas-api/src/test/java/io/openbas/utils/mockUser/WithMockAdminUser.java
savacano28: I tried but I had this error: java.lang.ClassCastException: class org.springframework.security.core.userdetails.User cannot be cast to ( are in unnamed module of loader 'app')

openbas-front/tests_e2e/model/contracts/contract-api.ts
savacano28: Because the response from back can change so I prefer test with stable data :)

RomuDeuxfois: After reflection, I think we need to dig into why the data changes rather than mocking the back-end.
Otherwise, I have the feeling that it becomes a front-end test and not e2e

RomuDeuxfois (approved):

Issue #569, opened 2024-02-04
Handle 404 properly within admin


SamuelHassine:
## Use case

OpenBAS:

![image](https://github.com/OpenEx-Platform/openex/assets/1334279/0d5d2d6b-19b2-4c77-bbd6-d7245fb28690)

OpenCTI:

![image](https://github.com/OpenEx-Platform/openex/assets/1334279/6b628f47-559e-412d-8a7a-ed7c74375d46)

PR #635, opened 2024-03-05
[frontend] Handle 404 properly within admin (#569)

RomuDeuxfois:
### Related issues

* https://github.com/OpenBAS-Platform/openbas/issues/569


openbas-front/src/admin/components/assets/Index.tsx
```diff
@@ -22,6 +23,8 @@ const Index = () => {
           <Route path="" element={<Navigate to="endpoints" replace={true} />} />
           <Route path="endpoints" element={errorWrapper(Endpoints)()} />
           <Route path="asset_groups" element={errorWrapper(AssetGroups)()} />
+          {/* Not found */}
+          <Route path="*" element={<NotFound/>}/>
```
guillaumejparis, line 27: Pretty sure you do not need these because the one you put in admin/Index.tsx will trigger if no route match in this file

RomuDeuxfois: Unfortunately, this redirect system is necessary everywhere.
In the case of assets, the admin/index redirects all URLs that start with assets to the assets/index file.
So, all /assets/ URLs are managed by this file, like the URL assets/toto.
So, management at this level is required.

guillaumejparis (approved):

PR #639, opened 2024-03-11
[backend] Fix migration issue for inject expectation group

RomuDeuxfois:
### Related issues

* https://github.com/OpenBAS-Platform/openbas/issues/637

### DIFF
--- a/openbas-api/src/main/java/io/openbas/migration/V2_71__Modify_inject.java
+++ b/openbas-api/src/main/java/io/openbas/migration/V2_71__Modify_inject.java
@@ -41,7 +41,7 @@ public class V2_71__Modify_inject extends BaseJavaMigration {
         """);
     // Add asset to inject expectation
     select.execute("""
-        ALTER TABLE injects_expectations ADD COLUMN inject_expectation_group bool;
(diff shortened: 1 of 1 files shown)


guillaumejparis (approved):

openbas-api/src/main/java/io/openbas/migration/V2_71__Modify_inject.java
```diff
@@ -41,7 +41,7 @@ constraint injects_asset_groups_pkey primary key (inject_id, asset_group_id)
         """);
     // Add asset to inject expectation
     select.execute("""
-        ALTER TABLE injects_expectations ADD COLUMN inject_expectation_group bool;
+        ALTER TABLE injects_expectations ADD COLUMN inject_expectation_group bool default false;
```
savacano28, line 44: Why not add another migration script with this update? That should generate a checksum error, right?

RomuDeuxfois: Because we don't have a release with this migration in it.
So we can do the changes direclty here.

savacano28: thanks!

PR #733, opened 2024-04-05
[backend/frontend] Global search

RomuDeuxfois:
### Related issues

* https://github.com/OpenBAS-Platform/openbas/issues/505

### Checklist

### Further comments

⚠️ Dev environment should be in 21 java version


openbas-api/src/main/java/io/openbas/fulltextsearch/FullTextSearchService.java
```diff
@@ -0,0 +1,216 @@
...
+        .orElseThrow(() -> new IllegalArgumentException(clazz + " is not handle by full text search"));
+
+    JpaSpecificationExecutor<T> repository = repositoryMap.get(clazzT);
+
+    return buildPaginationJPA(
+        repository::findAll,
+        searchPaginationInput,
+        clazzT
+    ).map(this::transform);
+  }
```
savacano28, line 80: If all the results of the transform() are null, we can handle it by returning something like: new PageImpl<>(Collections.emptyList(), pageable, 0) (line 64), in order to send a same type of response ?

openbas-api/src/main/java/io/openbas/fulltextsearch/FullTextSearchService.java
```diff
@@ -0,0 +1,216 @@
...
+  @Data
+  public static class FullTextSearchResult {
+
+    @NotBlank
+    private String id;
+    @NotBlank
+    private String name;
+    @NotBlank
+    private String clazz;
+
```
savacano28, line 213: Maybe add some additional information for differentiation? tags, short description?

openbas-api/src/main/java/io/openbas/fulltextsearch/FullTextSearchService.java
RomuDeuxfois: If the result is null, it will be handle in the buildPaginationJPA method

openbas-api/src/main/java/io/openbas/fulltextsearch/FullTextSearchService.java
RomuDeuxfois: Good point, I will add some infos

guillaumejparis (approved):

PR #788, opened 2024-05-01
[backend/frontend] Add import scenario

RomuDeuxfois:


RomuDeuxfois:
Missing assets & asset groups. How to handle this case ?

guillaumejparis (approved):

PR #873, opened 2024-05-15
[frontend]Fix users list header

johanah29:
https://github.com/OpenBAS-Platform/openbas/issues/892

### DIFF
--- a/openbas-front/src/admin/components/settings/users/Users.js
+++ b/openbas-front/src/admin/components/settings/users/Users.js
@@ -170,7 +170,7 @@ const Users = () => {
                   <div className={classes.bodyItem} style={inlineStyles.user_organization}>
                     {organizationsMap[user.user_organization]?.organization_name}
                   </div>
-                  <div className={classes.itemAdmin} style={inlineStyles.user_admin}>
+                  <div className={classes.bodyItem} style={inlineStyles.user_admin}>
                     {user.user_admin ? (<CheckCircleOutlined fontSize="small" />) : ('-')}
                   </div>
                   <div className={classes.bodyItem} style={inlineStyles.user_tags}>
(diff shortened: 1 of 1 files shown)


RomuDeuxfois:
what is the purpose of this fix?
If the goal is to center the admin icon, we have to think about what else you want to center: the headers, ect.
My concern is about adding a padding instead of have a generic flex class

openbas-front/src/admin/components/settings/users/Users.js
```diff
@@ -43,6 +43,13 @@ const useStyles = makeStyles(() => ({
     textOverflow: 'ellipsis',
     paddingRight: 10,
   },
+  itemAdmin: {
+    fontSize: 13,
+    whiteSpace: 'nowrap',
+    overflow: 'hidden',
+    textOverflow: 'ellipsis',
+    paddingLeft: 40,
+  },
```
RomuDeuxfois, line 52:
```suggestion
```

openbas-front/src/admin/components/settings/users/Users.js
```diff
@@ -169,7 +177,7 @@ const Users = () => {
                   <div className={classes.bodyItem} style={inlineStyles.user_organization}>
                     {organizationsMap[user.user_organization]?.organization_name}
                   </div>
-                  <div className={classes.bodyItem} style={inlineStyles.user_admin}>
+                  <div className={classes.itemAdmin} style={inlineStyles.user_admin}>
```
RomuDeuxfois, line 180:
```suggestion
                  <div className={classes.bodyItem} style={inlineStyles.user_admin}>
```

RomuDeuxfois (changes requested):

RomuDeuxfois (approved):

Issue #841, opened 2024-05-13
RabbitMQ version is not correct.


SamuelHassine:
## Description

![image](https://github.com/OpenBAS-Platform/openbas/assets/1334279/a81c5c61-5b93-46c6-8ae9-95cd29c43027)

Dimfacion:
Needs to be checked in an testing environment before merging.

SamuelHassine:
??? @Dimfacion @RomuDeuxfois?

SamuelHassine:
Should have been tested locally.

Dimfacion:
@SamuelHassine That's the issue, it works in local, not in a prod-like environment.

PR #880, opened 2024-05-15
[frontend/backend] Adding the correct version of RabbitMQ

Dimfacion:
### Proposed changes

* Adding a call to the management plugin of rabbitMQ to get the right version
* Adding a property to convey the right version to the frontend

### Related issues

* issue #841 
*

### Checklist


johanah29 (approved):

openbas-api/src/main/java/io/openbas/rest/settings/PlatformSettingsApi.java
```diff
@@ -157,6 +159,39 @@ private Setting resolveFromMap(Map<String, Setting> dbSettings, String themeKey,
...
+
+            // Init the headers
+            HttpHeaders headers = new HttpHeaders();
+            headers.setAccept(Collections.singletonList(MediaType.APPLICATION_JSON));
+            headers.setBasicAuth(this.openBASConfig.getRabbitmqUser(), this.openBASConfig.getRabbitmqPass());
+            HttpEntity<String> entity = new HttpEntity<>("parameters", headers);
+
+            // Make the call
+            ResponseEntity<?> result =
+                    restTemplate.exchange(uri, HttpMethod.GET, entity, String.class);
```
RomuDeuxfois, line 185: Can we move this logic to a dedicated service or client class for Rabbit ? 
What do you think ?

RomuDeuxfois:
Tested 🆗

openbas-api/src/main/java/io/openbas/rest/settings/PlatformSettingsApi.java
Dimfacion: I was trying to reduce as much as possible the footprint of this code as this is not very elegant but that would make sense to have a dedicated class (helper ?) for this

RomuDeuxfois: I vote for it, the code can be more readable/separated.

SamuelHassine (approved):

Issue #867, opened 2024-05-15
Avoid raising stack trace when entity is not found, just returning 404


SamuelHassine:
## Description

Avoid raising stack trace when entity is not found, just return 404.

Present everywhere on simple find like "scenarios/ID", "exercises/ID" in the API, etc.

PR #885, opened 2024-05-15
[backend] Avoid raising stack trace instead of returning a 404

Dimfacion:
### Proposed changes

* Instead of raising IllegalStateException or the generic NoSuchElementException, we're raising a custom defined Exception to keep the hand on where it is raised
* Adding a global exception handler to take care of the new exception and return a proper 404. The handler could be extended to add more intel when an exception is raised.

### Related issues

* issue #867 
*

### Checklist


RomuDeuxfois:
We need to rethink the management of 404 on the front side.
image

The user has an infinite loader when the entity does not exist right now, we need to display a 404 error message.
image

RomuDeuxfois (approved):

PR #915, opened 2024-05-21
[frontend] Fix warnings and fix linter

guillaumejparis:


RomuDeuxfois:
image
UI problem

openbas-front/src/admin/components/atomic_testings/InjectList.tsx
```diff
@@ -104,7 +104,7 @@ const InjectList: FunctionComponent<Props> = ({
       value: (injectDto: InjectResultDTO) => injectDto.inject_title,
     },
     {
-      field: 'inject_status',
+      field: 'inject_status.tracking_sent_date',
```
RomuDeuxfois, line 107: UI problem
image

openbas-front/src/admin/components/atomic_testings/InjectList.tsx
```diff
@@ -104,7 +107,7 @@ const InjectList: FunctionComponent<Props> = ({
...
```
RomuDeuxfois, line 110: This filter is not working
We need to modify the back-end for that

openbas-front/src/admin/components/atomic_testings/InjectList.tsx
```diff
@@ -118,7 +121,7 @@ const InjectList: FunctionComponent<Props> = ({
       },
     },
     {
-      field: 'inject_status',
+      field: 'inject_status.status_name',
```
RomuDeuxfois, line 124: This filter is not working
We need to modify the back-end for that

RomuDeuxfois (changes requested):

RomuDeuxfois (approved):

PR #967, opened 2024-05-24
[backend/frontend] Improv performance on pagination

RomuDeuxfois:
- /api/scenarios/search
- /api/teams/search
- /api/players/search
- /api/asset_groups/search
- /api/documents/search


RomuDeuxfois:
Use Set instead of List

PR #1011, opened 2024-05-29
[backend] Fix exercises list

RomuDeuxfois:
### Related issues

* https://github.com/OpenBAS-Platform/openbas/issues/1008

### DIFF
--- a/openbas-model/src/main/java/io/openbas/database/model/Inject.java
+++ b/openbas-model/src/main/java/io/openbas/database/model/Inject.java
@@ -421,45 +421,64 @@ public class Inject implements Base, Injection {
 
     // Set a list of expectations
     inject.setExpectations(new ArrayList<>());
-    for(String expectationId: rawInject.getInject_expectations()) {
-      // Create a new expectation
-      InjectExpectation expectation = new InjectExpectation();
-      expectation.setId(rawInjectExpectationMap.get(expectationId).getInject_expectation_id());
-      expectation.setType(InjectExpectation.EXPECTATION_TYPE.valueOf(rawInjectExpectationMap.get(expectationId).getInject_expectation_type()));
-      expectation.setScore(rawInjectExpectationMap.get(expectationId).getInject_expectation_score());
+    for (String expectationId : rawInject.getInject_expectations()) {
+      RawInjectExpectation rawInjectExpectation = rawInjectExpectationMap.get(expectationId);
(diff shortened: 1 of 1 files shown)


openbas-model/src/main/java/io/openbas/database/model/Inject.java
```diff
@@ -421,45 +421,80 @@ public static Inject fromRawInject(RawInject rawInject,
...
-      expectation.setId(rawInjectExpectationMap.get(expectationId).getInject_expectation_id());
-      expectation.setType(InjectExpectation.EXPECTATION_TYPE.valueOf(rawInjectExpectationMap.get(expectationId).getInject_expectation_type()));
-      expectation.setScore(rawInjectExpectationMap.get(expectationId).getInject_expectation_score());
-
-      // Add the team of the expectation
-      Team team = new Team();
-      team.setId(rawInjectExpectationMap.get(expectationId).getTeam_id());
-      team.setName(rawTeams.get(rawInjectExpectationMap.get(expectationId).getTeam_id()).getTeam_name());
-      expectation.setTeam(team);
+    for (String expectationId : rawInject.getInject_expectations()) {
```
Christian-DONGMO, line 424: Personnaly I think this method is too long, you should split it if it is possible

openbas-model/src/main/java/io/openbas/database/model/Inject.java
```diff
@@ -421,45 +421,80 @@ public static Inject fromRawInject(RawInject rawInject,
...
+            AssetGroup assetGroup = new AssetGroup();
+            assetGroup.setId(rawAssetGroup.getAsset_group_id());
+            assetGroup.setName(rawAssetGroup.getAsset_group_name());
+            assetGroup.setAssets(new ArrayList<>());
+
+            // We add the assets to the asset group
+            for (String assetId : rawAssetGroup.getAsset_ids()) {
+              RawAsset rawAsset = mapOfAsset.get(assetId);
+              if (rawAsset != null) {
+                if(rawAsset.getAsset_type().equals("Endpoint")) {
```
Christian-DONGMO, line 458: When you use a litteral more than two times, its recommended to create a final attribute to store it and use it instead. In your case, as it's a class name, you schould the class name to prevent error like that "Endpoint.class.getName()"

Christian-DONGMO (approved):
Everything sound good. I just have two comments

PR #1042, opened 2024-05-31
[backend/frontend] Improv scenario pagination

RomuDeuxfois:
### Proposed changes

* Use Criteria builder to craft pagination query
* Use Specification for Text Search and FIltering

### DIFF
--- a/openbas-api/src/main/java/io/openbas/service/ScenarioService.java
+++ b/openbas-api/src/main/java/io/openbas/service/ScenarioService.java
@@ -3,6 +3,7 @@ package io.openbas.service;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.openbas.config.OpenBASConfig;
 import io.openbas.database.model.*;
+import io.openbas.database.raw.RawPaginationScenario;
 import io.openbas.database.raw.RawScenario;
 import io.openbas.database.repository.*;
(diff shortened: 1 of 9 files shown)


openbas-api/src/main/java/io/openbas/service/ScenarioService.java
```diff
@@ -119,6 +130,78 @@ public Page<Scenario> scenarios(SearchPaginationInput searchPaginationInput) {
...
+        ))
+        .toList();
+
+    // -- Count Query --
+    CriteriaQuery<Long> countQuery = cb.createQuery(Long.class);
+    Root<Scenario> countRoot = countQuery.from(Scenario.class);
+    countQuery.select(cb.count(countRoot));
+    Long total = entityManager.createQuery(countQuery).getSingleResult();
+
+    return new PageImpl<>(scenarios, pageable, total);
```
Dimfacion, line 202: Couldn't we skip the second query and just do a dumb .size() on the scenarios ?

RomuDeuxfois: scenarios contains only the X first element depending of the pagination

RomuDeuxfois:
@guillaumejparis ready for a second test

guillaumejparis (approved):

PR #1048, opened 2024-06-05
[backend] Change external reference resolution for OpenCTI integration

RomuDeuxfois:
### Proposed changes

OpenCTI should get the latest ended simulation based on scenarios instead of the latest simulation based on the first scenario.

### Related issues

* https://github.com/OpenBAS-Platform/openbas/issues/1047
*

### Testing

- Generate a scenario from OpenCTI to OpenBAS
- Launch a simulation with some expectations
- Wait for the end of this simulation
- Go to opencti and checked that the security posture is the same as that on OpenBAS
- Re launch the simulation, change the expectation result and verify the new posture in both platform

OpenCTI PR -> https://github.com/OpenCTI-Platform/opencti/pull/7233

### DIFF
--- a/openbas-api/src/main/java/io/openbas/rest/scenario/ScenarioApi.java
+++ b/openbas-api/src/main/java/io/openbas/rest/scenario/ScenarioApi.java
@@ -1,9 +1,6 @@
 package io.openbas.rest.scenario;
 
-import io.openbas.database.model.Scenario;
-import io.openbas.database.model.Team;
-import io.openbas.database.model.TeamSimple;
-import io.openbas.database.model.User;
+import io.openbas.database.model.*;
 import io.openbas.database.raw.RawPaginationScenario;
 import io.openbas.database.repository.*;
 import io.openbas.rest.exception.ElementNotFoundException;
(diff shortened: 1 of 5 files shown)


openbas-api/src/main/java/io/openbas/rest/scenario/ScenarioApi.java
```diff
@@ -176,15 +168,18 @@ public void importScenario(@RequestPart("file") @NotNull MultipartFile file) thr
...
-  // region scenarios
   @GetMapping(SCENARIO_URI + "/{scenarioId}/exercises")
   @PreAuthorize("isScenarioObserver(#scenarioId)")
   public Iterable<ExerciseSimple> scenarioExercises(@PathVariable @NotBlank final String scenarioId) {
     Scenario scenario = this.scenarioService.scenario(scenarioId);
     return scenario.getExercises().stream().map(ExerciseSimple::fromExercise).toList();
   }
 
-  // endregion
+  @GetMapping(SCENARIO_URI + "/external_reference/{externalReferenceId}")
```
RomuDeuxfois, line 178: Prefix by /opencti and add versioning (v1)
Keep the old URL with deprecated method
Add swagger doc to explain which endpoint is supported

SouadHadjiat (approved):
tested locally

PR #1054, opened 2024-06-06
[backend/frontend] Fix reply to function in animation>mail section

RomuDeuxfois:
### Related issues

* https://github.com/OpenBAS-Platform/openbas/issues/979

### DIFF
--- a/openbas-api/src/main/java/io/openbas/config/GlobalExceptionHandler.java
+++ b/openbas-api/src/main/java/io/openbas/config/GlobalExceptionHandler.java
@@ -19,18 +19,4 @@ public class GlobalExceptionHandler {
         return new ResponseEntity<>(message, HttpStatus.NOT_FOUND);
     }
 
-    @ExceptionHandler(RuntimeException.class)
-    public ResponseEntity<ErrorMessage> handleRuntimeException(RuntimeException ex) {
-        log.severe("RuntimeException: " + ex.getMessage());
-        ErrorMessage message = new ErrorMessage("Internal server error: " + ex.getMessage());
-        return new ResponseEntity<>(message, HttpStatus.INTERNAL_SERVER_ERROR);
-    }
-
-    @ExceptionHandler(Exception.class)
-    public ResponseEntity<ErrorMessage> handleException(Exception ex) {
-        log.severe("Exception: " + ex.getMessage());
-        ErrorMessage message = new ErrorMessage("An unexpected error occurred: " + ex.getMessage());
-        return new ResponseEntity<>(message, HttpStatus.INTERNAL_SERVER_ERROR);
-    }
-
 }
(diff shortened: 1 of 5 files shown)


openbas-api/src/main/java/io/openbas/config/GlobalExceptionHandler.java
```diff
@@ -13,9 +13,10 @@
 public class GlobalExceptionHandler {
 
     @ExceptionHandler(ElementNotFoundException.class)
-    public ResponseEntity<ErrorMessage> handleElementNotFoundException() {
-        ErrorMessage message = new ErrorMessage("Not found");
-        return new ResponseEntity<ErrorMessage>(message, HttpStatus.NOT_FOUND);
+    public ResponseEntity<ErrorMessage> handleElementNotFoundException(ElementNotFoundException ex) {
+        ErrorMessage message = new ErrorMessage("Element not found: " + ex.getMessage());
```
guillaumejparis, line 17: More meaningful logs <3

openbas-api/src/main/java/io/openbas/config/GlobalExceptionHandler.java
```diff
@@ -13,9 +13,10 @@
 public class GlobalExceptionHandler {
 
     @ExceptionHandler(ElementNotFoundException.class)
-    public ResponseEntity<ErrorMessage> handleElementNotFoundException() {
-        ErrorMessage message = new ErrorMessage("Not found");
-        return new ResponseEntity<ErrorMessage>(message, HttpStatus.NOT_FOUND);
+    public ResponseEntity<ErrorMessage> handleElementNotFoundException(ElementNotFoundException ex) {
```
Christian-DONGMO, line 16: What if your exception is null?

openbas-api/src/main/java/io/openbas/rest/inject/InjectApi.java
Christian-DONGMO, line 1: 1 - You should use "private final" on all the attributes ofthis class in order to take benefit of lombok annotation (@RequiredArgsConstructor). That will clean some setters
2 - The "mapper" attribute is inherited from RestBehavior, no need to set it again.

Christian-DONGMO (changes requested):

Christian-DONGMO:
For this point : "The response should be logged in OBAS."
I was not able to check this part. I will appreciate if you could show me.

guillaumejparis:
I've added a small ui change for the reply button

guillaumejparis (approved):

openbas-api/src/main/java/io/openbas/config/GlobalExceptionHandler.java
RomuDeuxfois: Is not possible, this is the purpose of this kind of handler.
https://www.baeldung.com/exception-handling-for-rest-with-spring

openbas-api/src/main/java/io/openbas/rest/inject/InjectApi.java
RomuDeuxfois: I take action for the second point, the first one wil be in an other step

Christian-DONGMO (approved):

PR #1061, opened 2024-06-07
[backend/frontend] Improv injects perf

RomuDeuxfois:
https://github.com/OpenBAS-Platform/openbas/issues/898

### DIFF
--- a/openbas-api/src/main/java/io/openbas/rest/inject/output/InjectOutput.java
+++ b/openbas-api/src/main/java/io/openbas/rest/inject/output/InjectOutput.java
@@ -1,0 +1,85 @@
+package io.openbas.rest.inject.output;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import io.openbas.database.model.InjectorContract;
+import io.openbas.helper.InjectModelHelper;
+import jakarta.validation.constraints.Min;
+import jakarta.validation.constraints.NotBlank;
+import jakarta.validation.constraints.NotNull;
+import lombok.Data;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+@Data
+public class InjectOutput {
+
+  @JsonProperty("inject_id")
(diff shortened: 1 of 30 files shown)


openbas-api/src/main/java/io/openbas/rest/inject/output/InjectOutput.java
```diff
@@ -0,0 +1,85 @@
...
+
+  @JsonProperty("inject_exercise")
+  private String exercise;
+
+  @JsonProperty("inject_scenario")
+  private String scenario;
+
+  @JsonProperty("inject_depends_duration")
+  @NotNull
+  @Min(value = 0L, message = "The value must be positive")
```
Dimfacion, line 37: Since this can only be used for outputs, are validation annotation still needed ?

RomuDeuxfois: This is usefull for api-types file
Besides, I forgot to update it, thank you!

openbas-front/src/admin/components/common/injects/InjectDefinition.js
```diff
@@ -349,12 +350,12 @@ class InjectDefinition extends Component {
     super(props);
     this.state = {
       allTeams: props.inject.inject_all_teams,
-      teamsIds: props.inject.inject_teams,
-      assetIds: props.inject.inject_assets,
-      assetGroupIds: props.inject.inject_asset_groups,
+      teamsIds: props.inject.inject_teams || [],
```
Dimfacion, line 353: Nitpick : Shouldn't we make sure that arrays sent back from the backend are not null instead of doing this ?

RomuDeuxfois: This is not a back-end problem but a store level problem.
When the component is created, we are not sure that the inject object is still fully populated and therefore some properties may be undefined.

Hence the code in the componentDidUpdate method

We can discuss it in person to brainstorm on this type of problem.

Dimfacion (approved):

Issue #860, opened 2024-05-14
Improvement of select date widget


jborozco:
The current time picker is too clunky, the set up is not intuitive and lead to frustration when doing a set up, we need to improve the overall experience.

PR #1062, opened 2024-06-07
[Frontend]Improve select date widgets

johanah29:


johanah29:
#860

openbas-front/src/utils/Localization.js
```diff
@@ -846,7 +846,7 @@ const i18n = {
       Friday: 'Vendredi',
       Saturday: 'Samedi',
       Sunday: 'Dimanche',
-      Hour: 'Heure',
+      Scheduling_time: 'Heure',
```
RomuDeuxfois, line 849: Why not put it directly as a **time** label in the code? to avoid additional translation at the trad file level

openbas-front/src/admin/components/simulations/simulation/ExerciseDateForm.tsx
```diff
@@ -0,0 +1,188 @@
...
+import { z } from 'zod';
+import type { ExerciseUpdateStartDateInput } from '../../../../utils/api-types';
+import { useFormatter } from '../../../../components/i18n';
+import { zodImplement } from '../../../../utils/Zod';
+import { minutesInFuture } from '../../../../utils/Time';
+
+interface Props {
+  onSubmit: SubmitHandler<ExerciseUpdateStartDateInput>;
+  initialValues?: ExerciseUpdateStartDateInput;
+  editing?: boolean;
```
RomuDeuxfois, line 15: Seems to not be used

openbas-front/src/admin/components/simulations/simulation/ExerciseDateForm.tsx
```diff
@@ -0,0 +1,188 @@
...
+  });
+
+  useEffect(() => {
+    if (initialValues?.exercise_start_date != null) {
+      reset({
+        date: initialValues.exercise_start_date,
+        time: initialValues.exercise_start_date,
+      });
+    }
+  });
```
RomuDeuxfois, line 104: When my simulation has not start date, the dialog has the manual launch option checked -> OK
When my simulation has a start date, the dialog has the manual launch option checked -> NOK

Should have the good property set
image

RomuDeuxfois: Bug found
When I change the Time, is not updated

RomuDeuxfois (changes requested):

openbas-front/src/utils/Localization.js
johanah29: The label 'time' is already used in the code with another french translation.

RomuDeuxfois:
One last bug found

[screen-capture (3).webm](https://github.com/OpenBAS-Platform/openbas/assets/53513584/4a5ebc33-ac25-4280-b790-0fd47356386d)

In a simulation, when I update time to the future it's ok.
When I update tume to the past, the time is not updated (good behavior) BUT I have no error message to explain that is not possible.

RomuDeuxfois (approved):

PR #1064, opened 2024-06-07
[backend] Improv performance on articles

RomuDeuxfois:
https://github.com/OpenBAS-Platform/openbas/issues/898

### DIFF
--- a/openbas-api/src/main/java/io/openbas/rest/channel/output/ArticleOutput.java
+++ b/openbas-api/src/main/java/io/openbas/rest/channel/output/ArticleOutput.java
@@ -1,0 +1,86 @@
+package io.openbas.rest.channel.output;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.openbas.database.model.Article;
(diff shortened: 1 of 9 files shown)


openbas-api/src/main/java/io/openbas/rest/channel/output/ArticleOutput.java
```diff
@@ -0,0 +1,86 @@
...
+  public Instant getVirtualPublication() {
+    return this.virtualPublication;
+  }
+
+  @JsonProperty("article_is_scheduled")
+  public boolean isScheduledPublication() {
+    return this.virtualPublication != null;
+  }
+
+  public static ArticleOutput from(@org.jetbrains.annotations.NotNull final Article article) {
```
Christian-DONGMO, line 69: I think you should use jakarta.validation.constraints.NotNull instead of org.jetbrains.annotations.NotNull

Christian-DONGMO (approved):
I just have a comment about annotation. Otherwise, everything else looks good to me

Issue #1044, opened 2024-06-02
When modifying an injector contract, the attack pattern field "+" is not in the right position


SamuelHassine:
## Description

![image](https://github.com/OpenBAS-Platform/openbas/assets/1334279/54afa1ef-b459-4956-9daf-cb6e1f7a0489)

jborozco:
Reproducible here https://testing.obas.staging.filigran.io/admin/integrations/injectors/0097265b-0515-48a5-9bff-71d0f375fcc4

PR #1073, opened 2024-06-12
[Frontend] Adjust plus sign on attack pattern field

johanah29:
#1044


openbas-front/src/components/Autocomplete.js
```diff
@@ -56,7 +73,8 @@ const renderAutocomplete = ({
         <IconButton
           onClick={() => openCreate()}
           edge="end"
-          style={{ position: 'absolute', top, right: 35 }}
+          className={injectorContractUpdate ? classes.iconInjector : classes.iconDefault}
```
RomuDeuxfois, line 76: Is not a good practice to add a specific variable like this in a generic component

RomuDeuxfois: And the add and delete icons overlap
image

RomuDeuxfois (changes requested):

RomuDeuxfois:
I think we can dig on endAdornment option

https://stackoverflow.com/questions/70475707/mui-autocomplete-endadornment-blowing-away-default-clearicon

RomuDeuxfois (approved):

Issue #1017, opened 2024-05-29
Full text search wrong indicators


SamuelHassine:
## Description

https://testing.obas.staging.filigran.io/admin/fulltextsearch?search=cccd98ea-64cb-4350-a3e6-18aee01f530b

=> Numbers are showed in the left menu, but nothing in the list

PR #1075, opened 2024-06-12
[frontend/backend] Fixing full text search wrong indicators

Dimfacion:
### Proposed changes

* Fixing the wrong indicators
* Adding a way to search for an id
* Fixing the search bar not updating when going on the search from a url 

### Related issues

* #1017 
*

### Checklist

### Further comments

### DIFF
--- a/openbas-api/src/main/java/io/openbas/utils/pagination/PaginationUtils.java
+++ b/openbas-api/src/main/java/io/openbas/utils/pagination/PaginationUtils.java
@@ -46,4 +46,26 @@ public class PaginationUtils {
     return findAll.apply(filterSpecifications.and(searchSpecifications), pageable);
   }
 
+  /**
+   * Build PaginationJPA with a specified specifications
+   * @param findAll the find all method
+   * @param input the search inputs
+   * @param specification the specified specification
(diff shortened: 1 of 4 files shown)


openbas-api/src/main/java/io/openbas/utils/pagination/PaginationUtils.java
```diff
@@ -46,4 +46,26 @@ public static <T, U> Page<U> buildPaginationCriteriaBuilder(
...
   }
 
+  /**
+   * Build PaginationJPA with a specified specifications
+   * @param findAll the find all method
+   * @param input the search inputs
+   * @param specification the specified specification
+   * @param clazz the class that we're looking for
+   * @return a Page of results
+   */
```
RomuDeuxfois, line 56: I think of other developers and I say to myself that we should be more specific on the nomenclature, the doc or the name of the variables to clearly highlight the fact that here the specification given as a parameter replaces the default search specification.

Dimfacion: I made changes to try to be more explicit

PR #1082, opened 2024-06-13
Issue/950 inject broken filter for the attck matrix

Christian-DONGMO:
### Proposed changes

*
*

### Related issues

*
*

### Checklist

### Further comments


RomuDeuxfois:
The build is broken

openbas-api/src/main/java/io/openbas/rest/injector_contract/service/InjectorContractService.java
```diff
@@ -0,0 +1,117 @@
...
+import java.time.Instant;
+import java.util.List;
+
+import static io.openbas.helper.DatabaseHelper.updateRelation;
+import static io.openbas.helper.StreamHelper.fromIterable;
+import static io.openbas.utils.pagination.PaginationUtils.buildPaginationJPA;
+
+@Service
+@RequiredArgsConstructor
+public class InjectorContractService {
```
Dimfacion, line 33: I get what you want to do here by migrating functional into a service layer but I don't think we should do such impactful changes during a bug fix.
Besides, since we have a team discussion about refactoring the structure of the project, we should wait for it before doing anything on that part.

Christian-DONGMO: I think we should start with a migration as small as that. I also think we all agree that we need to make these changes. But you're also right, it can be very impactful. I did the rollback

openbas-api/src/main/java/io/openbas/rest/injector_contract/InjectorContractApi.java
```diff
@@ -89,8 +71,7 @@ public Page<InjectorContract> injectorContracts(@RequestBody @Valid final Search
             }
         }
         return buildPaginationJPA(
-                (Specification<InjectorContract> specification, Pageable pageable) -> this.injectorContractRepository.findAll(
-                        specification, pageable),
+                this.injectorContractRepository::findAll,
```
Dimfacion, line 74: Why are we not using the Specification anymore ? Wouldn't that impact the sorting and paging ?

Christian-DONGMO: This is a method reference expression. We are still using specification and pageable

PR #1084, opened 2024-06-14
[backend] Improv performance on atomic testing pagination list

RomuDeuxfois:
### Related issues

* https://github.com/OpenBAS-Platform/openbas/issues/898


openbas-api/src/main/java/io/openbas/service/InjectService.java
```diff
@@ -71,22 +88,165 @@ public Inject updateInjectStatus(String injectId, InjectUpdateStatusInput input)
     return injectRepository.save(inject);
   }
 
+  public Page<AtomicTestingOutput> atomicTestings(Specification<Inject> specification, Pageable pageable) {
```
RomuDeuxfois, line 91: Need to be put in AtomicTestingService direclty

Dimfacion:
image
It looks like there is an issue here

RomuDeuxfois:
> image It looks like there is an issue here

Fix

Dimfacion (approved):

PR #1090, opened 2024-06-17
[backend/frontend] Add pagination in exercise list and improv performance

RomuDeuxfois:
### Proposed changes

* Add pagination in exercise list
* Improv performance on exercise list


Christian-DONGMO (approved):
Everything is good.

Issue #1100, opened 2024-06-18
Wrong alignement in add assets to inject for platform icons


SamuelHassine:
## Description

Wrong alignement in add assets to inject for platform icons

![image](https://github.com/OpenBAS-Platform/openbas/assets/1334279/abb807f8-cb51-4114-a30a-bc9dd443a4c3)

jborozco:
![340639732-abb807f8-cb51-4114-a30a-bc9dd443a4c3](https://github.com/OpenBAS-Platform/openbas/assets/35300800/4f8360a2-fd14-421d-84ee-1c5ee42e9c48)

The platform icon is not vertically centered.

PR #1117, opened 2024-06-19
[Frontend]Fix alignment of platform icons in add assets to inject

johanah29:
#1100

### DIFF
--- a/openbas-front/src/admin/components/assets/endpoints/EndpointsDialogAdding.tsx
+++ b/openbas-front/src/admin/components/assets/endpoints/EndpointsDialogAdding.tsx
@@ -180,7 +180,7 @@ const EndpointsDialogAdding: FunctionComponent<Props> = ({
                     onClick={() => addEndpoint(endpoint.asset_id)}
                   >
                     <ListItemIcon>
-                      <DevicesOtherOutlined color="primary"/>
+                      <DevicesOtherOutlined color="primary" />
                     </ListItemIcon>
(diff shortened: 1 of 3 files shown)


johanah29:
image

openbas-front/src/admin/components/assets/endpoints/EndpointsDialogAdding.tsx
```diff
@@ -189,13 +189,13 @@ const EndpointsDialogAdding: FunctionComponent<Props> = ({
                             {endpoint.asset_name}
                           </div>
                           <div className={classes.bodyItem} style={{ width: '25%' }}>
-                            <PlatformIcon platform={endpoint.endpoint_platform} width={20} marginRight={10} /> {endpoint.endpoint_platform}
+                            <PlatformIcon platform={endpoint.endpoint_platform} width={20} position={'relative'} marginRight={10} top={5} /> {endpoint.endpoint_platform}
```
RomuDeuxfois, line 192: This component should be atomic.
So, if the component itself is not align (text + icon), it should be changed directly in this component <Platform Icon />

if this is the case, then it is the parent component that must manage to align the interior of this component
Via a display: flex and align-items: center for example.

RomuDeuxfois (changes requested):

RomuDeuxfois:
No activity for more than 10 days ...

RomuDeuxfois (approved):

Issue #1128, opened 2024-06-20
[Ask AI] Greyed-out AI button not usable


Lhorus6:
## Description

When I want to use the AI button when creating a media pressure in a scenario, I can't. The button is grayed out, unclickable.

![Screenshot 2024-06-20 170508](https://github.com/OpenBAS-Platform/openbas/assets/70956706/5d3a885b-43a4-4dd7-a052-d7067808f5f5)

## Environment

OBAS 1.1.0 on testing and filigran (prod)

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Create a scenario
2. Go on the "Definition" tab
3. Click on the "+" next to "media pressure"
4. Try to use the AI
-> You can't, the AI button is grayed out (see screenshot in description)

savacano28:
There is a condition to disable the AI icon for text lengths less than 10. I noticed that on other pages, there is default text provided. Could I set a default text here as well?
![Image](https://github.com/OpenBAS-Platform/openbas/assets/4162296/222665b6-d12e-4302-956f-1f4ccf21762e)

PR #1175, opened 2024-07-01
[frontend] Enable AI button on media pressure article (#1128)

savacano28:
### Proposed changes

* Remove limit for showing ask AI button
* Add limit to generate media

### Related issues

* https://github.com/OpenBAS-Platform/openbas/issues/1128
*

### Checklist

### Further comments

### DIFF
--- a/openbas-front/src/admin/components/common/form/TextFieldAskAI.tsx
+++ b/openbas-front/src/admin/components/common/form/TextFieldAskAI.tsx
@@ -124,44 +124,36 @@ const TextFieldAskAI: FunctionComponent<TextFieldAskAiProps> = ({
     switch (action) {
       case 'spelling':
         await askFixSpelling();
-        setDisableResponse(false);
         break;
       case 'shorter':
         await askMakeShorter();
-        setDisableResponse(false);
         break;
       case 'longer':
         await askMakeLonger();
-        setDisableResponse(false);
         break;
       case 'tone':
         await askChangeTone();
-        setDisableResponse(false);
         break;
       case 'summarize':
         await askSummarize();
-        setDisableResponse(false);
         break;
       case 'explain':
         await askExplain();
-        setDisableResponse(false);
         break;
       case 'genMessage':
         await askGenMessage();
-        setDisableResponse(false);
         break;
       case 'genSubject':
         await askGenSubject();
-        setDisableResponse(false);
         break;
       case 'genMedia':
         await askGenMedia();
-        setDisableResponse(false);
         break;
       default:
         // do nothing
-        setDisableResponse(false);
+        break;
     }
+    setDisableResponse(false);
   };
   const renderButton = () => {
     return (
--- a/openbas-front/src/admin/components/common/form/TextFieldAskAI.tsx
+++ b/openbas-front/src/admin/components/common/form/TextFieldAskAI.tsx
@@ -156,8 +156,6 @@ const TextFieldAskAI: FunctionComponent<TextFieldAskAiProps> = ({
     setDisableResponse(false);
   };
   const renderButton = () => {
-    console.log('çurrent :', currentValue);
-    console.log('message : ', messageInput);
     return (
       <>
         <EETooltip forAi={true} title={t('Ask AI')}>
--- a/openbas-front/src/admin/components/common/form/TextFieldAskAI.tsx
+++ b/openbas-front/src/admin/components/common/form/TextFieldAskAI.tsx
@@ -194,22 +194,22 @@ const TextFieldAskAI: FunctionComponent<TextFieldAskAiProps> = ({
               {t('Generate an article')}
             </MenuItem>
           )}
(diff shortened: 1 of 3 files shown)


Dimfacion:
This does fix the issue but I noticed several others (not linked to your code) that might need to be fixed at the same time : 
- If I try to change tone without opening another popover before, the Generate button is always greyed out.
- The cancel button of the generate an article is not working at all.
- If I start generate something and then generate something else really quickly, the 2 calls collide and fight for the ownership of the paragraph input.
@jborozco Should we fix this before shipping the smaller fix ?

jborozco:
Hi @Dimfacion @savacano28 Yes we should fix these 2 before shipping, otherwise the inconsistency might confuse the user. 

- If I try to change tone without opening another popover before, the Generate button is always greyed out.
- The cancel button of the generate an article is not working at all.

The last one about prompt colliding is nice to have if it is easy to fix but otherwise we can skip it and open another ticket for it.

savacano28:
Hi! I made the changes.

openbas-front/src/admin/components/common/form/TextFieldAskAI.tsx
```diff
@@ -194,22 +187,22 @@ const TextFieldAskAI: FunctionComponent<TextFieldAskAiProps> = ({
               {t('Generate an article')}
             </MenuItem>
           )}
-          <MenuItem onClick={() => handleAskAi('spelling')}>
+          <MenuItem onClick={() => handleAskAi('spelling')} disabled={messageInput.length === 0}>
```
Dimfacion, line 190: I'm unsure we should make a test on messageInput. My understanding is that messageInput is the input field of the popover when generating a new article. So here, we can't have this button active until we open the generate an article popover and type something in the Input field. I think we want to be able to check the spelling of what's inside the content field so the correct test should be currentValue.length === 0, don't you think ?

savacano28: Yes! I'll change it!

savacano28: In the code: messageInput is initialize with 'current value':   
const [messageInput, setMessageInput] = useState<string>(currentValue);

openbas-front/src/admin/components/common/form/TextFieldAskAI.tsx
```diff
@@ -194,22 +187,22 @@ const TextFieldAskAI: FunctionComponent<TextFieldAskAiProps> = ({
               {t('Generate an article')}
             </MenuItem>
           )}
-          <MenuItem onClick={() => handleAskAi('spelling')}>
+          <MenuItem onClick={() => handleAskAi('spelling')} disabled={currentValue.length === 0}>
```
RomuDeuxfois, line 190: It would be nice to add a popover to explain to the user why it is grayed out. What do you think ?

openbas-front/src/admin/components/common/form/TextFieldAskAI.tsx
```diff
@@ -418,7 +412,7 @@ const TextFieldAskAI: FunctionComponent<TextFieldAskAiProps> = ({
               {t('Cancel')}
             </Button>
             <Button
-              disabled={messageInput.length < 5}
+              disabled={currentValue.length === 0}
```
RomuDeuxfois, line 415: Why here it's currentValue.lenght et not messageInput.length ?

RomuDeuxfois:
For me, apart from the two three comments, the bug is fixed.

openbas-front/src/admin/components/common/form/TextFieldAskAI.tsx
savacano28: Because this tone generator corresponds to the main menu, the action should be applied to the currentMessage and not to the MessageInput (which is the text field used when you want to generate text).

openbas-front/src/admin/components/common/form/TextFieldAskAI.tsx
savacano28: Sure, but I have a doubt because when you try to generate some text, you get this message: '**Input (describe what you want).**' So maybe that is enough?

RomuDeuxfois: When did you get this message? I miss it I think

openbas-front/src/admin/components/common/form/TextFieldAskAI.tsx
RomuDeuxfois: Oh great !
Thanks for the explanation

openbas-front/src/admin/components/common/form/TextFieldAskAI.tsx
savacano28: Ah, sorry! You're right. I could add a tooltip to explain the status of the button :)

RomuDeuxfois (approved):

PR #1201, opened 2024-07-08
If Caldera executor is enabled and Caldera is down, the platform should fail to start

RomuDeuxfois:
### Related issues

* https://github.com/OpenBAS-Platform/openbas/issues/1034


SamuelHassine:
@RomuDeuxfois do you think this PR is a good occasion to implement also: https://github.com/OpenBAS-Platform/openbas/issues/1206?

openbas-api/src/main/java/io/openbas/executors/caldera/service/CalderaExecutorService.java
```diff
@@ -87,8 +88,10 @@ public CalderaExecutorService(
             } else {
                 executorService.remove(config.getId());
             }
+            platformSettingsService.cleanMessage();
         } catch (Exception e) {
             log.log(Level.SEVERE, "Error creating caldera executor: " + e);
+            platformSettingsService.errorMessage("Executor Caldera is not responding, your exercises may be impacted.");
```
Dimfacion, line 94: So that means that if the service is not available at start, we'll always have the message until Caldera is up and the backend has been restarted ? Do you think we could have a process trying every minute for instance until it's up to avoid having to restart the backend ?

Jipegien:
> @RomuDeuxfois I hate to be that guy but the banner is displayed next to and not above the scrollbar : image It might be worth talking with @jborozco or @Jipegien and see if that's a deal breaker or not.

something to align with what is currently implemented at OpenCTI side.

RomuDeuxfois:
> @RomuDeuxfois I hate to be that guy but the banner is displayed next to and not above the scrollbar : image It might be worth talking with @jborozco or @Jipegien and see if that's a deal breaker or not.

You can still scroll without worries
image

RomuDeuxfois:
> > @RomuDeuxfois I hate to be that guy but the banner is displayed next to and not above the scrollbar : image It might be worth talking with @jborozco or @Jipegien and see if that's a deal breaker or not.
> 
> something to align with what is currently implemented at OpenCTI side.

It's align
image

openbas-api/src/main/java/io/openbas/executors/caldera/service/CalderaExecutorService.java
RomuDeuxfois: Good point, it's changed in the last commit

RomuDeuxfois:
@Dimfacion ready for a second round

PR #1212, opened 2024-07-15
[backend] Change trace level when store message in imap

RomuDeuxfois:
The current mechanism allows us to make three attempts to store messages in the imap.

But by setting the traces to error, the status changes to partial, even if after the third attempt the storage is done.
By passing the trace into info, we make sure to keep the status close to reality.


Christian-DONGMO (approved):
I'm not able to test this case in local but I agree with the code.

PR #1216, opened 2024-07-16
[frontend] add delete option for scenarios/exercises

savacano28:
### Proposed changes

* Delete a list of injects in scenarios/exercise.
* Some fix de style

### Related issues

* https://github.com/OpenBAS-Platform/openbas/issues/1157
*

### Checklist

### Further comments

### DIFF
--- a/openbas-api/src/main/java/io/openbas/rest/exercise/ExerciseService.java
+++ b/openbas-api/src/main/java/io/openbas/rest/exercise/ExerciseService.java
@@ -6,8 +6,6 @@ import io.openbas.config.OpenBASConfig;
 import io.openbas.database.model.*;
 import io.openbas.database.repository.ArticleRepository;
 import io.openbas.database.repository.ExerciseRepository;
-import io.openbas.database.repository.TagRepository;
-import io.openbas.rest.exercise.form.ExerciseCreateInput;
 import io.openbas.rest.exercise.form.ExerciseSimple;
 import io.openbas.rest.inject.service.InjectDuplicateService;
 import io.openbas.service.GrantService;
@@ -37,7 +35,6 @@ import java.util.stream.Collectors;
 
 import static io.openbas.database.criteria.ExerciseCriteria.countQuery;
 import static io.openbas.helper.StreamHelper.fromIterable;
-import static io.openbas.helper.StreamHelper.iterableToSet;
 import static io.openbas.utils.AtomicTestingUtils.getExpectationResultByTypes;
 import static io.openbas.utils.Constants.ARTICLES;
 import static io.openbas.utils.Constants.MAX_SIZE_OF_STRING;
@@ -62,7 +59,6 @@ public class ExerciseService {
 
     private final ArticleRepository articleRepository;
     private final ExerciseRepository exerciseRepository;
-    private final TagRepository tagRepository;
 
     // region properties
     @Value("${openbas.mail.imap.enabled}")
--- a/openbas-api/src/main/java/io/openbas/rest/exercise/ExerciseService.java
+++ b/openbas-api/src/main/java/io/openbas/rest/exercise/ExerciseService.java
@@ -2,13 +2,18 @@ package io.openbas.rest.exercise;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ArrayNode;
+import io.openbas.config.OpenBASConfig;
 import io.openbas.database.model.*;
 import io.openbas.database.repository.ArticleRepository;
 import io.openbas.database.repository.ExerciseRepository;
+import io.openbas.database.repository.TagRepository;
+import io.openbas.rest.exercise.form.ExerciseCreateInput;
 import io.openbas.rest.exercise.form.ExerciseSimple;
 import io.openbas.rest.inject.service.InjectDuplicateService;
+import io.openbas.service.GrantService;
 import io.openbas.service.InjectService;
 import io.openbas.service.VariableService;
+import jakarta.annotation.Resource;
 import jakarta.persistence.EntityManager;
 import jakarta.persistence.PersistenceContext;
 import jakarta.persistence.Tuple;
@@ -16,6 +21,7 @@ import jakarta.persistence.TypedQuery;
 import jakarta.persistence.criteria.*;
 import jakarta.validation.constraints.NotBlank;
 import lombok.RequiredArgsConstructor;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.data.domain.Page;
 import org.springframework.data.domain.PageImpl;
 import org.springframework.data.domain.Pageable;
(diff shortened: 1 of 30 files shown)


openbas-api/src/main/java/io/openbas/rest/exercise/ExerciseService.java
```diff
@@ -168,6 +186,23 @@ private List<ExerciseSimple> execution(TypedQuery<Tuple> query) {
                 .toList();
     }
 
+    // -- CREATION --
+
+    @Transactional
```
Dimfacion, line 191: This is great 👍

openbas-api/src/test/java/io/openbas/rest/InjectApiTest.java
```diff
@@ -1,175 +1,360 @@
...
+    private CommunicationRepository communicationRepository;
+    @Autowired
+    private InjectExpectationRepository injectExpectationRepository;
+    @Autowired
+    private TeamRepository teamRepository;
+    @Autowired
+    private InjectorContractRepository injectorContractRepository;
+
+    @BeforeAll
+    void beforeAll() {
```
Dimfacion, line 65: I might miss something but wouldn't it be better to mock the repositories instead of creating and deleting elements in a @BeforeAll and @AfterAll annotation ?

openbas-front/src/actions/Inject.js
```diff
@@ -55,6 +55,11 @@ export const deleteInjectForExercise = (exerciseId, injectId) => (dispatch) => {
   return delReferential(uri, 'injects', injectId)(dispatch);
 };
 
+export const bulkDeleteInjectsForExercise = (exerciseId, injectIds) => (dispatch) => {
+  const uri = `/api/exercises/${exerciseId}/injects/delete`;
```
Dimfacion, line 59: I'm wondering if we shouldn't have a file with all URI ?

openbas-api/src/test/java/io/openbas/rest/InjectApiTest.java
savacano28: In this case, it is an integration test, so I usually don't use mocks (just for 3rd services)... but, it's true that I'd prefer another way to load and clean data for the tests, perhaps a sql script...

Dimfacion: Ok, I usually mock DB calls and 3rd party services in my integration tests but I understand why you'd want to keep them. Not for this PR but maybe we could think of a throwable DB used only for tests.

Dimfacion:
From reading your code, I was expecting the dialog to switch from 'these' to 'this' if only one inject is selected. Did I misunderstood it ?
image

openbas-front/src/actions/Inject.js
savacano28: open-api? /api-docs

Dimfacion: No, I meant having a file in the frontend with all the URI for API calls as constant so they are all in the same place.

savacano28:
@Dimfacion Yes, it should be: this inject, I'm trying to reproduce it. ty!

openbas-front/src/actions/Inject.js
savacano28: I don't think we have a file with all URLs, but we usually have a file for each resource.

Dimfacion (dismissed):

openbas-api/src/main/java/io/openbas/service/InjectService.java
```diff
@@ -129,8 +129,16 @@ public Inject updateInjectStatus(String injectId, InjectUpdateStatusInput input)
         return injectRepository.save(inject);
     }
 
-    public List<InjectOutput> injects(Specification<Inject> specification) {
-        CriteriaBuilder cb = this.entityManager.getCriteriaBuilder();
+
+  @Transactional(rollbackOn = Exception.class)
+  public void deleteAllByIds(List<String> injectIds) {
+    if (injectIds != null && !injectIds.isEmpty()) {
```
RomuDeuxfois, line 135: You can use !CollectionUtils.isEmpty(injectIds)

openbas-model/src/main/java/io/openbas/database/repository/InjectRepository.java
```diff
@@ -23,6 +23,11 @@ public interface InjectRepository extends CrudRepository<Inject, String>, JpaSpe
   @NotNull
   Optional<Inject> findById(@NotNull String id);
 
+  List<Inject> findByExerciseId(@NotNull String exerciseId);
+
+  @NotNull
+  List<Inject> findByScenarioId(@NotNull String scenarioId);
```
RomuDeuxfois, line 29: Oh that's great
If you have time, you can use it in InjectAPI.scenarioInjects method
It will be clearer

openbas-model/src/main/java/io/openbas/database/repository/InjectRepository.java
```diff
@@ -23,6 +23,11 @@ public interface InjectRepository extends CrudRepository<Inject, String>, JpaSpe
   @NotNull
   Optional<Inject> findById(@NotNull String id);
 
+  List<Inject> findByExerciseId(@NotNull String exerciseId);
```
RomuDeuxfois, line 26: Oh that's great
If you have time, you can use it in InjectAPI.exerciseInjects method
It will be clearer

openbas-front/src/admin/components/settings/users/Users.js
```diff
@@ -47,7 +47,7 @@ const useStyles = makeStyles(() => ({
 
 const inlineStyles = {
   user_email: {
-    width: '25%',
+    width: '20%',
```
RomuDeuxfois, line 50: Why are you changing this ?

RomuDeuxfois:
Tested 🆗

openbas-api/src/main/java/io/openbas/rest/inject/InjectApi.java
```diff
@@ -331,6 +331,12 @@ public void deleteInject(@PathVariable String exerciseId, @PathVariable String i
         injectRepository.deleteById(injectId);
     }
 
+    @PostMapping(EXERCISE_URI + "/{exerciseId}/injects/delete")
```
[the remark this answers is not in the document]
savacano28: I'd prefer to use a POST request for the request body, especially considering the limitations of requestparam. However, I can use @DeleteMapping and still include a request body... so...

openbas-front/src/admin/components/settings/users/Users.js
savacano28: I have some tags that were closely spaced in users, so I adjusted the column size for that. However, I could also display fewer tags to avoid the problem. :)

RomuDeuxfois (approved):

PR #1222, opened 2024-07-18
[frontend] 1205 - Delete an item from a list (inject, atomic testing,…

Christian-DONGMO:
… scenario, simlation)

### Proposed changes

*
*

### Related issues

*
*

### Checklist

### Further comments


openbas-front/src/components/common/pagination/PaginationComponent.tsx
```diff
@@ -41,6 +41,7 @@ const useStyles = makeStyles(() => ({
 const ROWS_PER_PAGE_OPTIONS = [20, 50, 100];
 
 interface Props<T> {
+  key?: React.Key;
```
RomuDeuxfois, line 44: This new param seems to not be used

openbas-front/src/components/common/pagination/PaginationComponent.tsx
```diff
@@ -101,7 +102,7 @@ const PaginationComponent = <T extends object>({
   // Filters
   const [openMitreFilter, setOpenMitreFilter] = React.useState(false);
 
-  useEffect(() => {
+  const fetchData = () => {
```
RomuDeuxfois, line 105: Why this change ?

If fetchData is not called by any other component.

openbas-api/src/main/java/io/openbas/utils/pagination/SearchPaginationInput.java
```diff
@@ -22,7 +22,7 @@ public class SearchPaginationInput {
   @Min(0) int page = 0;
 
   @Schema(description = "Element number by page")
-  @Max(1000) int size = 10;
+  @Max(1000) int size = 20;
```
RomuDeuxfois, line 25: Why this change ?

openbas-front/src/admin/components/Dashboard.tsx
```diff
@@ -222,6 +226,7 @@ const Dashboard = () => {
             exercises={exercises}
             searchPaginationInput={searchPaginationInput}
             setSearchPaginationInput={setSearchPaginationInput}
+            onOperationSuccess={refreshExercises}
```
RomuDeuxfois, line 229: Can you explain why we need this new mechanism ?
I don't understand

openbas-front/src/admin/components/scenarios/Scenarios.tsx
```diff
@@ -194,22 +194,14 @@ const Scenarios = () => {
...
-  const handleOpenExport = (scenarioId: string) => {
-    setOpenExportId(scenarioId);
-  };
-  const handleCloseExport = () => {
-    setOpenExportId(null);
+  const refreshScenarios = () => {
+    searchScenarios(searchPaginationInput).then((result) => {
+      const { data } = result;
+      setScenarios(data.content);
+      setModificationTrigger((prev) => !prev);
```
RomuDeuxfois, line 203: I don't understand the usage of modificationTrigger

openbas-front/src/components/common/pagination/PaginationComponent.tsx
Christian-DONGMO: This param is used in Scenarios.tsx to force the pagination to refresh in the case of a duplication or deletion of an element from the list

openbas-front/src/components/common/pagination/PaginationComponent.tsx
Christian-DONGMO: Forgetting to go back during a test

openbas-api/src/main/java/io/openbas/utils/pagination/SearchPaginationInput.java
Christian-DONGMO: I found that when I modify an element in the list, and I propagate the change to update the list, it is reloaded but uses the default value set in the backend. And there is an inconsistency between the minimum value of the back which is 10 and that of the front which is 20

openbas-front/src/admin/components/Dashboard.tsx
Christian-DONGMO: When you add or delete a simulation from the Dashboard, the list is not updated. This new mechanism allows me to propagate the change to the parent and forces it to update the list

openbas-front/src/admin/components/scenarios/Scenarios.tsx
Christian-DONGMO: modificationTrigger just allows me to signal a change of state which I transmit to the PaginationComponent component

openbas-front/src/components/common/pagination/PaginationComponent.tsx
RomuDeuxfois: Ok but the parameter is not used in the class. It looks like magic.

And by making this change, we bypass the use of the store.

What is the use case?

openbas-api/src/main/java/io/openbas/utils/pagination/SearchPaginationInput.java
RomuDeuxfois: well seen

openbas-front/src/admin/components/Dashboard.tsx
RomuDeuxfois: Can you show me where it's possible to add or delete a simulation from the Dashboard screen ?

RomuDeuxfois (changes requested):

openbas-front/src/admin/components/simulations/simulation/ExercisePopover.tsx
```diff
@@ -152,7 +160,7 @@ const ExercisePopover: FunctionComponent<ExercisePopoverProps> = ({
   // Button Popover
   const entries = [];
   if (actions.includes('Update')) entries.push({ label: 'Update', action: () => handleOpenEdit(), disabled: !permissions.canWriteBypassStatus });
-  if (actions.includes('Delete')) entries.push({ label: 'Delete', action: () => handleOpenDelete(), disabled: !permissions.canWriteBypassStatus });
```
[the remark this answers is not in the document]
RomuDeuxfois: We need to talk with the product team about this.

PR #1239, opened 2024-07-23
[frontend] Change header on inject details

RomuDeuxfois:
### Related issues

* https://github.com/OpenBAS-Platform/openbas/issues/1209

### DIFF
--- a/openbas-front/src/admin/components/injects/InjectHeader.tsx
+++ b/openbas-front/src/admin/components/injects/InjectHeader.tsx
@@ -1,0 +1,22 @@
+import { Tooltip, Typography } from '@mui/material';
+import React, { FunctionComponent } from 'react';
+import { truncate } from '../../../utils/String';
+import type { InjectResultDTO } from '../../../utils/api-types';
+
+interface Props {
+  inject: InjectResultDTO;
+}
+
+const InjectHeader: FunctionComponent<Props> = ({
+  inject,
+}) => {
+  return (
+    <Tooltip title={inject.inject_title}>
+      <Typography variant="h1" gutterBottom >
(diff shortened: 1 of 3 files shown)


savacano28:
@RomuDeuxfois For this PR, we should take into account ? 
Jp:
* Remove the simulation/scenario header when in an Inject overview and display the name of the inject **: OK**
* Ensure that the breadcrumbs is consistent: currently, the breadcrum does not contain the name of the Simulation/Scenario anymore when the user display the inject overview (it is replaced by a clickable "Animation"). It should be Simulations/[nameof Simu...]/**Animation**/etc. with the name of the simu/scenar clickable to land on its Overview: **the modif of /Animation/ is not in the pr.**
image

RomuDeuxfois:
> @RomuDeuxfois For this PR, we should take into account ? Jp:
> 
> * Remove the simulation/scenario header when in an Inject overview and display the name of the inject **: OK**
> * Ensure that the breadcrumbs is consistent: currently, the breadcrum does not contain the name of the Simulation/Scenario anymore when the user display the inject overview (it is replaced by a clickable "Animation"). It should be Simulations/[nameof Simu...]/**Animation**/etc. with the name of the simu/scenar clickable to land on its Overview: **the modif of /Animation/ is not in the pr.**
>   image

For the second point, I don't update the issue sorry
We should have two breadcrumbs:
- when we came from the inject in the overview: scenario/SCENARIO_NAME/injects/INJECT_NAME
- when we came from the inject in the animation: scenario/SCENARIO_NAME/animation/injects/INJECT_NAME

like that the user can go back with the breadcrumbs

savacano28 (approved):

PR #1243, opened 2024-07-23
[backend/frontend] Last fixes on import

Dimfacion:
### Proposed changes

* Fix on import of relative time
* Better readibility on log message/injects when testing a mapper
* Comprehensive error message in case of critical error during import

### Related issues

*
*

### Checklist

### Further comments

### DIFF
--- a/openbas-front/src/admin/components/scenarios/scenario/ScenarioContext.ts
+++ b/openbas-front/src/admin/components/scenarios/scenario/ScenarioContext.ts
@@ -1,4 +1,4 @@
-import type { Inject, InjectsImportInput } from '../../../../utils/api-types';
+import type { Inject, InjectsImportInput, ImportMessage } from '../../../../utils/api-types';
 import {
   addInjectForScenario,
   bulkDeleteInjectsForScenario,
@@ -11,7 +11,6 @@ import { useAppDispatch } from '../../../../utils/hooks';
 import type { ScenarioStore } from '../../../../actions/scenarios/Scenario';
 import type { InjectStore } from '../../../../actions/injects/Inject';
 import { fetchScenario, fetchScenarioTeams, importXls } from '../../../../actions/scenarios/scenario-actions';
-import type { ImportMessage } from '../../../../utils/api-types';
 import { MESSAGING$ } from '../../../../utils/Environment';
 import { useFormatter } from '../../../../components/i18n';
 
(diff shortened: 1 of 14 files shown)


openbas-front/src/admin/components/scenarios/scenario/ScenarioContext.ts
```diff
@@ -32,10 +35,14 @@ const injectContextForScenario = (scenario: ScenarioStore) => {
       return dispatch(deleteInjectScenario(scenario.scenario_id, injectId));
     },
     onImportInjectFromXls(importId: string, input: InjectsImportInput): Promise<void> {
-      return importXls(scenario.scenario_id, importId, input).then((_value) => new Promise((resolve, _reject) => {
+      return importXls(scenario.scenario_id, importId, input).then((response) => new Promise((resolve, _reject) => {
         dispatch(fetchScenarioInjects(scenario.scenario_id));
         dispatch(fetchScenario(scenario.scenario_id));
         dispatch(fetchScenarioTeams(scenario.scenario_id));
+        const criticalMessages = response.data.import_message.filter((value: ImportMessage) => value.message_level === 'CRITICAL');
```
RomuDeuxfois, line 42: Should be put in scenario-action to show success or error depending on result

openbas-front/src/utils/Environment.js
```diff
@@ -8,7 +8,7 @@ import { useLocation } from 'react-router-dom';
 const MESSENGER$ = new Subject().pipe(debounce(() => timer(500)));
 export const MESSAGING$ = {
   messages: MESSENGER$,
-  notifyError: (text) => MESSENGER$.next([{ type: 'error', text }]),
+  notifyError: (text, sticky) => MESSENGER$.next([{ type: 'error', text, sticky }]),
```
RomuDeuxfois, line 11: Should we have a default value for sticky ?
Or it's already handle in the class ?

Dimfacion: As of now, we're using the fact that sticky is undefined and therefore falsy. That would be better to have a default value tho, so I'm changing it.

Dimfacion: Done

openbas-front/src/admin/components/scenarios/scenario/ScenarioContext.ts
Dimfacion: Should be better now (although I'm unsure how clean the part in openbas-front/src/admin/components/common/injects/ImportUploaderInjectFromXls.tsx is)

RomuDeuxfois (approved):

openbas-front/src/admin/components/scenarios/scenario/ScenarioContext.ts
RomuDeuxfois: I don't understand what your mean

RomuDeuxfois:
I just found out with the doc
But the menu names are xls formatters, it should be xls mappers instead, right?
image

Dimfacion:
@RomuDeuxfois Good call, I made the change

Issue #1246, opened 2024-07-24
At XLS import, allow user to configure a launch time for the Scenario/Simulation


Jipegien:
## Use case

I want to import from an xls sheet a list of inject for my Scenario or my Simulation. Injects in my sheet have trigger time in absolute time (ex.: 9:30, 10h34, etc.). I have set according date pattern in my xls mapper. 
When I import the file and choose the right xls mapper, I want to be able to set a launch date for my Scenario/Simulation on-the-fly (if not already set). Otherwise, the import will fail because the platform cannot those absolute times into relative ones.

PR #1249, opened 2024-07-24
[backend/frontend] Allow user to configure a launch time during import if needed

Dimfacion:
### Proposed changes

* Allow user to configure a launch time during import if needed
*

### Related issues

* #1246 
*

### Checklist

### Further comments

### DIFF
--- a/openbas-front/src/admin/components/common/injects/ImportUploaderInjectFromXlsInjects.tsx
+++ b/openbas-front/src/admin/components/common/injects/ImportUploaderInjectFromXlsInjects.tsx
@@ -127,6 +127,7 @@ const ImportUploaderInjectFromXlsInjects: FunctionComponent<Props> = ({
   const checkNeedLaunchDate = () => {
     const formValues = getValues();
     if (formValues.importMapperId && formValues.sheetName && formValues.timezone) {
+      setNeedLaunchDate(false);
       const input: InjectsImportInput = {
         import_mapper_id: formValues.importMapperId,
         sheet_name: formValues.sheetName,
@@ -138,7 +139,6 @@ const ImportUploaderInjectFromXlsInjects: FunctionComponent<Props> = ({
           setNeedLaunchDate(true);
         }
       });
-      setNeedLaunchDate(false);
     }
   };
 
--- a/openbas-front/src/admin/components/common/injects/ImportUploaderInjectFromXlsInjects.tsx
+++ b/openbas-front/src/admin/components/common/injects/ImportUploaderInjectFromXlsInjects.tsx
@@ -83,7 +83,7 @@ const ImportUploaderInjectFromXlsInjects: FunctionComponent<Props> = ({
         importMapperId: z.string().min(1, { message: t('Should not be empty') }),
         timezone: z.string().min(1, { message: t('Should not be empty') }),
         startDate: z.string().optional(),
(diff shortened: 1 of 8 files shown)


openbas-front/src/admin/components/scenarios/scenario/ScenarioContext.ts
```diff
@@ -39,6 +39,11 @@ const injectContextForScenario = (scenario: ScenarioStore) => {
         resolve(response.data);
       }));
     },
+    onDryImportInjectFromXls(importId: string, input: InjectsImportInput): Promise<ImportTestSummary> {
+      return dryImportXls(scenario.scenario_id, importId, input).then((response) => new Promise((resolve, _reject) => {
+        resolve(response.data);
+      }));
+    },
     onBulkDeleteInjects(injectIds: string[]): void {
```
savacano28, line 47: I think you could just write: dryImportXls(scenario.scenario_id, importId, input)**.then**(response => response.data); (without wrapper into another promise)

openbas-front/src/admin/components/scenarios/scenario/ScenarioContext.ts
```diff
@@ -39,6 +39,11 @@ const injectContextForScenario = (scenario: ScenarioStore) => {
         resolve(response.data);
```
savacano28, line 39: I didn't see that before. Here, I'd prefer to use **await** rather than creating another promise within a then block (l34-l39)

openbas-front/src/admin/components/common/injects/ImportUploaderInjectFromXlsInjects.tsx
```diff
@@ -109,8 +124,26 @@ const ImportUploaderInjectFromXlsInjects: FunctionComponent<Props> = ({
...
+        timezone_offset: moment.tz(formValues.timezone).utcOffset(),
+      };
+      injectContext.onDryImportInjectFromXls?.(importId, input).then((value: ImportTestSummary) => {
+        const criticalMessages = value.import_message?.filter((importMessage: ImportMessage) => importMessage.message_level === 'CRITICAL');
+        if (criticalMessages && criticalMessages?.filter((message) => { return message.message_code === 'ABSOLUTE_TIME_WITHOUT_START_DATE'; }).length > 0) {
+          setNeedLaunchDate(true);
+        }
+      });
+      setNeedLaunchDate(false);
+    }
```
savacano28, line 142: This is okay, but I'd prefer to add a catch block or use await and include the conditionals for setNeedLaunchDate(false). You could move also setNeedLaunchDate(false) to the beginning of the function to show that it is the default value (it s just for better readability).

RomuDeuxfois:
I have a question: Should trigger time property in csv mapper be required ?

RomuDeuxfois:
I am not sure how to test it.
I have this xls
[MaSuperSimulation2.xlsx](https://github.com/user-attachments/files/16387749/MaSuperSimulation2.xlsx)
And this mapper 
image

And I don't need to set a start date when I am importing my xls in a scenario

Dimfacion:
@RomuDeuxfois I don't think the trigger should be mandatory. We should be able to create injects with as little informations as possible.

Your mapper pattern should be HH'h'mm. Since it does not match, it switch to relative and does not have the issue.

RomuDeuxfois (approved):

Issue #1143, opened 2024-06-22
Payload form attack patterns / tags fields not correctly displayed


SamuelHassine:
## Description

Correct (for instance in endpoint / create):

![image](https://github.com/OpenBAS-Platform/openbas/assets/1334279/de96f8ab-19ef-466d-b053-3fd49998e6cc)

Incorrect, in payload / create:

![image](https://github.com/OpenBAS-Platform/openbas/assets/1334279/31a27d61-9735-4d8b-9a4b-8bb5e4764596)

PR #1251, opened 2024-07-25
[Frontend|Backend]Rewriting payloadForm

johanah29:
#1143


openbas-front/src/admin/components/payloads/OldPayloadForm.js
```diff
@@ -22,7 +22,7 @@ const useStyles = makeStyles(() => ({
   },
 }));
 
-const PayloadForm = (props) => {
+const OldPayloadForm = (props) => {
```
RomuDeuxfois, line 25: This file is no longer used, it must be deleted

openbas-front/src/components/fields/SelectField.jsx
```diff
@@ -40,9 +43,9 @@ const SelectField = (props) => {
           </MUISelect>
         )}
       />
-      {helperText && (
-        <FormHelperText variant={others.variant}>
-          {helperText}
+      {!!error && (
+        <FormHelperText variant={others.variant} error={!!error}>
+          {helperText ?? error?.message}
```
RomuDeuxfois, line 48: I think helperText can be removed
Is no longer in the params of the component and it's handle by zod with the error.message

openbas-front/src/components/OldPlatformField.js
```diff
@@ -21,7 +21,7 @@ const styles = () => ({
   },
 });
 
-class PlatformField extends Component {
+class OldPlatformField extends Component {
```
RomuDeuxfois, line 24: This file is no longer used, it must be deleted

openbas-front/src/admin/components/payloads/PayloadForm.tsx
```diff
@@ -0,0 +1,513 @@
...
+            value={value}
+            error={error}
+          />
+        )}
+      />
+
+      <TextField
+        name="payload_description"
+        multiline={true}
+        fullWidth={true}
```
RomuDeuxfois, line 177: You can avoid ={true] because it's the default value

RomuDeuxfois:
Creation
- When we create a DNS resolution payload, the hostname field is required but never in red when is missing
  image

Update
- create DNS resolution and try to update id -> not possible
- update a file drop -> not possible
image
- update a payload with a tag, the tag is not in the form

openbas-front/src/components/fields/SelectField.jsx
johanah29: HelperText needs to be mentionned, else no text is displayed.

RomuDeuxfois (approved):

PR #1256, opened 2024-07-26
Add multi-language&simplified chinese translation

yolylight:
1.Add multi-language support
2.Add simplified chinese translation


RomuDeuxfois:
Thanks a lot for this PR !
The changes suit us.
In order to pass it, it would be necessary to resolve the conflicts and pass the linter on the front side: yarn lint --fix

Issue #1260, opened 2024-07-29
Ability to import / export XLS mappers


SamuelHassine:
## Use case

Ability to import / export XLS mappers

PR #1263, opened 2024-07-29
[frontend/backend]Adding the ability to import/export mappers

Dimfacion:
### Proposed changes

* Adding the ability to import/export mappers
*

### Related issues

* #1260 
*

### Checklist

### Further comments

### DIFF
--- a/openbas-api/src/main/java/io/openbas/rest/mapper/form/ImportMapperUpdateInput.java
+++ b/openbas-api/src/main/java/io/openbas/rest/mapper/form/ImportMapperUpdateInput.java
@@ -15,15 +15,15 @@ import static io.openbas.config.AppConfig.MANDATORY_MESSAGE;
 public class ImportMapperUpdateInput {
 
     @NotBlank(message = MANDATORY_MESSAGE)
-    @JsonProperty("mapper_name")
+    @JsonProperty("import_mapper_name")
     private String name;
 
     @Pattern(regexp="^[A-Z]{1,2}$")
-    @JsonProperty("mapper_inject_type_column")
+    @JsonProperty("import_mapper_inject_type_column")
     @NotBlank
     private String injectTypeColumn;
 
-    @JsonProperty("mapper_inject_importers")
+    @JsonProperty("import_mapper_inject_importers")
     @NotNull
     private List<InjectImporterUpdateInput> importers = new ArrayList<>();
 }
--- a/openbas-api/src/main/java/io/openbas/service/MapperService.java
+++ b/openbas-api/src/main/java/io/openbas/service/MapperService.java
@@ -1,14 +1,19 @@
 package io.openbas.service;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import io.openbas.database.model.ImportMapper;
 import io.openbas.database.model.InjectImporter;
 import io.openbas.database.model.InjectorContract;
 import io.openbas.database.model.RuleAttribute;
 import io.openbas.database.repository.ImportMapperRepository;
 import io.openbas.database.repository.InjectorContractRepository;
+import io.openbas.helper.ObjectMapperHelper;
 import io.openbas.rest.exception.ElementNotFoundException;
+import io.openbas.rest.mapper.export.MapperExportMixins;
 import io.openbas.rest.mapper.form.*;
 import lombok.RequiredArgsConstructor;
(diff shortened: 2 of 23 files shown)


openbas-front/src/actions/mapper/mapper-actions.ts
```diff
@@ -39,3 +47,15 @@ export const testXlsFile = (importId: string, input: InjectsImportTestInput) =>
...
 };
+
+export const exportMapper = (input: ExportMapperInput) => {
+  const uri = `${XLS_MAPPER_URI}/export`;
+  return simplePostCall(uri, input).then((response) => {
+    return { data: response.data, filename: response.headers['content-disposition'].split('filename=')[1] };
+  });
+};
+
+export const importMapper = (formData: FormData) => (dispatch: Dispatch) => {
```
RomuDeuxfois, line 58: Why are you using a dispatch here ?

openbas-front/src/admin/components/settings/data_ingestion/ImportUploaderMapper.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { useNavigate } from 'react-router-dom';
+import { useAppDispatch } from '../../../../utils/hooks';
+import ImportUploader from '../../../../components/common/ImportUploader';
+import { importMapper } from '../../../../actions/mapper/mapper-actions';
+
+const ImportUploaderMapper = () => {
+  // Standard hooks
+  const dispatch = useAppDispatch();
```
RomuDeuxfois, line 9: We don't have mapper in store right ?
So we don't need to dispatch it

openbas-api/src/main/java/io/openbas/service/MapperService.java
```diff
@@ -176,4 +177,34 @@ private void updateInjectImporter(List<InjectImporterUpdateInput> injectImporter
     });
   }
 
+  public List<ImportMapperAddInput> exportMappers(List<String> idsToExport) {
```
RomuDeuxfois, line 180: I don't understand why we need to recreate an object here

Dimfacion: It's kind of a shortcut : instead of exporting objects that match what is in the database, we export the object that would allow us to recreate the same mapper (which is why the import code is so small). This is why we're converting it into ImportMapperAddInput.
It result in a smaller file but to be honest, I'm kind of unsure if that's the best way to do it as it assumes that the export is made so that it can be imported. What do you think of it ?

openbas-front/src/admin/components/settings/data_ingestion/ImportUploaderMapper.tsx
Dimfacion: Good point, I based this on ImportUploaderExercise which does have one. I'm removing it

Dimfacion: Done :)

openbas-front/src/actions/mapper/mapper-actions.ts
Dimfacion: Same as https://github.com/OpenBAS-Platform/openbas/pull/1263/files#r1698685702

openbas-api/src/main/java/io/openbas/service/MapperService.java
RomuDeuxfois: We need to talk about it.
We have an existing mechanism to export based on Json annotation. We can challenge this too to check if there are a modern way to do it.

RomuDeuxfois:
@Dimfacion I used Mixin to homogenise the export mechanism.
I let you give your opinion about the new code.

openbas-api/src/main/java/io/openbas/rest/mapper/form/ImportMapperUpdateInput.java
```diff
@@ -15,15 +15,15 @@
 public class ImportMapperUpdateInput {
```
savacano28, line 16: ImportMapperAdd and ImportMapperUpdate are the same?

Dimfacion: No, the importers property is not the same between the two. One use InjectImporterAddInput and the other InjectImporterUpdateInput. Those two objects are different

savacano28:
Code: ok, Tested: ok

savacano28 (approved):

PR #1372, opened 2024-08-30
[github] Adding a way to have the labels automatically set

Dimfacion:
### Proposed changes

* Adding a way to have the labels automatically set so I can stop forgetting them
*

### Related issues

*
*

### Checklist

### Further comments

### DIFF
--- a/.github/workflows/auto-set-label.yml
+++ b/.github/workflows/auto-set-label.yml
@@ -1,0 +1,14 @@
+name: Assign PR team labels
+on:
+  pull_request:
+    branches:
+      - main
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setting labels
+        uses: FiligranHQ/auto-label@1.0.0
+        with:
+          labels_by_organization: "{\"FiligranHQ\":[\"filigran team\"]}"


savacano28 (approved):

.github/workflows/auto-set-label.yml
```diff
@@ -0,0 +1,14 @@
...
+      - main
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setting labels
+        uses: FiligranHQ/auto-label@1.0.0
+        with:
+          labels_by_organization: "{\"FiligranHQ\":[\"filigran team\"]}"
```
guillaumejparis, line 14: Does our bots considered in our organization ?
And if yes maybe it is not really a problem that they get the label

Dimfacion: Yes. It should be quite easy to remove them though as I think there is a type in the user information. If we feel it's annoying (like with the renovate bot having the labels), we can make sure they do not get the label and only the real users do. Or even better, since the github action is in our hands, we could add a new parameter where users have some kind of label and bots have another one ...

RomuDeuxfois: So, when a PR is open, we add a label if the user is in our organization ? This is the purpose of FiligranHQ/auto-label@1.0.0 script ?

Dimfacion: Yes, exactly : FiligranHQ/auto-label@1.0.0 is an action that can be found here : https://github.com/FiligranHQ/auto-label You can have a look at the index.js, the code is less than 100 lines and is pretty straightforward. Basically, it'll get the list of the members of the organization and if the author is in it, it'll add the labels to the PR.

RomuDeuxfois (approved):

PR #1377, opened 2024-08-30
[backend/frontend] Fix inject test condition

RomuDeuxfois:
- Reduce time execution for bulk test inject
- Remove check on teams -> when we test an inject, emails or SMS are send to the current user. Teams are not used.
- Remove InjectTestStatus on Inject class -> is not used

### DIFF
--- a/openbas-front/src/actions/injects/inject-action.ts
+++ b/openbas-front/src/actions/injects/inject-action.ts
@@ -11,7 +11,7 @@ export const testInject = (injectId: string) => {
 export const bulkTestInjects = (injectIds: string[]) => {
   const data = injectIds;
   const uri = '/api/injects/bulk/test';
-  return simplePostCall(uri, data, 'The inject does not have any team defined');
+  return simplePostCall(uri, data, "This inject can't be test");
 };
 
 // -- EXERCISES --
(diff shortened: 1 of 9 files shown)


johanah29 (commented):
image
This error message prompts when the user tries to test a technical inject using the bulk actions toolbar. Can we have a more appropriate message ?

RomuDeuxfois:
> image This error message prompts when the user tries to test a technical inject using the bulk actions toolbar. Can we have a more appropriate message ?

Fix with "This inject can't be test"

openbas-front/src/actions/injects/inject-action.ts
```diff
@@ -11,7 +11,7 @@ export const testInject = (injectId: string) => {
 export const bulkTestInjects = (injectIds: string[]) => {
   const data = injectIds;
   const uri = '/api/injects/bulk/test';
-  return simplePostCall(uri, data, 'The inject does not have any team defined');
+  return simplePostCall(uri, data, "This inject can't be test");
```
damgouj, line 14: can't be tested ?

damgouj (approved):

openbas-front/src/actions/injects/inject-action.ts
guillaumejparis: missing translation

guillaumejparis:
This message does not sense anymore ,no ?
`Only SMS and emails related injects with teams defined will be tested`

PR #1379, opened 2024-09-02
[backend/frontend] fix expectation status when updating score

MarineLeM:
issue :
https://github.com/OpenBAS-Platform/openbas/issues/1350


savacano28:
Tested OK,
Code OK,

PR #1389, opened 2024-09-05
Atomic testing list is slow on demo environment 

RomuDeuxfois:
### Proposed changes

* Add new indexes to speed up SQL Join

### Related issues

* https://github.com/OpenBAS-Platform/openbas/issues/1184

### DIFF
--- a/openbas-api/src/main/java/io/openbas/migration/V3_33__Add_Index_inject.java
+++ b/openbas-api/src/main/java/io/openbas/migration/V3_33__Add_Index_inject.java
@@ -1,0 +1,21 @@
+package io.openbas.migration;
+
+import org.flywaydb.core.api.migration.BaseJavaMigration;
+import org.flywaydb.core.api.migration.Context;
+import org.springframework.stereotype.Component;
+
+import java.sql.Statement;
+
+@Component
+public class V3_33__Add_Index_inject extends BaseJavaMigration {
+
+    @Override
+    public void migrate(Context context) throws Exception {
+        Statement select = context.getConnection().createStatement();
+        select.execute(""
+            + "CREATE INDEX idx_inject_inject_injector_contract on injects (inject_injector_contract);"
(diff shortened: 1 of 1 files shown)


openbas-api/src/main/java/io/openbas/migration/V3_33_ADD_Index_inject.java
```diff
@@ -0,0 +1,22 @@
...
+public class V3_33__Add_index extends BaseJavaMigration {
+
+    @Override
+    public void migrate(Context context) throws Exception {
+        Statement select = context.getConnection().createStatement();
+        select.execute(""
+            + "CREATE INDEX idx_inject_inject_injector_contract on injects (inject_injector_contract);"
+            + "CREATE INDEX idx_injector_contract_injector on injectors_contracts (injector_id);"
+            + "CREATE INDEX idx_injector_inject on injectors (injector_id);"
+            + ""
```
savacano28, line 19: The last one is already present (pk). I think we should generally add indexes on foreign keys.

RomuDeuxfois: Fix

RomuDeuxfois:
Ready for a second round

openbas-api/src/main/java/io/openbas/migration/V3_33_ADD_Index_inject.java
```diff
@@ -0,0 +1,21 @@
...
+
+import org.flywaydb.core.api.migration.BaseJavaMigration;
+import org.flywaydb.core.api.migration.Context;
+import org.springframework.stereotype.Component;
+
+import java.sql.Statement;
+
+@Component
+public class V3_33__Add_index extends BaseJavaMigration {
+
```
savacano28, line 11: I think the file and the class have to be named the same.

savacano28 (approved):

PR #1411, opened 2024-09-11
[frontend] Fix inject action update in inject list

RomuDeuxfois:
When we are on the inject list on the inject (scenario or simulation) and we do an action in the popover, the list is not updated.


Dimfacion (approved):
LGTM :)

PR #1461, opened 2024-09-17
[frontend] Add no data available placeholder text in empty simulations charts

isselparra:
### Proposed changes
- Add no data available placeholder text in empty scenario results chart

### Related issues

Close https://github.com/OpenBAS-Platform/openbas/issues/1371

### Checklist

- ~~[ ] I wrote test cases for the relevant uses case~~
- ~~[ ] I added/update the relevant documentation (either on github or on notion)~~

Before:
image

After:
image

### DIFF
--- a/openbas-front/src/utils/Charts.js
+++ b/openbas-front/src/utils/Charts.js
@@ -260,7 +260,6 @@ export const areaChartOptions = (
  * @param {boolean} isResult
  * @param {boolean} isFakeData
  * @param {number} max
- * @param {boolean} isChartAPlaceholder
  * @param {string} emptyChartText
  */
 export const verticalBarsChartOptions = (
@@ -275,7 +274,6 @@ export const verticalBarsChartOptions = (
   isResult = false,
   isFakeData = false,
   max = undefined,
-  isChartAPlaceholder = false,
   emptyChartText = '',
 ) => ({
   chart: {
@@ -287,10 +285,10 @@ export const verticalBarsChartOptions = (
     width: '100%',
     height: '100%',
     zoom: {
-      enabled: !isChartAPlaceholder,
+      enabled: !isFakeData,
     },
     animations: {
-      enabled: !isChartAPlaceholder,
+      enabled: !isFakeData,
     },
   },
   theme: {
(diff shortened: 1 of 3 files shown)


Dimfacion:
Seems to be working but I'm unsure that's what we want. 
@jborozco can you confirm that you want an empty block with just "No data to display" ?

jborozco:
Hello viewed on Teams with Issel, We want to have the text but also a background to look nicer

isselparra:
PR updated @jborozco @Dimfacion

guillaumejparis:
Build failed https://drone.filigran.io/OpenBAS-Platform/openbas/7119/1/7 
I think you maybe just missed some intellij plugin for eslint

isselparra:
@guillaumejparis the plugin was already installed, an intellij config needed to be updated

openbas-front/src/utils/Charts.js
```diff
@@ -273,6 +275,8 @@ export const verticalBarsChartOptions = (
   isResult = false,
   isFakeData = false,
   max = undefined,
+  isChartAPlaceholder = false,
```
Dimfacion, line 278: Can we maybe use the existing isFakeData property instead of adding a new one ?

Dimfacion:
I sometimes have an issue where 'No data to display' appears in front of the graph even though there are datas. 
image
Disabling the cache seems to fix the issue so maybe a problem with the store ?

openbas-front/src/utils/Charts.js
isselparra: If we use `isFakeData` directly we are going to also impact the `Simulations` chart in the home dashboard.
The scope described in the issue is only to impact the `Simulations results` chart at the scenario dashboard level.

Should we also modify the `Simulations` chart ? cc @jborozco

isselparra:
@Dimfacion I corrected the problem with the "No data to display" text appearing event when the data is present.

I added a loader and I don't see the issue anymore.

Let me know if it's OK on your side :pray:

openbas-front/src/utils/Charts.js
isselparra: Discussed with @jborozco, we are updating also the `Simulations` chart

guillaumejparis (approved):
LGTM

Issue #1476, opened 2024-09-18
When adding a team with multiple players, it can lead to duplicate inserted in database which generate an error


Dimfacion:
## Description

When adding a team with multiple players, it can lead to duplicate inserted in database which generate an error.

## Environment

1. OS (where OpenBAS server runs): { e.g. Mac OS 10, Windows 10, Ubuntu 16.4, etc. } All
2. OpenBAS version: { e.g. OpenBAS 1.0.2 } 1.6.1
3. OpenBAS client: { e.g. frontend or python } Frontend
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. { e.g. Run ... } Add a team with at least 1 player already existing and added to the scenario
2. { e.g. Click ... } Import injects with that same team
3. { e.g. Error ... } Error 409 ""A different object with the same identifier value was already associated with the session : [io.openbas.database.model.ScenarioTeamUser#ScenarioTeamUserId(scenarioId=xxx, teamId=xxx, userId=xxx)]"
"

## Expected Output

Import Sucessful

## Actual Output

Import in error

PR #1477, opened 2024-09-18
[backend] Small fix on adding a tuple scenario/teams/users

Dimfacion:
### Proposed changes

* Only add the tuple scenario/team/user if does not already exists
*

### Related issues

* #1476 
*

### Checklist

### Further comments

### DIFF
--- a/openbas-api/src/test/java/io/openbas/service/InjectServiceTest.java
+++ b/openbas-api/src/test/java/io/openbas/service/InjectServiceTest.java
@@ -274,6 +274,55 @@ class InjectServiceTest {
         }
     }
 
+    @DisplayName("Import an XLS file with absolute date")
+    @Test
+    void testImportXlsAbsoluteDateWithExistingScenarioTeamUser() throws IOException {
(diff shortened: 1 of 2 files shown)


openbas-api/src/test/java/io/openbas/service/InjectServiceTest.java
```diff
@@ -250,11 +253,15 @@ void testImportXlsAbsoluteDate() throws IOException {
             when(userRepository.findById(any())).thenReturn(Optional.of(mockedUser));
             Team team1 = new Team();
             team1.setName("team1");
+            team1.setUsers(List.of(new User()));
```
RomuDeuxfois, line 256: I am not sure how you cover the bug with this test

Dimfacion: That's a good point, this test will only test one of the two if fork from https://github.com/OpenBAS-Platform/openbas/pull/1477/files#diff-f41ecd0d8c78018477a00f131a786b67748d1421903f22a472a06fc9a563f20aR367
I added a test to test the other one so we should be better now.

RomuDeuxfois (approved):

PR #1495, opened 2024-09-19
[backend/frontend] Improv performance on teams

RomuDeuxfois:
### Proposed changes

* Change the query mechanism to use Criteria Builder instead of JPA

### Related issues

* https://github.com/OpenBAS-Platform/openbas/issues/1409

### Metrics

Page : Scenario ID - Defeinition
URL : /admin/scenarios/SCENARIO_ID/definition
Api Path : /api/teams
Mechanism | Jaeger time
-- | --
JPA → RawPaginationTeam with pagination (buildPaginationJPA) | 90 < 110 ms
CriteriaBuilder → TeamOutput with pagination (buildPaginationCriteriaBuilder) | 20 < 25 ms

Page : Teams
URL : /admin/teams/teams
Api Path : /api/teams/search
Mechanism | Jaeger time
-- | --
JPA → RawPaginationTeam with pagination (buildPaginationJPA) | 90 < 110 ms
CriteriaBuilder → TeamOutput with pagination (buildPaginationCriteriaBuilder) | 50 < 60 ms


openbas-front/src/components/ItemTags.js
```diff
@@ -52,6 +56,9 @@ const ItemTags = (props) => {
     const allTags = helper.getTags() ?? [];
     return allTags.filter((tag) => (tags ?? []).includes(tag.tag_id));
   });
+  useDataLoader(() => {
+    dispatch(fetchTags());
```
RomuDeuxfois, line 60: Don't do that, fetch tags on the parent side

openbas-api/src/main/java/io/openbas/service/TeamService.java
```diff
@@ -1,21 +1,142 @@
...
+    select(cb, cq, teamRoot);
+
+    // -- Specification --
+    if (specification != null) {
+      Predicate predicate = specification.toPredicate(teamRoot, cq, cb);
+      if (predicate != null) {
+        cq.where(predicate);
+      }
+    }
+
```
MarineLeM, line 103: May be we could encapsulate lines 90 to 102 into a separate function to avoid code duplication with the find function. What do you think ?

openbas-api/src/main/java/io/openbas/rest/team/ExerciseTeamApi.java
```diff
@@ -0,0 +1,45 @@
...
+
+@RequiredArgsConstructor
+@RestController
+@Secured(ROLE_USER)
+public class ExerciseTeamApi extends RestBehavior {
+
+  private final TeamService teamService;
+
+  @PostMapping(EXERCISE_URI + "/{exerciseId}/teams/search")
+  @PreAuthorize("isScenarioObserver(#exerciseId)")
```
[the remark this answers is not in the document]
RomuDeuxfois: well seen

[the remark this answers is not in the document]
RomuDeuxfois: Fix

MarineLeM:
Is it normal that I don’t have the same behavior as in the master branch? When adding a team to an exercise, I can no longer create teams, because these button doesn't appears anymore : 

image

openbas-api/src/main/java/io/openbas/service/TeamService.java
RomuDeuxfois: I agree but this is not so easy
The method should return cb, teamRoot and cq.

And I don't want to have a big structure with everything inside

MarineLeM:
i found another different behavior from 'master' branch. When you already have teams assigned to an exercise and click on 'add team', on 'master' branch nothing appears in the right square (see the image below to compare with the new behavior implemented). This component is intended solely for adding new teams, not for removing them from the exercise.

If this new behavior is intentional, please be cautious: when you attempt to remove a chip and then click 'add', you receive a message indicating that everything has been updated successfully, but in reality, the team remains assigned.
image

openbas-front/src/actions/exercises/exercise-teams-action.ts
```diff
@@ -0,0 +1,9 @@
+import type { Scenario, SearchPaginationInput } from '../../utils/api-types';
+import { simplePostCall } from '../../utils/Action';
+import { EXERCISE_URI } from './exercise-action';
+
+// eslint-disable-next-line import/prefer-default-export
+export const searchExerciseTeams = (exerciseId: Scenario['scenario_id'], paginationInput: SearchPaginationInput) => {
+  const uri = `${EXERCISE_URI}/${exerciseId}/teams/search`;
```
MarineLeM, line 7: be careful EXERCISE_URI = '/api/exercises/'  so here there is two //

openbas-api/src/main/java/io/openbas/rest/team/ExerciseTeamApi.java
```diff
@@ -0,0 +1,45 @@
...
+
+  @PostMapping(EXERCISE_URI + "/{exerciseId}/teams/search")
+  @PreAuthorize("isExerciseObserver(#exerciseId)")
+  @Transactional(readOnly = true)
+  @Tracing(name = "Paginate teams for exercise", layer = "api", operation = "POST")
+  public Page<TeamOutput> searchTeams(
+      @PathVariable @NotBlank final String exerciseId,
+      @RequestBody @Valid SearchPaginationInput searchPaginationInput) {
+    final Specification<Team> teamSpecification = contextual(false).or(fromExercise(exerciseId).and(contextual(true)));
+    return this.teamService.teamPagination(searchPaginationInput, teamSpecification);
```
MarineLeM, line 42: should be final Specification<Team> teamSpecification =fromExercise(exerciseId); 
nevermind about contextual teams because the path of your request is already /{exerciseId}/teams/search 

What do you think ?

RomuDeuxfois: We are looking teams that can be set to this exercise
We have two cases:
1. platform team -> not contextual
2. contextual team in this exercise -> contextual & exerciseId

RomuDeuxfois:
I forget this button... It should be there.

> Is it normal that I don’t have the same behavior as in the master branch? When adding a team to an exercise, I can no longer create teams, because these button doesn't appears anymore :
> 
> image

You are right.
I will change the behavior to have a replace method insteaf of an add method.
To be able to remove multiple teams easily.
Please, test it and tell me if it's better.

> i found another different behavior from 'master' branch. When you already have teams assigned to an exercise and click on 'add team', on 'master' branch nothing appears in the right square (see the image below to compare with the new behavior implemented). This component is intended solely for adding new teams, not for removing them from the exercise.
> 
> If this new behavior is intentional, please be cautious: when you attempt to remove a chip and then click 'add', you receive a message indicating that everything has been updated successfully, but in reality, the team remains assigned. image

openbas-front/src/admin/components/components/teams/AddTeams.tsx
```diff
@@ -125,90 +130,25 @@ const AddTeams: React.FC<Props> = ({ addedTeamIds, onAddTeams }) => {
...
-          <Button onClick={() => {
-            setOpen(false);
-            setKeyword('');
-            setTeamIds([]);
-          }}
-          >{t('Cancel')}</Button>
-          <Button color="secondary" onClick={submitAddTeams}>
+          <Button onClick={handleClose}>{t('Cancel')}</Button>
+          <Button color="secondary" onClick={handleSubmit}>
             {t('Add')}
```
MarineLeM, line 152: The component's behavior has changed: we can now add or remove teams. It might be more appropriate to update the label to 'Update', what do you think ? 
And also to change the DialogTitle ?

MarineLeM:
The 'remove from the context' button does not work anympre
image

openbas-front/src/admin/components/components/teams/AddTeams.tsx
RomuDeuxfois: Well seen

RomuDeuxfois:
Fix

> The 'remove from the context' button does not work anympre image

MarineLeM (approved):

PR #1510, opened 2024-09-23
[frontend/backend] remove updateAttributes and isListened from api-ty…

MarineLeM:
Remove updateAttributes and isListened from api-types file

### DIFF
--- a/openbas-front/src/utils/api-types.d.ts
+++ b/openbas-front/src/utils/api-types.d.ts
@@ -63,8 +63,6 @@ export interface Article {
   article_updated_at: string;
   /** @format date-time */
   article_virtual_publication?: string;
-  listened?: boolean;
-  updateAttributes?: object;
 }
 
 export interface ArticleCreateInput {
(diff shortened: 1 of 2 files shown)


RomuDeuxfois (approved):

openbas-front/src/utils/api-types.d.ts
```diff
@@ -1111,8 +1065,6 @@ export interface Inject {
   inject_user?: User;
   /** @format int64 */
   inject_users_number?: number;
-  listened?: boolean;
```
Dimfacion, line 1114: listened is actually used by the frontend at least in this use case. Thing is, whenever an object is updated, an event is sent to the stream indicating to the frontend that it needs to be updated. When you do something that impacts a lot of object (say when you import hundreds of injects. AFAIK, it's the only use case where we use it so far), you get an hundred events and so an hundred of individual updates which makes the frontend freeze.

The listened property is a quick fix for this issue (but not ideal). It is set to false on the backend when there are a lot of updates to avoid this issue and let the frontend do a manual global fetch instead of lots of updates after the call it made.

PR #1512, opened 2024-09-23
[backend] Fix null field in import

RomuDeuxfois:
### Proposed changes

* Improv property extraction to avoid null values


isselparra (approved):
Functional test OK

Regression detected: there seems to be a regression regarding the creation of an Scenario, when a severity is specified, on the UI it appears as "Unknown". cc @RomuDeuxfois

RomuDeuxfois:
I fix the regression @isselparra

isselparra (approved):
Test OK

- The Severity regression is not present anymore
- On import/export the severity was taken into account

PR #1519, opened 2024-09-23
[frontend/backend] add inject result inside report

MarineLeM:
issue :  https://github.com/OpenBAS-Platform/openbas/issues/1080
Chunk 2 

add result inject inside report 
image


openbas-api/src/main/java/io/openbas/service/ReportService.java
```diff
@@ -50,6 +53,24 @@ public Report updateReport(@NotNull final Report report, @NotNull final ReportIn
         return this.reportRepository.save(report);
     }
 
+    public List<ReportInjectComment> updateReportInjectComment(@NotNull final Report report, @NotNull final Inject inject, @NotNull final ReportInjectCommentInput input){
+        ReportInjectComment reportInjectComment = report.getReportInjectsComments().stream()
+                .filter(c -> c.getInject().getId().equals(input.getInjectId()))
+                .findFirst()
+                .orElse(null);
```
RomuDeuxfois, line 60: Can we have the direct query to do that ? 
To avoid hibernate to lazy load too much datas

openbas-front/src/utils/hooks/useDataLoader.js
```diff
@@ -18,7 +18,7 @@ const ERROR_30S_DELAY = 30000;
 let sseClient;
 let lastPingDate = new Date().getTime();
 const listeners = new Map();
-const useDataLoader = (loader = () => {}) => {
+const useDataLoader = (loader = () => {}, refetchArg = []) => {
```
RomuDeuxfois, line 21: Can you explain this changes ?
Thank you

openbas-front/src/admin/components/simulations/simulation/reports/InjectReportResult.tsx
```diff
@@ -0,0 +1,183 @@
...
+    },
+  ];
+
+  return (
+    <div style={style}>
+      <Typography variant="h4" gutterBottom>
+        {t('Injects results')}
+      </Typography>
+
+      <TableContainer component={Paper}>
```
RomuDeuxfois, line 143: I am not sure this is the UI we want
The background and the border are differents
image

openbas-front/src/admin/components/simulations/simulation/reports/InjectReportResult.tsx
```diff
@@ -0,0 +1,183 @@
...
+        <Table aria-label="injects results">
+          <TableHead>
+            <TableRow>
+              {columns.map((col) => (
+                <TableCell
+                  sx={col.label === 'Comment' ? { padding: '0px', width: '35%', flexGrow: 1 } : {}}
```
RomuDeuxfois, line 149:
```suggestion
                  sx={col.label === 'Comments' ? { padding: '0px', width: '35%', flexGrow: 1 } : {}}
```

openbas-front/src/admin/components/simulations/simulation/reports/InjectReportResult.tsx
```diff
@@ -0,0 +1,183 @@
...
+      label: 'Targets',
+      render: (inject: InjectResultDTO) => <ItemTargets targets={inject.inject_targets} />,
+    },
+    {
+      label: 'Comments',
+      render: (inject: InjectResultDTO) => {
+        const currentInjectComment = findInjectCommentsByInjectId(inject.inject_id);
+
+        return (editInjectId === inject.inject_id
+          ? <TextField
```
RomuDeuxfois, line 106: This is a new behavior in the OpenBAS platform.
Are we sure this is what we want ?

openbas-model/src/main/java/io/openbas/database/model/ReportInjectComment.java
```diff
@@ -0,0 +1,48 @@
...
+    @JsonProperty("report_inject_comment")
+    private String comment;
+
+    @JsonProperty("inject_id")
+    public String getInjectId() {
+        return inject != null ? inject.getId() : null; // Customize serialization to return ID
+    }
+
+    @JsonProperty("report_id")
+    public String getReportId() {
```
RomuDeuxfois, line 45: reportId is String here and UUID in ReportInjectCommentId ?
I don't understand

openbas-front/src/utils/hooks/useDataLoader.js
MarineLeM: Let’s return to the exerciseReportContent folder. I need the useDataLoader function to refetch data every time the report changes. To achieve this, I want to pass a refetch argument into the useEffect hook inside the useDataLoader function. 

By implementing this approach, I ensure that only the necessary data for the report is loaded, rather than fetching all available data. Whenever I update my report configuration, the useDataLoader function will refetch only the specific data required for the updated report.

openbas-api/src/main/java/io/openbas/service/ReportService.java
```diff
@@ -50,6 +55,33 @@ public Report updateReport(@NotNull final Report report, @NotNull final ReportIn
...
+
+    private ReportInjectComment findReportInjectComment(String reportId, String injectId) {
+        String jpql = "SELECT r FROM ReportInjectComment r WHERE r.report.id = :reportId AND r.inject.id = :injectId";
+
+        return this.entityManager.createQuery(jpql, ReportInjectComment.class)
+            .setParameter("reportId", UUID.fromString(reportId))
+            .setParameter("injectId", injectId)
+            .getResultStream()
+            .findFirst()
+            .orElse(null);
```
savacano28, line 82: Maybe return an optional: private Optional<ReportInjectComment>?

openbas-api/src/main/java/io/openbas/service/ReportService.java
```diff
@@ -50,6 +53,33 @@ public Report updateReport(@NotNull final Report report, @NotNull final ReportIn
...
+            injectComment.setReport(report);
+            injectComment.setComment(input.getComment());
+            report.getReportInjectsComments().add(injectComment);
+        }
+        this.reportRepository.save(report);
+        return report.getReportInjectsComments();
+    }
+
+    private Optional<ReportInjectComment> findReportInjectComment(String reportId, String injectId) {
+        String jpql = "SELECT r FROM ReportInjectComment r WHERE r.report.id = :reportId AND r.inject.id = :injectId";
```
RomuDeuxfois, line 74: Could be better to have this directly in the ReportRepository

    @Query(value = "SELECT r FROM ReportInjectComment r WHERE r.report.id = :reportId AND r.inject.id = :injectId", nativeQuery = true)
    Optional<ReportInjectComment> method(@Param("PARAM")...);

openbas-front/src/utils/hooks/useDataLoader.js
RomuDeuxfois: Great, thanks for the explanation

RomuDeuxfois:
The report takes a lot of time to show up...

openbas-front/src/components/fields/MarkDownField.tsx
```diff
@@ -1,41 +1,22 @@
 import React, { useState } from 'react';
-import { useController, useFormContext } from 'react-hook-form';
 import MDEditor, { commands } from '@uiw/react-md-editor/nohighlight';
-import { Box, FormHelperText, InputLabel, Typography } from '@mui/material';
 import { ICommand } from '@uiw/react-md-editor';
-import TextFieldAskAI from '../../admin/components/common/form/TextFieldAskAI';
+import { Typography } from '@mui/material';
```
RomuDeuxfois, line 4: It's a good idea to break the component down like that. Thanks!

openbas-api/src/main/java/io/openbas/rest/report/ReportApi.java
```diff
@@ -23,6 +25,7 @@ public class ReportApi extends RestBehavior {
 
     private final ExerciseService exerciseService;
     private final ReportService reportService;
+    private final InjectService injectService;
 
     @GetMapping("/api/reports/{reportId}")
```
savacano28, line 30: Maybe we could include exerciseId /exercises/exerciseId/reports..

savacano28:
I have a question about the path: /reports/exerciseId/reportId, maybe could be better: /exercises/exerciseId/reports/reportId ? ty!

savacano28:
Maybe we could manage better when a report no longer exists. Currently, we have : 
image

savacano28 (approved):

RomuDeuxfois (approved):

PR #1533, opened 2024-09-25
[frontend] Added space to avoid banner goes over the drawer title

RomuDeuxfois:
### Proposed changes

* Just add space logic in Drawer like we do on leftbar, topbar, ect
Before
image

After
image


savacano28 (approved):

guillaumejparis:
There are other places where this is a problem I will fix and commit on this PR

PR #1546, opened 2024-09-26
[frontend/backend] add global observation

MarineLeM:
issue : https://github.com/OpenBAS-Platform/openbas/issues/1080

Chunk 3 
Add global observation on report
image


openbas-front/src/admin/components/simulations/simulation/reports/ExerciseReportContent.tsx
```diff
@@ -129,10 +129,38 @@ const ExerciseReportContent: React.FC<Props> = ({ report, exerciseId, canWrite =
...
+                  value={globalObservationRef.current}
+                  onChange={(value: string) => {
+                    globalObservationRef.current = value;
+                  }}
+                  style={{ height: 200, width: '100%' }}
+                  onBlur={() => updateReportGlobalObservation(exerciseId, report.report_id, {
+                    report_informations: report.report_informations,
+                    report_global_observation: globalObservationRef.current,
+                    report_name: report.report_name,
+                  } as ReportInput)
```
savacano28, line 152: Maybe extract: handleBlur() ?

openbas-front/src/admin/components/simulations/simulation/reports/ExerciseReportContent.tsx
```diff
@@ -129,10 +129,38 @@ const ExerciseReportContent: React.FC<Props> = ({ report, exerciseId, canWrite =
...
+                }
+                /> : <Paper variant="outlined" sx={{ padding: '10px 15px 10px 15px' }}>
+                  {globalObservationRef.current
+                    ? <div dangerouslySetInnerHTML={{ __html: globalObservationRef.current }}/>
+                    : <div style={{ textTransform: 'none' }}>
```
savacano28, line 157: Do we need to sanitize dangerouslySetInnerHTML? 
We use it for the articles with <ExpandableMarkdown source={article.article_content ?? ''}/>. Maybe we could use it here as well?

savacano28 (approved):

PR #1584, opened 2024-09-30
Improve latency on endpoint teams

RomuDeuxfois:
### Related issues

* https://github.com/OpenBAS-Platform/openbas/issues/1581

### DIFF
--- a/openbas-api/src/main/java/io/openbas/injectors/email/service/ImapService.java
+++ b/openbas-api/src/main/java/io/openbas/injectors/email/service/ImapService.java
@@ -70,7 +70,7 @@ public class ImapService {
     private FileService fileService;
     private final PlatformSettingsService platformSettingsService;
 
-    public ImapService(Environment env, @Autowired PlatformSettingsService platformSettingsService) throws Exception {
+    public ImapService(Environment env, @Autowired PlatformSettingsService platformSettingsService) {
         this.platformSettingsService = platformSettingsService;
         try {
             initStore(env);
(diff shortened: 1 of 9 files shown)


openbas-api/src/main/java/io/openbas/injectors/email/service/ImapService.java
```diff
@@ -294,6 +294,7 @@ private void synchronizeBox(Folder inbox, Boolean isSent) throws Exception {
         }
         currentState.setValue(String.valueOf(messageCount));
         settingRepository.save(currentState);
+        inbox.close();
```
RomuDeuxfois, line 297: Maybe it should be better to use the aucloseable interface and change the syncFolders method to

```private void syncFolders() throws Exception {
    try {
        // Sync sent
        try (Folder sentBox = imapStore.getFolder(sentFolder)) {
            sentBox.open(Folder.READ_ONLY);
            synchronizeBox(sentBox, true);
        }

        // Sync received
        for (String listeningFolder : inboxFolders) {
            try (Folder inbox = imapStore.getFolder(listeningFolder)) {
                inbox.open(Folder.READ_ONLY);
                synchronizeBox(inbox, false);
            }
        }
    } catch (MessagingException e) {
        log.warning("Connection failure: " + e.getMessage());
        // Retry logic or rethrow the exception
        retrySyncFolders();
    }
}```

What do you think ?

File: openbas-api/src/main/java/io/openbas/injectors/email/service/ImapService.java [hunk 2]
```diff
@@ -325,6 +326,9 @@ private void retrySyncFolders() throws Exception {
                 Thread.sleep(2000);
             }
         }
+        if(imapStore != null && imapStore.isConnected()) {
+            imapStore.close();
```
RomuDeuxfois, line 330: I am not sure we understand each other on this point.
For me, we can close imapstore only if we can reconnect with it

```
    private void retrySyncFolders() throws Exception {
        for (int i = 0; i < 3; i++) {
            try {
                syncFolders();
                break;
            } catch (MessagingException e) {
                log.warning("Retrying connection..." + e.getMessage());
                Thread.sleep(2000);
                if (i == 2) {
                        imapStore.close();
                }
            }
        }
    }
```
But, we want to have the imapstore opened for all the application life. So I don't know exactly the good solution on this one.

damgouj (approved):

PR #1589, opened 2024-10-01
[backend] Delete duplicated joins on default query

### LINKED ISSUE #1583: Improve latency on page: admin/scenarios
## Use case

API call spotted with default filters: 
- /api/scenarios/search

savacano28:
### Proposed changes

Delete duplicated joins when any filter is added:
Execution Time: 36.735 ms
Network response time: 45.28 ms
*     select
        distinct s1_0.scenario_id,
        s1_0.scenario_name,
        s1_0.scenario_severity,
        s1_0.scenario_category,
        s1_0.scenario_recurrence,
        s1_0.scenario_updated_at,
        array_remove(array_agg(t1_0.tag_id), null),
        array_union_agg(ic1_0.injector_contract_platforms) 
        from
            scenarios s1_0 
        left join
            scenarios_tags t1_0 
                on s1_0.scenario_id=t1_0.scenario_id 
        left join
            injects i1_0 
                on s1_0.scenario_id=i1_0.inject_scenario 
        left join
            injectors_contracts ic1_0 
                on ic1_0.injector_contract_id=i1_0.inject_injector_contract 
        where
            1=1 
        group by
            1 

### Related issues

* Closes #1583 

### Checklist

### Further comments

### DIFF
--- a/openbas-api/src/main/java/io/openbas/rest/scenario/utils/ScenarioUtils.java
+++ b/openbas-api/src/main/java/io/openbas/rest/scenario/utils/ScenarioUtils.java
@@ -37,6 +37,23 @@ public class ScenarioUtils {
     // Existence of the filter
     Optional<Filters.Filter> scenarioRecurrenceFilterOpt = ofNullable(searchPaginationInput.getFilterGroup())
         .flatMap(f -> f.findByKey(SCENARIO_RECURRENCE_FILTER));
+    Optional<Filters.Filter> scenarioKillChainsFilterOpt = ofNullable(searchPaginationInput.getFilterGroup())
+        .flatMap(f -> f.findByKey(SCENARIO_KILL_CHAIN_PHASES_FILTER));
+    Optional<Filters.Filter> scenarioTagsFilterOpt = ofNullable(searchPaginationInput.getFilterGroup())
+        .flatMap(f -> f.findByKey(SCENARIO_TAGS_FILTER));
+
+    if (scenarioKillChainsFilterOpt.isPresent()) {
+    searchPaginationInput.getFilterGroup().removeByKey(SCENARIO_KILL_CHAIN_PHASES_FILTER);
+      if (!scenarioKillChainsFilterOpt.get().getValues().isEmpty()) {
+        return (Specification<Scenario> specification) -> specification;
+      }
+    }
+    if (scenarioTagsFilterOpt.isPresent()) {
+    searchPaginationInput.getFilterGroup().removeByKey(SCENARIO_TAGS_FILTER);
+      if (!scenarioTagsFilterOpt.get().getValues().isEmpty()) {
+        return (Specification<Scenario> specification) -> specification;
+      }
+    }
 
     if (scenarioRecurrenceFilterOpt.isPresent()) {
       // Purge filter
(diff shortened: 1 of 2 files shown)


openbas-model/src/main/java/io/openbas/database/model/Scenario.java
```diff
@@ -31,15 +31,6 @@
 @Entity
 @Table(name = "scenarios")
 @EntityListeners(ModelBaseListener.class)
-@NamedEntityGraphs({
```
RomuDeuxfois, line 34: Good move

openbas-api/src/main/java/io/openbas/rest/scenario/utils/ScenarioUtils.java
```diff
@@ -21,37 +22,69 @@ private ScenarioUtils() {
...
-      @NotNull final SearchPaginationInput searchPaginationInput) {
-    // Existence of the filter
-    Optional<Filters.Filter> scenarioRecurrenceFilterOpt = ofNullable(searchPaginationInput.getFilterGroup())
-        .flatMap(f -> f.findByKey(SCENARIO_RECURRENCE_FILTER));
+      @NotNull SearchPaginationInput searchPaginationInput) {
+
+    // Extract filters from the input
+    Optional<Filters.Filter> scenarioRecurrenceFilterOpt = getFilter(searchPaginationInput, SCENARIO_RECURRENCE_FILTER);
+    Optional<Filters.Filter> scenarioKillChainsFilterOpt = getFilter(searchPaginationInput, SCENARIO_KILL_CHAIN_PHASES_FILTER);
+    Optional<Filters.Filter> scenarioTagsFilterOpt = getFilter(searchPaginationInput, SCENARIO_TAGS_FILTER);
```
RomuDeuxfois, line 42: Do you think we can have a generic system to remove double join ?
Check if the join is already there and don't do it again ?


I think we have this kind of problem on exercise and teams too

savacano28: Yes, I think it’s better to find another generic solution, but the double joins are normal and are indeed created based on our requests. For example, if we ask for two different attributes from 'inject,' the database has to perform double joins with 'inject.' In this case (scenarios), the query also has a raw format, which requires including the joins for tags and platform in the query base.

RomuDeuxfois: We have a join from the toPath method in the resolution of filter
And we have a join in the criteriabuilder when we build the scenario object

The way you propose can be a good solution if we can't arrive to generalize something. 

We can try to reuse the join from the scenario object when we resolve filters. Maybe by having a request context to populate with a map of Join object.

savacano28: Yeah, I initially tried to use the same join inject.contract for both platforms and kill chains (with modified paths and move getters) but I couldn't get it to work. I’m going to re-try the two options you suggested: 1. Verify duplicated joins, and 2. Force the use of the same joins. Thanks!

savacano28:
Replace by PR: https://github.com/OpenBAS-Platform/openbas/pull/1619

PR #1628, opened 2024-10-08
[backend/frontend] fix bulk update injects

savacano28:
### Proposed changes

* Add other endpoint to bulk update in order to just update properties available on bulk update.
*

### Related issues

* Closes #https://github.com/OpenBAS-Platform/openbas/issues/1627 
*

### Checklist

### Further comments

### DIFF
--- a/openbas-api/src/main/java/io/openbas/rest/inject/InjectApi.java
+++ b/openbas-api/src/main/java/io/openbas/rest/inject/InjectApi.java
@@ -539,7 +539,7 @@ public class InjectApi extends RestBehavior {
       @PathVariable @NotBlank final String injectId,
       @Valid @RequestBody @NotNull InjectInput input) {
     Scenario scenario = this.scenarioService.scenario(scenarioId);
-    Inject inject = bulkUpdateInject(injectId, input);
+    Inject inject = updateInject(injectId, input);
 
     // If Documents not yet linked directly to the exercise, attached it
     inject.getDocuments().forEach(document -> {
--- a/openbas-api/src/main/java/io/openbas/rest/inject/InjectApi.java
+++ b/openbas-api/src/main/java/io/openbas/rest/inject/InjectApi.java
@@ -519,6 +519,19 @@ public class InjectApi extends RestBehavior {
   }
 
   @Transactional(rollbackFor = Exception.class)
+  @PutMapping(INJECT_URI + "/{scenarioId}/injects/{injectId}/bulk")
+  @PreAuthorize("isScenarioPlanner(#scenarioId)")
+  public Inject bulkUpdateInjectForScenario(
+      @PathVariable String scenarioId,
+      @PathVariable String injectId,
+      @Valid @RequestBody InjectInput input) {
+    Scenario scenario = this.scenarioService.scenario(scenarioId);
+    Inject inject = bulkUpdateInject(injectId, input);
+    this.scenarioService.updateScenario(scenario);
+    return injectRepository.save(inject);
+  }
+
+  @Transactional(rollbackFor = Exception.class)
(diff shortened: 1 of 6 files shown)


openbas-api/src/main/java/io/openbas/rest/inject/InjectApi.java
```diff
@@ -505,6 +518,19 @@ public Inject scenarioInject(
...
+  @Transactional(rollbackFor = Exception.class)
+  @PutMapping(INJECT_URI + "/{scenarioId}/injects/{injectId}/bulk")
+  @PreAuthorize("isScenarioPlanner(#scenarioId)")
+  public Inject bulkUpdateInjectForScenario(
+      @PathVariable String scenarioId,
+      @PathVariable String injectId,
+      @Valid @RequestBody InjectInput input) {
+    Scenario scenario = this.scenarioService.scenario(scenarioId);
+    Inject inject = bulkUpdateInject(injectId, input);
+    this.scenarioService.updateScenario(scenario);
```
Dimfacion, line 530: Is it really needed ? Just from a shallow view, it looks like we get the scenario and update it right after but is it really having an effect ?

openbas-api/src/main/java/io/openbas/rest/inject/InjectApi.java
```diff
@@ -177,6 +177,19 @@ public Inject updateInject(
...
+  @Transactional(rollbackFor = Exception.class)
+  @PutMapping(INJECT_URI + "/{exerciseId}/{injectId}/bulk")
+  @PreAuthorize("isExercisePlanner(#exerciseId)")
+  public Inject bulkUpdateInject(
+      @PathVariable String exerciseId,
+      @PathVariable String injectId,
+      @Valid @RequestBody InjectInput input) {
+    Exercise exercise = exerciseRepository.findById(exerciseId).orElseThrow(ElementNotFoundException::new);
+    Inject inject = bulkUpdateInject(injectId, input);
+    this.exerciseRepository.save(exercise);
```
Dimfacion, line 189: Same as https://github.com/OpenBAS-Platform/openbas/pull/1628/files#r1796824058

openbas-api/src/main/java/io/openbas/rest/inject/InjectApi.java
```diff
@@ -505,6 +518,19 @@ public Inject scenarioInject(
     return injectRepository.findById(injectId).orElseThrow(ElementNotFoundException::new);
   }
 
+  @Transactional(rollbackFor = Exception.class)
+  @PutMapping(INJECT_URI + "/{scenarioId}/injects/{injectId}/bulk")
+  @PreAuthorize("isScenarioPlanner(#scenarioId)")
+  public Inject bulkUpdateInjectForScenario(
+      @PathVariable String scenarioId,
+      @PathVariable String injectId,
```
Dimfacion, line 526: I'm gonna be super annoying but in a bulk update endpoint, shouldn't we send the list of injects to update in one call instead of having many calls viewed as bulk ? Since this endpoint is dedicated to those kind of calls, that would make sense and prevent us from calling a lot of times the API for almost the same call with just the inject id changing ...

savacano28: You are right! I'll modify it, thank you!

openbas-api/src/main/java/io/openbas/rest/inject/InjectApi.java
savacano28: Ty!

openbas-api/src/main/java/io/openbas/rest/inject/InjectApi.java
savacano28: nop, you are right :), ty!

Dimfacion (approved):

Issue #1485, opened 2024-09-19
[ Unit testing - simulation scope ] - inject creation/update


EllynBsc:
Write unit tests in the context of a simulation for the inject creation/update
 ex: update simulation launch time, make sure we don’t lose the content of the inject

PR #1636, opened 2024-10-14
[backend] Inject creation/update in a simulation (#1485)

johanah29:
#1485

### DIFF
--- a/openbas-api/src/test/java/io/openbas/rest/InjectApiTest.java
+++ b/openbas-api/src/test/java/io/openbas/rest/InjectApiTest.java
@@ -15,7 +15,6 @@ import org.junit.jupiter.api.*;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.MediaType;
 import org.springframework.test.web.servlet.MockMvc;
-import org.springframework.web.bind.annotation.PutMapping;
 
 import java.time.Instant;
 import java.util.List;
--- a/openbas-api/src/test/java/io/openbas/rest/InjectApiTest.java
+++ b/openbas-api/src/test/java/io/openbas/rest/InjectApiTest.java
@@ -40,7 +40,6 @@ class InjectApiTest extends IntegrationTest {
   static Document DOCUMENT2;
   static Team TEAM;
   static String SCENARIO_INJECT_ID;
-  static String EXERCISE_INJECT_ID;
 
   @Autowired
   private MockMvc mvc;
@@ -292,7 +291,6 @@ class InjectApiTest extends IntegrationTest {
     // -- ASSERT --
     assertNotNull(response);
     assertEquals("Test inject", JsonPath.read(response, "$.inject_title"));
-    EXERCISE_INJECT_ID = JsonPath.read(response, "$.inject_id");
 
   }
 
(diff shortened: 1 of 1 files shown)


openbas-api/src/test/java/io/openbas/rest/InjectApiTest.java
```diff
@@ -265,6 +267,71 @@ void deleteInjectsForScenarioTest() throws Exception {
 
   // -- EXERCISES --
 
+
+  @DisplayName("Add an inject for simulation")
+  @Test
+  @Order(9)
```
MarineLeM, line 273: this annotation is unnecessary unit testing function should be independant so they should be processed at any order

openbas-api/src/test/java/io/openbas/rest/InjectApiTest.java
```diff
@@ -265,6 +267,71 @@ void deleteInjectsForScenarioTest() throws Exception {
...
+        .andExpect(status().is2xxSuccessful())
+        .andReturn()
+        .getResponse()
+        .getContentAsString();
+
+    // -- ASSERT --
+    assertNotNull(response);
+    EXERCISE_INJECT_ID = JsonPath.read(response, "$.inject_id");
+    response = mvc
+        .perform(get(EXERCISE_URI + "/" + EXERCISE.getId())
```
MarineLeM, line 297: If this request fails, your test will fail even if the add inject request succeeds. This is unit testing the add inject request test should be independant from the fetch injects request

openbas-api/src/test/java/io/openbas/rest/InjectApiTest.java
```diff
@@ -265,6 +267,71 @@ void deleteInjectsForScenarioTest() throws Exception {
...
+        .andExpect(status().is2xxSuccessful())
+        .andReturn()
+        .getResponse()
+        .getContentAsString();
+    assertEquals(EXERCISE_INJECT_ID, JsonPath.read(response, "$.exercise_injects[0]"));
+  }
+
+  @DisplayName("Update inject for simulation")
+  @Test
+  @Order(10)
```
MarineLeM, line 308: Same as above

openbas-api/src/test/java/io/openbas/rest/InjectApiTest.java
```diff
@@ -265,6 +267,71 @@ void deleteInjectsForScenarioTest() throws Exception {
...
+  @WithMockPlannerUser
+  void updateInjectForSimulationTest() throws Exception {
+    // -- PREPARE --
+    Inject inject = injectRepository.findById(EXERCISE_INJECT_ID).orElseThrow();
+    InjectInput input = new InjectInput();
+    String injectTitle = "A new title";
+    input.setTitle(injectTitle);
+    input.setInjectorContract(inject.getInjectorContract().map(InjectorContract::getId).orElse(null));
```
MarineLeM, line 316: Why are you setting the injector contract, if you don't check it in the assert part ?

openbas-api/src/test/java/io/openbas/rest/InjectApiTest.java
```diff
@@ -39,6 +40,7 @@ class InjectApiTest extends IntegrationTest {
   static Document DOCUMENT2;
   static Team TEAM;
   static String SCENARIO_INJECT_ID;
+  static String EXERCISE_INJECT_ID;
```
MarineLeM, line 43: I don't think it's necessary to create a global variable since it's only used once. What do you think ?

johanah29: Since it is not used as it was anymore, it is better to delete it.

MarineLeM (approved):

PR #1640, opened 2024-10-14
[frontend] Chaining injects logically - UI Improvements

### LINKED ISSUE #1385: Chaining inject
chunk 3

Dimfacion:
### Proposed changes

* Better node positionning
* Labels on edges

### Related issues

* Closes #1385 
*

### Checklist

### Further comments

### DIFF
--- a/openbas-front/src/components/ChainedTimeline.tsx
+++ b/openbas-front/src/components/ChainedTimeline.tsx
@@ -121,6 +121,8 @@ const ChainedTimelineFlow: FunctionComponent<Props> = ({
   ];
   const gapSize = 125;
   const newNodeSize = 50;
+  const nodeHeightClearance = 200;
+  const nodeWidthClearance = 350;
 
   let startDate: string | undefined;
 
@@ -171,8 +173,8 @@ const ChainedTimelineFlow: FunctionComponent<Props> = ({
       const dependencies = nodesAvailable.filter((dependencyNode) => nodesId.includes(dependencyNode.id));
       const minX = Math.min(currentNode.position.x, ...dependencies.map((value) => value.data.boundingBox!.topLeft.x));
       const minY = Math.min(currentNode.position.y, ...dependencies.map((value) => value.data.boundingBox!.topLeft.y));
-      const maxX = Math.max(currentNode.position.x + 250, ...dependencies.map((value) => value.data.boundingBox!.bottomRight.x));
-      const maxY = Math.max(currentNode.position.y + 200, ...dependencies.map((value) => value.data.boundingBox!.bottomRight.y));
+      const maxX = Math.max(currentNode.position.x + nodeWidthClearance, ...dependencies.map((value) => value.data.boundingBox!.bottomRight.x));
+      const maxY = Math.max(currentNode.position.y + nodeHeightClearance, ...dependencies.map((value) => value.data.boundingBox!.bottomRight.y));
       return {
         topLeft: { x: minX, y: minY },
         bottomRight: { x: maxX, y: maxY },
@@ -180,7 +182,7 @@ const ChainedTimelineFlow: FunctionComponent<Props> = ({
     }
     return {
       topLeft: currentNode.position,
-      bottomRight: { x: currentNode.position.x + 250, y: currentNode.position.y + 200 },
+      bottomRight: { x: currentNode.position.x + nodeWidthClearance, y: currentNode.position.y + nodeHeightClearance },
     };
   };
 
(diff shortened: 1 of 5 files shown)


openbas-front/src/components/common/chaining/ChainingUtils.tsx
```diff
@@ -15,6 +22,24 @@ const fromInjectDependencyToInputDependency = (dependencies: InjectDependency[])
...
+      label += ' OR ';
+    } else {
+      const splittedValues = splittedConditions[i].trim().split(breakpointValue);
+      const key = Array.from(splittedValues[0].trim().matchAll(typeFromName), (m) => m[1]);
+      label += `${key} is ${splittedValues[1].trim() === 'true' ? 'Success' : 'Failure'}`;
+    }
+  }
+
+  return label;
+};
```
isselparra, line 41: **suggestion (non-blocking):** when possible, it's preferable to use a functional approach, since immutability and no side-effects reduce bugs and it can be more easy to understand code

here we could use `reduce` (not tested):
```
const fromInjectDependencyToLabel = (dependency: string) => {
  const splittedConditions = dependency.split(breakpointAndOr);

  return splittedConditions.reduce((label, condition) => {
    const trimmedCondition = condition.trim();

    if (trimmedCondition === '&&') {
      return label + ' AND ';
    } else if (trimmedCondition === '||') {
      return label + ' OR ';
    } else {
      const splittedValues = trimmedCondition.split(breakpointValue);
      const key = Array.from(splittedValues[0].trim().matchAll(typeFromName), (m) => m[1]);
      return label + `${key} is ${splittedValues[1].trim() === 'true' ? 'Success' : 'Failure'}`;
    }
  }, '');
};
```

openbas-front/src/components/ChainedTimeline.tsx
```diff
@@ -148,33 +150,102 @@ const ChainedTimelineFlow: FunctionComponent<Props> = ({
...
+          && nextNode.data.inject?.inject_depends_on !== undefined
+          && nextNode.data.inject?.inject_depends_on !== null
+          && nodeInjects[i].id in nextNode.data.inject!.inject_depends_on);
+
+      childrens = childrens.sort((a, b) => a.data.inject!.inject_depends_duration - b.data.inject!.inject_depends_duration);
+
+      for (let j = 0; j < childrens.length; j += 1) {
+        reorganizedInjects = moveItem(reorganizedInjects, i + j + 1, reorganizedInjects.indexOf(childrens[j], i));
+      }
+    }
```
isselparra, line 207: **nitpick (non-blocking):** it's possible to access the index with `forEach`

not-tested:
```
let reorganizedInjects = nodeInjects;

nodeInjects.forEach((node, i) => {
  let childrens = reorganizedInjects.slice(i).filter((nextNode) => 
    nextNode.id !== node.id &&
    nextNode.data.inject?.inject_depends_on !== undefined &&
    nextNode.data.inject?.inject_depends_on !== null &&
    node.id in nextNode.data.inject!.inject_depends_on
  );

  childrens = childrens.sort((a, b) => 
    a.data.inject!.inject_depends_duration - b.data.inject!.inject_depends_duration
  );

  childrens.forEach((child, j) => {
    reorganizedInjects = moveItem(reorganizedInjects, i + j + 1, reorganizedInjects.indexOf(child, i));
  });
});
```

isselparra:
A few observations after testing:

- On the injects tab, with the interactive view activated, when I add a new inject it doesn't appear immediately on chart, I need to go else where and when I come back it appears.
- At one point I had three injects, the first two with a logical chain, then when I clicked to add a logical chain between injects 2 and 3, the logical chain between 1 and 2 disappeared.

openbas-front/src/components/common/chaining/ChainingUtils.tsx
Dimfacion: Good point ! Doing :)

openbas-front/src/components/ChainedTimeline.tsx
Dimfacion: Yes, I'm always a bit reluctant to use functional when I know I might touch the array itself but in this case, I don't so you're right, that makes totally sense :) Doing !

Dimfacion:
@isselparra I haven't had issues since the refactoring on #1551. Could you try again and confirm you cannot reproduce these issues ?

isselparra:
@Dimfacion 

- On the injects tab, with the interactive view activated, when I add a new inject it doesn't appear immediately on chart, I need to go else where and when I come back it appears. **=> OK, working fine**
- At one point I had three injects, the first two with a logical chain, then when I clicked to add a logical chain between injects 2 and 3, the logical chain between 1 and 2 disappeared.**=> OK, I don't see the problem anymore**

Now I found another issue
- I have 3 `Injects`
- I added a logic chain between 1 and 2, and 2 and 3, all OK
- I updated the logical chain in 3 to include Prevention and Detention, and visually the logical chain between 1 and 2 "disappeared"
Screenshot from 2024-10-22 14-36-07
- After a refresh the logical chain between 1 and 2 re-appeared
Screenshot from 2024-10-22 14-36-21

Dimfacion:
@isselparra This should be ok now :)

isselparra (approved):
Test OK

PR #1641, opened 2024-10-15
[frontend] Documents are not duplicated when an inject is duplicated

### LINKED ISSUE #1629: Documents are not duplicated when an inject is duplicated
## Reproducible Steps

1. Duplicate an inject with documents
2. Click on duplicated inject
3. Modify any property (name, assets)
4. Click on update button

## Expected Output

Documents should be still linked to duplicated inject

## Actual Output

Documents are deleted from duplicated inject

## Additional information

If you open the update drawer and click outside of it, the duplicated inject retains the documents. However, if you click the update button, the documents are deleted.

## Screenshots (optional)
We send an object without documents, so documents are set as null:
image

isselparra:
### Proposed changes
- Wait for Inject to load before being able to edit

### Related issues

- Closes #1629 

### Checklist


savacano28:
The documents are not deleted after updating the injection, but the drawer displays information more slowly. Maybe we should create a ticket to address this issue?

savacano28 (approved):

isselparra:
Yes @savacano28 this is a quick fix, we need another ticker to address the issue in more depth.

PR #1709, opened 2024-10-21
[frontend] add pagination on simulation inject list + update list whe…

### LINKED ISSUE #1633: Import injects: Pagination hides injects over 100 + action only works if the user refreshes
## Description
Go to the injects tab inside of a simulation and import a XLS file of 99 injects 
[testchronogramme.xls](https://github.com/user-attachments/files/17344010/testchronogramme.xls)

**2 problems appear:**

1. If you don't refresh the page, you won't see any injects getting imported => which leads the user to redo the same process => And then leads to think that the feature does not work and import doublons. See [this issue ](https://github.com/OpenBAS-Platform/openbas/issues/1630)
3. The current pagination reduces the injects display to a number less than 100. If you want to check that the inject is properly imported, you can look for it in the search bar and find it

MarineLeM:
Close #1633


damgouj (commented):
Ok for me, just a comment, when you create a new simulation, the default element number of the inject pagination is 100 and is 20 for the other screen ? Could you fix it to put the default ?

damgouj (approved):

Issue #1778, opened 2024-10-30
Non-admin user granted for a simulation not able to access it 


EllynBsc:
## Description

A non-admin user granted for a simulation is not able to access it

I reproduced the error and I get a 401.

https://www.loom.com/share/62bf820965b64d79a72ec2de6d0e99cc

Suggestion solution:
-Bring clarity on the error page - create a 401 unauthorized error page to specify to our user that they're unauthorized.

PR #1790, opened 2024-10-31
[frontend/backend] correct logout redirection

MarineLeM:
Close : #1778 

The user creation functionality worked correctly. 
However, the issue was when clicking the logout button: the page displayed the login form, but the URL did not update accordingly.

Moreover i update the status code in the back 
when a user is connected but don't have access the api return 404
when a user is not connected anymore and want to have access to some information the api return 401

### DIFF
--- a/openbas-api/src/main/java/io/openbas/rest/helper/RestBehavior.java
+++ b/openbas-api/src/main/java/io/openbas/rest/helper/RestBehavior.java
@@ -22,6 +22,7 @@ import org.springframework.dao.DataIntegrityViolationException;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.security.access.AccessDeniedException;
+import org.springframework.security.core.AuthenticationException;
 import org.springframework.validation.FieldError;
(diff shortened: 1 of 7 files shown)


damgouj (approved):

openbas-api/src/main/java/io/openbas/rest/helper/RestBehavior.java
```diff
@@ -92,7 +93,7 @@ public ValidationErrorBag handleBadRequestExceptions(ImportException ex) {
   }
 
   @ResponseStatus(HttpStatus.UNAUTHORIZED)
-  @ExceptionHandler(AccessDeniedException.class)
+  @ExceptionHandler(AuthenticationException.class)
   public ValidationErrorBag handleValidationExceptions() {
     ValidationErrorBag bag =
         new ValidationErrorBag(HttpStatus.UNAUTHORIZED.value(), "ACCESS_DENIED");
```
guillaumejparis, line 99: Maybe change default message for something more meaningful as UNATHORIZED is described as : "it lacks valid authentication credentials for the requested resource"

guillaumejparis (approved):

PR #1808, opened 2024-11-04
[backend] Allow for use of AWS IAM Role

### LINKED ISSUE #1453: Ability to support IAM roles for Amazon S3 / MinIO configuration
## Use case

Ability to deploy on AWS 

- equivalent to OpenCTI's MINIO__USE_AWS_ROLE
- requirement AWS

Dimfacion:
### Proposed changes

* Added a boolean in the config to use aws iam role
* Added the possibility to use a custom sts endpoint

### Related issues

* Closes #1453 
*

### Checklist

### Further comments

### DIFF
--- a/openbas-api/src/main/resources/application.properties
+++ b/openbas-api/src/main/resources/application.properties
@@ -123,6 +123,9 @@ minio.bucket=openbas
 minio.port=9000
 minio.access-key=<key>
 minio.access-secret=<secret>
+# S3 configuration
+openbas.s3.use_aws_role=false
+openbas.s3.sts_endpoint=<sts_endpoint>
 
 # Logging
 logging.level.root=fatal
--- a/openbas-api/src/main/resources/application.properties
+++ b/openbas-api/src/main/resources/application.properties
@@ -124,8 +124,8 @@ minio.port=9000
 minio.access-key=<key>
 minio.access-secret=<secret>
 # S3 configuration
-openbas.s3.use_aws_role=false
-openbas.s3.sts_endpoint=<sts_endpoint>
+openbas.s3.use-aws-role=false
+openbas.s3.sts-endpoint=<sts_endpoint>
 
 # Logging
 logging.level.root=fatal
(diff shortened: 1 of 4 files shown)


openbas-api/src/main/resources/application.properties
```diff
@@ -123,6 +123,9 @@ minio.bucket=openbas
 minio.port=9000
 minio.access-key=<key>
 minio.access-secret=<secret>
+# S3 configuration
+openbas.s3.use_aws_role=false
+openbas.s3.sts_endpoint=<sts_endpoint>
```
RomuDeuxfois, line 128: **question:**  Wy _ instead of - ? I see that we have - for all the other properties

openbas-model/src/main/java/io/openbas/driver/MinioDriver.java
```diff
@@ -3,27 +3,41 @@
...
-        MinioClient.builder()
-            .endpoint(minioConfig.getEndpoint(), minioConfig.getPort(), minioConfig.isSecure())
-            .credentials(minioConfig.getAccessKey(), minioConfig.getAccessSecret())
-            .build();
+    MinioClient minioClient;
+    if (s3Config.isUseAwsRole()) {
+      String stsEndpoint = null;
+      if (s3Config.getStsEndpoint() != null && !s3Config.getStsEndpoint().isEmpty()) {
+        stsEndpoint = s3Config.getStsEndpoint();
+      }
```
RomuDeuxfois, line 26: **question:** should we throw an exception if there are null on StsEndpoint ?

Dimfacion: No, it's on purpose. StsEndpoint is only if you want to use a custom one (say you implement your own STS endpoint) but if you don't, the library will use either the generic global url of AWS or a regional url of AWS if your AWS_REGION variable is set.

openbas-api/src/main/resources/application.properties
Dimfacion: Good point, I'll change it

RomuDeuxfois (approved):

PR #1825, opened 2024-11-06
[backend/frontend] enforce null consistency with executor and command (#1621)

### LINKED ISSUE #1621: Payload - Enforce required cleanup command AND executor when choosing one of the two
## Description

When I create a payload, I can specify a cleanup executor without a cleanup command.
And vice versa.

## Expected Output

If I specify a cleanup executor or a cleanup command, the second property should be defined too.

antoinemzs:
### Proposed changes

* Enforce having a value for both the Cleanup Executor and the Cleanup Command in Payload definition

### Related issues

* Closes #1621
* Must be merged at the same time: https://github.com/OpenBAS-Platform/collectors/pull/49

### Checklist

### Further comments

On the payload definition panel, it is now not possible to only define the cleanup executor or the cleanup command. Both must be defined at the same time or none of them (i.e. blank fields).

The backend part was done with a "pure" null philosophy, unfortunately the frontend is not yet able to handle null values in react field components: the frontend special cases empty string for fields to submit null instead.

### DIFF
--- a/openbas-api/src/main/java/io/openbas/migration/V3_47__Enforce_payload_command_definition_consistency.java
+++ b/openbas-api/src/main/java/io/openbas/migration/V3_47__Enforce_payload_command_definition_consistency.java
@@ -4,7 +4,6 @@ import java.sql.SQLException;
 import java.sql.Statement;
 import org.flywaydb.core.api.migration.BaseJavaMigration;
 import org.flywaydb.core.api.migration.Context;
-import org.springframework.stereotype.Component;
 
 public class V3_47__Enforce_payload_command_definition_consistency extends BaseJavaMigration {
 
--- a/openbas-api/src/main/java/io/openbas/migration/V3_47__Enforce_payload_command_definition_consistency.java
+++ b/openbas-api/src/main/java/io/openbas/migration/V3_47__Enforce_payload_command_definition_consistency.java
@@ -1,0 +1,38 @@
+package io.openbas.migration;
+
+import org.flywaydb.core.api.migration.BaseJavaMigration;
+import org.flywaydb.core.api.migration.Context;
+
+import java.sql.SQLException;
+import java.sql.Statement;
+
+public class V3_47__Enforce_payload_command_definition_consistency extends BaseJavaMigration {
+
+    @Override
+    public void migrate(Context context) throws SQLException {
+        Statement migrator = context.getConnection().createStatement();
+        migrator.execute("""
+                    UPDATE payloads
+                    SET command_executor = NULL, command_content = NULL
+                    WHERE (command_executor = '') IS NOT FALSE OR (command_content = '') IS NOT FALSE
+                    """);
+        migrator.execute("""
+                    ALTER TABLE payloads
+                    ADD CONSTRAINT chk_payload_cmd_consistency
+                    CHECK ((command_executor IS NULL AND command_content IS NULL)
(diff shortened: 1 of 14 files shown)


openbas-front/src/utils/Localization.js
```diff
@@ -2680,6 +2681,7 @@ const i18n = {
       'Update the security platform': '更新安全平台',
       'Security Platforms': '安全平台',
       'Content should not be empty': '内容不可为空',
+      'Cleanup command and executor must be defined together or none at all': '清理命令和执行器必须一起定义，否则根本无法定义',
```
antoinemzs, line 2684: Deepl translation

antoinemzs:
Issue found: the new constraint prevents importing the atomic red team database

antoinemzs:
> Issue found: the new constraint prevents importing the atomic red team database

fixed

savacano28:
**nitpick:** When I try to update a payload (image 1), I receive errors for all the mandatory fields, but not for the scenario where only one of two fields is filled, so maybe it is possible to add this verification?

image

openbas-api/src/test/java/io/openbas/rest/PayloadApiTest.java
```diff
@@ -158,4 +279,20 @@ private PayloadCreateInput getExecutablePayloadCreateInput() {
     input.setExecutableArch(Endpoint.PLATFORM_ARCH.x86_64);
     return input;
   }
+
+  private PayloadCreateInput getCommandLinePayloadCreateInput() {
+    PayloadCreateInput input = new PayloadCreateInput();
```
savacano28, line 284: **nitpick**: Maybe use a fixture?

openbas-api/src/main/java/io/openbas/migration/V3_47__Enforce_payload_command_definition_consistency.java
```diff
@@ -0,0 +1,43 @@
...
+  @Override
+  public void migrate(Context context) throws SQLException {
+    Statement migrator = context.getConnection().createStatement();
+    migrator.execute(
+        """
+                    UPDATE payloads
+                    SET command_executor = NULL, command_content = NULL
+                    WHERE (command_executor = '') IS NOT FALSE OR (command_content = '') IS NOT FALSE
+                    """);
+    migrator.execute(
```
savacano28, line 21: **praise**: Thanks for adding this constraint

openbas-api/src/test/java/io/openbas/rest/PayloadApiTest.java
antoinemzs: Ah, I see that this is the *Input DTO-type object, not the actual model. This is probably why I haven't used the fixture in the first place, which returns a Payload model (not a PayloadInput DTO). I'll see if I can "fixturise" this.

antoinemzs:
> **Question**: When I create a payload with a type other than 'executable type,' I receive all the expected errors, but for 'executable type,' I don't. Could you please verify why the behavior is different?

It seems that if the `Executable` field is in error, the validation is not run on the remainder of the form fields, e.g. the Cleanup command executor/command tuple.

antoinemzs:
> **nitpick:** When I try to update a payload (image 1), I receive errors for all the mandatory fields, but not for the scenario where only one of two fields is filled, so maybe it is possible to add this verification?

@savacano28 could not reproduce this behaviour locally, unsure whether it's still occurring?

savacano28 (approved):

PR #1826, opened 2024-11-06
[backend] Add matching on parent process name for inject expectation signature

RomuDeuxfois:
Add matching on process parent name to be more accurate when we match prevention & detection expectation.

Linked to https://github.com/OpenBAS-Platform/collectors/pull/48


isselparra (approved):
Test OK

PR #1902, opened 2024-11-21
[backend] Fix NoSuchBeanDefinitionException in retrieving command line

RomuDeuxfois:
https://github.com/OpenBAS-Platform/openbas/issues/1903

### DIFF
--- a/openbas-api/src/main/java/io/openbas/utils/InjectUtils.java
+++ b/openbas-api/src/main/java/io/openbas/utils/InjectUtils.java
@@ -7,16 +7,19 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
 import lombok.RequiredArgsConstructor;
+import lombok.extern.java.Log;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.poi.ss.usermodel.Cell;
 import org.apache.poi.ss.usermodel.CellType;
 import org.apache.poi.ss.usermodel.Row;
 import org.hibernate.Hibernate;
+import org.springframework.beans.factory.NoSuchBeanDefinitionException;
 import org.springframework.context.ApplicationContext;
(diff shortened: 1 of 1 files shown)


damgouj (approved):

openbas-api/src/main/java/io/openbas/utils/InjectUtils.java
```diff
@@ -46,11 +48,15 @@ public InjectStatusCommandLine getCommandsLinesFromInject(final Inject inject) {
...
-            context.getBean(
-                injectorContract.getInjector().getType(), io.openbas.execution.Injector.class);
-        return executor.getCommandsLines(injectorContract.getId());
+        try {
+          // Inject comes from Caldera ability and tomorrow from other(s) Executor(s)
+          io.openbas.execution.Injector executor =
+              context.getBean(
+                  injectorContract.getInjector().getType(), io.openbas.execution.Injector.class);
+          return executor.getCommandsLines(injectorContract.getId());
+        } catch (NoSuchBeanDefinitionException e) {
```
antoinemzs, line 56: issue: You should log here that we are swallowing the exception

antoinemzs (changes requested):

openbas-api/src/main/java/io/openbas/utils/InjectUtils.java
RomuDeuxfois: True, I add some details.
Is this consistent with the idea you had?

antoinemzs: I'd probably escalate to Warning I suppose, it looks like a misconfiguration if the requested class is not available for loading.
Otherwise yes it's what I am referring to: no silently dropping an exception, so that we at least know from the logs that something is happening

PR #1907, opened 2024-11-22
[backend] Fix status traces comparisons

savacano28:
### Proposed changes

* Fix cases of comparison between InjectStatus vs IjectTracestatus

### Related issues

* Closes #ISSUE-NUMBER
*

### Checklist

### Further comments

### DIFF
--- a/openbas-api/src/main/java/io/openbas/injectors/caldera/service/CalderaInjectorService.java
+++ b/openbas-api/src/main/java/io/openbas/injectors/caldera/service/CalderaInjectorService.java
@@ -95,8 +95,7 @@ public class CalderaInjectorService {
                       platform.equals(PLATFORM_TYPE.MacOS)
                           ? "darwin"
                           : platform.name().toLowerCase());
-                  executor.put(
-                      "name", platform.equals(PLATFORM_TYPE.Windows.name()) ? "psh" : "sh");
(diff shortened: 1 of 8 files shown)


openbas-api/src/main/java/io/openbas/injectors/caldera/service/CalderaInjectorService.java
```diff
@@ -95,8 +95,7 @@ public Ability createAbility(Payload payload) {
                       platform.equals(PLATFORM_TYPE.MacOS)
                           ? "darwin"
                           : platform.name().toLowerCase());
-                  executor.put(
-                      "name", platform.equals(PLATFORM_TYPE.Windows.name()) ? "psh" : "sh");
+                  executor.put("name", PLATFORM_TYPE.Windows.equals(platform) ? "psh" : "sh");
```
antoinemzs, line 98: Question: why isn't the comparison style consistent with, e.g. the case above for MacOS? It looks like we could be open to a NPE in case platform ends up null for whatever reason; in which case the style you used is more robust (using a constant for the basis of comparison).

Probably more of a nitpick since `platform` being null would cause problems earlier than this part of the code.

antoinemzs (approved):
Good move to use constants as the basis for string comparison, evading a potential NPE

PR #1950, opened 2024-11-28
[backend] Adding a new "health" endpoint (#1850)

### LINKED ISSUE #1850: Introduce /health endpoint
## Use case

Introduce /health endpoint: for monitoring & troubleshooting, we need an API endpoint for health check.

This endpoint must check connectivity with all backends (postgresql, rabbitmq...), and give a clear response of the instance status (http 200: all good, 5xx: error)
It must be protected by an access key (new configuration option, defaults to changeme)

Example http://openbas/health?health_access_key=test

## Additional Information

You may have a look to the implementation in OpenCTI

heditar:
### Proposed changes

* Adding a new "health" endpoint that will return 200 if the service is healthy otherwise 5xx

### Related issues

* Closes #1850 

### Checklist

- [X] I added/update the relevant documentation (either on github or on notion)


openbas-api/src/main/java/io/openbas/rest/health_check/HealthCheckApi.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  @GetMapping(HEALTH_CHECK_URI)
+  @Operation(
+      summary = "Run an healthcheck ",
+      description = "Tries to connect to dependencies (DB/Minio/RabbitMQ)")
+  @ApiResponses(
+      value = {
+        @ApiResponse(responseCode = "200", description = "Service is healthy"),
+        @ApiResponse(responseCode = "503", description = "Service is not running properly")
```
antoinemzs, line 47: I see you make use of constants further in the code, would you consider using constants here too?

antoinemzs: nitpick: Additionally, would you consider tying the documentation to the actual code with an in-class constant? i.e.
```
private static final HttpStatus ALL_GOOD_STATUS = HttpStatus.OK;
private static final HttpStatus BROKEN_STATUS = HttpStatus.SERVICE_UNAVAILABLE;
...
@ApiResponse(responseCode = ALL_GOOD_STATUS.value(), description = "Service is healthy"),
@ApiResponse(responseCode = BROKEN_STATUS.value(), description = "Service is not running properly")
...
    throw new ResponseStatusException(
          HttpStatusCode.valueOf(BROKEN_STATUS.value()), message);
    }
    return new ResponseEntity<>("success", ALL_GOOD_STATUS);
```

openbas-api/src/main/java/io/openbas/rest/health_check/HealthCheckApi.java
```diff
@@ -0,0 +1,65 @@
...
+  public ResponseEntity<?> healthCheck(
+      @RequestParam("health_access_key") String requestHealthAccessKey) {
+    if (StringUtils.isEmpty(requestHealthAccessKey)
+        || StringUtils.isEmpty(healthCheckKey)
+        || !healthCheckKey.equals(requestHealthAccessKey)) {
+      throw new ResponseStatusException(HttpStatusCode.valueOf(HttpStatus.UNAUTHORIZED.value()));
+    }
+    try {
+      healthCheckService.runHealthCheck();
+    } catch (HealthCheckFailureException e) {
```
antoinemzs, line 58: Question: any other exception would result in a code 500 (not 503), will Platform handle this gracefully? i.e. have you checked with them that they are looking for a status code range (`>= 500`) and not an exact status code (`==503`).

openbas-api/src/main/java/io/openbas/rest/health_check/HealthCheckApi.java
```diff
@@ -0,0 +1,65 @@
...
+      description = "Tries to connect to dependencies (DB/Minio/RabbitMQ)")
+  @ApiResponses(
+      value = {
+        @ApiResponse(responseCode = "200", description = "Service is healthy"),
+        @ApiResponse(responseCode = "503", description = "Service is not running properly")
+      })
+  public ResponseEntity<?> healthCheck(
+      @RequestParam("health_access_key") String requestHealthAccessKey) {
+    if (StringUtils.isEmpty(requestHealthAccessKey)
+        || StringUtils.isEmpty(healthCheckKey)
```
antoinemzs, line 52: praise: good catch to block access if not configured

openbas-api/src/main/java/io/openbas/rest/health_check/HealthCheckApi.java
```diff
@@ -0,0 +1,65 @@
...
+      summary = "Run an healthcheck ",
+      description = "Tries to connect to dependencies (DB/Minio/RabbitMQ)")
+  @ApiResponses(
+      value = {
+        @ApiResponse(responseCode = "200", description = "Service is healthy"),
+        @ApiResponse(responseCode = "503", description = "Service is not running properly")
+      })
+  public ResponseEntity<?> healthCheck(
+      @RequestParam("health_access_key") String requestHealthAccessKey) {
+    if (StringUtils.isEmpty(requestHealthAccessKey)
```
antoinemzs, line 51: nitpick: `StringUtils.isEmpty` returns false even when string only consists of whitespace. Consider using `isBlank` to avoid an accidental misconfiguration with spaces (e.g. typo in config) https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#isEmpty(java.lang.CharSequence)

openbas-api/src/main/java/io/openbas/service/HealthCheckService.java
```diff
@@ -0,0 +1,97 @@
...
+
+  @VisibleForTesting
+  protected void runFileStorageCheck() throws HealthCheckFailureException {
+    try {
+      // we get a new client instance to avoid to update the client injected by Spring
+      MinioClient minioClient = minioDriver.getMinioClient();
+      minioClient.setTimeout(2000L, 2000L, 2000L);
+      minioClient.bucketExists(BucketExistsArgs.builder().bucket(minioConfig.getBucket()).build());
+    } catch (Exception e) {
+      throw new HealthCheckFailureException("FileStorage check failure", e);
```
antoinemzs, line 94: Question: why do you consider a blanket handling of any exception here but not for e.g. rabbitMQ?

openbas-api/src/test/java/io/openbas/rest/HealthCheckApiTest.java
```diff
@@ -0,0 +1,69 @@
...
+    ResponseStatusException exceptionThrown =
+        assertThrows(
+            ResponseStatusException.class,
+            () -> {
+              healthCheckApi.healthCheck(KEY);
+            });
+    assertEquals(
+        HttpStatusCode.valueOf(HttpStatus.SERVICE_UNAVAILABLE.value()),
+        exceptionThrown.getStatusCode());
+  }
```
antoinemzs, line 68: nitpick: There may be scenarios when the exception raised up to the rest endpoint handler is not of type HealthCheckFailureException; would you consider covering this?

openbas-model/src/main/java/io/openbas/database/repository/HealthCheckRepository.java
```diff
@@ -0,0 +1,13 @@
...
+import io.openbas.database.model.User;
+import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.stereotype.Repository;
+
+/** Repository used to run database health check */
+@Repository
+public interface HealthCheckRepository extends JpaRepository<User, String> {
+  @Query("select 1")
+  void healthCheck();
```
antoinemzs, line 12: Simple, effective :)

antoinemzs (approved):
A couple nitpicks but solid work overall, thanks!

openbas-api/src/main/java/io/openbas/rest/health_check/HealthCheckApi.java
heditar: yes that's better, I will update it in the next commit

openbas-api/src/main/java/io/openbas/service/HealthCheckService.java
heditar: Thanks for catching that, that was done because the  minio client before I split the method was throwing an Exception, which is not the case in this last version. That will be update in the next commit.

openbas-api/src/main/java/io/openbas/rest/health_check/HealthCheckApi.java
heditar: Yes, that makes sense I will update it

heditar: I would need to add a constant string so it doesn't make too much sense or there is another way but I didn't spent too much time thinking about that

openbas-api/src/test/java/io/openbas/rest/HealthCheckApiTest.java
heditar: added in the next commit

antoinemzs (approved):

antoinemzs (approved):

PR #1967, opened 2024-12-02
[backend] Overriding spring's postgresql dependency

### LINKED ISSUE #1771: [security] Upgrade of postgresql JDBC driver
## Description

Upgrade of postgresql JDBC driver to 42.6.1

Dimfacion:
### Proposed changes

* Overriding spring's postgresql dependency to prevent a critical CVE
*

### Related issues

* Closes #1771 
*

### Checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,6 +64,12 @@
             <artifactId>lombok</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.postgresql</groupId>
+            <artifactId>postgresql</artifactId>
+            <scope>runtime</scope>
+            <version>42.6.2</version>
+        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -64,6 +64,12 @@
             <artifactId>lombok</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.postgresql</groupId>
+            <artifactId>postgresql</artifactId>
+            <scope>runtime</scope>
+            <version>42.6.2</version>
+        </dependency>
```
RomuDeuxfois, line 72: **question**

Why moving this from the model package ?
Postgres is direclty used by the repository class.

Dimfacion: An older version of Postgres is pulled by openbas-api and openbas-framework because of other dependencies. Setting it in the root pom.xml will make sure we override the older version we don't want.

RomuDeuxfois: Ok understood
Thanks for the explanation.

RomuDeuxfois (approved):

PR #1978, opened 2024-12-03
[tool] update drone & circle for prerelease platform (#1909)

### LINKED ISSUE #1909: Handle prerelease platform on our CI
# Description

We aim to follow the new OpenCTI branching model and require a pre-release platform to do so.

We need to improv our CI to handle the deployment on prerelease platform on ALL repositories.

Ressources: 
- https://github.com/OpenCTI-Platform/opencti/blob/402d6bca3a988246a5da299e07957ee999c3ac60/.circleci/config.yml#L460

# Improvment

Think about a script to automatically create release branch on ALL repositories.

guillaumejparis:
### Proposed changes

* Add build & deploy for prerelease when merging on release/*
* Add a way to fetch latests/prelease agents/implants when on testing/prerelease

### Related issues

* Closes #1909 
*

### Checklist

### Further comments

### DIFF
--- a/openbas-api/src/main/resources/application.properties
+++ b/openbas-api/src/main/resources/application.properties
@@ -254,6 +254,3 @@ openbas.expectation.human.expiration-time=86400
 # Max value: 100
 # Default value: 50
 openbas.expectation.manual.default-score-value=50
-
-#
-openbas.
--- a/openbas-api/src/main/resources/application.properties
+++ b/openbas-api/src/main/resources/application.properties
@@ -158,7 +158,15 @@ executor.tanium.windows-package-id=
 executor.tanium.unix-package-id=
 
 # Executor OpenBAS
-executor.openbas.version=local
+
+# valid values: local | repository
+# default: local
+executor.openbas.binaries.origin=
+
+# if executor.openbas.agent.binaries.origin is set to "local",
+# this config is ignored
+# default: the OpenBAS instance's version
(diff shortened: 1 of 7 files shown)


openbas-api/src/main/resources/application.properties
```diff
@@ -157,6 +157,9 @@ executor.tanium.action-group-id=4
 executor.tanium.windows-package-id=
 executor.tanium.unix-package-id=
 
+# Executor OpenBAS
+executor.openbas.version=local
+
```
antoinemzs, line 162: I feel this config's current name can be confusing; I'd alter its use further though.

As we discussed, the aim of this config is to allow bypassing the locally contained agent binaries and have OpenBAS serve a binary directly from Artifactory given a specific version.

I'd suggest changing the naming semantics:
```suggestion
# Executor OpenBAS

# valid values: local | repository
# default: local
executor.openbas.agent.binaries.origin=

# if executor.openbas.agent.binaries.origin is set to "local",
# this config is ignored
# default: the OpenBAS instance's version
executor.openbas.agent.binaries.version=

```
Leaving them both blank, hence the default behaviour should just be locally hosted binaries.

I know it's two configs rather than a single one, but I'm uncomfortable with a single config that triggers multiple behaviours (here that would be changing the origin, and the version of the binaries).

antoinemzs (approved):

PR #1999, opened 2024-12-05
Improv swagger documentation

RomuDeuxfois:
### Proposed changes

Generating the swagger doc takes a lot of time due to the object nesting that we have in our back-end.
However, these object nestings do not reflect reality, since we use Deserializers to return the IDs and not the entire object.

Example:
```
public class Exercise implements Base {
...
  @JsonSerialize(using = MonoIdDeserializer.class)
  @JsonProperty("exercise_scenario")
  private Scenario scenario;
...
}
```
The api get exercise_scenario as a string, but the swagger take the full bean.
The idea is to use the swagger api annotations to reflect reality (based on @MarineLeM work & troubleshooting by @savacano28)

```
public class Exercise implements Base {
...
  @JsonSerialize(using = MonoIdDeserializer.class)
  @JsonProperty("exercise_scenario")
  @Schema(type = "string")
  private Scenario scenario;
...
}
```

### Next steps

The next step is to remove the Model**Store** types, which made it possible to reflect the reality of the back side on the front side, in favor of using the types in the api-types file which are now clean.

```
export type ArticleStore = Omit<Article, 'article_channel' | 'article_documents'> & {
  article_channel: string | undefined;
  article_documents: string[] | undefined;
};
```

### DIFF
--- a/openbas-model/src/main/java/io/openbas/database/model/Group.java
+++ b/openbas-model/src/main/java/io/openbas/database/model/Group.java
@@ -6,6 +6,8 @@ import io.openbas.annotation.Queryable;
 import io.openbas.database.audit.ModelBaseListener;
 import io.openbas.helper.MultiIdListDeserializer;
 import io.openbas.helper.MultiModelDeserializer;
+import io.swagger.v3.oas.annotations.media.ArraySchema;
+import io.swagger.v3.oas.annotations.media.Schema;
 import jakarta.persistence.*;
 import jakarta.validation.constraints.NotBlank;
 import java.util.ArrayList;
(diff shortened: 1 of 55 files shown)


openbas-model/src/main/java/io/openbas/database/model/Group.java
```diff
@@ -69,6 +71,7 @@ public class Group implements Base {
   @Fetch(value = FetchMode.SUBSELECT)
   private List<Grant> grants = new ArrayList<>();
 
+  @ArraySchema(schema = @Schema(type = "string"))
```
MarineLeM, line 74: you should remove from it because FetchType is EAGER

RomuDeuxfois: I am not sure to understand.
FetchType is EAGER but we still returning only IDs

MarineLeM (approved):

PR #2001, opened 2024-12-05
[backend] handle cmd variables

MarineLeM:
### Proposed changes

Fix this problem : 

Instantiating a `cmd `variable and using it in the same line doesn't work as expected. 
For example, the command 
```
set x=test & echo %x%
``` 
won't display the value of x because the variable hasn't been instantiated yet. To fix this, you can use the command 
```
cmd /v /c "set x=test & echo !x!"
``` 
 which enables delayed variable expansion, allowing the variable to be properly referenced
 
 
 In this PR i replace the %test% by !test!
 In this PR https://github.com/OpenBAS-Platform/implant/pull/26 i add the  /v arguments

### Related issues

* Related to this issue https://github.com/OpenBAS-Platform/openbas/issues/1604

### DIFF
--- a/openbas-api/src/main/java/io/openbas/rest/inject/service/ExecutableInjectService.java
+++ b/openbas-api/src/main/java/io/openbas/rest/inject/service/ExecutableInjectService.java
@@ -18,6 +18,7 @@ public class ExecutableInjectService {
 
   private final InjectRepository injectRepository;
   private static final Pattern argumentsRegex = Pattern.compile("#\\{([^#{}]+)}");
+  private static final Pattern cmdVariablesRegex = Pattern.compile("%(\\w+)%");
 
   private List<String> getArgumentsFromCommandLines(String command) {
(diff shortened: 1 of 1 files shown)


openbas-api/src/main/java/io/openbas/rest/inject/service/ExecutableInjectService.java
```diff
@@ -88,7 +103,8 @@ public Payload getExecutablePayloadInject(String injectId) {
         .getPrerequisites()
         .forEach(
             prerequisite -> {
-              if (prerequisite.getCheckCommand() != null) {
+              if (prerequisite.getCheckCommand() != null
+                  && !prerequisite.getCheckCommand().isEmpty()) {
```
damgouj, line 107: Maybe put blank instead of empty ?

damgouj (changes requested):

damgouj (approved):

Issue #2000, opened 2024-12-05
Technical payload with cmd and if statement does not work


MarineLeM:
## Description

The command with if statement for cmd does not work

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Create a payload for windows
2. add a command with a if statement on multiline, the executor must be `cmd`
for example :
```
if exist "C:\Users\Tom\Pictures\img.png" (
    echo File exists
) else (
    echo File does not exist
)
``` 
3. Create an inject with this payload and launch it

## Expected Output
The command works

## Actual Output
The command does not work

## Additional information
The problem come from how we handle multilines for cmd

PR #2002, opened 2024-12-05
[backend] handle multilines with cmd

MarineLeM:
Close: #2000


damgouj (changes requested):
Tested with complicated loops and not working, maybe we could try to find a generic way to do like powershell (if it exists for cmd !) ? Thanks !
Capture d’écran 2024-12-06 085942
Capture d’écran 2024-12-06 085649

MarineLeM:
@damgouj ready for a second round, it should now handle complicated loops

PR #2003, opened 2024-12-06
[backend] Apply the right platform & architecture when generating a scenario from OCTI (#1713)

### LINKED ISSUE #1713: Apply the right platform & architecture when generating a scenario from OCTI
## Use case

- When I create a simulation based on a report that was targeting windows, I want my simulation to only have injects for windows
- When I have a TTP that is not covered for a platform I want to be aware that we couldn't leverage all phases 
- When I have more that one platform in my report I want my simulation simulate both by default 
- When I have an architecture specified in my report I want only injects targrting this architecture to be populated 

## Current Workaround

Rework manually my list of injects 

## Proposed Solution
OpenCTI Side:
- Add in the form two new selectfields: platforms (multiple) & architecture (single)
- Can be optional

OpenBAS Side:
- Filter the injects by platform & architecture based on OpenCTI form
- Have a way to alert that some ttp were not covered

savacano28:
### Proposed changes

* Add non-regression tests for endpoints used by the Octi feature
* Sync values for architectures

### Related issues

* Closes #1713 
* Octi [#8820](https://github.com/OpenCTI-Platform/opencti/issues/8820)

### Checklist

### Further comments


openbas-api/src/main/java/io/openbas/migration/V3_52__Sync_archi_names.java
```diff
@@ -0,0 +1,21 @@
...
+
+@Component
+public class V3_52__Sync_archi_names extends BaseJavaMigration {
+
+  @Override
+  public void migrate(Context context) throws Exception {
+    Connection connection = context.getConnection();
+    Statement statement = connection.createStatement();
+    statement.execute(
+        "UPDATE payloads SET payload_execution_arch = 'x86_64' WHERE payload_execution_arch = 'X86_64';");
```
RomuDeuxfois, line 17: **question:**

Why we have this changes after this migration ?
V3_51__Add_architecture_to_payload

openbas-api/src/main/java/io/openbas/utils/ArchitectureFilterUtils.java
```diff
@@ -31,8 +31,8 @@ public static SearchPaginationInput handleArchitectureFilter(
 
     filterOpt.ifPresent(
         payloadFilter -> {
-          if (payloadFilter.getValues().contains("X86_64")
-              || payloadFilter.getValues().contains("ARM64")) {
+          if (payloadFilter.getValues().contains("x86_64")
```
RomuDeuxfois, line 34: **todo:**
You should use the enum

openbas-api/src/test/java/io/openbas/rest/PayloadApiSearchTest.java
```diff
@@ -204,7 +204,7 @@ void given_filter_input_by_source_should_return_a_page_of_payloads_filter_by_sou
                     .contentType(MediaType.APPLICATION_JSON)
                     .content(asJsonString(searchPaginationInput)))
             .andExpect(status().is2xxSuccessful())
-            .andExpect(jsonPath("$.numberOfElements").value(1));
+            .andExpect(jsonPath("$.numberOfElements").value(3));
```
RomuDeuxfois, line 207: **question:**

Why this changes ?

openbas-api/src/test/java/io/openbas/rest/PayloadApiTest.java
```diff
@@ -70,7 +70,7 @@ void createExecutablePayload() throws Exception {
         .andExpect(jsonPath("$.payload_source").value("MANUAL"))
         .andExpect(jsonPath("$.payload_status").value("VERIFIED"))
         .andExpect(jsonPath("$.payload_platforms.[0]").value("Linux"))
-        .andExpect(jsonPath("$.payload_execution_arch").value("X86_64"));
+        .andExpect(jsonPath("$.payload_execution_arch").value("x86_64"));
```
RomuDeuxfois, line 73: **Todo:**

Use enum here

openbas-api/src/test/java/io/openbas/rest/PayloadApiTest.java
```diff
@@ -119,7 +119,7 @@ void updateExecutablePayload() throws Exception {
             .andExpect(status().is2xxSuccessful())
             .andExpect(jsonPath("$.payload_name").value("My Executable Payload"))
             .andExpect(jsonPath("$.payload_platforms.[0]").value("Linux"))
-            .andExpect(jsonPath("$.payload_execution_arch").value("X86_64"))
+            .andExpect(jsonPath("$.payload_execution_arch").value("x86_64"))
```
RomuDeuxfois, line 122: **Todo:**

Use enum here

openbas-api/src/test/java/io/openbas/rest/PayloadApiTest.java
```diff
@@ -136,7 +136,7 @@ void updateExecutablePayload() throws Exception {
         .andExpect(status().is2xxSuccessful())
         .andExpect(jsonPath("$.payload_name").value("My Updated Executable Payload"))
         .andExpect(jsonPath("$.payload_platforms.[0]").value("MacOS"))
-        .andExpect(jsonPath("$.payload_execution_arch").value("ARM64"));
+        .andExpect(jsonPath("$.payload_execution_arch").value("arm64"));
```
RomuDeuxfois, line 139: **Todo:**

Use enum here

RomuDeuxfois:
Tested 🆗

openbas-api/src/test/java/io/openbas/rest/PayloadApiSearchTest.java
savacano28: Because, in the beforeAll(), we create 3 payloads: dns(all_arch), command(all_arc), and executable(arm64). The all_arch payload is considered to include both arm64 and x86_64. So, when we search for a payload with this architecture, we now get arm64 along with all_arch.

openbas-api/src/main/java/io/openbas/migration/V3_52__Sync_archi_names.java
savacano28: Because, in the previous migration, I set all values in uppercase (enum), but after reflection, it’s better to keep the values as they were before (such as x86_64, arm64), rather than introducing new values like AMD64 or ARM64.

RomuDeuxfois (approved):

PR #2018, opened 2024-12-09
[backend] Fix import/export with unknown contract ID

RomuDeuxfois:
### Related issues

* https://github.com/OpenBAS-Platform/openbas/issues/1992


antoinemzs (approved):
Thanks for fixing this and for the last minute changes; lgtm

PR #2040, opened 2024-12-12
[frontend] add button to navigate back to scenario from simulation (#1611)

### LINKED ISSUE #1611: Be able to access to the original scenario from simulation (going to parent)
## Use case

Be able to access to the original scenario from simulation (going to parent)

antoinemzs:
### Proposed changes

* Adding a way to navigate back to the parent scenario from a simulation
*

### Related issues

* Closes #1611
*

### Checklist

### Further comments
image
image
image
image


openbas-front/src/admin/components/simulations/simulation/Index.tsx
```diff
@@ -173,6 +174,9 @@ const Index = () => {
       setLoading(false);
     });
   });
+  useDataLoader(() => {
+    dispatch(fetchScenario(exercise.exercise_scenario));
```
savacano28, line 178: Do we really need to fetch the entire scenario just to display a name? Why not have a button that says 'Go to Parent Scenario' instead?

openbas-front/src/admin/components/simulations/simulation/ExerciseMainInformation.tsx
```diff
@@ -18,11 +23,18 @@ interface Props {
 const ExerciseMainInformation: React.FC<Props> = ({ exercise }) => {
   const { t } = useFormatter();
   const sortByOrder = R.sortWith([R.ascend(R.prop('phase_order'))]);
+  const scenarioBaseUri = '/admin/scenarios';
+  const scenarioId: string = exercise.exercise_scenario || '';
+  const { scenario } = scenarioId
```
savacano28, line 28: If I remove the parent scenario, I get this error on exercise side:
image
image

savacano28 (changes requested):

openbas-front/src/admin/components/simulations/simulation/ExerciseMainInformation.tsx
antoinemzs: 💀

openbas-front/src/admin/components/simulations/simulation/Index.tsx
antoinemzs: I've checked with product re: UX and design and the outcome was that it was an ok spot with displaying the scenario name. Fetching a single scenario is probably fine in this instance (i.e. it's not a collection).

openbas-front/src/admin/components/simulations/simulation/ExerciseMainInformation.tsx
antoinemzs: Ok I had made a mistake in null checking, I fixed it in the latest commit

savacano28:
Tested OK

PR #2065, opened 2024-12-13
[frontend] enable navigating to scenario with no extra load (#1611)

### LINKED ISSUE #1611: Be able to access to the original scenario from simulation (going to parent)

antoinemzs:
### Proposed changes

* Add a button back to parent scenario when available

### Related issues

* Closes #1611

### Checklist

### Further comments

Second take on a change for #1611 without loading an extra resource (scenario), with a generic wording; born of comment https://github.com/OpenBAS-Platform/openbas/pull/2040#discussion_r1883478393
The display is more stable this way since we are not subjected to varying string lengths to display in a button.

With available scenario:
image

Without available scenario (e.g. scenario was deleted):
image


savacano28:
I think they have different sizes ? (Back To Parent is bigger than Pause)
image

savacano28 (approved):

antoinemzs:
Thanks ! Product went with the other version eventually. We'll revisit this if it needs further improvement !

PR #2089, opened 2024-12-17
[tests] Explicitly mock Instant.now to fix test flakiness

antoinemzs:
### Proposed changes

* Mock `Instant.now` and `Clock.systemDefaultZone` to get predictable times when production code calls `Instant.now()` and `LocalDateTime.now()`

### Checklist

### Further comments
I only changed the tests that depended on waiting for some time and fetch back the injects (i.e. time based tests); other tests don't seem to be affected by passing time.

### DIFF
--- a/openbas-api/src/test/java/io/openbas/rest/exercise/ExerciseApiStatusTest.java
+++ b/openbas-api/src/test/java/io/openbas/rest/exercise/ExerciseApiStatusTest.java
@@ -214,8 +214,9 @@ public class ExerciseApiStatusTest {
         mockedInstant.when(Instant::now).thenReturn(instantAfterSetRun);
 
         List<ExecutableInject> injects = injectHelper.getInjectsToRun();
-        Instant nextMinute = instantBeforeSetRun.truncatedTo(MINUTES).plus(1, MINUTES);
-        assertEquals(nextMinute.toString(), JsonPath.read(response, "$.exercise_start_date"));
+        Instant expectedStartTime = instantBeforeSetRun.truncatedTo(MINUTES).plus(1, MINUTES);
+        assertEquals(
+            expectedStartTime.toString(), JsonPath.read(response, "$.exercise_start_date"));
         assertEquals(
             Arrays.asList(ExerciseStatus.CANCELED.name(), ExerciseStatus.PAUSED.name()),
             JsonPath.read(response, "$.exercise_next_possible_status"));
--- a/openbas-api/src/test/java/io/openbas/rest/exercise/ExerciseApiStatusTest.java
+++ b/openbas-api/src/test/java/io/openbas/rest/exercise/ExerciseApiStatusTest.java
@@ -9,6 +9,8 @@ import static java.time.temporal.ChronoUnit.MINUTES;
 import static org.junit.jupiter.api.Assertions.*;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.TestInstance.Lifecycle.PER_CLASS;
+import static org.mockito.Mockito.CALLS_REAL_METHODS;
+import static org.mockito.Mockito.mockStatic;
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.put;
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
 
@@ -24,11 +26,14 @@ import io.openbas.utils.fixtures.*;
 import io.openbas.utils.mockUser.WithMockAdminUser;
 import jakarta.annotation.Resource;
 import jakarta.servlet.ServletException;
+import java.time.Clock;
 import java.time.Instant;
+import java.time.ZoneId;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
 import org.junit.jupiter.api.*;
+import org.mockito.MockedStatic;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc;
 import org.springframework.boot.test.context.SpringBootTest;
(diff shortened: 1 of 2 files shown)


openbas-api/src/test/java/io/openbas/rest/exercise/ExerciseApiStatusTest.java
```diff
@@ -176,28 +184,49 @@ void manualStartExerciseTest() throws Exception {
...
+    Instant instantAfterSetRun = REFERENCE_TIME.plus(1, MINUTES);
+    Clock clock = Clock.fixed(REFERENCE_TIME, ZoneId.of("UTC"));
+
+    try (MockedStatic<Instant> mockedInstant = mockStatic(Instant.class, CALLS_REAL_METHODS)) {
+      try (MockedStatic<Clock> mockedClock = mockStatic(Clock.class, CALLS_REAL_METHODS)) {
+
+        mockedInstant.when(Instant::now).thenReturn(instantBeforeSetRun);
+        // we need this other mock because the production code
+        // inconsistently calls Instant.now() and LocalDateTime.now()
+        mockedClock.when(Clock::systemDefaultZone).thenReturn(clock);
```
impolitepanda, line 197: question: isn't there a risk of timezone mismatch if we mock a systemDefaultZone that isn't UTC by the UTC clock defined l189?

openbas-api/src/test/java/io/openbas/rest/exercise/ExerciseApiStatusTest.java
```diff
@@ -176,28 +184,49 @@ void manualStartExerciseTest() throws Exception {
...
-    Thread.sleep(2000);
-    List<ExecutableInject> injects = injectHelper.getInjectsToRun();
-    Instant nextMinute = now().truncatedTo(MINUTES).plus(1, MINUTES);
-    assertEquals(nextMinute.toString(), JsonPath.read(response, "$.exercise_start_date"));
-    assertEquals(
-        Arrays.asList(ExerciseStatus.CANCELED.name(), ExerciseStatus.PAUSED.name()),
-        JsonPath.read(response, "$.exercise_next_possible_status"));
-    assertEquals(1, injects.size());
+    Instant instantBeforeSetRun = REFERENCE_TIME;
+    Instant instantAfterSetRun = REFERENCE_TIME.plus(1, MINUTES);
```
impolitepanda, line 188: question: what's the average time the test takes to run ? Will 1 minute always be enough ?

openbas-api/src/test/java/io/openbas/rest/exercise/ExerciseApiStatusTest.java
```diff
@@ -176,28 +184,49 @@ void manualStartExerciseTest() throws Exception {
...
+                .getResponse()
+                .getContentAsString();
+
+        // -- ASSERT --
+
+        // NOTE: we are changing the time of Instant.now() here to fast forward in the future
+        mockedInstant.when(Instant::now).thenReturn(instantAfterSetRun);
+
+        List<ExecutableInject> injects = injectHelper.getInjectsToRun();
+        Instant nextMinute = instantBeforeSetRun.truncatedTo(MINUTES).plus(1, MINUTES);
```
impolitepanda, line 217: question: could we use `instantAfterSetRun.truncatedTo(MINUTES)` instead of adding 1 minute again ?

openbas-api/src/test/java/io/openbas/rest/exercise/ExerciseApiStatusTest.java
```diff
@@ -263,37 +292,52 @@ void runExerciseAfterPauseTest() throws Exception {
...
+                        .contentType(MediaType.APPLICATION_JSON)
+                        .accept(MediaType.APPLICATION_JSON))
+                .andExpect(status().is2xxSuccessful())
+                .andReturn()
+                .getResponse()
+                .getContentAsString();
+
+        // --ASSERT--
+        // NOTE: we are changing the time of Instant.now() here to fast forward in the future
+        mockedInstant.when(Instant::now).thenReturn(instantAfterSetRun);
```
impolitepanda, line 321: question: I don't see any assertion requiring a time comparison. I take it that it's still necessary to mock it somehow, but is there a way to create an assertion around it ?

openbas-api/src/test/java/io/openbas/rest/exercise/ExerciseApiStatusTest.java
antoinemzs: All mocked times are specifically UTC

openbas-api/src/test/java/io/openbas/rest/exercise/ExerciseApiStatusTest.java
antoinemzs: It's not a duration, they are two different points in time to simulate a real-world scenario when we check if an inject is ready to run: "Now" must be later or equal to "startTime". Here I'm forcing it to be slightly later.

impolitepanda (approved):

openbas-api/src/test/java/io/openbas/rest/exercise/ExerciseApiStatusTest.java
antoinemzs: This check is to assert that when we set the RUNNING status at, say, 10:30:45 (in this instance, the variable `instantBeforeSetRun`), the inject has been set to run exactly at 10:31:00 (so the next minute at 0 seconds). This is because the engine currently behaves this way, injects are set to RUNNING every minute on the minute.

As discussed we agreed the code is correct.

openbas-api/src/test/java/io/openbas/rest/exercise/ExerciseApiStatusTest.java
antoinemzs: We don't sepcifically assert on times except the expected `exercise_start_date` as discussed here https://github.com/OpenBAS-Platform/openbas/pull/2089#discussion_r1888712580

PR #2127, opened 2024-12-26
[frontend] Change "Add teams..." dialog into "Modify teams..." and alter behaviour (#2112)

### LINKED ISSUE #2112: Modifying targets in an inject is not working properly
## Description

Two issues were detected:
A. When I am in the animation tab and I want to modify the teams of my inject, I do not have any teams available although I do have them in the definition tab.
B. When I am in the injects tab and I want to modify the teams of an inject, I can only add them and not delete them via the popup.

## Reproducible Steps

Steps to create the smallest reproducible sim:

A. 
1. Go to a current or finished simulation and check that in the definition tab there are teams (if not add some)
2. Go to the animation tab
3. Modify an inject to add a team
4. I have nothing in the list of the popup that appears -> I should have the list of teams that are present in the definition tab

B.
1. Go to a current or finished simulation and check that in the definition tab there are teams (if not add some)
2. Go to the inject tab
3. Modify an inject to remove a team (make sure before this inject already contains teams)
4. You can delete the team via the small cross on its name to the right of the popup. But when we click on the add button, the team remains present.

Depending on what you want: you should either make it possible to delete a team when you click on the add button and change the name of this button to update.
Either, do not make it possible to have the cross on the name of the team on the list on the right.

RomuDeuxfois:
### Proposed changes

* "Add teams..." controls in Inject definitions is changed to "Modify teams"
* The Modify teams dialog allows for adding or removing teams to/from inject
* Allowed team list is scoped down to the teams added to a scenario/simulation accordingly
  *  If no teams were added to a scenario or simulation, no teams are available to add to injects

### Related issues

* Fix #2112 

### Checklist

### Further comments

Based on @antoinemzs work


antoinemzs:
The scope looks to have increased from the issue's description, just to be clear, this solves the issue and introduces a similar behaviour for Target Assets ?

RomuDeuxfois:
> The scope looks to have increased from the issue's description, just to be clear, this solves the issue and introduces a similar behaviour for Target Assets ?

Exactly, I realized that we also had to align the behavior with assets and asset groups in order to have something consistent for the user.

damgouj (approved):

PR #2137, opened 2024-12-30
[backend] Fix Regular expression injection

RomuDeuxfois:
### Related issues

* Closes https://github.com/OpenBAS-Platform/openbas/issues/1783

### DIFF
--- a/openbas-api/src/main/java/io/openbas/rest/helper/RestBehavior.java
+++ b/openbas-api/src/main/java/io/openbas/rest/helper/RestBehavior.java
@@ -207,11 +207,11 @@ public class RestBehavior {
     }
   }
 
-  protected void validateUUID(final String id) {
+  protected void validateUUID(final String id) throws InputValidationException {
     try {
       UUID.fromString(id);
     } catch (IllegalArgumentException e) {
-      throw new IllegalArgumentException("ID invalid : " + id, e);
+      throw new InputValidationException("id", "The ID is not a valid UUID: " + id);
     }
   }
 }
--- a/openbas-api/src/main/java/io/openbas/rest/helper/RestBehavior.java
+++ b/openbas-api/src/main/java/io/openbas/rest/helper/RestBehavior.java
@@ -206,4 +206,12 @@ public class RestBehavior {
       }
     }
   }
+
+  protected void validateUUID(final String id) {
+    try {
+      UUID.fromString(id);
+    } catch (IllegalArgumentException e) {
+      throw new IllegalArgumentException("ID invalid : " + id, e);
+    }
+  }
 }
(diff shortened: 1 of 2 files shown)


openbas-api/src/main/java/io/openbas/rest/helper/RestBehavior.java
```diff
@@ -206,4 +206,12 @@ public void checkOrganizationAccess(UserRepository userRepository, String organi
       }
     }
   }
+
+  protected void validateUUID(final String id) {
+    try {
+      UUID.fromString(id);
+    } catch (IllegalArgumentException e) {
+      throw new IllegalArgumentException("ID invalid : " + id, e);
```
antoinemzs, line 214: Will this cause HTTP 500, where it should be HTTP 400 IMO?

antoinemzs: How about throwing https://www.baeldung.com/spring-response-status-exception instead, wrapping the original exception?

RomuDeuxfois: You are right
I can throwing an ElementNotFoundException and it will be handled by the RestBehavior class

  @ExceptionHandler(ElementNotFoundException.class)
  public ResponseEntity<ErrorMessage> handleElementNotFoundException(ElementNotFoundException ex) {
    ErrorMessage message = new ErrorMessage("Element not found: " + ex.getMessage());
    log.warning("ElementNotFoundException: " + ex.getMessage());
    return new ResponseEntity<>(message, HttpStatus.NOT_FOUND);
  }

antoinemzs: That's for HTTP 404, but perhaps there is a readymade handler for 400 as well ?N

RomuDeuxfois: I can use InputValidationException

  @ResponseStatus(HttpStatus.BAD_REQUEST)
  @ExceptionHandler(InputValidationException.class)
  public ValidationErrorBag handleInputValidationExceptions(InputValidationException ex) {
    ValidationErrorBag bag = new ValidationErrorBag();
    ValidationError errors = new ValidationError();
    Map<String, ValidationContent> errorsBag = new HashMap<>();
    errorsBag.put(ex.getField(), new ValidationContent(ex.getMessage()));
    errors.setChildren(errorsBag);
    bag.setErrors(errors);
    return bag;
  }

antoinemzs: sounds like a plan

RomuDeuxfois: I made the changes.

antoinemzs (approved):

Issue #1490, opened 2024-09-19
[ Unit testing - people scope ] - players creation/update


EllynBsc:
Write unit tests for players creation and players update.

PR #2139, opened 2024-12-31
[Backend]Test player creation/update

johanah29:
#1490

### DIFF
--- a/openbas-api/src/test/java/io/openbas/rest/PlayerApiTest.java
+++ b/openbas-api/src/test/java/io/openbas/rest/PlayerApiTest.java
@@ -51,8 +51,8 @@ public class PlayerApiTest {
   @Autowired private TagRepository tagRepository;
   @Autowired private UserRepository userRepository;
 
-  @BeforeAll
-  void beforeAll() {
+  @BeforeEach
+  void beforeEach() {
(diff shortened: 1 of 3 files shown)


RomuDeuxfois:
Can you take care of upsertPlayer as well ? thx

antoinemzs (approved):

openbas-api/src/test/java/io/openbas/rest/PlayerApiTest.java
```diff
@@ -0,0 +1,249 @@
...
+    ORGANIZATION = organizationRepository.save(OrganizationFixture.createOrganization());
+    TAG = tagRepository.save(TagFixture.getTag());
+  }
+
+  @AfterEach
+  void afterEach() {
+    organizationRepository.delete(ORGANIZATION);
+    tagRepository.delete(TAG);
+    userRepository.delete(USER);
+  }
```
antoinemzs, line 60: If we mark the class as @Transactional here we don't need this method

antoinemzs (commented):
Good coverage, however we seem to not validate bevhaviour on invalid inputs (e.g. HTTP 400). Could do with a small test case? Or is it validated as a blanket error somewhere else? (i.e. as a generic deserialisation error code)

Issue #1494, opened 2024-09-19
 [ Unit testing - assets scope ] - Asset groups creation/update


EllynBsc:
Write unit tests for asset groups creation/update

PR #2150, opened 2025-01-02
[Backend]Test asset groups creation/update (#1494)

johanah29:
#1494

### DIFF
--- a/openbas-api/src/test/java/io/openbas/rest/asset_group/AssetGroupTest.java
+++ b/openbas-api/src/test/java/io/openbas/rest/asset_group/AssetGroupTest.java
@@ -1,0 +1,110 @@
+package io.openbas.rest.asset_group;
+
+import com.jayway.jsonpath.JsonPath;
+import io.openbas.database.model.AssetGroup;
+import io.openbas.database.model.Tag;
+import io.openbas.database.repository.AssetGroupRepository;
+import io.openbas.database.repository.TagRepository;
+import io.openbas.rest.asset_group.form.AssetGroupInput;
+import io.openbas.utils.fixtures.AssetGroupFixture;
+import io.openbas.utils.fixtures.TagFixture;
+import io.openbas.utils.mockUser.WithMockAdminUser;
+import org.junit.jupiter.api.*;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.http.MediaType;
+import org.springframework.test.web.servlet.MockMvc;
+
+import static io.openbas.utils.JsonUtils.asJsonString;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post;
+
+import java.util.List;
+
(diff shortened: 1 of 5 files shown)


openbas-api/src/test/java/io/openbas/rest/asset_group/AssetGroupTest.java
```diff
@@ -0,0 +1,105 @@
...
+                    .content(asJsonString(ASSET_GROUP_INPUT))
+                    .contentType(MediaType.APPLICATION_JSON)
+                    .accept(MediaType.APPLICATION_JSON))
+            .andExpect(status().is2xxSuccessful())
+            .andReturn()
+            .getResponse()
+            .getContentAsString();
+
+    // --ASSERT--
+    assertEquals("Asset group updated", JsonPath.read(response, "$.asset_group_name"));
```
antoinemzs, line 101: trivial: `newName` was set to the expected value, so it's preferrable to use the local variable here instead of repeating the actual string literal :)

openbas-api/src/test/java/io/openbas/rest/asset_group/AssetGroupTest.java
```diff
@@ -0,0 +1,105 @@
...
+                    .content(asJsonString(ASSET_GROUP_INPUT))
+                    .contentType(MediaType.APPLICATION_JSON)
+                    .accept(MediaType.APPLICATION_JSON))
+            .andExpect(status().is2xxSuccessful())
+            .andReturn()
+            .getResponse()
+            .getContentAsString();
+
+    // --ASSERT--
+    assertEquals("Asset group", JsonPath.read(response, "$.asset_group_name"));
```
antoinemzs, line 72: trivial: As it stands, the expected value here is not tied to what's been set in the fixture factory, so if someone changes the value there, and not here, the test breaks. It's an obvious breakage of course, so it's not going to cause a major issue. Although it's adhering more closely to the AAA pattern to setup your expectations within the test itself so that they are close to the assertions and don't drift accidentally.

antoinemzs (commented):
Good start, thanks !

Apart from a few nitpicks, perhaps also consider adding more tests cases, e.g.:
* for when the asset group does not exist (for update)
* for other types of properties, eg description, tags, filter (create, update).

RomuDeuxfois:
> Good start, thanks !
> 
> Apart from a few nitpicks, perhaps also consider adding more tests cases, e.g.:
> 
> * for when the asset group does not exist (for update)
> * for other types of properties, eg description, tags, filter (create, update).

Johanah being on leave, I take over on her PRs.

I took your comments into account and added the Transactional annotation.

I also took the opportunity to change the name of the methods to follow this pattern: given this should have that.
What do you think of this pattern?

antoinemzs:
> What do you think of this pattern?

Great pattern IMO for self explanatory test scopes. "Given this then that" or "when/with this it does that" are two widely used speech-like patterns that are also useful for the developer to limit the scope of their own tests.

antoinemzs (approved):
lgtm

Dimfacion (commented):
LGTM too :) Just a small nitpick but I think we should return a 404 if deleting a non existent element. It seems more logical to me as it can hide an issue in the frontend : 
Imagine you have a bad code on your frontend or a client of the API, you have the id incorrectly set. You'll get back a 200 but the element won't be correctly deleted. Debugging this might prove annoying.

Dimfacion (approved):

Issue #1998, opened 2024-12-05
Define and apply default assets for OCTI generated simulation


jborozco:
## Use case

As a user I want to be able to set default endpoint so when I generate a scenario from a report, all inject have already a target and I can directly launch the scenario from OpenCTI.

PR #2170, opened 2025-01-08
[backend] Refactored TagRule to use AssetGRoup instead of Assets Issue/1998

heditar:
### Proposed changes

* Refactored the code so the TagRule use AssetGroup instead of Assets
*

### Related issues

* #1998
*

### Checklist


openbas-api/src/main/java/io/openbas/rest/inject/service/InjectService.java
```diff
@@ -142,42 +142,42 @@ public void delete(String id) {
...
-        defaultAssetsToRemove.stream()
-            .filter(asset -> !defaultAssetsToAdd.contains(asset))
-            .map(Asset::getId)
+    // remove/add default asset groups and remove duplicates
+    List<AssetGroup> currentAssetGroups = inject.getAssetGroups();
+    // Get the Id of the asset groups to remove and filter the assets that are in both lists
+    List<String> assetGroupIdsToRemove =
+        defaultAssetGroupsToRemove.stream()
+            .filter(asset -> !defaultAssetGroupsToAdd.contains(asset))
+            .map(AssetGroup::getId)
```
savacano28, line 168: agroup?

openbas-api/src/main/java/io/openbas/migration/V3_57__TagRule_Update.java
```diff
@@ -0,0 +1,35 @@
...
+  @Override
+  public void migrate(Context context) throws Exception {
+    Connection connection = context.getConnection();
+    Statement select = connection.createStatement();
+    select.execute(
+        """
+                   CREATE TABLE tag_rule_asset_groups (
+                       tag_rule_id varchar(255) not null
+                            constraint tag_rule_id_fk
+                               references tag_rules,
```
savacano28, line 21: we could set also here : on delete cascade?

openbas-api/src/main/java/io/openbas/rest/inject/service/InjectService.java
```diff
@@ -142,42 +142,42 @@ public void delete(String id) {
...
             .toList();
     Set<String> uniqueAssetsIds = new HashSet<>();
```
savacano28, line 170: agroups

openbas-api/src/main/java/io/openbas/rest/inject/service/InjectService.java
heditar: thanks, I will update it in the next commit

openbas-api/src/main/java/io/openbas/migration/V3_57__TagRule_Update.java
heditar: yes it should be on cascade I will update it, thanks

openbas-api/src/main/java/io/openbas/rest/inject/service/InjectService.java
heditar: thanks, I will update it in the next commit

PR #2186, opened 2025-01-09
[backend/frontend] Delete legacy dryinjects and dryruns (#1985)

### LINKED ISSUE #1985: Remove Code Associated with dryInject Feature (Backend, Frontend, DB)
## Description

Currently, we have legacy code for the dryInject feature, so we need to remove this feature from the backend, frontend, and database: 
* InjectorHelper
* ScenarioSettings.js
* DryInject* tables
* ...

savacano28:
### Proposed changes

* Delete tables dryinject*, dryruns*
* Remove backend legacy 
* Remove frontend legacy 

### Related issues

* Close #1985 

### Checklist

### Further comments


openbas-api/src/main/java/io/openbas/helper/InjectHelper.java
```diff
@@ -53,9 +50,7 @@ private List<Team> getInjectTeams(@NotNull final Inject inject) {
   // -- INJECTION --
 
   private Stream<Tuple2<User, String>> getUsersFromInjection(Injection injection) {
-    if (injection instanceof DryInject dryInject) {
-      return dryInject.getRun().getUsers().stream().map(user -> Tuples.of(user, "Dryrun"));
-    } else if (injection instanceof Inject inject) {
+    if (injection instanceof Inject inject) {
```
impolitepanda, line 53: question: now that we only have one type of inject, couldn't we change the method signature to take directly an Inject instead of an Injection, thus eliminating the instanceof check and the throw of unsupported operation?

openbas-api/src/main/java/io/openbas/scheduler/jobs/InjectsExecutionJob.java
```diff
@@ -108,16 +106,10 @@ private void executeExternal(ExecutableInject executableInject) {
               Injection source = executableInject.getInjection();
               Inject executingInject = null;
               InjectStatus injectRunningStatus = null;
-              DryInjectStatus dryInjectRunningStatus = null;
               if (source instanceof Inject) {
```
impolitepanda, line 109: question: Same as previous comment in regards to the instanceof check / change of signature to avoid using Injection interface

openbas-front/src/actions/scenarios/scenario-actions.ts
```diff
@@ -117,11 +108,6 @@ export const removeScenarioTeamPlayers = (scenarioId: Scenario['scenario_id'], t
 
 // -- EXERCISES --
 
-export const fetchScenarioExercises = (scenarioId: Scenario['scenario_id']) => (dispatch: Dispatch) => {
```
impolitepanda, line 120: question: I don't know the difference between this fetch function and the search one just below. I suppose it was only used for scenario information? The fact that it updated the referential and that the search doens't isn't a problem, right ?

savacano28: This is a unused constant

impolitepanda (approved):

PR #2198, opened 2025-01-13
[backend] Add trigger before delete injector contract no custom

savacano28:
### Proposed changes

* Add trigger "before delete"on injector contract in order to avoid delete injector contract no custom (specifically: email)

### Related issues

* Closes #ISSUE-NUMBER
*

### Checklist

### Further comments


impolitepanda:
For me, this is a temporary workaround solution to the recent issue we had with emails and sms ijector contracts being deleted.
It will prevent any further deletion, but we definitely need to find the root cause for an actual fix, implement it, and then remove this trigger so that we don't have business rules implemented in the DB.

So in short, I'm ok with having the trigger to preven the issue from reoccuring, but temporarily.

savacano28:
Fixed by modification in code logic: https://github.com/OpenBAS-Platform/openbas/issues/2203

Issue #1489, opened 2024-09-19
[ Unit testing - simulation scope ] - import/export of a simulation (JSON)


EllynBsc:
Write unit tests in the context of a simulation for:

-Export of a simulation (JSON)
-Import of a simulation (JSON)

antoinemzs:
Will fix #2152 at the same time

EllynBsc:
Can we close this one then @antoinemzs and add the label 'duplicate' ?

PR #2210, opened 2025-01-14
[frontend] Add tests to Environment.exportData (#1489)

antoinemzs:
### Proposed changes

* Adding unit tests to Environment.exportData

### Related issues

* Contributes to #1489 

### Checklist

### Further comments

This is just focused on the actual function that flattens objects in a collection into Excel-able structures (i.e. flat objects). This function only special-cases a few specific keys in a candidate object though.
This is part of the XLS aspect of the linked issue.


openbas-front/src/__tests__/utils/Environment.test.tsx
```diff
@@ -0,0 +1,442 @@
...
+
+import { exportData } from '../../utils/Environment';
+import {
+  createExercisesMap,
+  createOrganisationsMap,
+  createScenarioMap,
+  createTagMap,
+} from '../fixtures/api-types.fixtures';
+
+/* eslint-disable  @typescript-eslint/no-explicit-any */
```
antoinemzs, line 12: This should be OK as this is really what I want for a "test type"

openbas-front/src/__tests__/fixtures/api-types.fixtures.ts
```diff
@@ -0,0 +1,60 @@
+/* eslint-disable  import/no-extraneous-dependencies */
```
antoinemzs, line 1: I think I was yelled at by eslint because the file itself is not named with the pattern *.test.ts; but at the same time I don't want to give it such a name because otherwise vitest picks it up and fails (no tests in file), and I don't want to introduce exclusions based on the name of the file if it ever gets renamed... open to thoughts.

guillaumejparis: fixed by updating eslint conf

antoinemzs: cheers

.circleci/config.yml
```diff
@@ -16,6 +16,9 @@ jobs:
       - run:
           working_directory: ~/openbas/openbas-front
           command: yarn install
+      - run:
+          working_directory: ~/openbas/openbas-front
+          command: yarn run test
```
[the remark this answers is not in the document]
antoinemzs: A bit confused as to which CI instance does what. I guess this makes it not crucial to have it here as long as we have a contract that droneci passes before the circle run indeed.

PR #2236, opened 2025-01-17
adding @Transactional to the ExecuteInject method to fix issue/2220

### LINKED ISSUE #2220: File upload payloads executed by the default OpenBAS agent on Windows do not appear to function correctly
## Description

When creating a file drop payload and attempting to execute via an OpenBAS agent installed on a Windows host, the file is not uploaded and the task hangs. The behaviour was recreated on the OpenBAS demo instance.

Relevant snipped log file appears to be:
```":{"className":"org.hibernate.LazyInitializationException","message":"failed to lazily initialize a collection of role: io.openbas.database.model.Document.tags: could not initialize proxy - no Session","stepArray":[{"className":"org.hibernate.collection.spi.AbstractPersistentCollection","methodName":"throwLazyInitializationException","fileName":"AbstractPersistentCollection.java","lineNumber":636},{"className":"org.hibernate.collection.spi.AbstractPersistentCollection","methodName":"withTemporarySessionIfNeeded","fileName":"AbstractPersistentCollection.java","lineNumber":219}```

## Environment

1. OS (where OpenBAS server runs): Docker
2. OpenBAS version: 1.10.1
3. OpenBAS client: Default Windows agent
4. Other environment details: N/A

## Reproducible Steps

1. Create a payload which performs a file drop
2. Create an atomic to execute the file drop payload on a Windows system
3. The atomic will be stuck in a queue state

## Expected Output

The file to be dropped in the execution subdirectory

## Actual Output

Agent stuck in queue state, no relevant debug information on the agent side - error messages on the server side

## Additional information

As a workaround, you can create a payload to download the file directly from the OpenBAS server via the API, however a valid API key needs to be included:
```
$headers = @{
    Authorization = "Bearer <snip>"
}

Invoke-WebRequest -Uri "https://<snip>/api/documents/<snip>/file" -Headers $headers -OutFile asd.exe
```

## Screenshots (optional)

Image

Image

### PR DESCRIPTION
**Pull request by heditar**
### Proposed changes

* adding the @Transactionnal annotation to the InjectsExecutionJob.execute method to fix an issue with the Lazy initialization that was happening when an InjectStatus was saved

* adding the @JsonIgnore annotation to Base.isListened method as the class StatusPayload as a an attribute payloadOutput that needs to be serialized and deserialized and having a property with a get a no set breaks the deserialization.
This issue is only impacting FileDrop payload as the issue with serialization deserialization is on the Document class.

This PR doesn't include test as having a test that would be relevant to this bug would cost too much and would require refactoring. 

### Related issues

* Closes #2220 
*

### Checklist

### Further comments

### DIFF
--- a/openbas-api/src/main/java/io/openbas/scheduler/jobs/InjectsExecutionJob.java
+++ b/openbas-api/src/main/java/io/openbas/scheduler/jobs/InjectsExecutionJob.java
@@ -32,6 +32,7 @@ import org.springframework.expression.Expression;
 import org.springframework.expression.ExpressionParser;
 import org.springframework.expression.spel.standard.SpelExpressionParser;
 import org.springframework.stereotype.Component;
+import org.springframework.transaction.annotation.Transactional;
 
 @Component
 @DisallowConcurrentExecution
@@ -380,6 +381,7 @@ public class InjectsExecutionJob implements Job {
     exerciseRepository.save(exercise);
   }
 
+  @Transactional
   @Override
   public void execute(JobExecutionContext jobExecutionContext) throws JobExecutionException {
     try {
--- a/openbas-model/src/main/java/io/openbas/database/model/Base.java
+++ b/openbas-model/src/main/java/io/openbas/database/model/Base.java
@@ -23,6 +23,7 @@ public interface Base {
     BeanUtils.copyProperties(input, this);
   }
 
+  @JsonIgnore
   default boolean isListened() {
     return true;
   }

### CODE REVIEW

File: openbas-api/src/main/java/io/openbas/scheduler/jobs/InjectsExecutionJob.java [hunk 1]
```diff
@@ -380,6 +381,7 @@ public void updateExercise(String exerciseId) {
     exerciseRepository.save(exercise);
   }
 
+  @Transactional
```
RomuDeuxfois, line 384: It seems risky to position it that high.
Were you able to analyze the impacts in relation to all the methods involved? We can work on this together if needed.

MarineLeM:
i think I encounter the same problem with my task because InjectStatus gets updated throughout the process (e.g., adding traces), but in the catch block, the InjectStatus fetched from inject.getStatus() isn't in its latest state..

openbas-api/src/main/java/io/openbas/scheduler/jobs/InjectsExecutionJob.java
heditar: I didn't spend too much time investigating potential impact as it seems to make sense to me that one "execute" was part of on transaction. What kind of impact are you worried about ?

RomuDeuxfois: Rather side effects due to the fact that a lot of things are wrapped in the transaction.
Let's take a moment to discuss this.

openbas-api/src/main/java/io/openbas/rest/inject/service/InjectService.java
```diff
@@ -377,6 +380,36 @@ public <T extends Base> void isPlanner(
     }
   }
 
+  @Transactional
+  public void initializeInjectStatus(
```
RomuDeuxfois, line 384: Can we have an InjectStatusService ?
What do you think ?

openbas-api/src/test/java/io/openbas/rest/inject/service/InjectServiceTest.java
```diff
@@ -599,4 +603,35 @@ void testIsPlannerNoInjects() {
     // Assert
     verify(securityExpression, times(0)).isSimulationPlanner("exercise1");
   }
+
+  @Test
+  void testInitializeInjectStatus_WITH_nonexisting_id() {
```
RomuDeuxfois, line 608: Should follow the nomenclature: given_that_should_this

openbas-api/src/main/java/io/openbas/rest/inject/service/InjectService.java
```diff
@@ -377,6 +380,36 @@ public <T extends Base> void isPlanner(
     }
   }
 
+  @Transactional
```
RomuDeuxfois, line 383: @Transactional(rollbackFor = Exception.class)

openbas-api/src/main/java/io/openbas/rest/inject/service/InjectService.java
```diff
@@ -377,6 +380,36 @@ public <T extends Base> void isPlanner(
...
+  public void initializeInjectStatus(
+      @NotNull final String injectId,
+      @NotNull final ExecutionStatus status,
+      @Nullable final InjectStatusExecution trace) {
+
+    Inject inject =
```
RomuDeuxfois, line 389: This method exists in the InjectService class: inject(injectId)

RomuDeuxfois:
Tested 🆗

openbas-api/src/main/java/io/openbas/rest/inject/service/InjectService.java
MarineLeM: I already create one, maybe i can just move this function inside when i will rebase my branch ? Or do you prefer to create this file now ?

impolitepanda:
Quick question, as I don't know the whole scope of the problem and PR, but are we sure we want to add transaction management at service level ? That means that if there are multiple transactional annotation called in succession (like from controller plus service methods), we might end up with some data persisted and some rollback when an error occurs.

heditar:
> Quick question, as I don't know the whole scope of the problem and PR, but are we sure we want to add transaction management at service level ? That means that if there are multiple transactional annotation called in succession (like from controller plus service methods), we might end up with some data persisted and some rollback when an error occurs.

Yes that makes sense for me to add it at the service level and actually I think that's where it should be. Service methods should be very clear about what is doing and we should know if we want to add the @Tansactionnal annotation.  If by controller you mean the Api class then I don't see why that can be an issue. If a transactional annotation is added on a method that call this method (or somewhere higher) with or without this annotation here it would be the same thing.  And if an annotation is added somewhere in a method called by initializeInjectStatus then this annotation won't impact the annotation on initializeInjectStatus.

impolitepanda:
You're absolutely correect ! 
To confirm, here is the default propagation value for transactional:

Propagation REQUIRED(TransactionDefinition.PROPAGATION_REQUIRED)
// Enum constant ordinal: 0
 
Support a current transaction, create a new one if none exists. Analogous to EJB transaction attribute of the same name.
This is the default setting of a transaction annotation.

So no issue at all :)

openbas-api/src/main/java/io/openbas/rest/inject/service/InjectService.java
RomuDeuxfois: Like you want.
If it goes in a good file at the end, it's ok for me.

RomuDeuxfois:
@heditar let me know if I can do a second round :)

RomuDeuxfois (approved):

PR #2251, opened 2025-01-20
[backend] Add default values to is ready computation

RomuDeuxfois:
### Related issues

* Closes https://github.com/OpenBAS-Platform/openbas/issues/2250


openbas-api/src/main/java/io/openbas/executors/tanium/openbas-tanium-packages.json
```diff
@@ -1 +1,56 @@
...
\ No newline at end of file
+{
```
impolitepanda, line 1: Is the commit of this file normal? (really don't know but it seems generated so sorry for asking)

openbas-model/src/main/java/io/openbas/helper/InjectModelHelper.java
```diff
@@ -28,47 +32,89 @@ public static boolean isReady(
...
+    ArrayNode injectContractFields;
+
+    try {
+      injectContractFields =
+          (ArrayNode)
+              mapper
+                  .readValue(injectorContract.getContent(), ObjectNode.class)
+                  .get(CONTACT_CONTENT_FIELDS);
+    } catch (JsonProcessingException e) {
+      throw new RuntimeException("Error parsing injector contract content", e);
```
impolitepanda, line 49: Would that really be a runtime exception ? The given JSON is client geven, so wouldn't it be more a badrequest ?

openbas-model/src/main/java/io/openbas/helper/InjectModelHelper.java
```diff
@@ -28,47 +32,89 @@ public static boolean isReady(
...
-              ready.set(false);
-            }
-          });
+
+    AtomicBoolean isReady = new AtomicBoolean(true);
+    for (JsonNode jsonField : contractMandatoryFields) {
+      String key = jsonField.get(CONTACT_ELEMENT_CONTENT_KEY).asText();
+
+      switch (key) {
+        case CONTACT_ELEMENT_CONTENT_KEY_TEAMS -> {
```
impolitepanda, line 73: Is there a reason you used lambdas here instead of a normal block ? It would avoid the use of an atomic boolean

openbas-api/src/main/java/io/openbas/executors/tanium/openbas-tanium-packages.json
RomuDeuxfois: Good questions: I only formatted this file to see it more clearly.
we can pass it, no other changes

openbas-model/src/main/java/io/openbas/helper/InjectModelHelper.java
RomuDeuxfois: It was simpler for me to use a runtime rather than having to propagate the exception in the methods above.
But I'm open to suggestions.

openbas-model/src/main/java/io/openbas/helper/InjectModelHelper.java
RomuDeuxfois: No reason, I'm going to change it.

openbas-model/src/main/java/io/openbas/helper/InjectModelHelper.java
impolitepanda: Let's go with RTE for now and we will see if we need to change it in the future, then :)

RomuDeuxfois:
Ready for a second round @impolitepanda :)

impolitepanda (approved):

PR #2287, opened 2025-01-22
[backend/frontend] Set feature flag policy to opt-in (#1824)

### LINKED ISSUE #1824: Align Feature Flag with the changes done on OCTI
## Use case

Switch to an enabled_dev_features properties instead of a disabled_dev_features to be much clearer and align ourselves with the changes made on the OCTI side.

antoinemzs:
### Proposed changes

* Change logic of feature flags to be opt-in, not opt-out

### Related issues

* Closes #1824

### Checklist

### Further comments

Note this introduces an enum for feature flags to guide which preview features exist


antoinemzs:
Lacking frontend tests because store is still hard to mock properly (and settings do come from the store in the relevant function on the front side)

openbas-api/src/main/java/io/openbas/service/PlatformSettingsService.java
```diff
@@ -217,7 +218,7 @@ public PlatformSettings findSettings() {
       platformSettings.setXtmOpenctiEnable(openCTIConfig.getEnable());
       platformSettings.setXtmOpenctiUrl(openCTIConfig.getUrl());
       platformSettings.setAiEnabled(aiConfig.isEnabled());
-      platformSettings.setAiHasToken(!aiConfig.getToken().isBlank());
+      platformSettings.setAiHasToken(aiConfig.getToken() != null && !aiConfig.getToken().isBlank());
```
impolitepanda, line 221: use StringUtils.isNotBlank

openbas-api/src/main/java/io/openbas/rest/settings/PreviewFeatureEnum.java
```diff
@@ -0,0 +1,14 @@
+package io.openbas.rest.settings;
+
+public enum PreviewFeatureEnum {
+  _RESERVED;
```
impolitepanda, line 4: Can you describe what thie enum value mean in a comment?

openbas-api/src/main/java/io/openbas/service/PlatformSettingsService.java
```diff
@@ -250,9 +251,23 @@ public PlatformSettings findSettings() {
     platformSettings.setPolicies(policies);
 
     // FEATURE FLAG
-    if (!openBASConfig.getDisabledDevFeatures().isEmpty()) {
-      platformSettings.setDisabledDevFeatures(
-          Arrays.stream(openBASConfig.getDisabledDevFeatures().split(",")).toList());
+    if (openBASConfig.getEnabledDevFeatures() == null
```
impolitepanda, line 254: use StringUtils.isBlank

openbas-api/src/main/java/io/openbas/service/PlatformSettingsService.java
```diff
@@ -250,9 +251,23 @@ public PlatformSettings findSettings() {
...
+        || openBASConfig.getEnabledDevFeatures().isEmpty()) {
+      platformSettings.setEnabledDevFeatures(List.of());
```
impolitepanda, line 256: List.of() creates an immutable list. Are we sure we want one here ?

openbas-api/src/main/java/io/openbas/service/PlatformSettingsService.java
```diff
@@ -250,9 +251,23 @@ public PlatformSettings findSettings() {
...
+    } else {
+      platformSettings.setEnabledDevFeatures(
+          Arrays.stream(openBASConfig.getEnabledDevFeatures().split(","))
+              .map(
+                  featureStr -> {
+                    try {
+                      return PreviewFeatureEnum.fromStringIgnoreCase(featureStr.strip());
+                    } catch (IllegalArgumentException e) {
+                      log.warning("Invalid feature flag: " + e.getMessage());
```
impolitepanda, line 265: Only a warning ?

openbas-api/src/test/java/io/openbas/service/PlatformServiceSettingsTest.java
```diff
@@ -0,0 +1,71 @@
...
+  @BeforeAll
+  public void beforeAll() {
+    // some repetitive setup necessary to mock config
+    rabbitmqConfig.setUser("admin");
+    rabbitmqConfig.setPass("pass");
+  }
+
+  @Test
+  @WithMockAdminUser
+  public void given_config_has_null_flags_enabled_features_is_empty() {
```
impolitepanda, line 43: not a problem at all, but why the sudden change to snake case ? :)

openbas-api/src/test/java/io/openbas/service/PlatformServiceSettingsTest.java
```diff
@@ -0,0 +1,71 @@
...
+    openbasConfig.setEnabledDevFeatures(null);
+
+    PlatformSettings settings = platformSettingsService.findSettings();
+
+    assertThat(settings.getEnabledDevFeatures(), is(equalTo(List.of())));
```
impolitepanda, line 48: I would have compared size to 0 instead of equals to list.of. But no idea which i's better. What do you think ?

openbas-api/src/test/java/io/openbas/service/PlatformServiceSettingsTest.java
```diff
@@ -0,0 +1,71 @@
...
+    openbasConfig.setEnabledDevFeatures("non existing feature flag");
+
+    PlatformSettings settings = platformSettingsService.findSettings();
+
+    assertThat(settings.getEnabledDevFeatures(), is(equalTo(List.of())));
+  }
+
+  @Test
+  @WithMockAdminUser
+  public void given_config_has_valid_flags_enabled_features_accounts_for_these_flags() {
```
impolitepanda, line 63: could you add a tests to verify what happens if you do openbasConfig.setEnabledDevFeatures(PreviewFeatureEnum._RESERVED.name() + ", " + PreviewFeatureEnum._RESERVED.name());

(to test what happens with a CSV string, but also what happens if you put multiple time the same value)

impolitepanda (changes requested):

openbas-api/src/main/java/io/openbas/service/PlatformSettingsService.java
antoinemzs: It seems it's called `hasText` in the version we have available: https://docs.spring.io/spring-framework/docs/6.1.16/javadoc-api/org/springframework/util/StringUtils.html#hasText(java.lang.String)

Thanks for the heads up!

openbas-api/src/main/java/io/openbas/service/PlatformSettingsService.java
antoinemzs: Current design mandates the `PlatformSettings` be updated by the service via the current `findSettings` method, so I don't see a benefit with a mutable collection in the absence of any set flag as they are to be computed all at once like here.
Note that the `toList` below is also creating an immutable list.

openbas-api/src/test/java/io/openbas/service/PlatformServiceSettingsTest.java
antoinemzs: i didn't think it through and it's force of habit, I still have to actively remind myself from time to time to camel case the names in Java.
I'll fix it, although I now remember that I have certainly committed the same mistake elsewhere :crying_cat_face:

openbas-api/src/test/java/io/openbas/service/PlatformServiceSettingsTest.java
antoinemzs: You got me think about it twice and actually there's the [`empty` matcher](https://hamcrest.org/JavaHamcrest/javadoc/2.2/org/hamcrest/Matchers.html#empty--) that's more expressive.

Btw the Hamcrest [`equalTo` matcher](https://hamcrest.org/JavaHamcrest/javadoc/2.2/org/hamcrest/Matchers.html#equalTo-T-) does support array matching with symmetrical indices so it works fine for empty lists

openbas-api/src/main/java/io/openbas/service/PlatformSettingsService.java
impolitepanda: Let's keep it as a warn for now, then

openbas-api/src/main/java/io/openbas/rest/settings/PreviewFeatureEnum.java
antoinemzs: It's in the latest commit [095b061](https://github.com/OpenBAS-Platform/openbas/pull/2287/commits/095b0614dcddb2dd7fa2f0965f8b3f6185ec2909)

PR #2293, opened 2025-01-24
[tool] Release artifacts to package expanded application.properties (docs#91)

antoinemzs:
### Proposed changes

* Release artifacts should package application.properties with expanded variables, not the placeholders

### Related issues

* Closes https://github.com/OpenBAS-Platform/docs/issues/91

### Checklist

### Further comments

The raw application.properties file contains placeholders to help with development (as to not disseminate the same data everywhere such as product version), but this can't be used to run a production server from a prebuilt jar.
Even though it seems unnecessary purely for running (a version of application.properties with expanded vars is embedded into the prebuilt jar), it is useful for users to have a quasi exhaustive config file to peruse possible switches and keys. It seems appropriate that the provided file would allow for running the jar with the same default values as set in the embedded version.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -136,7 +136,7 @@ jobs:
       - attach_workspace:
           at: ~/
       - run: mkdir -p release/openbas
-      - run: cp -a ./openbas-api/target/openbas-api.jar release/openbas/ && cp -a ./openbas-api/src/main/resources/application.properties release/openbas/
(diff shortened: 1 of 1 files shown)


antoinemzs:
Note that I don't propose a retroactive fix for all past releases. A few artifacts from 1.10.0 to the current 1.11.4 are affected with a "wrong" application.properties file. Users may figure out that they can ditch the file (or set it aside), or fix or remove the few offending placeholders.

impolitepanda (approved):

PR #2303, opened 2025-01-27
Rework executor system

RomuDeuxfois:


openbas-api/src/main/java/io/openbas/utils/PayloadMapper.java
```diff
@@ -91,8 +96,18 @@ public StatusPayloadOutput getStatusPayloadOutputFromInject(Optional<Inject> inj
...
               .prerequisites(statusPayload.getPrerequisites())
               .externalId(statusPayload.getExternalId())
-              .executableFile(statusPayload.getExecutableFile())
-              .fileDropFile(statusPayload.getFileDropFile())
+              .executableFile(
+                  Optional.ofNullable(statusPayload.getExecutableFile())
+                      .orElse(
+                          Optional.ofNullable(statusPayload.getExecutableFileId())
+                              .flatMap(this.documentRepository::findById)
+                              .orElse(null)))
```
impolitepanda: Won't returning null here cause an NPE on the next instruction?

openbas-api/src/main/java/io/openbas/utils/PayloadMapper.java
```diff
@@ -91,8 +96,18 @@ public StatusPayloadOutput getStatusPayloadOutputFromInject(Optional<Inject> inj
...
+              .fileDropFile(
+                  Optional.ofNullable(statusPayload.getFileDropFile())
+                      .orElse(
+                          Optional.ofNullable(statusPayload.getFileDropFileId())
+                              .flatMap(this.documentRepository::findById)
```
impolitepanda: Call to documentRepo findById is done twice. Could we maybe check the optional logic before and only call it once if needed?

openbas-api/src/main/java/io/openbas/utils/PayloadMapper.java
```diff
@@ -91,8 +96,18 @@ public StatusPayloadOutput getStatusPayloadOutputFromInject(Optional<Inject> inj
...
+                              .orElse(null)))
```
impolitepanda: Same as above regarding returning null value

impolitepanda (changes requested):

openbas-api/src/main/java/io/openbas/executors/Executor.java
```diff
@@ -12,102 +14,102 @@
...
+        injectorRepository
+            .findByType(injectorContract.getInjector().getType())
+            .orElseThrow(
+                () ->
+                    new IllegalStateException(
+                        "Injector not found for type: "
+                            + injectorContract.getInjector().getType()));
+    // Status
+    InjectStatus injectStatus = this.injectService.upsertInjectStatus(inject, EXECUTING, null);
+    inject.setStatus(injectStatus);
```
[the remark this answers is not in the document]
RomuDeuxfois: Fixed

openbas-api/src/main/java/io/openbas/executors/Executor.java
```diff
@@ -12,102 +14,102 @@
...
+        injectStatusRepository.save(injectStatus);
+        return injectStatus;
+      }
+    }
+    if (injector.isExternal()) {
+      injectStatus = executeExternal(executableInject, injector);
+    } else {
+      injectStatus = executeInternal(executableInject, injector);
+    }
+    return injectStatus;
```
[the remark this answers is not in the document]
RomuDeuxfois: Fixed

openbas-model/src/main/java/io/openbas/database/model/StatusPayload.java
```diff
@@ -79,12 +83,42 @@ public StatusPayload(
...
     this.externalId = externalId;
     this.prerequisites = prerequisites;
     this.arguments = arguments;
     this.payloadCommandBlocks = payloadCommandBlocks;
     this.cleanupExecutor = cleanupExecutor;
   }
+
+  @Data
+  @JsonIgnoreProperties(ignoreUnknown = true)
+  public static class StatusPayloadDocument {
```
heditar, line 101: should we move that to its own file ?

openbas-model/src/main/java/io/openbas/database/model/StatusPayload.java
```diff
@@ -29,10 +33,10 @@ public class StatusPayload {
   private String externalId;
 
   @JsonProperty("executable_file")
-  private Document executableFile;
+  private StatusPayloadDocument executableFile;
 
   @JsonProperty("file_drop_file")
-  private Document fileDropFile;
+  private StatusPayloadDocument fileDropFile;
```
heditar, line 39: can we use a builder here ?

openbas-api/src/main/java/io/openbas/rest/inject/service/InjectService.java
```diff
@@ -398,31 +407,28 @@ public <T extends Base> void isPlanner(
...
-      injectStatus.getTraces().add(trace);
+    if (traces != null) {
+      injectStatus.getTraces().addAll(traces);
     }
     injectStatus.setName(status);
     injectStatus.setTrackingSentDate(Instant.now());
     injectStatus.setPayloadOutput(injectUtils.getStatusPayloadFromInject(inject));
-    injectStatusRepository.save(injectStatus);
+    InjectStatus injectStatusSaved = injectStatusRepository.save(injectStatus);
+    inject.setStatus(injectStatusSaved);
```
heditar, line 431: As you I don't like this approach of updating the object in the method as it can have non explicit impact. I know that is not an universal opinion but having both "return" + updating the inject seems more confusing to me.

RomuDeuxfois:
@heditar & @impolitepanda ready for a secound round

openbas-api/src/main/java/io/openbas/executors/Executor.java
```diff
@@ -39,51 +38,71 @@ public class Executor {
...
+      queueService.publish(injector.getType(), jsonInject);
+      InjectStatus injectStatus = this.injectStatusRepository.findByInject(inject).orElseThrow();
+      injectStatus.addInfoTrace(
           "The inject has been published and is now waiting to be consumed.",
-          ExecutionTraceAction.EXECUTION,
-          null);
-      queueService.publish(injectorContract.getInjector().getType(), jsonInject);
+          ExecutionTraceAction.EXECUTION);
+      return this.injectStatusRepository.save(injectStatus);
     } catch (Exception e) {
```
impolitepanda, line 51: Add a log here ?

openbas-api/src/main/java/io/openbas/executors/Executor.java
```diff
@@ -39,51 +38,71 @@ public class Executor {
...
+                            + injectorContract.getInjector().getType()));
+
+    // Status
+    InjectStatus injectStatus =
+        this.injectStatusService.initializeInjectStatus(inject.getId(), EXECUTING);
+    if (Boolean.TRUE.equals(injectorContract.getNeedsExecutor())) {
+      try {
+        this.executionExecutorService.launchExecutorContext(inject);
+      } catch (Exception e) {
+        injectStatus.setName(ERROR);
```
impolitepanda, line 94: Add a log here ?

openbas-api/src/main/java/io/openbas/rest/inject/service/InjectService.java
```diff
@@ -92,23 +91,15 @@ public void deleteAll(List<Inject> injects) {
 
   public Map<Asset, Boolean> resolveAllAssetsToExecute(@NotNull final Inject inject) {
     Map<Asset, Boolean> assets = new HashMap<>();
-    inject
-        .getAssets()
-        .forEach(
-            (asset -> {
-              assets.put(asset, false);
-            }));
+    inject.getAssets().forEach((asset -> assets.put(asset, false)));
```
impolitepanda, line 94: Map<Asset, Boolean> assets = inject.getAssets().stream().collect(Collectors.toMap(Asset, false));

(forEach is often slower than streams)

openbas-api/src/main/java/io/openbas/scheduler/jobs/InjectsExecutionJob.java
```diff
@@ -274,15 +180,11 @@ private Optional<List<String>> getErrorMessagesPreExecution(String exerciseId, I
...
-              .forEach(
-                  condition -> {
-                    mapCondition.put(condition.getKey(), false);
-                  });
-        });
+        injectDependency ->
+            injectDependency
+                .getInjectDependencyCondition()
+                .getConditions()
+                .forEach(condition -> mapCondition.put(condition.getKey(), false)));
```
impolitepanda, line 187: cf comment above about foreach/stream

impolitepanda (approved):

heditar (approved):

openbas-api/src/main/java/io/openbas/executors/Executor.java
MarineLeM: already add log in the function failInjectStatus

openbas-api/src/main/java/io/openbas/executors/Executor.java
```diff
@@ -39,51 +38,71 @@
...
-    status.setInject(inject);
+  private InjectStatus executeExternal(ExecutableInject executableInject, Injector injector) {
+    Inject inject = executableInject.getInjection().getInject();
     try {
       String jsonInject = mapper.writeValueAsString(executableInject);
-      status.setName(ExecutionStatus.PENDING); // FIXME: need to be test with HTTP Collector
-      status.addTrace(
-          ExecutionTraceStatus.INFO,
+      queueService.publish(injector.getType(), jsonInject);
+      InjectStatus injectStatus = this.injectStatusRepository.findByInject(inject).orElseThrow();
```
MarineLeM, line 46: Place that line before calling queueService.publish(injector.getType(), jsonInject);. If you execute it afterward and an error is thrown due to a missing injectStatus, you'll miss logging that the inject was published. This ensures the publish log is recorded regardless of later errors.

openbas-api/src/main/java/io/openbas/executors/Executor.java
MarineLeM: yes you should call  injectStatusService.failInjectStatus(...);

openbas-api/src/main/java/io/openbas/executors/Executor.java
```diff
@@ -39,51 +38,71 @@ public class Executor {
...
+    return injectStatusRepository.save(completeStatus);
+  }
+
+  public InjectStatus execute(ExecutableInject executableInject) {
+    Inject inject = executableInject.getInjection().getInject();
     InjectorContract injectorContract =
         inject
             .getInjectorContract()
             .orElseThrow(
                 () -> new UnsupportedOperationException("Inject does not have a contract"));
```
MarineLeM, line 75: If you throw an error here the injectStatus will remain in 'QUEING' and no logs will be added because the catch error in the parent level also throw an error. May be you should call failInjectStatus function here or directly in the parent level in the InjectExecutionJob.execute function

openbas-api/src/main/java/io/openbas/executors/Executor.java
```diff
@@ -39,51 +38,71 @@ public class Executor {
...
-    Inject executedInject = injectRepository.findById(inject.getId()).orElseThrow();
-    InjectStatus completeStatus = injectStatusService.fromExecution(execution, executedInject);
-    return injectStatusRepository.save(completeStatus);
+    // Depending on injector type (internal or external) execution must be done differently
+    Injector injector =
+        injectorRepository
+            .findByType(injectorContract.getInjector().getType())
+            .orElseThrow(
+                () ->
+                    new IllegalStateException(
```
MarineLeM, line 83: same as above

openbas-api/src/main/java/io/openbas/rest/inject/service/InjectService.java
RomuDeuxfois: I didn't know that. Thx !

MarineLeM:
Test ok ✔️

openbas-api/src/main/java/io/openbas/executors/Executor.java
RomuDeuxfois: I interpret that the other way around.  
If I publish and it works well, then I can log "It's published now."

heditar:
I tested a filedrop with tags and I confirm that is also fixing our deserialization issue as you remove the Documents in StatusPayload.

openbas-api/src/main/java/io/openbas/executors/Executor.java
heditar: Yes we need a try catch around setInjectStatusAndExecuteInject in InjectsExecutionJob.executeInject

openbas-api/src/main/java/io/openbas/scheduler/jobs/InjectsExecutionJob.java
```diff
@@ -163,48 +103,14 @@ private void executeInject(ExecutableInject executableInject) {
   }
 
   private void setInjectStatusAndExecuteInject(ExecutableInject executableInject, Inject inject) {
```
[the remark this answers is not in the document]
RomuDeuxfois: I will remove this method

MarineLeM (approved):

PR #2319, opened 2025-01-29
[backend] Account for all relevant pauses in inject start date computation (#2282)

### LINKED ISSUE #2282: The first inject after a pause in a simulation doesn't get executed
## Description

Create a simulation with some sms/email injects.
Pause it for a few minutes.
The first inject after the pause does not get executed and gets 'error' status
Image

antoinemzs:
### Proposed changes

* Correctly account for pause times in inject start date computation

### Related issues

* Closes #2282 

### Checklist

### Further comments

Tests are better implemented with composers hence I propose this PR to release branch. Another PR with just the code fix is going to master.

There's a drive-by fix on the schema because current pause was being persisted assuming the provided instant was local time, when in reality it was provided as UTC => the backend would compute an incorrect UTC time to persist. It was low impact but needed to be addressed.

### DIFF
--- a/openbas-api/src/test/java/io/openbas/database/model/ExerciseTest.java
+++ b/openbas-api/src/test/java/io/openbas/database/model/ExerciseTest.java
@@ -1,0 +1,51 @@
+package io.openbas.database.model;
+
+import static org.junit.jupiter.api.TestInstance.Lifecycle.PER_CLASS;
+
+import io.openbas.database.raw.RawExercise;
+import io.openbas.database.repository.ExerciseRepository;
+import io.openbas.utils.fixtures.ExerciseFixture;
+import io.openbas.utils.fixtures.composers.ExerciseComposer;
+import jakarta.persistence.EntityManager;
+import java.time.Instant;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.DisplayName;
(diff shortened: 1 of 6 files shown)


openbas-api/src/main/java/io/openbas/migration/V3_62__make_exercise_pause_tz_aware.java
```diff
@@ -0,0 +1,25 @@
...
+    Connection connection = context.getConnection();
+    Statement statement = connection.createStatement();
+
+    statement.execute(
+        """
+        ALTER TABLE exercises ADD COLUMN exercise_pause_date_tempwithtz TIMESTAMP WITH TIME ZONE;
+        UPDATE exercises SET exercise_pause_date_tempwithtz = exercise_pause_date;
+        ALTER TABLE exercises DROP COLUMN exercise_pause_date;
+        ALTER TABLE exercises RENAME COLUMN exercise_pause_date_tempwithtz TO exercise_pause_date;
+    """);
```
antoinemzs, line 23: There's no migration of existing data (i.e. recompute of timestamps) because I lack information on the relevant timezone that originated the Instant.
But it's not really important as it only affects currently paused simulations and would not impact them further. Assumed TZ after migration is UTC.

guillaumejparis (approved):

PR #2321, opened 2025-01-29
[backend] Converting return lines to HTML tags when importing rich text fields

### LINKED ISSUE #2317: Correct handling of return line when importing into an email content
## Context

When using the import mapper to import cell into the content of an email, the content of the cell is not properly formatted. 
The content of the email is expecting HTML while the content of a cell is extracted as plain text. 
As a first step, we should at least convert return line into HTML line break as it's the most impactful as well as the easiest to do.

Dimfacion:
### Proposed changes

* Converting return lines to HTML tags when importing rich text fields
*

### Related issues

* Closes #2317 
*

### Checklist

### Further comments

### DIFF
--- a/openbas-api/src/main/java/io/openbas/service/InjectImportService.java
+++ b/openbas-api/src/main/java/io/openbas/service/InjectImportService.java
@@ -809,6 +809,14 @@ public class InjectImportService {
               Arrays.stream(ruleAttribute.getColumns().split("\\+"))
                   .map(column -> getValueAsString(row, column))
                   .collect(Collectors.joining());
+
+          // Given that richText fields are editable using CKEditor which expects HTML,
+          // we're converting return line into <br>.
+          // TODO : convert properly the whole cell into HTML including formatting (bold, ...)
+          if (mapFieldByKey.get(ruleAttribute.getName()).get("richText") != null
+              && mapFieldByKey.get(ruleAttribute.getName()).get("richText").asBoolean()) {
+            columnValue = columnValue.replaceAll("\n", "<br>");
+          }
         }
         if (columnValue.isBlank()) {
           inject.getContent().put(ruleAttribute.getDefaultValue(), columnValue);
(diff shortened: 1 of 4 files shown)


openbas-api/src/main/java/io/openbas/service/InjectImportService.java
```diff
@@ -809,6 +809,14 @@ private List<ImportMessage> addFields(
               Arrays.stream(ruleAttribute.getColumns().split("\\+"))
                   .map(column -> getValueAsString(row, column))
                   .collect(Collectors.joining());
+
+          // Given that richText fields are editable using CKEditor which expects HTML,
+          // we're converting return line into <br>.
+          // TODO : convert properly the whole cell into HTML including formatting (bold, ...)
+          if (mapFieldByKey.get(ruleAttribute.getName()).get("richText") != null
+              && mapFieldByKey.get(ruleAttribute.getName()).get("richText").asBoolean()) {
+            columnValue = columnValue.replaceAll("\n", "<br>");
```
antoinemzs, line 818: A couple nitpicks:
* since `replaceAll` takes a regex, would you consider optionally matching `\r` (as in `\r\n`), e.g. `\r?\n`
* the HTML standard asks that standalone tags are autoclosed, i.e. `<br />`

antoinemzs, line 818: Would you consider extracting this as a function and write a few tests for it ? The method is already extra long and complex, and hard to test by itself.

antoinemzs (approved):
Nice quick fix, it's fine as it is, although I suggest a small extra mile perhaps achievable without too much effort?

openbas-api/src/main/java/io/openbas/service/InjectImportService.java
Dimfacion: That's a good point, doing right now :)

antoinemzs: On second thought, I have had it wrong re HTML and self closing tags, see https://developer.mozilla.org/en-US/docs/Glossary/Void_element#self-closing_tags

It looks like it's got to do with XHTML, not plain HTML. So perhaps not so important depending which one of the many standards we want to adhere to :sleepy:

Dimfacion: Good to know ! It's not so important but I feel it's cleaner when those are self closed so I'll do it anyway if that's also fine by you ;)

Dimfacion: I extracted a few functions into a utilitarian class and wrote tests for the extracted functions. It should be better now :)

antoinemzs (approved):
Cheers for the changes ! lgtm

PR #2330, opened 2025-01-30
[backend] Improvement creation inject expectation 

savacano28:
### Proposed changes

* Improvement of creation inject expectations with clean signatures : 
->Obas ImplantExecutor, we pass only signature parent process name for agent expectations

* Refact Caldera executor in order to generate expectations for every executed agent

### Checklist

### Further comments

### DIFF
--- a/openbas-api/src/main/java/io/openbas/injectors/caldera/CalderaExecutor.java
+++ b/openbas-api/src/main/java/io/openbas/injectors/caldera/CalderaExecutor.java
@@ -2,6 +2,7 @@ package io.openbas.injectors.caldera;
 
 import static io.openbas.database.model.InjectExpectationSignature.*;
 import static io.openbas.database.model.InjectStatusExecution.*;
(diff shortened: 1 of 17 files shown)


openbas-api/src/main/java/io/openbas/injectors/caldera/CalderaExecutor.java
```diff
@@ -340,126 +283,213 @@ private Map<Asset, Boolean> resolveAllAssets(@NotNull final ExecutableInject inj
...
+    } else {
+      injectExpectationSignatures.add(
+          InjectExpectationSignature.builder()
+              .type(EXPECTATION_SIGNATURE_TYPE_PROCESS_NAME)
+              .value(processName)
+              .build());
+    }
+
+    return injectExpectationSignatures;
+  }
```
antoinemzs, line 422: This looks like the signature creation should be in the Payload class, although it could be argued that a given signature is a function of the Payload and the Executor. Happy to discuss.

antoinemzs (commented):
First cursory pass, happy to discuss if needed

PR #2354, opened 2025-01-31
[doc] add comments to critical sections of the application.properties file (docs/#89)

antoinemzs:
### Proposed changes

* More comments to help setting up the mandatory dependencies in application.properties

### Related issues

* Contributes to https://github.com/OpenBAS-Platform/docs/issues/89

### DIFF
--- a/openbas-api/src/main/resources/application.properties
+++ b/openbas-api/src/main/resources/application.properties
@@ -19,10 +19,63 @@ openbas.with-proxy=false
 #openbas.admin.password=ChangeMe
 #openbas.admin.token=ChangeMe
 
-#Feature Flag
-openbas.disabled-dev-features=
+########################
+# RUNTIME DEPENDENCIES #
+########################
+# PostgreSQL, MinIO and RabbitMQ must be configured (and running) for OpenBAS to work properly
 
-# Server configuration
+### POSTGRESQL settings
+### see also: https://docs.openbas.io/latest/deployment/configuration/#postgresql
+# connection URL to the PostgreSQL server
+spring.datasource.url=jdbc:postgresql://...
+# Username: should be identical to the username set on the database instance
+spring.datasource.username=openbas
+# Password: the password for the username set above
+spring.datasource.password=<password>
+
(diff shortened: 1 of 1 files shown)


openbas-api/src/main/resources/application.properties
```diff
@@ -19,10 +19,63 @@ openbas.with-proxy=false
...
-openbas.disabled-dev-features=
+########################
+# RUNTIME DEPENDENCIES #
+########################
+# PostgreSQL, MinIO and RabbitMQ must be configured (and running) for OpenBAS to work properly
+
+### POSTGRESQL settings
+### see also: https://docs.openbas.io/latest/deployment/configuration/#postgresql
+# connection URL to the PostgreSQL server
+spring.datasource.url=jdbc:postgresql://...
```
Dimfacion, line 30: *thought :* I'm wondering if we shouldn't make this field blank and give an example in the comment above as I've seen people leaving this field as is

Dimfacion: Or maybe to be in line with other places in this .properties (but I'm afraid it might confuse even more users) 
```suggestion
spring.datasource.url=<jdbc:postgresql://...>
```

RomuDeuxfois: I agree, we can let a placeholder like <something>

antoinemzs: See: [`050462a`](https://github.com/OpenBAS-Platform/openbas/pull/2354/commits/050462aa120d4ba40418a7e4de3c5653da6d241c)

openbas-api/src/main/resources/application.properties
```diff
@@ -275,6 +308,7 @@ openbas.expectation.manual.default-score-value=50
 
 
 #Health check
+# a unique random string
```
Dimfacion, line 311:
```suggestion
# a unique random string used when calling the health endpoint 
# Example of a call : /api/health?health_access_key=ChangeMe
```

Dimfacion (approved):

Issue #2324, opened 2025-01-30
Error with MailGun


jborozco:
## Description

we noticed on a client an inject in error and it seems to be relative to mailgun.

Link to reproduce in project comments

RomuDeuxfois:
We need to improv our retry system.

RomuDeuxfois:
![Image](https://github.com/user-attachments/assets/294f08c3-9670-4b54-b5e5-4f5139c090a4)

johanah29:
We openend a PR with some code allowing to visualize more traces. However we recommend an investigation from the platform team to see if there was some traces from the Mailgun server.

PR #2367, opened 2025-02-04
[backend]Add traces for sending emails(#2324)

johanah29:
#2324

### DIFF
--- a/openbas-api/src/main/java/io/openbas/injectors/email/service/EmailService.java
+++ b/openbas-api/src/main/java/io/openbas/injectors/email/service/EmailService.java
@@ -13,10 +13,12 @@ import jakarta.mail.MessagingException;
 import jakarta.mail.Multipart;
 import jakarta.mail.internet.*;
 import jakarta.mail.util.ByteArrayDataSource;
+
 import java.io.ByteArrayOutputStream;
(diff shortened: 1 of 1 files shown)


openbas-api/src/main/java/io/openbas/injectors/email/service/EmailService.java
```diff
@@ -210,17 +210,18 @@ private MimeMessage getEncryptedMimeMessage(
...
   }
 
-  private void sendEmailWithRetry(MimeMessage mimeMessage) throws Exception {
+  private void sendEmailWithRetry(Execution execution, MimeMessage mimeMessage)
+      throws InterruptedException {
     for (int i = 0; i < 3; i++) {
       try {
         emailSender.send(mimeMessage);
-        break;
+        execution.addTrace(traceSuccess("Mail successfully sent"));
```
RomuDeuxfois, line 218: @MarineLeM what do you think ?

MarineLeM: I think this trace is not needed here because it's already in the parent function at line 108. Having it in both places would result in duplicate traces

johanah29:
Opening pull request from release/current branch

Issue #2349, opened 2025-01-31
Incorrect behaviour on field with on-the-fly creation (label is not correctly placed)


SamuelHassine:
## Description

Incorrect behaviour on field with on-the-fly creation (label is not correctly placed)

Correct:

![Image](https://github.com/user-attachments/assets/144da2c1-4768-4e4b-b651-2c902d5a778d)

Incorrect:

![Image](https://github.com/user-attachments/assets/bef43b2f-0708-41e0-8509-334704453972)

PR #2380, opened 2025-02-06
[frontend] Correctly placing label of fields with on-the-fly creation (#2349)

johanah29:
#2349


guillaumejparis (changes requested):
Please so the same as octi 
image

johanah29:
image

guillaumejparis (approved):

PR #2387, opened 2025-02-06
When I create a technical inject in a simulation and I remove the expectation, expectations still there

RomuDeuxfois:
### Related issues

* https://github.com/OpenBAS-Platform/openbas/issues/2316

### DIFF
--- a/openbas-api/src/main/java/io/openbas/rest/inject/service/InjectService.java
+++ b/openbas-api/src/main/java/io/openbas/rest/inject/service/InjectService.java
@@ -1,6 +1,8 @@
 package io.openbas.rest.inject.service;
 
-import static io.openbas.utils.FilterUtilsJpa.*;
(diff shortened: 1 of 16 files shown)


openbas-api/src/main/java/io/openbas/rest/inject/service/InjectService.java
```diff
@@ -61,9 +66,91 @@ public class InjectService {
...
+  private final TagService tagService;
+  private final DocumentService documentService;
 
   @Resource protected ObjectMapper mapper;
 
+  public Inject createInject(
+      @Nullable final Exercise exercise,
+      @Nullable final Scenario scenario,
+      @NotNull final InjectInput input) {
+    if (exercise == null & scenario == null) {
```
RomuDeuxfois, line 81: Sanity check need to be done on rest part.
Should have one exercise or one scenario, not both, at least one.

PR #2388, opened 2025-02-06
[frontend/backend] Filter injects on teams, assets and asset groups

### LINKED ISSUE #1958: Be able to filter inject tab by teams, asset and asset groups
## Use case

Be able to filter injects by target (player and teams) so I can target the replacement of an asset by an other when I edit in bulk.

impolitepanda:
### Proposed changes

* Add filters on inject tables on teams, assets, asset groups and assets in asset groups

### Related issues

* Closes #1958 

### Checklist


openbas-framework/src/main/java/io/openbas/utils/schema/SchemaUtils.java
```diff
@@ -171,10 +171,11 @@ private static void processAnnotations(
             .filterable(queryable.filterable())
             .dynamicValues(queryable.dynamicValues())
             .sortable(queryable.sortable())
-            .path(queryable.path());
+            .path(queryable.path())
+            .paths(queryable.paths());
         if (member instanceof Method) {
           builder.type(queryable.clazz()); // Override
-        } else if (member instanceof Field && hasText(queryable.path())) {
+        } else if (hasText(queryable.path()) || queryable.paths().length > 0) {
```
Dimfacion, line 178: **question :** Why removing member instanceof Field ?

impolitepanda: Because it was always true. On line 164, there is already a test on this. I can put it back if you want.

Dimfacion: Oh, right ! Thanks, it's fine that way

openbas-model/src/main/java/io/openbas/database/model/Inject.java
```diff
@@ -208,6 +210,9 @@ public class Inject implements Base, Injection {
       inverseJoinColumns = @JoinColumn(name = "asset_group_id"))
   @JsonSerialize(using = MultiIdListDeserializer.class)
   @JsonProperty("inject_asset_groups")
+  @Queryable(
+      filterable = true,
+      paths = {"assetGroups.name", "assetGroups.assets.name"})
```
Dimfacion, line 215: **question :** After rereading the notion page, I'm wondering if it wouldn't make more sense to move "assetGroups.assets.name" in the queryable of the assets.

impolitepanda: I think you're absolutely right. Doing.

impolitepanda: Done

Dimfacion (approved):
LGTM but since I worked a bit on this, I'd be nice if someone else also looked at it

savacano28:
Tested OK

impolitepanda:
Note: The ability to filter injects on assets from asset groups have been removed from the scope of this issue.

openbas-api/src/main/java/io/openbas/rest/asset/endpoint/EndpointApi.java
```diff
@@ -148,4 +151,25 @@
...
   }
+
+  // -- OPTION --
+
+  @GetMapping(ENDPOINT_URI + "/options")
+  public List<FilterUtilsJpa.Option> optionsByName(
+      @RequestParam(required = false) final String searchText,
+      @RequestParam(required = false) final String simulationOrScenarioId) {
+    return endpointRepository
+        .findAllBySimulationOrScenarioIdAndName(
```
MarineLeM, line 162: This function is used only for this request. Is it bad practice to create a query that returns only the name and id instead of fetching the entire entity and mapping it afterward?

MarineLeM (approved):

openbas-framework/src/main/java/io/openbas/utils/FilterUtilsJpa.java
```diff
@@ -91,6 +91,31 @@ private static <T, U> Specification<T> computeFilter(
       }
       List<PropertySchema> filterableProperties = getFilterableProperties(propertySchemas);
       PropertySchema filterableProperty = retrieveProperty(filterableProperties, filterKey);
+
+      // multiple paths case
+      if (filterableProperty.getPaths().length > 1) {
```
savacano28, line 96: We need still this code?

openbas-framework/src/main/java/io/openbas/utils/schema/PropertySchema.java
```diff
@@ -34,6 +34,7 @@ public class PropertySchema {
 
   private final JoinTable joinTable;
   private final String path;
+  private final String[] paths;
```
savacano28, line 37: paths is used?

savacano28 (approved):

openbas-api/src/main/java/io/openbas/rest/asset/endpoint/EndpointApi.java
impolitepanda: It's kind of a bad practice indeed. Repository method should almost always return the full object in order to make them more reusable in the future. Also, there would be no gain here in returning only names and ids: the scan of the DB wouldn't change and the cost of the mapping in the controller layer is minimal.

openbas-framework/src/main/java/io/openbas/utils/FilterUtilsJpa.java
impolitepanda: As seen together, it is not needed anymore. However, it will be usefull in the future and so will be commited

openbas-framework/src/main/java/io/openbas/utils/schema/PropertySchema.java
impolitepanda: As seen together, it is not needed anymore. However, it will be usefull in the future and so will be commited

RomuDeuxfois:
I did a test on prerelease and seems that empty and is empty operator is not working properly @impolitepanda

impolitepanda:
> I did a test on prerelease and seems that empty and is empty operator is not working properly @impolitepanda

will look at it asap

PR #2428, opened 2025-02-11
[backend] fix openapi annotations to restore compatible api types

antoinemzs:
### Proposed changes

* Fix some OpenAPI annotations to avoid conflict with established typescript code when regenerating the api type proxies in typescript


Dimfacion (approved):
LGTM :)

PR #2438, opened 2025-02-11
Improve release workflow

Dimfacion:
Creation of a new release workflow with each repo being owner of their release process

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,15 +6,15 @@ on:
       branch_platform:
         type: string
         default: 'main'
-        required: false
+        required: true
       previous_version:
         type: string
         default: ''
-        required: false
+        required: true
       new_version:
         type: string
         default: ''
-        required: false
+        required: true
 
 jobs:
   release:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,7 +5,7 @@ on:
     inputs:
       branch_platform:
         type: string
-        default: 'main'
+        default: 'master'
         required: true
       previous_version:
         type: string
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,4 +33,4 @@ jobs:
           if [ -f scripts/requirements.txt ]; then pip install -r scripts/requirements.txt; fi
       - name: execute py script # run file
         run: |
-          python scripts/release.py ${{ github.event.inputs.branch_platform }} ${{ github.event.inputs.previous_version }} ${{ github.event.inputs.new_version }}
+          python scripts/release.py ${{ github.event.inputs.branch_platform }} ${{ github.event.inputs.previous_version }} ${{ github.event.inputs.new_version }} ${{ github.token }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,36 @@
+name: "Release"
+
+on:
+  workflow_call:
+    inputs:
+      branch_platform:
+        type: string
+        default: 'main'
+        required: false
+      previous_version:
+        type: string
+        default: ''
+        required: false
+      new_version:
+        type: string
+        default: ''
+        required: false
+
+jobs:
+  release:
+    name: Release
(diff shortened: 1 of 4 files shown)


RomuDeuxfois:
There are some files unrelated to the release changes.

Dimfacion:
@RomuDeuxfois My bad, forgot to do a fetch before creating the branch. It's better now :)

scripts/release.py
```diff
@@ -0,0 +1,252 @@
...
+)
+time.sleep(1800)
+
+logging.info("[platform] Generating release")
+os.system("gren release > /dev/null 2>&1")
+
+# Modify the release note
+logging.info("[platform] Getting the current release note")
+release = requests.get(
+    "https://api.github.com/repos/OpenBAS-Platform/openbas/releases/latest",
```
antoinemzs, line 153: Most URLs in the script would benefit from being built from a set of constants in this script.
A better but more involved chage would be to pass variables from the build environment (e.g. `${{ github.server_url }}/${{ github.repository }}/`) but that's perhaps overkill for this first version.

scripts/release.py
```diff
@@ -0,0 +1,252 @@
...
+    new_release_note = release_body
+
+logging.info("[platform] Updating the release")
+requests.patch(
+    "https://api.github.com/repos/OpenBAS-Platform/openbas/releases/"
+    + str(release_data["id"]),
+    headers={
+        "Accept": "application/vnd.github+json",
+        "Authorization": "Bearer " + github_token,
+        "X-GitHub-Api-Version": "2022-11-28",
```
antoinemzs, line 193: Headers are always the same: constant ?

antoinemzs (approved):
It looks ok as it is, a few suggestions to reduce the amount of repetition in strings and dicts that are similar

scripts/release.py
```diff
@@ -0,0 +1,256 @@
...
+os.environ["GIT_AUTHOR_NAME"] = "Filigran Automation"
+os.environ["GIT_AUTHOR_EMAIL"] = "automation@filigran.io"
+os.environ["GIT_COMMITTER_NAME"] = "Filigran Automation"
+os.environ["GIT_COMMITTER_EMAIL"] = "automation@filigran.io"
+
+# Platform
+logging.info("[platform] Starting the release")
+
+# Pom
+with open(
```
RomuDeuxfois, line 44: Can we have a loop here based on file name ?
["pom.xml", "./openbas-api/pom.xml", ...] for each
open
write

scripts/requirements.txt
```diff
@@ -0,0 +1,3 @@
+PyYAML
+requests
+git+https://github.com/OpenBAS-Platform/release.git@main
```
RomuDeuxfois, line 3: Master on this repo

Dimfacion: No, we're targeting the main branch of the release repo, it does not have a master branch as far as I know

RomuDeuxfois: My bad

RomuDeuxfois (approved):

scripts/release.py
Dimfacion: Totally valid criticism but this code come from splitting the existing python script into several smaller scripts and we intend to move away from them soon so while it would make sense, I feel we should direct our energy toward removing these scripts rather than make them better. wdyt ?

scripts/release.py
Dimfacion: Same as https://github.com/OpenBAS-Platform/openbas/pull/2438#discussion_r1956405865

.github/workflows/release.yml
```diff
@@ -0,0 +1,36 @@
+name: "Release"
+
+on:
+  workflow_call:
+    inputs:
+      branch_platform:
+        type: string
+        default: 'main'
```
savacano28, line 8: main or master for openbas?

scripts/release.py
antoinemzs: Sure, thanks for the clarification :pray:

PR #2443, opened 2025-02-12
[backend] Clean ExerciseDetails model & query

RomuDeuxfois:
### Proposed changes

* Remove unused front-end properties from the ExerciseDetails model
* Keep all simple properties to keep a coherent model

For later:
* Remove all relationships properties and build dedicated API to fetch them (API & front-end)
example -> simulations/simulationID/lessons_learneds

### Related issues

* https://github.com/OpenBAS-Platform/openbas/issues/2326

### Testing

Create a simulation with a lot of dependencies and go to the overview page.
Every information should be present.


openbas-api/src/main/java/io/openbas/rest/exercise/ExerciseApi.java
```diff
@@ -380,20 +378,20 @@ public Exercise updateExerciseLessons(
...
   @DeleteMapping(EXERCISE_URI + "/{exerciseId}")
   @PreAuthorize("isExercisePlanner(#exerciseId)")
-  @Transactional(rollbackOn = Exception.class)
+  @Transactional(rollbackFor = Exception.class)
   public void deleteExercise(@PathVariable String exerciseId) {
     exerciseRepository.deleteById(exerciseId);
   }
 
   @GetMapping(EXERCISE_URI + "/{exerciseId}")
   @PreAuthorize("isExerciseObserver(#exerciseId)")
```
impolitepanda, line 387: Could we renme this isExerciseObserver function as well, maybe ? (or duplicate it like i did and mark the old one as deprecated?)

openbas-api/src/main/java/io/openbas/rest/exercise/ExerciseApi.java
```diff
@@ -380,20 +378,20 @@ public Exercise updateExerciseLessons(
...
-  @org.springframework.transaction.annotation.Transactional(readOnly = true)
-  public ExerciseDetails exercise(@PathVariable String exerciseId) {
+  @Transactional(readOnly = true)
+  public SimulationDetails exercise(@PathVariable String exerciseId) {
```
impolitepanda, line 389: Parameter name is still exerciseId. Do we wanna change it as well ?

impolitepanda (approved):

openbas-api/src/main/java/io/openbas/rest/exercise/ExerciseApi.java
```diff
@@ -446,59 +444,8 @@ public ExerciseDetails exercise(@PathVariable String exerciseId) {
...
-                                pause.setId(rawPause.getPause_id());
-                                pause.setDuration(rawPause.getPause_duration());
-                                return pause;
-                              })
-                          .toList());
-                  exercise.setCurrentPause(rawExercise.getExercise_pause_date());
-                  inject.setExercise(exercise);
-                  return inject;
-                })
-            .toList();
```
antoinemzs, line 497: I'm concerned this might have an impact in computing execution dates correctly on the front end. Right now I'm not sure these are displayed but there is a risk they might not be correctly computed if a "current pause" is not set.

antoinemzs: It looks fine from testing it, injects are fetched separately and the related simulation is loaded through the repository as epexected (not from a raw query)

antoinemzs (approved):

openbas-api/src/main/java/io/openbas/rest/exercise/ExerciseApi.java
RomuDeuxfois: Good point, I did it :)

PR #2581, opened 2025-03-03
[backend] clean trace from openTelemetry

MarineLeM:
### Proposed changes

*O penTelemetry tracing is no longer needed as it has been replaced by Pyroscope.


Dimfacion:
> Maybe we should also clean Jaeger from docker-compose? @Dimfacion

Yes, this is not relevant anymore

savacano28 (approved):

PR #2607, opened 2025-03-04
[frontend] fix bug in inject update issue/2603

### LINKED ISSUE #2603: Not able to add expectations on injects
## Environment

1. OS (where OpenBAS server runs): pre release / prod

1. Create a scenario
2. Add an inject without expectation and save
3. Update the inject you created by adding an expectation

## Expected Output

The expectation should be added to the inject

## Actual Output

The expectation is not added

heditar:
### Proposed changes

* The "pristine" flag should be false when the details is already open which is the case when updating an inject. This PR will update the "prisitine" flag default value depending on the openDetails.

### Related issues

* Closes #2603 

### Checklist


Dimfacion:
LGTM, thanks !

PR #2613, opened 2025-03-05
[backend] Fix missing parent reference for the asset expectations

### LINKED ISSUE #1829: Manage asset group expectations in the same way as team expectations
## Description

In the current system, the expectations generated for asset groups and their assets do not fill the "asset group" column for every asset expectation. This is different from the behavior with teams, where for each player expectation, the "team" column is filled with the ID of the player's parent team.

If we set the asset group for each asset expectation (if applicable), when an asset group is deleted, we will clean up the associated expectations.

To verify:
* The global score query needs to filter on assets, asset groups, and teams.
* All expectations related to this asset group are deleted when we delete this asset group

Example error generated for this:
* We can not distinguish expectations from an asset/asset group from expectations form an asset (without asset group)
Image

savacano28:
### Proposed changes

* Set asset group for expectations that are part of a assetgroup
* Update atomicTestingUtils (view atomicTesting results) to filter with assetgroup 
* Update TargetResults in order to fetch the correct expectations : UI
* Refact the update of expectations from ui/collectors

! The creation of expectations needs a refactoring, so I created an issue impr technique: https://github.com/OpenBAS-Platform/openbas/issues/2708

image

### Related issues

* Closes #1829 
*

### Checklist

### Further comments


openbas-api/src/main/java/io/openbas/collectors/expectations_expiration_manager/service/ExpectationsExpirationManagerService.java
```diff
@@ -70,6 +71,7 @@ private void computeExpectationsForAssets(@NotNull final List<InjectExpectation>
               this.injectExpectationService.expectationsForAgents(
```
RomuDeuxfois, line 71: **nitpick**: We can rework our signature method and give expectationAsset direclty.
Can be done later.

openbas-api/src/main/java/io/openbas/rest/inject/service/InjectService.java
```diff
@@ -181,19 +181,38 @@ public void deleteAll(List<Inject> injects) {
     }
   }
 
-  public Map<Asset, Boolean> resolveAllAssetsToExecute(@NotNull final Inject inject) {
-    Map<Asset, Boolean> assets =
-        inject.getAssets().stream().collect(Collectors.toMap(asset -> asset, asset -> false));
+  public List<AssetToExecute> resolveAllAssetsToExecute(@NotNull final Inject inject) {
```
RomuDeuxfois, line 184: **question**: Why changing this logic ? Because of the equals method ?
It complicates the code quite a bit to add an additional record.

openbas-api/src/main/java/io/openbas/rest/inject/service/AssetToExecute.java
```diff
@@ -0,0 +1,13 @@
+package io.openbas.rest.inject.service;
+
+import io.openbas.database.model.Asset;
+import io.openbas.database.model.AssetGroup;
+import java.util.ArrayList;
+import java.util.List;
+
+public record AssetToExecute(Asset asset, boolean targetByInject, List<AssetGroup> assetGroups) {
```
RomuDeuxfois, line 8: **question**: what's targetByInject ?

openbas-api/src/main/java/io/openbas/utils/ExpectationUtils.java
```diff
@@ -106,8 +110,185 @@ public static List<InjectExpectation> processByValidationType(
 
   // -- CALDERA EXPECTATIONS --
 
+  public static List<PreventionExpectation> getPreventionExpectations(
```
RomuDeuxfois, line 113: **nitpick:** can be mutualize with the other method getPreventionExpectations ?

openbas-front/src/admin/components/atomic_testings/atomic_testing/TargetResultsDetail.tsx
```diff
@@ -278,7 +280,7 @@ const TargetResultsDetailFlow: FunctionComponent<Props> = ({
           fontSize: 9,
         },
       })));
-      fetchTargetResult(inject.inject_id, target.id!, target.targetType!, parentTargetId).then(
+      fetchTargetResult(inject.inject_id, target.id!, target.targetType!, target.targetType === 'AGENT' ? previousParentTargetId : parentTargetId).then(
```
RomuDeuxfois, line 283: **question**: I don't understand why we have that.

openbas-api/src/main/java/io/openbas/rest/inject/service/InjectService.java
savacano28: Nop, I changed it because now I need to know which asset groups the asset belongs to, so for this reason now I have a record with 2 attributs: 1. asset is directly linked to inject and 2. List asset groups parents. Now, can I know if I have a asset linked to the inject and also an Asset group linked to this inject with this asset. 
For the execution, we just execute one time one asset (n agent so n executions))

openbas-api/src/main/java/io/openbas/rest/inject/service/AssetToExecute.java
savacano28: This boolean indicates whether the asset is directly linked for inject. I wasn't sure about the name either.

openbas-front/src/admin/components/atomic_testings/atomic_testing/TargetResultsDetail.tsx
savacano28: To get the correct expectation, I need to pass the asset group reference as well, so I can pass all the references: asset group, asset, agent, inject, but here I only pass: **Parent, target, inject**, because to filter the expectation by agent or assets, I need mostly the asset group instead of the parent directly...

openbas-api/src/main/java/io/openbas/utils/ExpectationUtils.java
savacano28: I have created an issue for this refact (these methods have slight differences depending on whether it is from Caldera or not) : https://github.com/OpenBAS-Platform/openbas/issues/2708

RomuDeuxfois (approved):

PR #2789, opened 2025-03-27
[frontend]Update inject form visual

### LINKED ISSUE #2523: Spacing is incorrect in inject creation form (and titles should be bold)
## Description

Spacing is incorrect in inject creation form.

Also, think the titles should be bold here, it is not clear enough.

Image

Should have a margin top of 20px here.

johanah29:
Closes #2523

### DIFF
--- a/openbas-front/src/admin/components/common/injects/form/InjectDefinition.js
+++ b/openbas-front/src/admin/components/common/injects/form/InjectDefinition.js
@@ -561,6 +561,7 @@ class InjectDefinition extends Component {
             <Typography
               variant="h5"
               style={{
+                marginTop: 20,
                 fontWeight: 500,
                 float: 'left',
               }}
--- a/openbas-front/src/admin/components/common/injects/form/InjectDefinition.js
+++ b/openbas-front/src/admin/components/common/injects/form/InjectDefinition.js
@@ -440,6 +440,7 @@ class InjectDefinition extends Component {
   render() {
     const {
       t,
+      theme,
       classes,
       setValue,
       values,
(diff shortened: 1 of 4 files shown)


openbas-front/src/admin/components/common/injects/form/InjectDefinition.js
```diff
@@ -561,6 +561,7 @@ class InjectDefinition extends Component {
             <Typography
               variant="h5"
               style={{
+                marginTop: 20,
```
RomuDeuxfois, line 564: **todo:** Don't forget to use theme spacing, is realy important to use theme as much as possible.

openbas-front/src/admin/components/common/injects/form/InjectForm.js
```diff
@@ -63,34 +65,58 @@ class InjectForm extends Component {
...
       <div>
+        <Typography
+          variant="h5"
+          style={{
+            fontWeight: 500,
+            marginTop: theme.spacing(2),
+          }}
+        >
+          {t('Title')}
+        </Typography>
```
MarineLeM, line 76: Are you sure it's what we want ? Because you change completly the behavior or the field Title description and Tags.
In the issue we only speak about the margin of the tags fiels

openbas-front/src/admin/components/common/injects/form/InjectForm.js
```diff
@@ -55,6 +56,7 @@ class InjectForm extends Component {
   render() {
     const {
       t,
+      theme,
```
MarineLeM, line 59: avoid props drilling you can get theme with useTheme

MarineLeM (changes requested):

openbas-front/src/admin/components/common/injects/form/InjectForm.js
johanah29: The declaration of useTheme in this js file results in hooks errors

MarineLeM (approved):

PR #2827, opened 2025-03-31
[backend] Add flyway validator to avoid same version at runtime

RomuDeuxfois:
### Proposed changes

* Crash as soon as possible when we have two migrations with the same versions
* This helps provide a clearer error message on Drone

### DIFF
--- a/openbas-api/src/main/java/io/openbas/tools/FlywayMigrationValidator.java
+++ b/openbas-api/src/main/java/io/openbas/tools/FlywayMigrationValidator.java
@@ -1,0 +1,66 @@
+package io.openbas.tools;
+
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Stream;
+
+public class FlywayMigrationValidator {
+
+  private FlywayMigrationValidator() {
+  }
+
+  private static final Pattern VERSION_PATTERN = Pattern.compile("^V(\\d+(?:_\\d+)*)__.*\\.class$");
+  public static final String OPENBAS_MIGRATION_PATH = "io/openbas/migration";
+
+  public static void validateFlywayMigrationNames() {
+    try {
+      Path migrationDir = Paths.get(
+          FlywayMigrationValidator.class.getProtectionDomain().getCodeSource().getLocation().toURI()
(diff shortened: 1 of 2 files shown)


openbas-api/src/main/java/io/openbas/tools/FlywayMigrationValidator.java
```diff
@@ -0,0 +1,73 @@
...
+            .filter(p -> p.getFileName().toString().startsWith("V"))
+            .filter(p -> p.toString().endsWith(".class"))
+            .filter(p -> !p.getFileName().toString().contains("$")) // Ignore internal classes
+            .forEach(
+                p -> {
+                  String fileName = p.getFileName().toString();
+                  Matcher matcher = VERSION_PATTERN.matcher(fileName);
+                  if (matcher.matches()) {
+                    String version = matcher.group(1).replace('_', '.');
+                    versionToFiles.computeIfAbsent(version, k -> new ArrayList<>()).add(fileName);
```
damgouj, line 52: Maybe you can do this in another way but I don't know if it's better (less code lines): 

List<String> versions = new ArrayList; -> instead of Map versionToFiles

In your files.forEach -> replace versionToFiles.compute by if(versions.contains(version) {throw new IllegalStateException...} else {versions.add(version)}

Delete your versionToFiles.entrySet()....... 

What do you think ?

damgouj (changes requested):

openbas-api/src/main/java/io/openbas/tools/FlywayMigrationValidator.java
RomuDeuxfois: Thats a good point !
What do you think of this

      List<String> versions = new ArrayList<>();

      try (Stream<Path> files = Files.walk(migrationDir)) {
        files
            .filter(Files::isRegularFile)
            .filter(p -> p.getFileName().toString().startsWith("V"))
            .filter(p -> p.toString().endsWith(".class"))
            .filter(p -> !p.getFileName().toString().contains("$")) // Ignore internal classes
            .forEach(
                p -> {
                  String fileName = p.getFileName().toString();
                  Matcher matcher = VERSION_PATTERN.matcher(fileName);
                  if (matcher.matches()) {
                    String version = matcher.group(1).replace('_', '.');
                    if (versions.contains(version)) {
                      throw new IllegalStateException("Duplicate Flyway migration version found: " + version);
                    } else {
                      versions.add(version);
                    }
                  }
                });
      }

damgouj: Great for me !

RomuDeuxfois: I made the changes

damgouj (approved):

PR #2896, opened 2025-04-08
[backend] add attack pattern and killchaine phase to scenario export

### LINKED ISSUE #2016: Kill chain phases & attack patterns are not taken into account in the import/export of scenario/simulation
## Description

When I export or import a simulation/scenario, the kill chain phases & attack patterns linked to any model (injector contract, payload, ect) are not present.

heditar:
### Proposed changes

* Added the KillChain phase and AttackPattern details to the export by adding another attribute containing the list of the whole objects to avoid to modify the existing attribute that only serialize the ids and is used by the front end
*In the import we now verify if the KillChainPhase / Attack Pattern exist and and we create it if it doesn't

For testing:
-> export a scenario containing payload with killchain phases
-> delete the payload and the killchain phases
-> import the scenario and verify that the payload and the killchainphase /attack pattern are created

### Related issues

* Closes #2016

### Checklist

### Further comments

### DIFF
--- a/openbas-api/src/main/java/io/openbas/importer/V1_DataImporter.java
+++ b/openbas-api/src/main/java/io/openbas/importer/V1_DataImporter.java
@@ -55,6 +55,7 @@ public class V1_DataImporter implements Importer {
   private final DocumentRepository documentRepository;
   private final TagRepository tagRepository;
   private final AttackPatternRepository attackPatternRepository;
+  private final KillChainPhaseRepository killChainPhaseRepository;
   private final ExerciseRepository exerciseRepository;
   private final ScenarioService scenarioService;
   private final TeamRepository teamRepository;
@@ -214,10 +215,13 @@ public class V1_DataImporter implements Importer {
                   this.attackPatternRepository.findAllByExternalIdInIgnoreCase(List.of(name));
               if (!existingAttackPattern.isEmpty()) {
                 baseIds.put(id, existingAttackPattern.getFirst());
-                attackPatternIds.add(id);
+                attackPatternIds.add(existingAttackPattern.getFirst().getId());
               } else {
                 AttackPattern attackPatternCreated =
-                    this.attackPatternRepository.save(createAttackPattern(nodeAttackPattern));
+                    this.attackPatternRepository.save(
+                        createAttackPattern(
+                            nodeAttackPattern,
+                            importKillChainPhase(nodeAttackPattern, "attack_pattern_", baseIds)));
                 baseIds.put(id, attackPatternCreated);
                 attackPatternIds.add(attackPatternCreated.getId());
               }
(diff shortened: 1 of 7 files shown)


openbas-model/src/main/java/io/openbas/database/model/Payload.java
```diff
@@ -205,11 +205,17 @@ public enum PAYLOAD_EXECUTION_ARCH {
   @NotNull
   private Instant updatedAt = now();
 
+  @JsonProperty("payload_attack_patterns_details")
```
RomuDeuxfois, line 208: **question:** Why not having the name  payload_attack_patterns here ?

openbas-model/src/main/java/io/openbas/database/model/AttackPattern.java
```diff
@@ -94,6 +94,11 @@ public class AttackPattern implements Base {
   @JsonProperty("attack_pattern_kill_chain_phases")
   private List<KillChainPhase> killChainPhases = new ArrayList<>();
 
+  @JsonProperty("attack_patterns_kill_chain_phases_details")
```
RomuDeuxfois, line 97: question: Why not having the name attack_patterns_kill_chain_phases here ?

openbas-model/src/main/java/io/openbas/database/model/Payload.java
heditar: payload_attack_patterns attributes already exists in Payload. It is serialized/deserialized using MultiIdListDeserializer which only serialize ids. I didn't want to update the existing attribute as it is breaking the front in several places so I added this method that will be serialized in the json with all the information

openbas-model/src/main/java/io/openbas/database/model/AttackPattern.java
heditar: see previous question on the payload_attack_patterns_details attribute

openbas-api/src/main/java/io/openbas/importer/V1_DataImporter.java
```diff
@@ -1058,6 +1064,70 @@ private String importPayload(@NotNull final JsonNode payloadNode, Map<String, Ba
...
+    attackPattern.setDescription(attackPatternNode.get("attack_pattern_description").textValue());
+    attackPattern.setStixId(attackPatternNode.get("attack_pattern_stix_id").textValue());
+    attackPattern.setPermissionsRequired(
+        jsonNodeToStringArray(attackPatternNode.get("attack_pattern_permissions_required")));
+    attackPattern.setPlatforms(
+        jsonNodeToStringArray(attackPatternNode.get("attack_pattern_platforms")));
+    return attackPattern;
+  }
+
+  private List<KillChainPhase> importKillChainPhases(@NotNull final JsonNode killChainPhasesNode) {
```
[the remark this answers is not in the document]
heditar: If we use the exporter it won't be null, just empty

RomuDeuxfois (approved):

PR #2916, opened 2025-04-10
[backend/frontend] Fix pagination on asset group management

RomuDeuxfois:
### Proposed changes

* Fix pagination on asset group management -> Specification **or** return a wrong total element
* Add loader

### DIFF
--- a/openbas-api/src/main/java/io/openbas/service/EndpointService.java
+++ b/openbas-api/src/main/java/io/openbas/service/EndpointService.java
@@ -1,37 +1,5 @@
 package io.openbas.service;
 
-import io.openbas.config.OpenBASConfig;
-import io.openbas.database.model.*;
-import io.openbas.database.repository.*;
-import io.openbas.executors.model.AgentRegisterInput;
-import io.openbas.rest.asset.endpoint.form.EndpointRegisterInput;
-import io.openbas.rest.asset.endpoint.form.EndpointUpdateInput;
-import io.openbas.rest.exception.ElementNotFoundException;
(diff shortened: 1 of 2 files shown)


openbas-api/src/main/java/io/openbas/service/EndpointService.java
```diff
@@ -147,26 +151,40 @@ public Page<Endpoint> searchManagedEndpoints(
...
+                  this.endpointRepository.findAll(specificationStatic.and(specification), pageable),
+              handleEndpointFilter(searchPaginationInput),
+              Endpoint.class);
+      List<Endpoint> mergedContent =
+          Stream.concat(dynamicResult.getContent().stream(), staticResult.getContent().stream())
+              .distinct()
+              .limit(searchPaginationInput.getSize())
+              .collect(Collectors.toList());
+
+      long total = dynamicResult.getTotalElements() + staticResult.getTotalElements();
```
savacano28, line 183: Question: total could have some duplicated values?

RomuDeuxfois: Hum, I think so...
I don't know how to avoir that

I am opened to suggestion

heditar: tu peux pas faire result.stream().distinct() ?

heditar: you can create a merged result without limit , get the total number and then limit the size to return

heditar (changes requested):

PR #2945, opened 2025-04-14
Added bulk create expectation traces endpoint

### LINKED ISSUE #2873: Performance improvement: bulk insert expectation traces and results
## Context

For tests with a large amount of assets, the insertion of traces and the update of results for every expectation can take too long.

## Use case

With a simulation of 500+ assets, results and traces will start to get very slow while inserting them. For 1000+, timeouts start to occur.

## Current Workaround

None

## Proposed Solution

Change expectation traces insert endpoint and expectation result endpoint from a unitary one to a bulk one.
Update collectors to call this new endpoint at the end of every iteration instead of at every match.

impolitepanda:
### Proposed changes

* Added bulk create expectation traces endpoint

### Related issues

* Closes #2873 

### Checklist


openbas-api/src/main/java/io/openbas/rest/inject_expectation_trace/InjectExpectationTraceApi.java
```diff
@@ -43,6 +63,38 @@ public InjectExpectationTrace createInjectExpectationTraceForCollector(
...
+    }
+    this.injectExpectationTraceService.bulkInsertInjectExpectationTraces(
+        inputs.getExpectationTraces());
+    Instant afterSelect = Instant.now();
+    log.warning(
+        "It took "
+            + Duration.between(start, afterSelect).toMillis()
+            + " ms to handle "
+            + inputs.getExpectationTraces().size()
+            + " traces");
```
antoinemzs, line 94: Unsure if this should log a warning everytime the bulk method runs; perhaps convert this to a debug trace.

openbas-api/src/main/java/io/openbas/rest/inject_expectation_trace/form/BulkInjectExpectationTraceInput.java
```diff
@@ -0,0 +1,15 @@
+package io.openbas.rest.inject_expectation_trace.form;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.openbas.config.AppConfig;
+import jakarta.validation.constraints.NotNull;
+import java.util.List;
+import lombok.Data;
+
+@Data
+public class BulkInjectExpectationTraceInput {
```
antoinemzs, line 10: I knwo we don't have a clear naming pattern, but in two similarly named operations, the name of the imput object follows a different pattern. Perhaps make them more similar for the devs' mental map sake?

openbas-api/src/test/java/io/openbas/rest/ExpectationApiTest.java
```diff
@@ -971,6 +972,110 @@ void updateInjectExpectationWithTwoSuccess() throws Exception {
...
+              savedAgent1, savedEndpoint, savedAssetGroup, EXPIRATION_TIME_SIX_HOURS, emptyList());
+
+      injectExpectationService.buildAndSaveInjectExpectations(
+          executableInject,
+          List.of(
+              detectionExpectationForAssetGroup,
+              detectionExpectationForAsset,
+              detectionExpectationAgent,
+              detectionExpectationAgent1,
+              detectionExpectation2Agent1));
```
antoinemzs, line 1013: Are you cleaning the saved entities? The test class is not transactional (I wish it were).

antoinemzs (changes requested):

openbas-api/src/main/java/io/openbas/rest/inject_expectation_trace/InjectExpectationTraceApi.java
impolitepanda: Testing code, i'll remove it, thanks!

openbas-api/src/test/java/io/openbas/rest/ExpectationApiTest.java
impolitepanda: Yes, ther is an afterEach which resets the repository ( injectExpectationRepository.deleteAll() ) as well as an afterAll which resets all repositories used in this test :)

openbas-api/src/main/java/io/openbas/service/InjectExpectationTraceService.java
```diff
@@ -59,4 +69,49 @@ public long getAlertLinksNumber(
...
+        });
+
+    // Dedupe from DB
+    List<SimpleRawExpectationTrace> rawsfromDB1 =
+        this.injectExpectationTraceRepository.findAllTracesNewerThan(
+            oldestAlertDate.get().truncatedTo(ChronoUnit.SECONDS));
+    Set<SimpleRawExpectationTrace> fromDB = new HashSet<>(rawsfromDB1);
+
+    // Removing duplicate traces
+    traces.keySet().removeAll(fromDB);
```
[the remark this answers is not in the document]
antoinemzs: I can't think of something better than a custom insert query with the ON CONFLICT DO NOTHING clause from postgres https://www.postgresql.org/docs/current/sql-insert.html#SQL-ON-CONFLICT

And the conflict would be on the alert_link column.

The JpaRepository does not expose this natively as it's a postgres feature (not standard SQL).

Here are examples as native queries:
* https://www.baeldung.com/spring-data-jpa-update-or-insert#3-using-database-built-in-feature
* https://www.baeldung.com/hibernate-insert-query-on-conflict-clause#2-do-nothing

Maybe worth a thought, especially if we're talking about five orders of magnitude?

The caveat here is that it would not account for alerts missing from subsequent batches (which would be deleted in the current code, not with the design discussed here).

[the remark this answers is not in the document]
impolitepanda: So I implemented what you asked. However, deduplication, on therefore conflict and unique key constraints must be on:
- Alert link for it's unique ID
- Alert Name, because according to my tests some alerts, notably on Defender, have the same global link but have different names depending on the "sub alerts" (not sure how to call it so I'm going with that)
- Security Platform ID, in case several collectors of the same kinds exists (should not happen, but there is nothing preventing having multiple Defender collector, for instance)
- Expectation ID, as several expectation can be filled with the same alert (should be extremely rare, but nonethe less, and is more future proof as well)

openbas-api/src/test/java/io/openbas/rest/InjectExpectationTraceApiTest.java
```diff
@@ -155,12 +156,24 @@ void getInjectExpectationTracesForCollector() throws Exception {
             .getContentAsString();
 
     // --ASSERT--
-    assertEquals(
-        savedInjectExpectationTrace1.getId(),
-        JsonPath.read(response, "$[0].inject_expectation_trace_id"));
-    assertEquals(
-        savedInjectExpectationTrace2.getId(),
+    assertNotEquals(
```
antoinemzs, line 159: Just a note: we have had a json assertions lib for a while now, example usage: https://github.com/OpenBAS-Platform/openbas/blob/7a5e499e7318d2a6ce73583fd165bf3050bac13a/openbas-api/src/test/java/io/openbas/rest/exercise/ExerciseApiExportTest.java#L384-L387

antoinemzs (approved):
Good stuff

good to know we have a schema constraint now :)

openbas-api/src/test/java/io/openbas/rest/InjectExpectationTraceApiTest.java
impolitepanda: Implemented the change, thanks!

PR #2998, opened 2025-04-18
[frontend] add ee highlight

### LINKED ISSUE #2838: Implementation of the EE lock
## Context

EE Licence implementation

## Use case

When I want to launch EE feature I have to agree the EE conditions to be able to launch the feature, if I don't then I can't launch.

## Proposed Solution

Add a lock to the agent's install for the CrowdStrike one by example with a message "You have to accept EE condition to use it", once the user agreed the conditions then he can add a Crowdstrike agent

MarineLeM:
### Proposed changes

on install agents :
-   Add EE highlights
-   Add an informational message in the installation crowdstrike modal

on endpoint page
-   Add EE highlights on crowdstrike

### Related issues

* Closes #2838 

### Further comments
image
image
image


openbas-front/src/admin/components/agents/Agents.tsx
```diff
@@ -84,6 +82,7 @@ const Executors = () => {
             <ExecutorSelector
               executor={executor}
               setSelectedExecutor={setSelectedExecutor}
+              isEEExecutor={executor.executor_type == OPENBAS_TANIUM || executor.executor_type == OPENBAS_CROWDSTRIKE}
```
impolitepanda, line 85: question (for the whole PR): shouldn't we add this as a more generic property somewhere ? Maybe we need to add a isEE propetry in all executors in addition to the executor_type ?

PR #3026, opened 2025-04-23
[backend] Batch insert of execution traces

### LINKED ISSUE #2830: Batching insert of execution traces
## Description

When executing a large number of agents, we will have a lot of calls to the API to send back execution traces. We need to put in place a batch system for inserting those execution traces to relieve the database and to speed up the endpoint to push execution traces.
To do that, the idea would be to be able to push data to deal with into a queue of a predetermined size. Once the queue is full or if we reach a timeout, we process in async the queue and insert the traces in batch.

Dimfacion:
### Proposed changes

* Use RabbitMQ to insert a large number of execution traces with a queue system
*

### Related issues

* Closes #2830
*

### Checklist

### Further comments


openbas-api/src/main/java/io/openbas/rest/helper/queue/BatchQueueService.java
```diff
@@ -0,0 +1,291 @@
...
+                // Ack the message
+                consumerChannel.basicAck(delivery.getEnvelope().getDeliveryTag(), true);
+
+                // Unmarshalling of our object and setting it in the queue for processing
+                T element = mapper.readValue(message, clazz);
+                queue.put(element);
+
+                // If we reach a critical mass, we take care of it immediately
+                if (queue.size() > this.queueConfig.getMaxSize()) {
+                  processBufferedBatch();
```
antoinemzs, line 156: There's a potential data loss if the server crashes or is otherwise shut down while there have been acked messages in `queue`.

I understand there's a balance between absolute safety (comusing the messages 1 by 1) and efficiency (batching several messages together for using bulk hibernate methods later).

Could we not ack any message before the batch is fully processed?

antoinemzs (changes requested):
There might be some changes needed to the ack/nack logic to account for incidents or unexpected exceptions.

Also there are no tests for covering the body of `DeliverCallback deliverCallback`; is it due to mocking?

openbas-api/src/main/java/io/openbas/rest/inject/service/InjectStatusService.java
```diff
@@ -119,66 +122,131 @@ private void computeExecutionTraceStatusIfNeeded(
...
+    // one.
+    for (InjectExecutionCallback injectExecutionCallback : sortedInjectExecutionCallbacks) {
 
-      // -- UPDATE STATUS --
-      updateInjectStatus(agent, inject, input);
+      // Getting the inject or throwing an exception
+      Inject inject = injects.get(injectExecutionCallback.getInjectId());
+      if (inject == null) {
+        log.log(Level.SEVERE, "Inject not found: {}", injectExecutionCallback.getInjectId());
+        break;
```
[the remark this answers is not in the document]
Dimfacion: Must not have been properly woken up when I wrote that :p Good call !

openbas-api/src/main/java/io/openbas/rest/inject/service/InjectStatusService.java
```diff
@@ -119,66 +122,131 @@ private void computeExecutionTraceStatusIfNeeded(
...
+                        sortedInjectExecutionCallbacks.stream()
+                            .map(InjectExecutionCallback::getAgentId)
+                            .filter(Objects::nonNull)
+                            .toList())
+                    .spliterator(),
+                false)
+            .collect(Collectors.toMap(Agent::getId, Function.identity()));
+
+    // Preparing a list of injects we need to save
+    List<Inject> injectsToSave = new ArrayList<>();
```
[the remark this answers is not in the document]
Dimfacion: Good point !

openbas-api/src/main/java/io/openbas/rest/helper/queue/BatchQueueService.java
```diff
@@ -0,0 +1,291 @@
...
+                }
+              } catch (Exception e) {
+                log.error("Error processing message: {}", e.getMessage());
+                // Nack the message and sending it back to the queue
+                consumerChannel.basicNack(delivery.getEnvelope().getDeliveryTag(), false, true);
```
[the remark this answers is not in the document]
Dimfacion: I reworked a bit the ack mechanism so that it happens in the processBufferedBatch. I also reduced the scope of exceptions and it turns out that we only have two exceptions that may happen : one when we have trouble filling the internal queue (=> so we nack here) and one if we have an IOException when trying to ack (=> meaning that we can't nack either)

openbas-api/src/main/java/io/openbas/rest/helper/queue/BatchQueueService.java
```diff
@@ -0,0 +1,307 @@
...
+   */
+  private void processBufferedBatch() {
+    // Draining the queue into the list with a max size
+    List<T> currentBatch = new ArrayList<>();
+    queue.drainTo(currentBatch, queueConfig.getMaxSize());
+
+    // If the list is not empty, we process it
+    if (!currentBatch.isEmpty()) {
+      log.info("Processing batch of {}", currentBatch.size());
+      queueExecution.perform(currentBatch);
```
antoinemzs, line 271: To discuss: should we nack the batch objects if processing it fails? It would leave a chance to accrue forever-failed messages, but would be more resilient to transient issues (eg. unreachable db?)

openbas-api/src/main/java/io/openbas/rest/helper/queue/BatchQueueService.java
```diff
@@ -0,0 +1,307 @@
...
+    try {
+      for (T element : currentBatch) {
+        consumerChannels
+            .get(rand.nextInt(consumerChannels.size()))
+            .basicAck(deliveryTable.remove(element), true);
+      }
+    } catch (IOException e) {
+      log.error(
+          String.format("Error processing batch - Cannot Ack the message: %s", e.getMessage()), e);
+    }
```
antoinemzs, line 284: I think the try-catch should only involve the single ack within the loop; otherwise if the ack fails on the first item, we fail to ack the entire batch.

Also we don't seem to deal with the possibility this would introduce duplication in the inject traces if the message is never acked? Is this an issue? I guess that would be for InjectStatusService to cope with this.

openbas-api/src/main/java/io/openbas/rest/helper/queue/BatchQueueService.java
Dimfacion: In an ideal world, I'd say yes but that would require to change our exception management (that needs to be improved anyway). As of now, I don't think this PR degrades the way it works but I do agree that it'd be better. Would you mind if I create an issue to handle this ?

PR #3033, opened 2025-04-24
[backend] fix TTP inside inject export

### LINKED ISSUE #3000: Inject export TTP link to UUID and not TTP ID
## Description

 it appears that the assigned TTP in the payload is not exported with the appropriate name (such as T1000.001), but rather a UID.

## Environment

Every environments

## Reproducible Steps

1. Go to an inject
2. Righ click export
3. Check the following variable: injector_contract_attack_patterns

## Expected Output

T1003

## Actual Output

ff629a00-103e-4e5d-a2bf-3d043ec5809f

## Screenshots (optional)

Image

Image

MarineLeM:
### Related issues

* Closes #3000


openbas-api/src/main/java/io/openbas/export/FileExportBase.java
```diff
@@ -43,6 +46,9 @@ protected FileExportBase(
     this.objectMapper.addMixIn(Channel.class, Mixins.Channel.class);
     this.objectMapper.addMixIn(Challenge.class, Mixins.Challenge.class);
     this.objectMapper.addMixIn(Tag.class, Mixins.Tag.class);
+    this.objectMapper.addMixIn(InjectorContract.class, InjectorContractMixin.class);
```
damgouj, line 49: You don't put the new classes in the file Mixins like the others, is it normal ? It will allow you to simplify your new classes I think :)

damgouj (changes requested):
Tested OK for the export. When you import again the inject, the attack pattern is not imported while tags are. However, it was the case before your development too.

antoinemzs:
> Tested OK for the export. When you import again the inject, the attack pattern is not imported while tags are. However, it was the case before your development too.

That's probably in scope with issue https://github.com/OpenBAS-Platform/openbas/issues/3000

openbas-api/src/main/java/io/openbas/export/Mixins.java
```diff
@@ -35,6 +38,33 @@ public static class ExerciseFileExport {}
...
+      value = {
+        "attack_pattern_id",
+        "attack_pattern_name",
+        "attack_pattern_description",
+        "attack_pattern_external_id",
+      })
+  public static class AttackPattern {}
+
+  @JsonIgnoreProperties(
+      ignoreUnknown = true,
```
damgouj, line 51: I think you can remove it because it's alreay in the Base class (ignoreUnknown + value) ?

openbas-api/src/main/java/io/openbas/export/Mixins.java
```diff
@@ -35,6 +38,33 @@ public static class ExerciseFileExport {}
...
+      value = {"listened"})
+  public abstract static class InjectorContract {
+
+    @JsonSerialize(using = JsonSerializer.None.class)
+    public abstract List<AttackPattern> getAttackPatterns();
+  }
+
+  @JsonIgnoreProperties(
+      value = {"listened"},
```
damgouj, line 60: Same here :)

damgouj (changes requested):
Sorry but tested KO, I have an error with the 2 injects you gave me last week. Available if you want to test with me or to explain if I do something wrong :).

damgouj (approved):

PR #3063, opened 2025-04-28
[frontend] add name colomn on findings

MarineLeM:
### Proposed changes

* Add column "name" on findings tables

image

### DIFF
--- a/openbas-front/src/admin/components/findings/FindingList.tsx
+++ b/openbas-front/src/admin/components/findings/FindingList.tsx
@@ -67,6 +67,11 @@ const FindingList = ({ searchFindings, filterLocalStorageKey, contextId, additio
       label: 'Type',
       isSortable: true,
       value: (finding: FindingOutput) => finding.finding_type,
+    }, {
+      field: 'finding_name',
+      label: 'Name',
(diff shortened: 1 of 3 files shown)


openbas-front/src/admin/components/findings/FindingList.tsx
```diff
@@ -96,9 +101,10 @@ const FindingList = ({ searchFindings, filterLocalStorageKey, contextId, additio
     ...additionalHeaders,
   ];
 
-  const basis = `${100 / headers.length}%`;
+  const basis = `${90 / (headers.length - 1)}%`;
```
RomuDeuxfois, line 104: **question:** Why not keeping the same logic as before ?

MarineLeM: I tried to reduce the width of the "Type" column to maximize the amount of information visible. However, with so many columns, the inject name is not clearly visible on smaller screens.
image

RomuDeuxfois: Allright, ok for me

RomuDeuxfois (approved):

PR #3067, opened 2025-04-28
[frontend] Store mitre matrix user preference

RomuDeuxfois:
### Proposed changes

* Mitre attack widget - Options show covered TTP only is not keep when page is reload

### DIFF
--- a/openbas-front/src/admin/components/workspaces/custom_dashboards/widgets/viz/SecurityCoverageContent.tsx
+++ b/openbas-front/src/admin/components/workspaces/custom_dashboards/widgets/viz/SecurityCoverageContent.tsx
@@ -1,7 +1,8 @@
 import { Box, Checkbox, FormControlLabel, Typography } from '@mui/material';
 import { useTheme } from '@mui/material/styles';
-import { type FunctionComponent, useState } from 'react';
+import { type FunctionComponent } from 'react';
 import { makeStyles } from 'tss-react/mui';
+import { useLocalStorage } from 'usehooks-ts';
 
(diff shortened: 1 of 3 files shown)


openbas-front/src/admin/components/workspaces/custom_dashboards/widgets/viz/SecurityCoverageContent.tsx
```diff
@@ -61,7 +65,11 @@ const SecurityCoverageContent: FunctionComponent<Props> = ({ data }) => {
   const resolvedDataSuccess = resolvedData(attackPatternMap, killChainPhaseMap, data.at(0)?.data ?? []);
   const resolvedDataFailure = resolvedData(attackPatternMap, killChainPhaseMap, data.at(1)?.data ?? []);
 
-  const [showCoveredOnly, setShowCoveredOnly] = useState(false);
+  const [showCoveredOnly, setShowCoveredOnly] = useLocalStorage<boolean>('widget-' + widgetId, false);
+  const handleShowCoveredOnly = (checked: boolean) => {
+    setShowCoveredOnly(checked);
+  };
```
guillaumejparis, line 71: Nitpick : we can avoid using an util function and calling directly setShowCoveredOnly

guillaumejparis (approved):

openbas-front/src/admin/components/workspaces/custom_dashboards/widgets/viz/SecurityCoverageContent.tsx
RomuDeuxfois: Of course, I made the changes. Thx !

PR #3077, opened 2025-04-29
[backend] import output parser

MarineLeM:
### Proposed changes

* Import output parser


openbas-api/src/main/java/io/openbas/importer/V1_DataImporter.java
```diff
@@ -1069,6 +1072,65 @@ private void importInjects(
     }
   }
 
+  public static ContractOutputType formatStringToContractOutputType(String value) {
+    for (ContractOutputType type : ContractOutputType.values()) {
```
Dimfacion, line 1076: Is there a reason to create this function ? It looks very similar to what 
```suggestion
    ContractOutputType.valueOf(value);
``` would do

Dimfacion (commented):
Aside from the question I left, it seems to be working.

Dimfacion (approved):

PR #3082, opened 2025-04-30
[frontend] CK editor is not dark anymore bugfix/ckeditor

EmilieFo17:
Bugfix for milestone 1.16.0  CK editor is not dark anymore

### DIFF
--- a/openbas-front/src/static/css/CKEditorDark.css
+++ b/openbas-front/src/static/css/CKEditorDark.css
@@ -1,5 +1,7 @@
 html[data-theme="dark"],
-body[data-theme="dark"] {
+body[data-theme="dark"],
+html[data-theme="auto"],
+body[data-theme="auto"]{
   --ck-z-modal: 3000 !important;
   --ck-z-panel: 3000 !important;
 


openbas-front/src/static/css/CKEditorDark.css
```diff
@@ -1,5 +1,7 @@
 html[data-theme="dark"],
-body[data-theme="dark"] {
+body[data-theme="dark"],
```
RomuDeuxfois, line 2: This one should be removed ?

openbas-front/src/static/css/CKEditorDark.css
```diff
@@ -1,5 +1,7 @@
 html[data-theme="dark"],
```
RomuDeuxfois, line 1: This one too

openbas-front/src/static/css/CKEditorDark.css
EmilieFo17: I think it's possible to have dark as a car, but at the moment there seems to be a regression on that, so I can't confirm it.

EmilieFo17:
I'm closing because the problem is linked to the regression on the theme update that @guillaumejparis  is fixing.

EmilieFo17:
I'm closing because the problem is linked to the regression on the theme update that @guillaumejparis  is fixing.

PR #3155, opened 2025-05-13
[backend] Avoiding parallel launch of parent/children

Dimfacion:
### Proposed changes

* Avoiding parallel launch of parent/children
*

### Testing Instructions

1. Create 3 injects and chains them together
2. Using the timeline, make them only a few seconds apart
3. Launch the simulation
4. Make sure they don't start at the same time but one at each new loop

### Related issues

* 
*

### Checklist


antoinemzs (changes requested):
Good job figuring this out! Two things however:
* Could you write a short, plain english comment to explain the logic behind the new filter clause? It looks like it could confuse newcomers.
* How would you go about writing an automated test to recreate the conditions of the bug and validate that it is fixed?

Dimfacion:
@antoinemzs Thanks for pointing that out :) That should be better now

antoinemzs (approved):
Thanks! LGTM

PR #3162, opened 2025-05-13
[backend] Define a logic for saving structured outputs

### LINKED ISSUE #2972: [EE] Improve chaining conditions of execution
## Context

Change the minimal condition of chaining is that the previous inject has ended

I can condition the launch of an inject based on:
- Structured output of an ancestor inject
- The expectations results
- The execution status
- a time to wait before execution

### LINKED ISSUE #3179: Define a logic for saving structured outputs
## Context

We currently store the raw result of an inject's execution in a trace (message). However, there are injects that can generate a result in a structured format, and these results aren't stored. To execute injects in a chained manner, we need to work with these structured results, so the code must be refactored to be able to store them.

## Use case

Use structured output as input for other injects.

savacano28:
### Proposed changes

* Add a new attribute to execution traces to save structured results (calculated from existing output parsers).
* Refact to calculate structured output when an output parser has been defined.
* Refact the compute of findings to use structured output as input instead of raw output.

### Testing Instructions

1. Create a payload with a output parser
Exemple PortScan: (TCP|UDP)\s+([\d\.\[\]:]+):(\d+)\s+([\d\.\[\]:]+):(\d+|\*)\s+(\S+) {host:$2, port: $3, service: $6}
image

3. Execute an inject with this paylaod
4. Verify in DDBB if there is any trace with a value in the column: structured_message (with the output parser structure)
image

5. If you have enabled findings, verify if any find was created

### Related issues

* Closes #3179 

### Checklist

### Further comments


MarineLeM:
Test : ok

Dimfacion:
To help understand what is done in this PR and improve readability, do you think you could add more comments and/or javadocs ?

Dimfacion:
I noticed that when no value is saved in execution_structured_output it's saved as the text 'null' instead of the null value. You can see the difference between value that were saved before the migration and new ones
image

Dimfacion (dismissed):

Dimfacion (approved):

Dimfacion:
LGTM :)

PR #3209, opened 2025-05-20
[backend] Asset group dynamic filter not null

### LINKED ISSUE #3147: Can't create a new atomic test with an asset group automatically created from Crowdstrike
## Description

When I try to create a new atomic test with an asset group automatically created, the attribute asset_group_dynamic_filter is not filled by default so the red message "internal error" is shown and the atomic test is not created.

## Environment

1. OpenBAS version: 1.16.1

## Reproducible Steps

1. Enable the Crowdstrike Executor in your conf
2. Check if the host group from CS is created on OpenBAS asset groups screen
3. Create a new atomic test with a technical payload and the CS asset group
4. Error when we click on validate

## Expected Output

Atomic test is created.

## Actual Output

Atomic test is created.

## Additional information

Suggested fix:
Add a migration to set the attribute "asset_group_dynamic_filter" from the table asset_groups not null and with a default value (like when it's created manually with assets linked and no dynamic filter).

damgouj:
### Proposed changes

* Add default value to dynamic filter for asset group as it's already done when you create an asset group manually

### Testing Instructions

1. Add a Crowdstrike config
2. Create manual asset groups
3. Create injects with the manual asset groups and the Crowdstrike asset group
4. Check it works

### Related issues

* Closes #3147 

### Checklist

### DIFF
--- a/openbas-api/src/main/java/io/openbas/migration/V3_88__Enforce_constraint_asset_group_dynamic_filter.java
+++ b/openbas-api/src/main/java/io/openbas/migration/V3_88__Enforce_constraint_asset_group_dynamic_filter.java
@@ -13,8 +13,8 @@ public class V3_88__Enforce_constraint_asset_group_dynamic_filter extends BaseJa
     try (Statement statement = context.getConnection().createStatement()) {
       statement.execute(
           """
-          ALTER TABLE asset_groups ALTER COLUMN asset_group_dynamic_filter SET DEFAULT '{"mode":"or","filters":[]}';
-          UPDATE asset_groups SET asset_group_dynamic_filter = '{"mode":"or","filters":[]}' where asset_group_dynamic_filter IS NULL;
+          ALTER TABLE asset_groups ALTER COLUMN asset_group_dynamic_filter SET DEFAULT '{"mode":"and","filters":[]}';
+          UPDATE asset_groups SET asset_group_dynamic_filter = '{"mode":"and","filters":[]}' where asset_group_dynamic_filter IS NULL;
           ALTER TABLE asset_groups ALTER COLUMN asset_group_dynamic_filter SET NOT NULL;
           """);
     }
--- a/openbas-api/src/main/java/io/openbas/migration/V3_88__Enforce_constraint_asset_group_dynamic_filter.java
+++ b/openbas-api/src/main/java/io/openbas/migration/V3_88__Enforce_constraint_asset_group_dynamic_filter.java
@@ -1,0 +1,22 @@
+package io.openbas.migration;
+
+import java.sql.Statement;
+import org.flywaydb.core.api.migration.BaseJavaMigration;
+import org.flywaydb.core.api.migration.Context;
+import org.springframework.stereotype.Component;
+
+@Component
+public class V3_88__Enforce_constraint_asset_group_dynamic_filter extends BaseJavaMigration {
+
+  @Override
+  public void migrate(Context context) throws Exception {
+    try (Statement statement = context.getConnection().createStatement()) {
+      statement.execute(
+          """
+          ALTER TABLE asset_groups ALTER COLUMN asset_group_dynamic_filter SET DEFAULT '{"mode":"or","filters":[]}';
+          UPDATE asset_groups SET asset_group_dynamic_filter = '{"mode":"or","filters":[]}' where asset_group_dynamic_filter IS NULL;
+          ALTER TABLE asset_groups ALTER COLUMN asset_group_dynamic_filter SET NOT NULL;
+          """);
+    }
+  }
+}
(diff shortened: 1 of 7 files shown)


openbas-api/src/main/java/io/openbas/migration/V3_88__Enforce_constraint_asset_group_dynamic_filter.java
```diff
@@ -0,0 +1,22 @@
...
+@Component
+public class V3_88__Enforce_constraint_asset_group_dynamic_filter extends BaseJavaMigration {
+
+  @Override
+  public void migrate(Context context) throws Exception {
+    try (Statement statement = context.getConnection().createStatement()) {
+      statement.execute(
+          """
+          ALTER TABLE asset_groups ALTER COLUMN asset_group_dynamic_filter SET DEFAULT '{"mode":"or","filters":[]}';
+          UPDATE asset_groups SET asset_group_dynamic_filter = '{"mode":"or","filters":[]}' where asset_group_dynamic_filter IS NULL;
```
antoinemzs, line 17: What made you choose `or` as the default operator on the Filter Group? That looks inconsistent with current practice that sets default mode as and in various places, see e.g.:
https://github.com/OpenBAS-Platform/openbas/blob/253d56c7e9ae0efd6bbfa1e2f838f9172b43654c/openbas-front/src/admin/components/common/injects/CreateInject.tsx#L157-L164

I suppose we'd better have this as `and`.

openbas-model/src/main/java/io/openbas/database/model/Filters.java
```diff
@@ -36,6 +36,13 @@ public static class FilterGroup {
     @NotNull private FilterMode mode; // Between filters
     private List<Filter> filters = new ArrayList<>();
 
+    public static FilterGroup defaultFilterGroup() {
+      FilterGroup filterGroup = new FilterGroup();
+      filterGroup.setMode(FilterMode.or);
+      filterGroup.setFilters(List.of());
```
impolitepanda, line 42: List.of here might be problematic as it creates an immutable list. Someone could want to create a default filter group and then add filters to it. This would wause a runtime error in this case. Maybe new ArrayList<>() instead ?

openbas-model/src/main/java/io/openbas/database/model/Filters.java
```diff
@@ -36,6 +36,13 @@ public static class FilterGroup {
     @NotNull private FilterMode mode; // Between filters
     private List<Filter> filters = new ArrayList<>();
 
+    public static FilterGroup defaultFilterGroup() {
+      FilterGroup filterGroup = new FilterGroup();
+      filterGroup.setMode(FilterMode.or);
```
antoinemzs, line 41: Mode should probably be `and`.

impolitepanda (approved):

openbas-api/src/test/java/io/openbas/rest/asset_group/AssetGroupApiTest.java
```diff
@@ -73,7 +74,81 @@ void given_validAssetGroupInput_should_createAssetGroupSuccessfully() throws Exc
...
+
+    // --ASSERT--
+    assertEquals(assetGroupInput.getName(), JsonPath.read(response2, "$.asset_group_name"));
+    assertEquals(
+        assetGroupInput.getDescription(), JsonPath.read(response2, "$.asset_group_description"));
+    // Check default because null in the input
+    assertEquals(
+        "{mode=or, filters=[]}",
+        JsonPath.read(response2, "$.asset_group_dynamic_filter").toString());
+    assertEquals(tag.getId(), JsonPath.read(response2, "$.asset_group_tags[0]"));
```
antoinemzs, line 102: `net.javacrumbs.jsonunit.assertj.JsonAssertions.assertThatJson` is more legible, would you consider using this lib instead?

openbas-model/src/main/java/io/openbas/database/model/Filters.java
```diff
@@ -36,6 +36,13 @@ public static class FilterGroup {
     @NotNull private FilterMode mode; // Between filters
     private List<Filter> filters = new ArrayList<>();
 
+    public static FilterGroup defaultFilterGroup() {
```
antoinemzs, line 39: There's a risk of inconsistency if this changes and the DB default doesn't.
Please do 2 things:
* leave a comment to hint at a developer to keep both in sync
* write a test that creates two asset groups: one through the entity (gets the default from this method), one through the entity but with the dynamicFilters field nulled (should get the default from the db); and then retrieve + compare both entities dynamicFilters (should be equals)

antoinemzs (changes requested):
Good job, thanks, however there are two levels of defaults and we need to assert that both are in sync, see comments :+1:

antoinemzs (approved):
Excellent thanks!

PR #3239, opened 2025-05-23
[backend] Only show findings of latest simulation for global, scenario, endpoint findings view (#3076)

### LINKED ISSUE #3076: Don't create a finding each time a scenario is run
## Context
When you run a scenario in OpenBAS, a new finding is generated each time the condition is matched—even if the same finding already exists for the same asset and inject (e.g., an open port like SSH on port 22). This leads to multiple redundant findings for the same problem across runs.

## Use case
We want to track issues like an open port (e.g., SSH on port 22) for a specific asset (e.g., an IP address). Instead of creating a new finding each time the scenario is run and the issue is still present, it would be more useful to update the existing finding to reflect that the issue persists. This allows for better historical tracking and avoids cluttering the findings list with duplicates.

## Proposed Solution
Update OpenBAS to:

Identify existing findings (based on asset, scenario) when a scenario is run.

Update the timestamp or metadata of the existing finding if the issue is still present, instead of creating a new one.

Close or expire the finding if it is no longer detected in subsequent scenario runs.

This would allow findings to represent the current state of an issue (present or resolved), rather than creating a new item for each scenario execution.

## Additional Information
This approach improves finding lifecycle management and aligns with the idea that a finding should represent the current state of an issue, not a historical log of every detection.

antoinemzs:
### Proposed changes

* Only show findings from the latest FINISHED simulation of any given scenario, in Global, Scenario and Endpoint fidings views
* [drive by change] optimise some queries for target search

### Testing Instructions

1. Create a scenario and equip it with an inject that will generate findings (nmap is a good candidate); preferrably an inject that will assign findings to an endpoint too
2. Run several simulations of the same scenario
3. In parallel, run a few atomic testings generating some findings, also preferrably assigning some to the same endpoint than the scenario inject.
4. Wait for simulations and atomics to end
5. Browse to global finding search: only latest finished simulation for each scenario are shown; additionally the findings from the atomic testings are also there
6. Browse to the Scenario findings: only the findings of the latest finished simulation of that scenario are shown
7. Browse to the Endpoint details page: only findings attached to that Endpoint from the latest finished simulation of each scenario (if relevant) are shown; also, the findings of any atomic testing attached to the Endpoint are shown

### Related issues

* Closes #3076

### Checklist

### Further comments

Note as to why there is a "launchOrder": this is to strictly sequentially order simulations; since simulations start on the minute, launching several simulations in a short time risks having them start at the (almost) exact same time.

Note that the exercise_start_date column has a resolution of 0 (meaning precise to the entire second), which is an almost guaranteed collision in the scenario described above.

Changing this resolution to eg the microsecond (the most precise possible in pg), overlooking possible side effects on various locations of the codebase, does not in effect guarantee the absence of collision; in fact I could already create a collision with launching a few simulations in a row.

Using a strictly increasing sequence for marking the "launch order" gets rid of that issue.

### DIFF
--- a/openbas-api/src/main/java/io/openbas/migration/V3_90__add_automatic_finding_hash_column.java
+++ b/openbas-api/src/main/java/io/openbas/migration/V3_90__add_automatic_finding_hash_column.java
@@ -1,0 +1,92 @@
+package io.openbas.migration;
+
+import org.flywaydb.core.api.migration.BaseJavaMigration;
+import org.flywaydb.core.api.migration.Context;
+import org.springframework.stereotype.Component;
+
+import java.sql.Statement;
+
+@Component
+public class V3_90__add_automatic_finding_hash_column extends BaseJavaMigration {
+
+  @Override
+  public void migrate(Context context) throws Exception {
+    try (Statement statement = context.getConnection().createStatement()) {
+
+      String updateInjectExpectationResult =
+          """
+          -- necessary for hashing functions
+          CREATE EXTENSION IF NOT EXISTS pgcrypto;
+          
+          CREATE OR REPLACE FUNCTION compute_finding_identity(id TEXT)
+          RETURNS TEXT
+          AS
+          $$
+            BEGIN
+              RETURN (
+                SELECT digest(
+                  concat(
+                      finding_field,
+                      finding_type,
+                      finding_type,
+                      finding_value,
+                      array_to_string(finding_labels, ''),
+                      finding_name,
+                      i.inject_content,
+                      i.inject_injector_contract,
+                      -- to force differentiating between identical atomic tests
+                      -- use an ad hoc random uuid instead of a scenario id.
+                      -- also assume a single scenario per simulation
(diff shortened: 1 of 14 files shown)


impolitepanda (approved):

openbas-api/src/main/java/io/openbas/migration/V3_90__add_automatic_finding_hash_column.java
```diff
@@ -0,0 +1,63 @@
...
+public class V3_90__add_automatic_finding_hash_column extends BaseJavaMigration {
+
+  @Override
+  public void migrate(Context context) throws Exception {
+    try (Statement statement = context.getConnection().createStatement()) {
+
+      String updateInjectExpectationResult =
+          """
+          -- necessary for hashing functions
+          CREATE SEQUENCE IF NOT EXISTS exercise_launch_order_seq
```
savacano28, line 18: Maybe we should add an index on this column?

savacano28:
Code OK: 
Juste maybe we should add an index on the new column?

Tested OK: :)
image
image
image

PR #3255, opened 2025-05-28
[frontend] update widget land you on the parameters stepper

### LINKED ISSUE #3091: Clicking on "Update widget" should always land you on the parameters
## Description

When I update a widget I can edit the parameter but if I go back to dimension I can’t edit the parameter again, if I get out of the modal and come back I still can’t access the parameter without validating the precedent steps. 

When I update and go back to the first step and cancel, if I re update I should be on the last step and not the first one.

Video are available on notion.

MarineLeM:
### Related issues

* Closes #3091


openbas-front/src/admin/components/workspaces/custom_dashboards/widgets/WidgetForm.tsx
```diff
@@ -135,16 +135,21 @@ const WidgetForm: FunctionComponent<Props> = ({
 
   const isLastStep = () => activeStep === steps.length - 1;
 
+  const onClose = () => {
+    setActiveStep(editing ? 2 : 0);
```
savacano28: Maybe add a cons like:  STEP_PARAMETER = 2 (magic number)

PR #3284, opened 2025-06-04
[backend] remove injects serialisation from asset group and team

antoinemzs:
### Proposed changes

* Fis perf issue introduced by serialising injects for Team and AssetGroup entities

### Testing Instructions

1. Snappy endpoints !

### DIFF
--- a/openbas-model/src/main/java/io/openbas/database/model/AssetGroup.java
+++ b/openbas-model/src/main/java/io/openbas/database/model/AssetGroup.java
@@ -3,6 +3,7 @@ package io.openbas.database.model;
 import static java.time.Instant.now;
 import static lombok.AccessLevel.NONE;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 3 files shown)


openbas-model/src/main/java/io/openbas/database/model/AssetGroup.java
```diff
@@ -108,8 +109,8 @@ public List<Asset> getDynamicAssets() {
       name = "injects_asset_groups",
       joinColumns = @JoinColumn(name = "asset_group_id"),
       inverseJoinColumns = @JoinColumn(name = "inject_id"))
-  @JsonSerialize(using = MultiIdListDeserializer.class)
   @JsonProperty("asset_group_injects")
```
savacano28, line 112: We should remove also jsonProperty? re generate api-types? and check if any page use this attribute (store)?

antoinemzs: No we need to define a "json name" for the search functionality. Yes I missed the api types ;) thanks

guillaumejparis (approved):

PR #3286, opened 2025-06-04
[backend] fix empty group case

MarineLeM:
Link to issue 2449

Fix : error when the group was empty


openbas-api/src/main/java/io/openbas/rest/inject/service/InjectAssistantService.java
```diff
@@ -452,6 +461,7 @@ private void handleAssetGroups(
    * @param injectNumberByTTP the maximum number of injects to create for each TTP
    * @return a list of generated injects
    */
+  @SneakyThrows
```
Dimfacion: **Question :** Why the SneakyThrows instead of adding it to the signature of the method ?

Issue #375, opened 2023-10-26
RBAC improvement 


nino-filigran:
## Use case

Ability to ensure that a specific team does not have access to other teams exercises. 
## Proposed Solution

RBAC solution to limit access to some teams only. 
## If the feature request is approved, would you be willing to submit a PR?

Yes / No (Help can be provided if you need assistance submitting a PR)

PR #3341, opened 2025-06-11
[backend] adding role/capa apis 

heditar:
### Proposed changes

* Adding Role and capa table+apis
* Adding the role to the user groups
* Adding the method to pull the capa In User so it is visible in the /me output

### Testing Instructions

1. Play with the /api/roles apis
2. Call /api/me to see the user's capa

### Related issues

*#375
*

### Checklist

- [ ] I added/update the relevant documentation (either on github or on notion) -> will be done in a separate task

### Further comments


openbas-api/src/main/java/io/openbas/rest/role/RoleApi.java
```diff
@@ -0,0 +1,116 @@
...
+    this.roleMapper = roleMapper;
+  }
+
+  @LogExecutionTime
+  @GetMapping(RoleApi.ROLE_URI + "/{roleId}")
+  @Operation(description = "Get Role by Id", summary = "Get Role")
+  @ApiResponses(value = {@ApiResponse(responseCode = "200", description = "The Role")})
+  public RoleOutput findRole(
+      @PathVariable @NotBlank @Schema(description = "ID of the role") final String roleId) {
+    return roleService.findById(roleId).map(roleMapper::toRoleOutput).orElse(null);
```
antoinemzs, line 47: It looks like this should return 404? Retruning a 200 code with a null body looks like it makes the API harder to use than necessary.

openbas-api/src/main/java/io/openbas/rest/role/RoleApi.java
```diff
@@ -0,0 +1,116 @@
...
+import jakarta.validation.Valid;
+import jakarta.validation.constraints.NotBlank;
+import java.util.List;
+import org.springframework.data.domain.Page;
+import org.springframework.security.access.annotation.Secured;
+import org.springframework.transaction.annotation.Transactional;
+import org.springframework.web.bind.annotation.*;
+
+@RestController
+@UserRoleDescription
```
antoinemzs, line 26: nitpick: There's a potential inconsistency here, as the `io.openbas.rest.helper.UserRoleDescriptionCustomizer` class is referencing the old semantics for what a required "User Role" is. Have we considered rewording at some point during the transition?

openbas-api/src/main/java/io/openbas/rest/role/RoleApi.java
```diff
@@ -0,0 +1,116 @@
...
+      @PathVariable @NotBlank @Schema(description = "ID of the role") final String roleId) {
+    roleService.deleteRole(roleId);
+  }
+
+  @Secured(ROLE_ADMIN)
+  @LogExecutionTime
+  @PostMapping(RoleApi.ROLE_URI)
+  @Transactional(rollbackFor = Exception.class)
+  @Operation(summary = "Create Role")
+  @ApiResponses(value = {@ApiResponse(responseCode = "200", description = "Role created")})
```
antoinemzs, line 78: maybe document also HTTP 409 CONFLICT for when the role already exists; I believe the framework automatically maps a datbase-level constraint violation exception to a 409 status code in the rest handler.

openbas-api/src/main/java/io/openbas/service/RoleService.java
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  private Set<Capability> getCapabilities(@NotNull final Set<String> capabilities) {
+    return capabilities.stream()
+        .map(
+            value -> {
+              try {
+                return Capability.valueOf(value);
+              } catch (IllegalArgumentException e) {
+                throw new UnsupportedOperationException("Invalid capability: " + value);
```
antoinemzs, line 78: nitpick: Why is `UnsupportedOperationException` better suited here than `IllegalArgumentException`? The try-ctach does not seem to serve a purpose other than swapping the exception type.

Perhaps consider chaining the original exception as cause for the new one though.

openbas-model/src/main/java/io/openbas/database/model/Group.java
```diff
@@ -92,6 +90,16 @@ public class Group implements Base {
   @JsonProperty("group_organizations")
   private List<Organization> organizations = new ArrayList<>();
 
+  @ArraySchema(schema = @Schema(type = "string"))
+  @ManyToMany(fetch = FetchType.EAGER)
```
antoinemzs, line 94: Consider whether it is necessary to use an eagerly loaded relationship here.

openbas-api/src/main/java/io/openbas/rest/role/RoleApi.java
heditar: Yes it's conflicting with the new logic but a task will take care off all that with the new annotation. For now I will update that to make it only admin access

openbas-api/src/main/java/io/openbas/rest/role/RoleApi.java
heditar: will be done in a next PR

openbas-api/src/main/java/io/openbas/service/RoleService.java
heditar: will be done in a next PR

openbas-api/src/main/java/io/openbas/rest/role/RoleApi.java
heditar: will be done in a next PR

PR #3351, opened 2025-06-12
[frontend] Fix dashboard layout update Issue/3090

### LINKED ISSUE #3090: Add a new widget on an existing dashboard lead to a reorganisation of all widgets
## Description

I don't have a specific workflow to reproduce it...
But sometines, if I have two widgets on a same line and I add a new widget.
The two first widgets are reorganized differently.

heditar:
### Proposed changes

* Setting dashboard state when the layout is updated to fix the issue when a widget is created it overrides previous layout modification

### Testing Instructions

1. Create a widget in a dashboard
2. Resize the widget
3. Without reloading the page create a new widget
4. Expected the widget resized doesn't go back to its previous size

### Related issues

* Closes #3090

### Checklist

### DIFF
--- a/openbas-front/src/admin/components/workspaces/custom_dashboards/CustomDashboard.tsx
+++ b/openbas-front/src/admin/components/workspaces/custom_dashboards/CustomDashboard.tsx
@@ -78,6 +78,21 @@ const CustomDashboardComponent: FunctionComponent<{ customDashboard: CustomDashb
         }),
       ),
     );
+    setCustomDashboardValue(prev => prev && {
+      ...prev,
+      custom_dashboard_widgets: prev.custom_dashboard_widgets?.map(widget => {
+        const existingLayout = layouts.find(x => x.i === widget.widget_id)!;
+        return {
+          ...widget,
+          widget_layout: {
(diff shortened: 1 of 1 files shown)


openbas-front/src/admin/components/workspaces/custom_dashboards/CustomDashboard.tsx
```diff
@@ -78,6 +78,21 @@ const CustomDashboardComponent: FunctionComponent<{ customDashboard: CustomDashb
         }),
       ),
     );
+    setCustomDashboardValue(prev => prev && {
+      ...prev,
+      custom_dashboard_widgets: prev.custom_dashboard_widgets?.map((widget) => {
+        const existingLayout = layouts.find(x => x.i === widget.widget_id)!;
```
EmilieFo17, line 84: A fallback can be provided if find fails, for example:
const existingLayout = layouts.find(x => x.i === widget.widget_id);
if (!existingLayout) return widget;

Otherwise, the rest is “ok” for me.

heditar: thanks I will add it

PR #3390, opened 2025-06-18
[frontend/backend] add asset payload argument

MarineLeM:
### Proposed changes

* Chunk1 https://github.com/OpenBAS-Platform/openbas/issues/3327

Front

- [x]  add target asset argument for the command payload form
- [x]  update inject form

Back

- [x]  Add in the injectorContract the field for the target assets
- [x]  replace in the command by the asset information 
- [x]  TUs

### DIFF
--- a/openbas-api/src/test/java/io/openbas/rest/inject/InjectApiTest.java
+++ b/openbas-api/src/test/java/io/openbas/rest/inject/InjectApiTest.java
@@ -633,7 +633,7 @@ class InjectApiTest extends IntegrationTest {
       payloadArguments.put(
           "asset-separate-by-coma", List.of(endpoint1Saved.getId(), endpoint2Saved.getId()));
       payloadArguments.put(
-          CONTRACT_ELEMENT_CONTENT_KEY_TARGETED_PROPERTY + "-asset-separate-by-space", "local-ip");
+          CONTRACT_ELEMENT_CONTENT_KEY_TARGETED_PROPERTY + "-asset-separate-by-space", "local_ip");
       payloadArguments.put(
           CONTRACT_ELEMENT_CONTENT_KEY_TARGETED_ASSET_SEPARATOR + "-asset-separate-by-space", " ");
 


openbas-api/src/main/java/io/openbas/rest/payload/service/PayloadService.java
```diff
@@ -115,6 +105,31 @@ private ContractChoiceInformation obfuscatorField() {
         "obfuscator", "Obfuscators", obfuscationInfo, obfuscationMap.getDefaultObfuscator());
   }
 
+  private List<ContractElement> targetedAssetFields(String key, PayloadArgument payloadArgument) {
+    ContractElement targetedAssetField = new ContractTargetedAsset(key, key);
+
+    Map<String, String> targetPropertySelectorMap = new HashMap<>();
+    targetPropertySelectorMap.put("hostname", "Hostname");
+    targetPropertySelectorMap.put("seen_ip", "Seen IP");
+    targetPropertySelectorMap.put("local_ip", "Local IP (first)");
```
RomuDeuxfois: **nitpick**: can be an enum and loop on it

openbas-api/src/main/java/io/openbas/rest/payload/service/PayloadService.java
```diff
@@ -145,11 +159,19 @@ private Contract buildContract(
           .getArguments()
           .forEach(
               payloadArgument -> {
-                builder.mandatory(
-                    textField(
-                        payloadArgument.getKey(),
-                        payloadArgument.getKey(),
-                        payloadArgument.getDefaultValue()));
+                if (ContractFieldType.Text.label.equals(payloadArgument.getType())) {
```
RomuDeuxfois, line 163: **question:** All text arguments defined in a payload is mandatory ? Why ?

openbas-front/src/admin/components/common/injects/form/InjectContentFieldComponent.tsx
```diff
@@ -30,7 +32,11 @@ export type InjectField = {
     key: string;
     label: string;
   }[];
-  type: 'textarea' | 'text' | 'select' | 'number' | 'checkbox' | 'dependency-select' | 'choice' | 'tags';
+  type: 'textarea' | 'text' | 'select' | 'number' | 'checkbox' | 'dependency-select' | 'choice' | 'tags' | 'targeted-asset';
+  linkedFields?: {
```
RomuDeuxfois, line 36: **nitpick:**Can you add java doc on linkedFields property to explain how we should use it ? Thank you

openbas-front/src/admin/components/common/injects/form/InjectContentFieldComponent.tsx
```diff
@@ -126,36 +131,64 @@ const InjectContentFieldComponent = ({
         } else {
           choices = field.choices as ChoiceItem[];
         }
+        const label = field.label?.toLowerCase().includes('targeted property')
+          ? `${field.label} - ${(field.key?.split('.')?.at(-1) ?? '').replace('targeted-property-', '')}`
+          : field.label;
```
RomuDeuxfois: **question:** Why not doing it on back-end side ?

openbas-front/src/admin/components/common/injects/form/InjectContentFieldComponent.tsx
```diff
@@ -126,36 +131,64 @@ const InjectContentFieldComponent = ({
...
+            />
+          </>
+        );
+      default:{
+        const label = field.linkedFields?.length && field.linkedFields?.length > 0
+          ? `${field.label} - ${field.linkedFields.map(f => f.key).join(', ')}`
+          : field.label;
+        if (field.key.includes('targeted-asset-separator')) {
+          return (
+            <SeparatorFieldController
```
RomuDeuxfois, line 163: **question:** I think this should be a choice field, and choices should be defined on back-end. What do you think ?

openbas-api/src/main/java/io/openbas/rest/payload/service/PayloadService.java
MarineLeM: it was already like that before, but moreover the field is crucial for replacing values in the command; if it's not defined, the command cannot execute.

openbas-front/src/admin/components/common/injects/form/InjectContentFieldComponent.tsx
MarineLeM: It's a free-solo autocomplete field, not a simple selector, allowing users to input any text. This flexibility is why its type is set to text.

openbas-front/src/admin/components/common/injects/form/InjectContentFieldComponent.tsx
MarineLeM: I cannot set it in the backend because the label is dynamically generated as 'Targeted property' + the targeted asset key, resulting in a unique label each time. This makes translation impossible since the label is a constant string.

RomuDeuxfois (approved):

PR #3393, opened 2025-06-18
Dash/list

antoinemzs:
### Proposed changes

*
*

### Testing Instructions

1. Step-by-step how to test

### Related issues

* Closes #ISSUE-NUMBER
*

### Checklist

### Further comments


openbas-front/src/admin/components/workspaces/custom_dashboards/widgets/viz/list/WidgetColumnsCustomizationInput.tsx
```diff
@@ -0,0 +1,184 @@
...
+                              height: 42,
+                            }}
+                          >
+                            <ListItemIcon {...providedDrag.dragHandleProps}>
+                              <DragIndicatorOutlined />
+                            </ListItemIcon>
+
+                            <ListItemText primary={formatColumnName(column)} />
+
+                            <ListItemSecondaryAction>
```
MarineLeM: It's deprecated, should be in secondaryAction props from ListItem

openbas-front/src/components/common/Accordion.jsx
```diff
@@ -0,0 +1,22 @@
+import { ArrowForwardIosSharp } from '@mui/icons-material';
```
MarineLeM, line 1: Why this component is in JS and not TS ?

openbas-front/src/admin/components/common/endpoints/fragments/elastic/AssetNameFragment.tsx
```diff
@@ -0,0 +1,9 @@
+import { type EsEndpoint } from '../../../../../../utils/api-types';
+
+type Props = { endpoint: EsEndpoint };
+
+const AssetNameFragment = (props: Props) => {
+  return props.endpoint.endpoint_name;
```
MarineLeM: Why creating a tsx component to only return the endpoint_name ?

openbas-front/src/admin/components/common/endpoints/fragments/elastic/AssetTypeFragment.tsx
```diff
@@ -0,0 +1,31 @@
...
+type Props = { endpoint: EsEndpoint };
+
+const AssetTypeFragment = (props: Props) => {
+  const useStyles = makeStyles()(() => ({
+    typeChip: {
+      height: 20,
+      borderRadius: 4,
+      textTransform: 'uppercase',
+      width: 100,
+      marginBottom: 5,
```
MarineLeM: Should use theme.spacing to define margin

openbas-front/src/admin/components/common/endpoints/fragments/elastic/AssetPlatformFragment.tsx
```diff
@@ -0,0 +1,24 @@
+import { useTheme } from '@mui/material/styles';
+
+import { useFormatter } from '../../../../../../components/i18n';
+import PlatformIcon from '../../../../../../components/PlatformIcon';
+import { type EsEndpoint } from '../../../../../../utils/api-types';
+
+type Props = { endpoint: EsEndpoint };
+
+const AssetPlatformFragment = (props: Props) => {
```
MarineLeM, line 9: It's a duplicated component from the AssetPlatformFragment of the output folder 
You have to remove one of it

openbas-front/src/admin/components/common/endpoints/fragments/elastic/AssetPlatformFragment.tsx
```diff
@@ -0,0 +1,24 @@
+import { useTheme } from '@mui/material/styles';
+
+import { useFormatter } from '../../../../../../components/i18n';
+import PlatformIcon from '../../../../../../components/PlatformIcon';
+import { type EsEndpoint } from '../../../../../../utils/api-types';
```
MarineLeM, line 5: Actually all the components from your elastic folder are duplicated components of the output folder

openbas-front/src/admin/components/common/endpoints/fragments/output/EndpointAgentsExecutorsFragment.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+const EndpointActiveFragment = (props: Props) => {
+  const dispatch = useAppDispatch();
+  // Fetching data
+  const { executorsMap } = useHelper((helper: ExecutorHelper) => ({ executorsMap: helper.getExecutorsMap() }));
+  useDataLoader(() => {
+    dispatch(fetchExecutors());
+  });
+  const { t } = useFormatter();
+  if (props.endpoint.asset_agents.length > 0) {
```
MarineLeM: for better understanding and readability, an early return it's preferable in this case

openbas-front/src/components/common/queryable/style/style.ts
```diff
@@ -10,7 +10,8 @@ const useBodyItemsStyles: () => {
   return ({
     bodyItems: {
       display: 'flex',
-      flexWrap: 'wrap',
+      overflow: 'scroll',
```
MarineLeM: please remove it, because it add scroll in every list
image

openbas-front/src/admin/components/workspaces/custom_dashboards/widgets/viz/list/elements/endpoint/EndpointListElement.tsx
```diff
@@ -0,0 +1,87 @@
...
+      default: return (endpoint: EsEndpoint) => {
+        const key = column as keyof typeof endpoint;
+        return endpoint[key];
+      };
+    }
+  };
+  /* eslint-enable react/display-name */
+
+  return (
+    <>
```
MarineLeM: <> => this tag is useless in this case

openbas-front/src/components/common/Accordion.jsx
antoinemzs: Stolen code from OpenCTI :shushing_face: Actually, typescript is giving me a hard time with this code, and I haven't spent too much time debugging it.

openbas-front/src/admin/components/workspaces/custom_dashboards/widgets/viz/list/WidgetColumnsCustomizationInput.tsx
antoinemzs: I ported this code verbatim from OpenCTI, I'll admit. Perhaps it could be improved further.

openbas-front/src/admin/components/common/endpoints/fragments/elastic/AssetPlatformFragment.tsx
antoinemzs: Yes, because they don't use the same input type in props, either `EndpointOutput` or `EsEndpoint`. Perhaps we ought to use an anonymous type with a basic contract and convert the output types before passing them.

openbas-front/src/admin/components/common/endpoints/fragments/output/EndpointAgentsExecutorsFragment.tsx
antoinemzs: That's an existing problem with the ported code, but I'll see to improve it, thanks

openbas-front/src/admin/components/common/endpoints/fragments/elastic/AssetPlatformFragment.tsx
antoinemzs: I have another chunk coming up right after this one and another, I propose to fix that then. Is it OK with you?

openbas-front/src/admin/components/common/endpoints/fragments/elastic/AssetNameFragment.tsx
antoinemzs: I want to separate the concerns from where it's being called, but I guess I could just let the call site fall back to the default component.

openbas-front/src/admin/components/common/endpoints/fragments/elastic/AssetPlatformFragment.tsx
antoinemzs: See other reply.

openbas-api/src/main/java/io/openbas/scheduler/jobs/ElasticSyncExecutionJob.java
```diff
@@ -11,18 +16,23 @@
 
 @Component
 @RequiredArgsConstructor
+@Slf4j
 public class ElasticSyncExecutionJob implements Job {
 
   private EsService esService;
+  private EsEngine esEngine;
 
   @Autowired
```
impolitepanda: this setter should be removable without issue as we have a RequiredArgsConstructor. esService and esEngine just need to be final as well.

openbas-api/src/test/java/io/openbas/rest/dashboard/DashboardApiTest.java
```diff
@@ -0,0 +1,137 @@
...
+              .persist()
+              .get();
+
+      // force persistence
+      entityManager.flush();
+      entityManager.clear();
+      esService.bulkProcessing(esEngine.getModels().stream());
+      // elastic needs to process the data; it does so async, so the method above
+      // completes before the data is available in the system
+      Thread.sleep(1000);
```
impolitepanda, line 77: Instead of using this kind of mechanism, couldn't we change the signature of `public <T extends EsBase> void bulkParallelProcessing()` to `public <T extends EsBase> CompletableFuture<Void> bulkParallelProcessing()` and use the future mechanism ?

openbas-dev/docker-compose.yml
```diff
@@ -104,6 +103,27 @@ services:
...
       - 9300:9300
+  openbas-test-elasticsearch:
+    container_name: openbas-test-elasticsearch
+    image: docker.elastic.co/elasticsearch/elasticsearch:8.18.2
+    environment:
+      - discovery.type=single-node
+      - xpack.ml.enabled=false
+      - xpack.security.enabled=false
+      # -XX:UseSVE=0 is necessary for Apple M4 architecture
+      - "ES_JAVA_OPTS=-Xms2G -Xmx2G -XX:+IgnoreUnrecognizedVMOptions -XX:UseSVE=0"
```
impolitepanda, line 114: As this is for test only, could me co bellow 2G, maybe 512 could be enough ? (no idea about actual requirements)

openbas-front/src/admin/components/workspaces/custom_dashboards/widgets/viz/list/elements/default/DefaultListElement.tsx
```diff
@@ -0,0 +1,67 @@
...
+    switch (column) {
+      default: return (element: EsBase) => {
+        const key = column as keyof typeof element;
+        return element[key];
+      };
+    }
+  };
+
+  return (
+    <>
```
MarineLeM: <> => is useless in this case

MarineLeM:
Too much data causes the component to overflow.:
image

openbas-api/src/main/java/io/openbas/rest/dashboard/DashboardApi.java
```diff
@@ -74,6 +76,17 @@ public List<EsSeries> series(@PathVariable final String widgetId) {
     throw new RuntimeException("Unsupported widget: " + widget);
   }
 
+  @GetMapping(DASHBOARD_URI + "/entities/{widgetId}")
+  public List<EsBase> entities(@PathVariable final String widgetId) {
+    Widget widget = this.widgetService.widget(widgetId);
+    ListConfiguration config = (ListConfiguration) widget.getWidgetConfiguration();
```
RomuDeuxfois, line 82: **nitpick;** Can we throw an exception if the configuration is not ListConfiguration type ?

openbas-front/src/admin/components/workspaces/custom_dashboards/widgets/viz/list/elements/default/DefaultListElement.tsx
```diff
@@ -0,0 +1,67 @@
...
+      <ListItemButton classes={{ root: classes.item }} inert={true}>
```
MarineLeM: Actually, DefaultListElement and EndpointListElement are duplicate components. The only differences are the ListItemIcon and the click action. Maybe they could be combined into a single component in your next chunk?

openbas-api/src/test/java/io/openbas/rest/dashboard/DashboardApiTest.java
```diff
@@ -0,0 +1,137 @@
...
+import org.junit.jupiter.api.*;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.MediaType;
+import org.springframework.test.web.servlet.MockMvc;
+
+@Transactional
+@WithMockAdminUser
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@DisplayName("Dashboard API tests")
+public class DashboardApiTest extends IntegrationTest {
```
RomuDeuxfois, line 37: **praise:** Big thanks for his !

openbas-api/src/test/resources/application.properties
```diff
@@ -45,7 +45,7 @@ spring.jpa.properties.hibernate.order_inserts=true
 ### ENGINE Configuration
 engine.index-prefix=openbas
 engine.index-suffix=-000001
-engine.url=http://localhost:9200
+engine.url=http://localhost:9201
```
RomuDeuxfois, line 48: **question:** Are you sure you want to commit this ?

openbas-model/src/main/java/io/openbas/service/EsService.java
```diff
@@ -499,6 +492,68 @@ public List<EsSeries> multiDateHistogram(RawUserAuth user, DateHistogramRuntime
         .toList();
   }
 
+  public List<EsBase> entities(RawUserAuth user, ListRuntime runtime) {
+    Filters.FilterGroup searchFilters = runtime.getWidget().getSeries().get(0).getFilter();
```
RomuDeuxfois, line 496: **question:** I am not sure what do you want to achieve here and why searching for the first element of the series.

openbas-model/src/main/java/io/openbas/engine/api/HistogramWidget.java
```diff
@@ -3,54 +3,29 @@
...
 
-  private String title;
   @NotBlank private String field;
   private boolean stacked;
 
   @JsonProperty("display_legend")
   private boolean displayLegend;
 
   HistogramWidget(String mode) {
+    super(
```
RomuDeuxfois, line 25: **nitpick:** Do we throw an exception if we are not temporal or structural ?

openbas-model/src/main/java/io/openbas/driver/ElasticDriver.java
```diff
@@ -323,12 +323,15 @@ public void cleanUpIndex(String indexName, ElasticsearchClient client) throws IO
       String fullIndexWithSuffix = fullIndexName + config.getIndexSuffix();
 
       // 1. Delete index and alias if they exist
-      try {
-        client.indices().delete(d -> d.index(fullIndexWithSuffix));
-        log.info("Deleted index: {}", fullIndexWithSuffix);
-      } catch (ElasticsearchException e) {
-        log.warn("Index {} does not exist or already deleted", fullIndexWithSuffix);
+      for (String name : List.of(fullIndexName, fullIndexWithSuffix)) {
```
RomuDeuxfois, line 326: **question:** Why adding fullIndexName here ?

openbas-front/src/admin/components/workspaces/custom_dashboards/widgets/viz/list/elements/endpoint/EndpointListElement.tsx
```diff
@@ -0,0 +1,83 @@
...
+  columns: string[];
+  element: EsEndpoint;
+  secondaryAction: (node: ReactNode) => void;
+};
+
+const EndpointListElement = (props: Props) => {
+  const { classes } = useStyles();
+  const bodyItemsStyles = useBodyItemsStyles();
+
+  const endpointUrl = `/admin/assets/endpoints/${props.element.base_id}`;
```
RomuDeuxfois: **nitpick:** Can we have a constant file to list this endpoints ?

openbas-front/src/admin/components/common/endpoints/EndpointListItemFragments.tsx
```diff
@@ -0,0 +1,15 @@
+enum EndpointListItemFragments {
+  ASSET_NAME = 'asset_name',
+  ENDPOINT_NAME = 'endpoint_name',
```
RomuDeuxfois: **todo:** Seems no used

openbas-front/src/admin/components/common/endpoints/EndpointListItemFragments.tsx
```diff
@@ -0,0 +1,15 @@
...
+  ASSET_TAGS = 'asset_tags',
+  BASE_TAGS_SIDE = 'base_tags_side',
+  ASSET_TYPE = 'asset_type',
+  ENDPOINT_ACTIVE = 'endpoint_active',
+  ENDPOINT_ARCH = 'endpoint_arch',
+  ENDPOINT_AGENTS_PRIVILEGE = 'endpoint_agents_privilege',
+  ENDPOINT_PLATFORM = 'endpoint_platform',
+  ASSET_PLATFORM = 'asset_platform',
+  ENDPOINT_AGENTS_EXECUTOR = 'endpoint_agents_executor',
```
RomuDeuxfois: **todo:** Seems no used

openbas-front/src/admin/components/common/endpoints/fragments/output/EndpointActiveFragment.tsx
```diff
@@ -0,0 +1,22 @@
+import { Tooltip } from '@mui/material';
```
RomuDeuxfois, line 1: **question:** Why having output on this uri ?

openbas-front/src/admin/components/workspaces/custom_dashboards/widgets/WidgetCreationParameters.tsx
```diff
@@ -17,70 +15,12 @@ const WidgetCreationParameters: FunctionComponent<{
...
-    name: 'widget_config.series',
-  });
-  const entities = series.map(v => getBaseEntities(v.filter)).flat();
-
-  // -- HANDLE MODE --
-  const availableModes = getAvailableModes(widgetType);
-  useEffect(() => {
-    if (availableModes.length === 1) {
-      setValue('widget_config.mode', availableModes[0]); // If only one mode is available, hide the field and set it automatically
+  const getParametersControl = (widgetType: Widget['widget_type']) => {
```
RomuDeuxfois, line 18: **praise:** Nice split

openbas-front/src/admin/components/workspaces/custom_dashboards/widgets/viz/list/ListWidgetParameters.tsx
```diff
@@ -0,0 +1,124 @@
...
+          <input
+            {...field}
+            type="hidden"
+            value={field.value ?? ''}
+          />
+        )}
+      />
+      <Controller
+        control={props.control}
+        name="widget_config.sorts.0.fieldName"
```
RomuDeuxfois, line 84: **question:** I am not a big fan of this. Can we loop on it or something else ?

openbas-front/src/admin/components/workspaces/custom_dashboards/widgets/viz/list/elements/default/DefaultListElement.tsx
RomuDeuxfois: **todo:** warning here -> A fragment with only one child is redundant.

openbas-front/src/admin/components/workspaces/custom_dashboards/widgets/viz/list/elements/default/DefaultListElement.tsx
antoinemzs: sure, I will try.

openbas-api/src/test/resources/application.properties
antoinemzs: yes, the tests should use the elastic test instance as configured in the dev compose file

RomuDeuxfois:
Testing:
- Can we categorize sort to have a clear view ?
image
- Missing trads
image
- EOL is not displayed
- Updated is not taking into account -> Update description, not displayed on list widget

openbas-dev/docker-compose.yml
antoinemzs: Good call, unsure about that too. I'd wait for some usage time before we make a change here. Would you be ok with keeping the dev instance settings and adjust from there ?

openbas-api/src/test/java/io/openbas/rest/dashboard/DashboardApiTest.java
antoinemzs: As discussed, this affects the production code for a test need. On paper, I'm 100% with your proposal. I'd propose that we do this as an individual tech enhancement to avoid scope creep. Would you agree?

antoinemzs: See #3483

openbas-model/src/main/java/io/openbas/driver/ElasticDriver.java
antoinemzs: Indexes were not being reliably cleared, sometimes the index was known with its name without suffix, and it would be retained (not cleared).

openbas-model/src/main/java/io/openbas/service/EsService.java
antoinemzs: Yes it's not the clearest. I want to obtain the entity name, which is contained as the "base_entity eq [entity name]" filter. This allows me to select the correct deserialiser for the entity, so that we get objects with the correct columns as a response.

On the java side this information is encoded in the series attribute.

Getting the first object is an assumption that the entities endpoint can only return a set of entities of the same type, as expressed by the fact that the elastic query is deserialised with a single target class.

antoinemzs: Perhaps there is a better way ?

openbas-api/src/main/java/io/openbas/rest/dashboard/DashboardApi.java
antoinemzs: The other endpoints don't do this; currently it will in practice crash (500) and that's probably the correct crash code: it's not the specific request's fault to have requested a earlier badly configured widget?

There would not be a reason for this though. Perhaps add one?

RomuDeuxfois: It's crash properyl, it's ok for me.

openbas-model/src/main/java/io/openbas/service/EsService.java
RomuDeuxfois: So, we can't have multiple filter right ?
Oh I see, we have the limitation on front-end

openbas-front/src/admin/components/workspaces/custom_dashboards/widgets/viz/list/ListWidgetParameters.tsx
antoinemzs: The UI is set for a single sort, in accordance to the mocked UI. There is no limitation for multiple sorts in the backend, but the intended use case for the UI is a single sort attribute.

openbas-model/src/main/java/io/openbas/service/EsService.java
antoinemzs: Let's defer to next chunk but yes, it should be a singular series.

openbas-model/src/main/java/io/openbas/engine/api/HistogramWidget.java
antoinemzs: Given the API does not restrict mode via enum, we decided to let this fallback in place and strive to improve the earlier input validation. Strong types (i.e. enums in this case) will ensure the validity of the input by design.

openbas-front/src/admin/components/common/endpoints/fragments/output/EndpointActiveFragment.tsx
antoinemzs: fragments are designed to take an input prop of either EsEndpoint or EndpointOutput; the `elastic` and `output` directories reflect this. I will reduce this split in a further chunk.

openbas-api/src/main/java/io/openbas/rest/dashboard/DashboardApi.java
RomuDeuxfois: Seen with Antoine: app crash properly

RomuDeuxfois (approved):

PR #3398, opened 2025-06-20
Add attack pattern filter on inject results

RomuDeuxfois:
### Related issues

* https://github.com/OpenBAS-Platform/openbas/issues/3340

### DIFF
--- a/openbas-model/src/main/java/io/openbas/database/model/Inject.java
+++ b/openbas-model/src/main/java/io/openbas/database/model/Inject.java
@@ -395,8 +395,8 @@ public class Inject implements Base, Injection {
   public List<KillChainPhase> getKillChainPhases() {
     return getInjectorContract()
         .map(
-            injectorContract ->
-                injectorContract.getAttackPatterns().stream()
+            ic ->
+                ic.getAttackPatterns().stream()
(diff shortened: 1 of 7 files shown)


openbas-model/src/main/java/io/openbas/database/model/Inject.java
```diff
@@ -395,15 +395,16 @@ public Instant getSentAt() {
   public List<KillChainPhase> getKillChainPhases() {
     return getInjectorContract()
         .map(
-            injectorContract ->
-                injectorContract.getAttackPatterns().stream()
+            ic ->
+                ic.getAttackPatterns().stream()
```
antoinemzs, line 399: nitpick: it looked clear enough but it's not worse after so ok why not!

antoinemzs (commented):
lgtm

openbas-model/src/main/java/io/openbas/database/model/Inject.java
RomuDeuxfois: InjectorContract is already defined in this class, and having duplicate variable names in Java is not ideal (I hate that).

antoinemzs: fair

PR #3430, opened 2025-06-24
[backend] Link Finding to targeted asset

### LINKED ISSUE #3327: Add asset as payload's argument
## Context
The solution to add the asset as a payload's argument is needed to enhance the accuracy and flexibility of data transactions within OpenBAS. By including the asset directly as an argument, users can ensure that all relevant asset data is transmitted in a single transaction, eliminating confusion and reducing errors. This change provides a technical improvement that enhances the user experience by simplifying data management and improving the precision of asset-related operations.

## Use case
The use case involves scenarios where users must submit transactions or interact with APIs that require detailed asset information. Currently, without the ability to include the asset as a direct argument in the payload, users face challenges in packaging all necessary data efficiently, which can lead to incomplete data processing or cumbersome manual data entry.

## Current Workaround
Currently, users may use additional fields or external documentation to track asset details, which they must reference separately during transactions. This approach is inefficient and error-prone, as it requires extra coordination outside the OpenBAS system and increases the potential for miscommunication.

## Proposed Solution
The proposed solution is to update OpenBAS to allow assets to be included directly as arguments within the payload of transactions. This would require modifications to the data handling processes and possibly updates to the API to accommodate the new parameter. This integration will streamline user workflows by consolidating all necessary information within a single, cohesive transaction process.

MarineLeM:
### Proposed changes

* Link Finding to targeted asset

 Closes #3327

### DIFF
--- a/openbas-api/src/test/java/io/openbas/rest/inject/service/InjectServiceTest.java
+++ b/openbas-api/src/test/java/io/openbas/rest/inject/service/InjectServiceTest.java
@@ -637,12 +637,14 @@ class InjectServiceTest {
       {
       "type": "defaultValue1",
       "key": "value1",
-      "defaultValue": "defaultValue1"
+      "defaultValue": ["defaultValue1"],
+       "cardinality":"1"
       },
       {
       "type": "asset",
       "key": "value2",
-      "defaultValue": "defaultValue2"
+      "defaultValue": ["defaultValue2"],
+      "cardinality":"1"
       }
     ]
   }


openbas-api/src/main/java/io/openbas/rest/finding/FindingUtils.java
```diff
@@ -1,86 +1,18 @@
...
+import io.openbas.injector_contract.outputs.InjectorContractContentOutputElement;
+import org.jetbrains.annotations.NotNull;
 
-@Slf4j
-@RequiredArgsConstructor
-@Component
 public class FindingUtils {
 
-  private final FindingRepository findingRepository;
+  private FindingUtils() {}
```
Dimfacion: **nitpick :** I don't think you need it given that the only method here is a static one

openbas-api/src/main/java/io/openbas/rest/finding/FindingService.java
```diff
@@ -82,53 +80,123 @@ public void deleteFinding(@NotNull final String id) {
     this.findingRepository.deleteById(id);
   }
 
+  public void buildFinding(
```
Dimfacion, line 92: **nitpick :** I know you just moved the existing method but could you add javadoc on this one please ? (Thanks for adding it so much on the rest of the PR btw :) )

PR #3439, opened 2025-06-25
[frontend] Handle not standard TTPs in security coverage

RomuDeuxfois:


Seb-MIGUEL:
Why don't we use the TTP-ID as Standard ID ? All the TTPs are presents into the Taxonomy we just have to mapped the TTP value as ID and every TTP could work as expected ?

RomuDeuxfois:
We can create TTP in the UI and craft or own matrix.

Seb-MIGUEL:
Not sure to understand we can talk about it ?

PR #3446, opened 2025-06-26
[backend]RBAC: Adding the "Launch" option in the grants  Issue/375 

heditar:
### Proposed changes

* Adding the "Launch" option in the grants 
*  -> this change won't be merge to release yet it will go to a feature branch first 

Capture d'écran 2025-06-26 142249

Chunk 3-1 of [Notion](https://www.notion.so/filigran/RBAC-Implement-roles-capability-and-apply-them-to-grants-7ae145c285b64762beddfa30d07e7f99)

### Testing Instructions

1. Create a group, on the popover select "Manage Grants"
2. Then play with the new option to add Launch grant

### Related issues

* #375

### Checklist

- [] I wrote test cases for the relevant uses case

### DIFF
--- a/openbas-front/src/admin/components/settings/groups/GroupPopover.js
+++ b/openbas-front/src/admin/components/settings/groups/GroupPopover.js
@@ -304,10 +304,13 @@ class GroupPopover extends Component {
                   <TableRow>
                     <TableCell>{t('Scenario')}</TableCell>
                     <TableCell style={{ textAlign: 'center' }}>
-                      {t('Read/Write')}
+                      {t('Access')}
                     </TableCell>
                     <TableCell style={{ textAlign: 'center' }}>
-                      {t('Read only')}
+                      {t('Manage')}
+                    </TableCell>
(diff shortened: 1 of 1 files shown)


openbas-front/src/admin/components/settings/groups/GroupPopover.js
```diff
@@ -304,10 +304,13 @@ class GroupPopover extends Component {
                   <TableRow>
                     <TableCell>{t('Scenario')}</TableCell>
                     <TableCell style={{ textAlign: 'center' }}>
-                      {t('Read/Write')}
```
damgouj, line 307: You can delete 'Read/Write' in the translations files I think :)

damgouj (changes requested):
Quick comment + Found a little UI issue: 
When I click on MANAGE, the first check is disabled and grey:
image
But when I click on LAUNCH, the two first checks are disabled but blue:
image
Thanks !

Issue #3352, opened 2025-06-12
Add Vulnerable Expectation


Seb-MIGUEL:
## Context

The goal is to add a new kind of expectation which help to easily show if assets vulnerable are discovered or no.

## Use case

I run a CVE scan arround assets and want to have a green or red color indicator whether assets are vulnerable or not

## Current Workaround

Nop

## Proposed Solution

Add new "vulnerable"expectation in the same logical than manual one which will be set in the inject we wants to check once it's done we want to update the status at the moment where a finding is discovered into the inject.

PR #3490, opened 2025-06-30
[backend|frontend]Add vulnerability expectation (#3352)

johanah29:


openbas-api/src/main/java/io/openbas/rest/inject/service/InjectExecutionService.java
```diff
@@ -70,6 +85,78 @@ private void processInjectExecution(
     }
   }
 
+  public void checkCveExpectation(
```
Dimfacion, line 97: **nitpick :** Please if possible add javadoc as much as possible, at least on the public method in your services

openbas-api/src/main/java/io/openbas/utils/ExpectationUtils.java
```diff
@@ -483,6 +486,67 @@ public static List<DetectionExpectation> getDetectionExpectations(
     return returnList;
   }
 
+  public static List<VulnerabilityExpectation> getVulnerabilityExpectations(
```
Dimfacion, line 497: **nitpick :** Same as https://github.com/OpenBAS-Platform/openbas/pull/3490/files#r2181024060

openbas-framework/src/main/java/io/openbas/model/expectation/VulnerabilityExpectation.java
```diff
@@ -0,0 +1,91 @@
...
+import jakarta.validation.constraints.NotBlank;
+import jakarta.validation.constraints.NotNull;
+import java.util.List;
+import java.util.Objects;
+import javax.annotation.Nullable;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
```
Dimfacion: **nitpick :** When you have @Getter and @Setter, you can replace them both with @Data

openbas-framework/src/main/java/io/openbas/model/expectation/VulnerabilityExpectation.java
```diff
@@ -0,0 +1,91 @@
...
+  private List<InjectExpectationSignature> injectExpectationSignatures;
+
+  private VulnerabilityExpectation() {}
+
+  @Override
+  public InjectExpectation.EXPECTATION_TYPE type() {
+    return VULNERABILITY;
+  }
+
+  public static VulnerabilityExpectation vulnerabilityExpectationForAgent(
```
Dimfacion: **nitpick :** I think it would be better for those three utilitarian methods to be in a dedicated utils class like VulnerabilityExpectationUtils.java instead of being declared inside the model itself

Dimfacion:
**nitpick :** Very nitpicky but on the screenshot, we can see that on the overview, the result section has expectations listed as "Prevention, Detection, Human response, Vulnerability" but in the inject results list, the global score is "Prevention, Detection, Vulnerability, Human response". Could we harmonize this quickly ? If not, leave it like that for now

openbas-framework/src/main/java/io/openbas/model/expectation/VulnerabilityExpectation.java
savacano28: I'm not sure if this a good practices for entities https://jpa-buddy.com/blog/lombok-and-jpa-what-may-go-wrong/#

Dimfacion (approved):

PR #3521, opened 2025-07-03
[frontend/backend] Misc fixes for List Widget

antoinemzs:
### Proposed changes

*
*

### Testing Instructions

1. Step-by-step how to test
2. Environment or config notes

### Related issues

* Closes #ISSUE-NUMBER
*

### Checklist

- [ ] I consider the submitted work as finished
- [ ] I tested the code for its functionality
- [ ] I wrote test cases for the relevant uses case
- [ ] I added/update the relevant documentation (either on github or on notion)
- [ ] Where necessary I refactored code to improve the overall quality
- [ ] For bug fix -> I implemented a test that covers the bug

### Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #3530, opened 2025-07-03
[backend/frontend] Fix limit 0

RomuDeuxfois:
### Proposed changes

* Fix limitation on front end for the limit field
* Fix this stack trace when limit is 0
co.elastic.clients.elasticsearch._types.ElasticsearchException: [es/search] failed: [x_content_parse_exception] [1:82] [terms] failed to parse field [size]


openbas-model/src/main/java/io/openbas/engine/api/StructuralHistogramWidget.java
```diff
@@ -14,7 +15,7 @@ public class StructuralHistogramWidget extends HistogramWidget {
 
   public static final String STRUCTURAL_MODE = "structural";
 
-  private int limit = 100;
+  @Positive private int limit = 100;
```
antoinemzs: TODO: if merging #3521 first, consider rebasing and using an engine constant for this.

openbas-model/src/main/java/io/openbas/service/EsService.java
```diff
@@ -425,18 +422,30 @@ public EsSeries termHistogram(
...
-                          new Aggregation.Builder().terms(termsAggregation).build()),
-              Void.class);
+
+      SearchRequest.Builder searchBuilder =
+          new SearchRequest.Builder()
+              .index(engineConfig.getIndexPrefix() + "*")
+              .size(0)
+              .query(query);
+
+      if (widgetConfig.getLimit() > 0) {
```
antoinemzs, line 435: Could this be constrained in the API too, e.g. via an annotation on the `StructuralHistogramWidget::limit` field?

antoinemzs, line 435: nitpick: Could use a small comment to explain that a `0` value crashes the query.

RomuDeuxfois: Let's limit the end user to 1 instead of 0 on all widgets in this case right ?

antoinemzs: Done

antoinemzs (approved):

PR #3534, opened 2025-07-03
Custom dashboard improvment

RomuDeuxfois:
### Proposed changes

* Fix -> Custom dashboard: filter can be longer
* Fix -> Custom Dashboard Options: load option is really long, we need to take a limit to display it

### DIFF
--- a/openbas-api/src/main/java/io/openbas/rest/simulation/SimulationService.java
+++ b/openbas-api/src/main/java/io/openbas/rest/simulation/SimulationService.java
@@ -2,11 +2,15 @@ package io.openbas.rest.simulation;
 
 import static io.openbas.database.specification.ExerciseSpecification.byName;
 import static io.openbas.helper.StreamHelper.fromIterable;
+import static io.openbas.utils.FilterUtilsJpa.PAGE_NUMBER_OPTION;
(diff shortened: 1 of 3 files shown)


openbas-api/src/main/java/io/openbas/rest/simulation/SimulationService.java
```diff
@@ -24,9 +28,9 @@ public class SimulationService {
    * @return list of {@link Option} objects containing exercise IDs and names
    */
   public List<Option> findAllAsOptions(final String searchText) {
-    return fromIterable(
-            exerciseRepository.findAll(byName(searchText), Sort.by(Sort.Direction.ASC, "name")))
-        .stream()
+    Pageable pageable =
+        PageRequest.of(PAGE_NUMBER_OPTION, PAGE_SIZE_OPTION, Sort.by(Sort.Direction.ASC, "name"));
```
antoinemzs: This will effectively truncate the number of results, and within that window, ordering just by name may return a different set across multiple calls (i.e. there's no guarantee that within a given "name cluster", the actual records will be ordered in the same way).

Can we order by both name, and then created_by?

RomuDeuxfois: Agree with your point, I added a new commit

antoinemzs (approved):

PR #3541, opened 2025-07-04
Bugfix/limit horizontal bar

EmilieFo17:
Fix bug : Horizontal bar: limit field is acting weird

### DIFF
--- a/openbas-front/src/admin/components/workspaces/custom_dashboards/widgets/HistogramParameters.tsx
+++ b/openbas-front/src/admin/components/workspaces/custom_dashboards/widgets/HistogramParameters.tsx
@@ -156,7 +156,7 @@ const HistogramParameters = ({ widgetType, control, setValue }: Props) => {
               label={t('Number of results')}
               sx={{ mt: 2 }}
               value={field.value ?? 10}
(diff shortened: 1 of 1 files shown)


RomuDeuxfois (approved):

openbas-front/src/admin/components/workspaces/custom_dashboards/widgets/HistogramParameters.tsx
```diff
@@ -156,7 +156,7 @@ const HistogramParameters = ({ widgetType, control, setValue }: Props) => {
               label={t('Number of results')}
               sx={{ mt: 2 }}
               value={field.value ?? 10}
-              onChange={e => field.onChange(e.target.value === '' ? undefined : Number(e.target.value))}
+              onChange={e => field.onChange(e.target.value === '' ? '' : Number(e.target.value))}
```
guillaumejparis, line 159: Why not 0 ?

EmilieFo17: because when you erase everything it's by doing a backspace it's weird that it adds you a number 0 like 100 besides it's annoying to use.

PR #3552, opened 2025-07-04
Issue/3550

### LINKED ISSUE #3550: My scenario can't run because injects stay in pending but it's executed
As you can see here sometimes into my scenario injects are successfully done but still pending and when I've got chaining that blocked my scenario

impolitepanda:
### Proposed changes

* The api will now check that the inject's status is EXECUTING before allowing the change to a complete state. Otherwise it will reply with a 409 Conflict error. The imlant should now handle this in it's code (PR to come)

### Testing Instructions

1. Launch a technical inject via Crowdstrike agent with an obas implant with a super fast payload.

### Related issues

* Closes #3550 

### Checklist

### DIFF
--- a/openbas-api/src/main/java/io/openbas/rest/inject/service/InjectExecutionService.java
+++ b/openbas-api/src/main/java/io/openbas/rest/inject/service/InjectExecutionService.java
@@ -10,6 +10,7 @@ import io.openbas.database.repository.InjectExpectationRepository;
 import io.openbas.database.repository.InjectRepository;
 import io.openbas.rest.exception.ElementNotFoundException;
 import io.openbas.rest.finding.FindingService;
+import io.openbas.rest.inject.form.InjectExecutionAction;
 import io.openbas.rest.inject.form.InjectExecutionInput;
 import io.openbas.rest.inject.form.InjectExpectationUpdateInput;
 import io.openbas.service.InjectExpectationService;
@@ -19,14 +20,14 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
-import java.util.logging.Level;
 import lombok.RequiredArgsConstructor;
-import lombok.extern.java.Log;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.dao.DataIntegrityViolationException;
 import org.springframework.stereotype.Service;
 
 @RequiredArgsConstructor
 @Service
-@Log
+@Slf4j
 public class InjectExecutionService {
 
   private final InjectRepository injectRepository;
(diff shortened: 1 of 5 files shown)


openbas-api/src/main/java/io/openbas/rest/inject/service/InjectExecutionService.java
```diff
@@ -45,6 +46,20 @@ public void handleInjectExecutionCallback(
...
+      if (input.getAction().equals(InjectExecutionAction.complete)
+          && (inject.getStatus().isEmpty()
+              || !inject.getStatus().get().getName().equals(ExecutionStatus.EXECUTING))) {
+        // If we receive a status update with a terminal state status, we must first check that the
+        // current status
+        // is in the Executing state
+        log.warn(
+            "Received a complete action for inject {} with status {}, but current status is not EXECUTING",
+            injectId,
+            inject.getStatus().map(is -> is.getName().toString()).orElse("UNKNOWN"));
```
antoinemzs: Should probably use an enum variant for the else case, don't you think? Even if it's just for the error message. Otherwise perhaps rephrase the message to not allude that UNKNOWN is a valid state?

impolitepanda: I would definitely agree normally. However, in this case, i disagree. 
First, this is only used here in a log message. 
Second, the UNKNOW value has no meaning in the status enumeration, and adding it could make it seem like it's a valid value in the state machine while it is not. 
And third this is a case that should never ever happen. There is no way for a status to not exist at this point in the inject execution pipeline. The orElse is only here because the status is an Optional and the compiler was screaming at me :)

Is that Ok for you or would you still prefer I add it to the enum ?

antoinemzs: Yeah that makes sense, my main thinking was that having a fully capitalised "UNKNOWN" would make one think it is actually a valid state but it'd not be found elsewhere. Lower case "unknown" would possibly avoid that confusion.

It's a nitpick of a nitpick since you explained that there is always a status and it's for compilation issues. Therefore the real issue is probably why the status is Optional but that's out of scope.

openbas-api/src/main/java/io/openbas/rest/inject/service/InjectStatusService.java
```diff
@@ -80,25 +80,18 @@ private int getCompleteTrace(Inject inject) {
...
+  public void updateFinalInjectStatus(InjectStatus injectStatus, Instant trackingEndDate) {
     log.info("[issue/2797] updateFinalInjectStatus 1: " + injectStatus.getId());
     ExecutionStatus finalStatus =
         computeStatus(
             injectStatus.getTraces().stream()
                 .filter(t -> ExecutionTraceAction.COMPLETE.equals(t.getAction()))
                 .toList());
     log.info("[issue/2797] updateFinalInjectStatus 2: " + injectStatus.getId());
-    injectStatus.setTrackingEndDate(Instant.now());
+    injectStatus.setTrackingEndDate(trackingEndDate == null ? Instant.now() : trackingEndDate);
```
antoinemzs: nitpick: why not require callers to pass an instance of Instant?

antoinemzs (approved):
Looks good, a few microscopic nitpicks

openbas-api/src/main/java/io/openbas/rest/inject/service/InjectExecutionService.java
impolitepanda: Let's go for a lowercase unknown, then :)
And agreed on status being an optional being strange ^^'

openbas-api/src/main/java/io/openbas/rest/inject/service/InjectStatusService.java
impolitepanda: To be honest, regression fear. But now that I think of it... it doens't change much. I'll fix that and make it notnull.

PR #3554, opened 2025-07-04
[backend] Add method to enable player in team (#3445)

### LINKED ISSUE #3445: Update Teams enable state by default
Currently when we add teams into our scenario there are by default disabled we have to click on it and enabled them. We would like to change that default behavior to enabled by default the teams we add and let the ability to update the teams if we would like to disabled it.

Image

Image

camrrx:
### Proposed changes

* When adding a team to a simulation, users are activated by default
Closes #3445

### Testing Instructions

1. Go to simulation page in "definition" tab
2. Add a team
3. All the users are supposed to be enabled
4. Do the same on scenario page 

### Checklist

### Further comments

### DIFF
--- a/openbas-api/src/main/java/io/openbas/rest/scenario/ScenarioApi.java
+++ b/openbas-api/src/main/java/io/openbas/rest/scenario/ScenarioApi.java
@@ -199,7 +199,8 @@ public class ScenarioApi extends RestBehavior {
       @PathVariable @NotBlank final String scenarioId,
       @PathVariable @NotBlank final String teamId,
       @Valid @RequestBody final ScenarioTeamPlayersEnableInput input) {
-    return this.scenarioService.enablePlayers(scenarioId, teamId, input.getPlayersIds());
+    return this.scenarioService.enableAddScenarioTeamPlayer(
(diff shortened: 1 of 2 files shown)


savacano28:
Tested ok: Scenario, simulation

openbas-api/src/main/java/io/openbas/rest/exercise/service/ExerciseService.java
```diff
@@ -684,6 +697,24 @@ public List<TeamOutput> replaceTeams(
     return teamService.find(fromIds(modifiedTeamIds));
   }
 
+  public Exercise enablePlayers(
+      @NotBlank final String exerciseId,
+      @NotBlank final String teamId,
+      @NotNull final List<String> playerIds) {
+    Exercise exercise =
+        exerciseRepository.findById(exerciseId).orElseThrow(ElementNotFoundException::new);
+    Team team = this.teamRepository.findById(teamId).orElseThrow();
```
savacano28: Maybe you can pass the team here and avoid re fetch? l683 =l706 (same for scenarioService)

PR #3557, opened 2025-07-07
[frontend] select at least one dimension widget

MarineLeM:
### Proposed changes

* Select at least one dimension to create a widget

### DIFF
--- a/openbas-front/src/admin/components/workspaces/custom_dashboards/widgets/WidgetForm.tsx
+++ b/openbas-front/src/admin/components/workspaces/custom_dashboards/widgets/WidgetForm.tsx
@@ -74,7 +74,7 @@ const WidgetForm: FunctionComponent<Props> = ({
       display_legend: z.boolean().optional(),
       series: z.array(z.object({
         name: z.string().optional(),
-        filter: z.any().optional(),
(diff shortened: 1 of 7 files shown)


openbas-front/src/admin/components/workspaces/custom_dashboards/widgets/WidgetForm.tsx
```diff
@@ -104,7 +104,7 @@ const WidgetForm: FunctionComponent<Props> = ({
       columns: z.array(z.string()),
       series: z.array(z.object({
         name: z.string().optional(),
-        filter: z.any().optional(),
+        filter: z.any().refine(val => val !== undefined, { message: 'Filter cannot be undefined' }),
```
guillaumejparis, line 107: "Filter cannot be undefined" is alrready translated ?

guillaumejparis (approved):

openbas-front/src/admin/components/workspaces/custom_dashboards/widgets/WidgetForm.tsx
MarineLeM: Nice catch i add it in the translations files, thanks

Issue #3359, opened 2025-06-13
Improve expectation signatures to support network detections matching with collectors


antoinemzs:
## Use case

SIEM collector should have sufficient metadata for matching network detections with injects. Collectors rely on "signatures" that are embedded in expectations for matching data from alerts. Currently, only a "parent process name" is proposed as a signature to match with alerts,

network detections need to rely on source and target ips and/or hostnames

## Current Workaround

No known workaround

## Proposed Solution

OpenBAS server should add Signatures in expectations based on inject contract data:
* for payload injects, use the "inject targets" IP as "source IP" (one signature each) (mind the dynamic asset groups)
* for special injectors such as nmap/nuclei: use the contract "target" as source IP or hostname

Also include signatures for specific inject start time, end time.

## If the feature request is approved, would you be willing to submit a PR?

Yes / No (Help can be provided if you need assistance submitting a PR)

jborozco:
@Seb-MIGUEL I'm moving this task to 1.19, does that mean we need to move splunk integration as well ? (https://github.com/OpenBAS-Platform/collectors/issues/15)

Seb-MIGUEL:
@jborozco Antoine's ETA for the List widget is tomorrow, we should let that task to allow Antoine begin to work on it as it's a really important one

PR #3610, opened 2025-07-16
[backend] add source and target expectation signatures

MarineLeM:
### Proposed changes

* add signatures for source and target asset 

### Related issues

* Chunk 1 #3359

### DIFF
--- a/openbas-api/src/main/java/io/openbas/utils/ExpectationUtils.java
+++ b/openbas-api/src/main/java/io/openbas/utils/ExpectationUtils.java
@@ -248,7 +248,8 @@ public class ExpectationUtils {
                         OBAS_IMPLANT_CALDERA,
                         executedAgent.getInject().getId(),
                         getIpsFromAsset(asset),
-                        executedAgent.getParent().getId(), List.of())))
+                        executedAgent.getParent().getId(),
+                        List.of())))
         .toList();
   }
 
@@ -273,7 +274,7 @@ public class ExpectationUtils {
                         executedAgent.getInject().getId(),
                         getIpsFromAsset(asset),
                         executedAgent.getParent().getId(),
-                            List.of())))
+                        List.of())))
         .toList();
   }
 
@@ -302,7 +303,7 @@ public class ExpectationUtils {
       AssetToExecute assetToExecute,
       Inject inject,
       io.openbas.model.inject.form.Expectation expectation,
-  List<String> targetedAssetValues) {
+      List<String> targetedAssetValues) {
     return getExpectations(
         assetToExecute,
         (AssetGroup assetGroup) ->
@@ -315,7 +316,11 @@ public class ExpectationUtils {
                 expectation.getExpirationTime()),
         (AssetGroup assetGroup, PreventionExpectation preventionExpectation) ->
             getPreventionExpectationList(
-                assetToExecute.asset(), assetGroup, inject, preventionExpectation, targetedAssetValues));
+                assetToExecute.asset(),
+                assetGroup,
+                inject,
+                preventionExpectation,
+                targetedAssetValues));
   }
 
   public static List<DetectionExpectation> getDetectionExpectations(
@@ -335,7 +340,11 @@ public class ExpectationUtils {
                 expectation.getExpirationTime()),
         (AssetGroup assetGroup, DetectionExpectation detectionExpectation) ->
             getDetectionExpectationList(
-                assetToExecute.asset(), assetGroup, inject, detectionExpectation, targetedAssetValues));
+                assetToExecute.asset(),
+                assetGroup,
+                inject,
+                detectionExpectation,
+                targetedAssetValues));
   }
 
   /**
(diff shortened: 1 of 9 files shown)


openbas-model/src/main/java/io/openbas/database/model/InjectExpectationSignature.java
```diff
@@ -9,6 +9,13 @@
...
 public class InjectExpectationSignature {
   public static final String EXPECTATION_SIGNATURE_TYPE_PARENT_PROCESS_NAME = "parent_process_name";
+  public static final String EXPECTATION_SIGNATURE_TYPE_SOURCE_IPV4_ADDRESS = "source_ipv4_address";
+  public static final String EXPECTATION_SIGNATURE_TYPE_SOURCE_IPV6_ADDRESS = "source_ipv6_address";
+  public static final String EXPECTATION_SIGNATURE_TYPE_TARGET_IPV4_ADDRESS = "target_ipv4_address";
+  public static final String EXPECTATION_SIGNATURE_TYPE_TARGET_IPV6_ADDRESS = "target_ipv6_address";
+  public static final String EXPECTATION_SIGNATURE_TYPE_TARGET_HOSTNAME_ADDRESS =
+      "target_hostname_address";
+
   public static final String EXPECTATION_SIGNATURE_TYPE_PROCESS_NAME = "process_name";
```
RomuDeuxfois: Can be removed

openbas-api/src/main/java/io/openbas/utils/ExpectationUtils.java
```diff
@@ -710,6 +537,19 @@ private static InjectExpectationSignature createSignature(
     return builder().type(signatureType).value(signatureValue).build();
   }
 
+  private static InjectExpectationSignature createIPSignature(
+      String ip, String ipv4Signature, String ipv6Signature) {
+    String ipv4Pattern =
+        "^(25[0-5]|2[0-4][0-9]|1[0-9]{2}|[1-9]?[0-9])"
+            + "(\\.(25[0-5]|2[0-4][0-9]|1[0-9]{2}|[1-9]?[0-9])){3}"
+            + "(?::([0-9]{1,5}))?$";
```
RomuDeuxfois: Should be great to have a constant or use the same system from Ipv4OrIpv6Validator

openbas-api/src/main/java/io/openbas/utils/ExpectationUtils.java
```diff
@@ -710,6 +537,19 @@ private static InjectExpectationSignature createSignature(
     return builder().type(signatureType).value(signatureValue).build();
   }
 
+  private static InjectExpectationSignature createIPSignature(
```
RomuDeuxfois: Can be moved to InjectExpectationSignature class, what do you think ?
Same as createSignature ?

openbas-api/src/main/java/io/openbas/utils/ExpectationUtils.java
```diff
@@ -694,14 +495,40 @@ public static List<ManualExpectation> getManualExpectationList(
   // COMPUTE SIGNATURES
 
   private static List<InjectExpectationSignature> computeSignatures(
-      String prefixSignature, String injectId, String agentId) {
+      String prefixSignature,
+      String injectId,
```
RomuDeuxfois, line 363: Can we change injectId to Inject and compute valueTargetedAssetsMap inside this method ?

openbas-api/src/main/java/io/openbas/utils/ExpectationUtils.java
```diff
@@ -694,14 +495,40 @@ public static List<ManualExpectation> getManualExpectationList(
...
+      List<String> sourceIps,
```
RomuDeuxfois: Same for sourceIps, we can use asset in args and compute sourceIps inside this method.

openbas-api/src/main/java/io/openbas/utils/ExpectationUtils.java
```diff
@@ -497,129 +358,52 @@ public static List<DetectionExpectation> getDetectionExpectations(
...
+                inject,
+                vulnerabilityExpectation,
+                valueTargetedAssetsMap));
   }
 
   public static List<ManualExpectation> getManualExpectations(
       AssetToExecute assetToExecute,
       Inject inject,
       io.openbas.model.inject.form.Expectation expectation) {
-    List<ManualExpectation> manualExpectationList = new ArrayList<>();
```
RomuDeuxfois, line 562: Not sure to understand this refacto. Method seems to not be the same but I suppose I missed something.

openbas-api/src/main/java/io/openbas/utils/ExpectationUtils.java
```diff
@@ -116,598 +119,271 @@ public static List<InjectExpectation> processByValidationType(
...
+                expectation.getScore(),
+                expectation.getName(),
+                expectation.getDescription(),
+                OBAS_IMPLANT_CALDERA.equals(implantType) ? agent.getParent() : agent,
+                assetToExecute.asset(),
+                assetGroup,
+                expectation.getExpirationTime(),
+                computeSignatures(
+                    implantType,
+                    OBAS_IMPLANT_CALDERA.equals(implantType) ? agent.getInject().getId() : injectId,
```
RomuDeuxfois, line 210: It's not possible to retrieve inject ID from an openbas implant agent ?

MarineLeM: An obas agent can execute many injects 
The inject is set only for the caldera case, to make difference between what we call agent caldera and implant caldera

RomuDeuxfois (approved):

Issue #3545, opened 2025-07-04
POC : Move away from dedicated index to configurable aggregation on dimensions


Dimfacion:
## Context

To solve an urgent need, we implemented a custom index with the columns we needed but we would like to avoid this kind of solution in the future. The goal of this task is to see if/how to move away from those custom index to configurable dimensions that are able to achieve the same result.

EmilieFo17:
Waiting priority product team

PR #3642, opened 2025-07-16
[backend/frontend] further refactor to better differentiate widget configuration types (#3545)

antoinemzs:
### Proposed changes

* Explicitly construct the `ListConfiguration` object and have specialised components to manage it as opposed to have it incidentally be mostly compatible with `DateHistogramConfiguration | StructuralHistogramConfiguration`
* Have a specific front code path for list widget, compared to other widgets; better stance for further UX departure

### Testing Instructions

1. everything should be backwards compatible

### Related issues

* related #3545 

### Checklist

### Further comments


openbas-api/src/main/java/io/openbas/migration/V4_10__List_Widget_config_update.java
```diff
@@ -0,0 +1,30 @@
...
+
+  @Override
+  public void migrate(Context context) throws Exception {
+    Connection connection = context.getConnection();
+    try (Statement statement = connection.createStatement()) {
+
+      // 1. Add new column for collector_type
+      statement.execute(
+          """
+            UPDATE widgets
```
damgouj, line 23: Could you explain those 2 SQL commands please (little complicated for me) ? I think you don't add a column for collector_type haha ;)

openbas-front/src/admin/components/workspaces/custom_dashboards/widgets/configuration/AuthorizedPerspectives.ts
```diff
@@ -0,0 +1,8 @@
+const getAuthorizedPerspectives = (): Map<string, string[]> => new Map([
```
damgouj, line 1: Could you explain here too please ?

openbas-model/src/main/java/io/openbas/engine/api/ListConfiguration.java
```diff
@@ -16,7 +16,7 @@
 @Setter
 public class ListConfiguration extends WidgetConfiguration {
 
-  @NotNull List<ListSeries> series = new ArrayList<>();
```
damgouj, line 19: Maybe I miss something but why do we lose the List<> here ?

openbas-front/src/admin/components/workspaces/custom_dashboards/widgets/configuration/AuthorizedPerspectives.ts
antoinemzs: It's the list of which entities you can select from the "Step 2" screen, and which filter names are allowed.

openbas-api/src/main/java/io/openbas/migration/V4_10__List_Widget_config_update.java
antoinemzs: The first one is a migration for a json field (widget_config) which adds `perspective` of type `Object` from the contents of the existing field  `series` of type `Array` (specifically, the first and only entry in the `series` array).

The second one drops the field `series`.

In effect this migrates an array to a single object. This is tied to how I represent the perspective in the server-side ListConfiguration class.

openbas-model/src/main/java/io/openbas/engine/api/ListConfiguration.java
antoinemzs: A List configuration can only let you select a single entity to display (for now). There is no point in letting it be a list.

Later plans may include having a nested structure, which still requires the head perspective to be a single object.

Issue #3479, opened 2025-06-29
Delete button on notification should be red (and title is incorrect)


SamuelHassine:
## Description

Delete button on notification should be red.

Also, the title of the drawer should be "Update a notification rule" (and not create), because it is already existing.

Correct:

![Image](https://github.com/user-attachments/assets/0c17949b-b077-4292-b01e-2f7bdb12db7e)

Incorrect:

![Image](https://github.com/user-attachments/assets/ef16b9b6-9d1e-4aca-8048-f3681d24e4d3)

PR #3653, opened 2025-07-17
[frontend]Correct scenario notifications update form(#3479)

johanah29:


openbas-front/src/admin/components/scenarios/scenario/notification_rule/NotificationRuleForm.tsx
```diff
@@ -96,6 +96,7 @@ const NotificationRuleForm: FunctionComponent<Props> = ({
               onClick={edition ? handleOpenDelete : handleClose}
               style={{ marginRight: theme.spacing(1) }}
               disabled={isSubmitting}
+              color={edition ? 'error' : 'primary'}
```
RomuDeuxfois, line 99: What is our standard? Is it documented somewhere in Notion?
Would it be wise to have a small component for this part of the UI? @guillaumejparis & @MarineLeM what do think ?

guillaumejparis: I'm not sure that we want to multiply components to the point to get a ValidateButton, a DeleteButton ...

Buthere  we need to review naming, because it is weird : edition ? handleOpenDelete : handleClose why edition true do an open delete

RomuDeuxfois: Ok
Can we add that to our internal documentation to share this knowledge ?

Issue #3468, opened 2025-06-29
Creation of team should be a drawer and not a dialog


SamuelHassine:
## Description

Creation of team should be a drawer and not a dialog

![Image](https://github.com/user-attachments/assets/f434137e-d724-437d-ab44-50610521c459)

PR #3654, opened 2025-07-17
[frontend]Change team creation/update to drawer(#3468)

johanah29:


johanah29:
image

I do not have the same issue when trying to create a team in the definition tab

openbas-front/src/admin/components/components/teams/TeamPopover.tsx
```diff
@@ -201,7 +201,7 @@ const TeamPopover: FunctionComponent<TeamPopoverProps> = ({
           </Button>
         </DialogActions>
       </MuiDialog>
-      <Dialog
+      <Drawer
```
RomuDeuxfois: Not sure we want a drawer on top of an other drawer
> create inject > add team > modify team
image

RomuDeuxfois (approved):

PR #3685, opened 2025-07-22
[backend/frontend] Fix delete partial documents (#3467)

### LINKED ISSUE #3467: Partial delete + error when deleting a file associated to a payload
## Description

Partial delete + error when deleting a file associated to a payload.

1. Create a payload of type "file drop"
2. Upload a new document as part of this payload
3. Try to delete the document

=> An error occured.

But, actually, the document was deleted in Minio (S3) but not the database entry. Which lead to an very inconsistant and not usable behavior.

What we should do:

1. A document associated to at least 1 payload cannot be deleted
2. When it is not associated to any payload it can be deleted
3. We should have a popover + dialog to display "where the file is used"

Like this:

Image

When clicking on number:

Image

savacano28:
### Proposed changes

* Avoid delete document when it is related to some payload

image
image

(This case is not possible, but the message is :)
image

### Testing Instructions

1. Create paylaod type executable or filedrop
2. Try delete this document

### Related issues

* Closes #3467 
*

### Checklist

### Further comments


openbas-api/src/test/java/io/openbas/rest/DocumentApiTest.java
```diff
@@ -65,7 +68,17 @@ void afterAll() {
   class CRUD {
 
     @Test
-    @DisplayName("Should delete a document")
+    @DisplayName("Should not delete a document when it is related to a payload")
+    void shouldNoDeleteDocumentWhenIsRelatedToAPayload() throws Exception {
```
MarineLeM: nitpick : we try to follow the pattern Given this then that for test method names

PR #3687, opened 2025-07-22
[backend] add start and end date signatures

### LINKED ISSUE #3359: Improve expectation signatures to support network detections matching with collectors

MarineLeM:
### Proposed changes

* add start and end date signatures

### Related issues

* Closes #3359

### DIFF
--- a/openbas-api/src/main/java/io/openbas/rest/inject/service/ExecutableInjectService.java
+++ b/openbas-api/src/main/java/io/openbas/rest/inject/service/ExecutableInjectService.java
@@ -203,7 +203,7 @@ public class ExecutableInjectService {
     this.injectStatusService.addStartImplantExecutionTraceByInject(
         injectId, agentId, "Implant is up and starting execution", startTime);
     this.injectExpectationService.addStartDateSignatureToInjectExpectationsByAgent(
-        injectId, agentId, startTime.toString());
+        injectId, agentId, startTime);
     return payloadToExecute;
   }
 
--- a/openbas-api/src/main/java/io/openbas/rest/inject/service/ExecutableInjectService.java
+++ b/openbas-api/src/main/java/io/openbas/rest/inject/service/ExecutableInjectService.java
@@ -13,6 +13,7 @@ import io.openbas.injectors.openbas.util.OpenBASObfuscationMap;
 import io.openbas.rest.document.DocumentService;
 import io.openbas.rest.exception.ElementNotFoundException;
 import io.openbas.rest.payload.service.PayloadService;
+import io.openbas.service.InjectExpectationService;
 import jakarta.annotation.Resource;
 import java.time.Instant;
 import java.util.ArrayList;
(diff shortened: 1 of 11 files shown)


openbas-api/src/test/java/io/openbas/utils/fixtures/InjectExpectationFixture.java
```diff
@@ -36,11 +36,11 @@ public static InjectExpectation createPreventionInjectExpectation(Team team, Inj
     return injectExpectation;
   }
 
-  public static InjectExpectation createDetectionInjectExpectation(Team team, Inject inject) {
+  public static InjectExpectation createDetectionInjectExpectation(Inject inject) {
```
RomuDeuxfois: **praise**: good point

openbas-api/src/main/java/io/openbas/rest/inject/service/ExecutableInjectService.java
```diff
@@ -200,6 +202,8 @@ public Payload getExecutablePayloadAndUpdateInjectStatus(String injectId, String
     Payload payloadToExecute = getExecutablePayloadInject(injectId);
     this.injectStatusService.addStartImplantExecutionTraceByInject(
         injectId, agentId, "Implant is up and starting execution", startTime);
+    this.injectExpectationService.addStartDateSignatureToInjectExpectationsByAgent(
```
RomuDeuxfois, line 205: **question**: Why not let the class date go to the InjectExpectationSignature method ?

openbas-api/src/main/java/io/openbas/rest/inject/service/InjectExecutionService.java
```diff
@@ -107,6 +110,35 @@ private void processInjectExecution(
...
+  private void addEndDateInjectExpectationTimeSignatureIfNeeded(
+      Inject inject, Agent agent, InjectExecutionInput input) {
+    if (agent != null
+        && ExecutionTraceAction.COMPLETE.equals(convertExecutionAction(input.getAction()))) {
+      InjectStatus injectStatus = inject.getStatus().orElseThrow();
+
+      Instant startTime =
+          injectStatus.getTraces().stream()
+              .filter(
+                  trace ->
```
RomuDeuxfois: question: same code visible on InjectStatusService.createExecutionTrace, can we have a refacto ?

RomuDeuxfois:
Tested: 🆗

Issue #3692, opened 2025-07-23
Tanium executor is not working properly


RomuDeuxfois:
## Description

When I run the tanium executor, endpoinds is not retrieve.

PR #3693, opened 2025-07-23
[backend] Fix Tanium executor endpoint retrieval (#3692)

RomuDeuxfois:


openbas-api/src/main/java/io/openbas/executors/tanium/client/TaniumExecutorClient.java
```diff
@@ -37,82 +42,86 @@ public class TaniumExecutorClient {
...
-              + "        eidLastSeen\n"
-              + "        os { platform }\n"
-              + "        processor { architecture }\n"
-              + "      }\n"
-              + "    }\n"
-              + "  }\n"
-              + "}";
+          String.format(
+              """
+                  query {
```
antoinemzs, line 55: You added the `query` keyword; hard to tell from the absence of PR description if that was part of the issue?

openbas-api/src/main/java/io/openbas/executors/tanium/client/TaniumExecutorClient.java
```diff
@@ -37,82 +42,86 @@ public class TaniumExecutorClient {
...
-          e);
-      throw new RuntimeException(e);
+
+      return response.data;
     } catch (IOException e) {
-      log.error(
-          String.format("I/O error occurred during API request. Error: %s", e.getMessage()), e);
-      throw new RuntimeException(e);
-    } catch (Exception e) {
-      log.error(String.format("Unexpected error occurred. Error: %s", e.getMessage()), e);
```
antoinemzs, line 87: Hard for me to judge whether we would have liked an explicit log of a crash or let spring log it as part of the ulterior runner crash (as you removed the catch-all).

I notice we may now throw an exception earlier, so the catch-all would have caught it immediately.

What do you think?

openbas-api/src/main/java/io/openbas/executors/tanium/client/TaniumExecutorClient.java
```diff
@@ -37,82 +42,86 @@ public class TaniumExecutorClient {
...
+
+      GraphQLResponse<DataEndpoints> response =
+          objectMapper.readValue(jsonResponse, new TypeReference<>() {});
+
+      if (response == null || response.data == null) {
+        throw new RuntimeException("API response malformed or empty");
       }
-      return this.objectMapper.readValue(jsonResponse, new TypeReference<>() {});
-    } catch (JsonProcessingException e) {
-      log.error(
```
antoinemzs: good catch that a `JsonProcessingException` is an `IOException`. Is the default message clear enough to not need the (now deleted) extra verbiage?

openbas-api/src/main/java/io/openbas/executors/tanium/client/TaniumExecutorClient.java
```diff
@@ -124,13 +133,32 @@ private String post(@NotNull final Map<String, Object> body) throws IOException
...
-      httpPost.setEntity(entity);
-      return httpClient.execute(httpPost, response -> EntityUtils.toString(response.getEntity()));
+      String json = this.objectMapper.writeValueAsString(body);
+      httpPost.setEntity(new StringEntity(json, StandardCharsets.UTF_8));
+
+      return httpClient.execute(
+          httpPost,
+          (ClassicHttpResponse response) -> {
+            int status = response.getCode();
+            String result = EntityUtils.toString(response.getEntity(), StandardCharsets.UTF_8);
```
antoinemzs, line 149: Charset selection could be part of the Content-Type (`Content-Type: application/json; charset=utf-8`) header but I don't know if the server honours it.

openbas-api/src/main/java/io/openbas/executors/tanium/client/TaniumExecutorClient.java
RomuDeuxfois: Not really, it works with and without the query word

openbas-api/src/main/java/io/openbas/executors/tanium/client/TaniumExecutorClient.java
RomuDeuxfois: Good point, I will rollback it

openbas-api/src/main/java/io/openbas/executors/tanium/client/TaniumExecutorClient.java
RomuDeuxfois: I am not sure to understand your point..

openbas-api/src/main/java/io/openbas/executors/tanium/client/TaniumExecutorClient.java
RomuDeuxfois: Can I do both ? "ceinture et bretelles"

openbas-api/src/main/java/io/openbas/executors/tanium/client/TaniumExecutorClient.java
antoinemzs: Any exception types other than JsonProcessingException and IOException are now explicitely handled, meanin g other exception types will now bubble upwards without the explicit logging (removed). It will be up to possible higher handlers to log (or not) the exception.

It's not the biggest deal since there was nothing else but a log in this handler.

antoinemzs: LEt's leave it

openbas-api/src/main/java/io/openbas/executors/tanium/client/TaniumExecutorClient.java
antoinemzs: in this scenario, specifying the charset in the request header would instruct the server to use utf8 to decode the request -- and we don't know whether this influences the response.

I'd say let's not bother if it's not broken.

openbas-api/src/main/java/io/openbas/executors/tanium/client/TaniumExecutorClient.java
RomuDeuxfois: Okay, I understand. Thanks for taking the time to re-explain your point.
I can re integrateall the previous exceptions if necessary; it's not a big deal.

PR #3708, opened 2025-07-24
[backend] fix TLS handling with custom certs

antoinemzs:
### Proposed changes

* Implement correct error handling and cert verification

### Testing Instructions

Demonstrated in automated tests.

### Checklist

### Further comments


openbas-api/src/main/java/io/openbas/authorisation/TlsConfig.java
```diff
@@ -93,42 +89,91 @@ public X509TrustManager tlsContextCustom() throws Exception {
...
+      @Override
+      public void checkServerTrusted(X509Certificate[] chain, String authType)
+          throws CertificateException {
+        try {
+          defaultX509CertificateTrustManager.checkServerTrusted(chain, authType);
+        } catch (Exception e) {
+          if (additionalTrustManager.isPresent()) {
+            additionalTrustManager.get().checkServerTrusted(chain, authType);
+          } else {
+            throw e;
```
damgouj, line 159: Don't log the exception ?

openbas-api/src/main/java/io/openbas/authorisation/TlsConfig.java
```diff
@@ -93,42 +89,91 @@ public X509TrustManager tlsContextCustom() throws Exception {
...
+      @Override
+      public void checkClientTrusted(X509Certificate[] chain, String authType)
+          throws CertificateException {
+        try {
+          defaultX509CertificateTrustManager.checkClientTrusted(chain, authType);
+        } catch (Exception e) {
+          if (additionalTrustManager.isPresent()) {
+            additionalTrustManager.get().checkClientTrusted(chain, authType);
+          } else {
+            throw e;
```
damgouj, line 173: Don't log the exception ?

openbas-api/src/test/java/io/openbas/config/tls/client/TlsTrustedCertificateTest.java
```diff
@@ -0,0 +1,58 @@
...
+@TestPropertySource(
+    properties = {"openbas.extra-trusted-certs-dir=src/test/resources/tls/localhost"})
+public class TlsTrustedCertificateTest extends IntegrationTest {
+  private ClientAndServer mockServer;
+  @Autowired private HttpClientFactory httpClientFactory;
+
+  @BeforeAll
+  public void before() {
+    mockServer =
+        new ClientAndServer(
```
damgouj, line 35: Suggestion: the before is the same in this class and the class below + there are only 2 tests, could you merge those two together ?

antoinemzs: I would but I'm currently stuck with how Spring Boot Test handles "Test properties". One class has a `@TestProperty` annotation, and not the other, causing one context to have extra certs (trusted) and the other not (untrusted).

I couldn't make them share the setup a `@Nested` classes -- the test properties are not well initialised (i.e. the "trusted cert" context would fail to load the certificates for lack of proper property initialisation).

openbas-api/src/main/java/io/openbas/authorisation/TlsConfig.java
antoinemzs: Unless this is handled and swallowed upwards (which it is not), it will show in the standard server logs. I usually recommend doing an explicit logging of an exception:
* if it is disappeared (swallowed)
* if there is anything interesting to add to the message

Are you thinking about something I missed?

PR #3713, opened 2025-07-25
[backend] fix: changed lock system for inject status update from implants

impolitepanda:
### Proposed changes

* Changed the lock system to a more robust one
* Added log for debug purposes

### Testing Instructions

1. Launch a salt typhoon simulation where every inject has 2 assets with only crowdstrike agents on them

### Related issues

* Relates #3550 

### Checklist

### DIFF
--- a/openbas-api/src/main/java/io/openbas/rest/inject/service/InjectStatusService.java
+++ b/openbas-api/src/main/java/io/openbas/rest/inject/service/InjectStatusService.java
@@ -18,8 +18,11 @@ import jakarta.validation.constraints.NotNull;
 import java.time.Instant;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
+import org.springframework.scheduling.annotation.Scheduled;
 import org.springframework.stereotype.Service;
 
 @RequiredArgsConstructor
@@ -80,6 +83,13 @@ public class InjectStatusService {
   public boolean isAllInjectAgentsExecuted(Inject inject) {
     int totalCompleteTrace = getCompleteTrace(inject);
     List<Agent> agents = this.injectService.getAgentsByInject(inject);
+    log.info(
+        "[issue/2797] Inputs for inject ID: "
+            + inject.getId()
+            + "::agents.size="
+            + agents.size()
+            + "::totalCompleteTrace="
+            + totalCompleteTrace);
     return agents.size() == totalCompleteTrace;
   }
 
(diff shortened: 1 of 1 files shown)


damgouj (approved):

openbas-api/src/main/java/io/openbas/rest/inject/InjectApi.java
```diff
@@ -345,7 +350,17 @@ public void injectExecutionCallback(
...
+    log.debug("Waiting for lock for inject ID: " + injectId);
+    Lock lock = stripedLocks.get(injectId);
+    lock.lock();
+    try {
+      log.debug("Acquired lock for inject ID: " + injectId);
+      injectExecutionService.handleInjectExecutionCallback(injectId, agentId, input);
+    } finally {
+      lock.unlock();
+      log.debug("Released lock for inject ID: " + injectId);
+    }
```
antoinemzs: In general I'm a bit wary of just putting locks on web-based polls, if only because congestion may cause these same polls to timeout, requiring the client to implement a retry mechanism. If concurrency is an issue we should eventually offload this to a background queue for quick endpoint responses and controlled serialised processing.

openbas-api/src/main/resources/application.properties
```diff
@@ -225,7 +225,7 @@ executor.crowdstrike.unix-script-name=OpenBAS Subprocessor (Unix)
 
 # valid values: local | repository
 # default: local
-executor.openbas.binaries.origin=local
+executor.openbas.binaries.origin=repository
```
antoinemzs: Unsure that this is what we want for a default setting

openbas-api/src/main/java/io/openbas/rest/inject/InjectApi.java
antoinemzs: I accept that this was effectively the case before, so from an API client POV, it's no worse than before.

antoinemzs (commented):
Also worth considering is that removing the lock in `InjectStatusService.updateFinalInjectStatus` may have impacts to clients of that method. Please ensure that this is safe to do.

openbas-api/src/main/java/io/openbas/rest/inject/InjectApi.java
savacano28: We have another ticket related to batching execution traces using queues (and that could be a better sln).

antoinemzs: Yes good point, I had reviewed the first iteration but the reimplementation is not near production just yet, so we need to cope with such locks until then... :/

impolitepanda: Agreed on all points, and indeed, this is definitely a temporary solution, while waiting for the queue system :)

openbas-api/src/main/resources/application.properties
impolitepanda: Absolutely not. It was used for testing purposes for a test feature branch. I removed it.

PR #3737, opened 2025-07-31
[backend] Code Refactoring

RomuDeuxfois:
### Proposed changes

* Code refactoring of this element to follow java/spring and our best practices:
* * regex group
* * output parser
* * contract output element
* * Payload: Update, Upsert, create

### How to test it

Payloads - Injects - Simulations
* Create/update payload with and without output parser, with and without remediation
* Upsert payload thanks to atomic red team collector
* Execution (inject with payload: command line, executable, drop file..)
* Import/export
* Duplication


openbas-api/src/main/java/io/openbas/rest/payload/PayloadUtils.java
```diff
@@ -121,6 +122,17 @@ public static void validateArchitecture(String payloadType, Payload.PAYLOAD_EXEC
     }
   }
 
+  public static Payload instantiatePayload(PayloadType type) {
+    return switch (type) {
```
impolitepanda: We could remove the switch here by encapsulating the payload directly in the enum (creating a payload var in the enum, with a getNewPayloadInstance, for example)

openbas-api/src/main/java/io/openbas/rest/payload/PayloadUtils.java
```diff
@@ -146,39 +158,51 @@ public <T extends Payload> void duplicateCommonProperties(
...
     }
   }
 
-  public Payload copyProperties(Object payloadInput, Payload target, boolean copyId) {
+  public Payload copyProperties(PayloadCreateInput payloadInput, Payload target) {
+    if (payloadInput == null) {
+      throw new IllegalArgumentException("Input payload cannot be null");
+    }
+    BeanUtils.copyProperties(
+        payloadInput, target, "outputParsers", "tags", "attackPatterns", "detectionRemediations");
```
impolitepanda, line 162: I don't have an alternative to propose, but this is dangerous. If in the future someone rename one of the ignored fields, it will not be ignored anymore.

openbas-api/src/main/java/io/openbas/rest/payload/output_parser/OutputParserUtils.java
```diff
@@ -0,0 +1,27 @@
...
+    return copy;
+  }
+
+  /**
+   * Copies properties from the given {@link OutputParserInput} into the target {@link
+   * OutputParser}.
+   */
+  public static OutputParser copyFromInput(
+      @NotNull final OutputParserInput input, @NotNull final OutputParser target) {
+    BeanUtils.copyProperties(input, target, "id", "contractOutputElements");
```
savacano28, line 24: Maybe we should check if we really need to add these ignored properties?

openbas-api/src/main/java/io/openbas/rest/payload/PayloadUtils.java
impolitepanda: Same for lots of places in the whole codebase, so could be discussed somewhere else.

openbas-api/src/main/java/io/openbas/rest/payload/PayloadUtils.java
```diff
@@ -146,39 +158,51 @@ public <T extends Payload> void duplicateCommonProperties(
...
-      detectionRemediationUtils.copy(
-          ((PayloadUpsertInput) payloadInput).getDetectionRemediations(), target, copyId);
-    } else {
-      throw new IllegalArgumentException("Unsupported payload input type");
+    outputParserService.copyOutputParsersFromInput((payloadInput).getOutputParsers(), target);
+    detectionRemediationUtils.copy((payloadInput).getDetectionRemediations(), target, true);
+    return target;
+  }
+
+  public Payload copyProperties(
```
impolitepanda, line 182: Could we try and merge all 3 copyProperties methods into 1 using generics ? like public < U extends PayloadInput > Payload copyProperties(U payloadInput,
      Payload target,
      boolean copyId) ?

openbas-api/src/main/java/io/openbas/rest/payload/output_parser/OutputParserUtils.java
RomuDeuxfois: We should avoid copying these properties due to Hibernate sensitivity:
* ID and relationships can cause detach errors.

openbas-api/src/main/java/io/openbas/rest/payload/PayloadUtils.java
RomuDeuxfois: Yes I am not a big fan either.
But I am not able to propose something else for now..

openbas-api/src/main/java/io/openbas/rest/payload/PayloadUtils.java
RomuDeuxfois: Something like that ?

public enum PayloadType {
    COMMAND {
        @Override
        public Payload getNewPayloadInstance() {
            return new Command();
        }
    },
    EXECUTABLE {
        @Override
        public Payload getNewPayloadInstance() {
            return new Executable();
        }
    },
    FILE_DROP {
        @Override
        public Payload getNewPayloadInstance() {
            return new FileDrop();
        }
    },
    DNS_RESOLUTION {
        @Override
        public Payload getNewPayloadInstance() {
            return new DnsResolution();
        }
    },
    NETWORK_TRAFFIC {
        @Override
        public Payload getNewPayloadInstance() {
            return new NetworkTraffic();
        }
    };

    public abstract Payload getNewPayloadInstance();
}

openbas-api/src/main/java/io/openbas/rest/payload/PayloadUtils.java
RomuDeuxfois: PayloadCreateInput, PayloadUpdateInput and PayloadUpsertInput are not linked at all..

savacano28:
Tested OK

openbas-api/src/main/java/io/openbas/rest/payload/PayloadUtils.java
impolitepanda: More like this, but same idea:
public enum PayloadType {
  COMMAND(Command::new),
  EXECUTABLE(Executable::new),
  FILE_DROP(FileDrop::new),
  DNS_RESOLUTION(DnsResolution::new),
  NETWORK_TRAFFIC(NetworkTraffic::new);

  private final Supplier<Payload> payloadSupplier;

  PayloadType(Supplier<Payload> payloadSupplier) {
    this.payloadSupplier = payloadSupplier;
  }

  public Payload getNewInstance() {
    return this.payloadSupplier.get();
  }
}

openbas-api/src/main/java/io/openbas/rest/payload/PayloadUtils.java
impolitepanda: Right, I assumed we could refactor them to have a superclass like PayloadInput as all three classes have a lot of fields in common. But I didn't check that the @ NotBlank where the same, sorry.

PR #3745, opened 2025-07-31
Issue/375 stream

heditar:
### Proposed changes

*
*

### Testing Instructions

1. Step-by-step how to test

### Related issues

* Closes #ISSUE-NUMBER
*

### Checklist

### Further comments


openbas-model/src/main/java/io/openbas/database/model/Agent.java
```diff
@@ -140,6 +140,8 @@ public boolean isActive() {
   @JsonProperty("agent_cleared_at")
   private Instant clearedAt = now();
 
+  @Transient private final ResourceType resourceType = ResourceType.AGENT;
```
impolitepanda: If we @ Transient this field to not save it in the database, should we also @ JsonIgnore it to remove it from the JSON view ? Or do you think it has value for the API clients ? (this comments apply to every resourceType fields in our entities added with this PR)

openbas-model/src/main/java/io/openbas/database/model/Base.java
```diff
@@ -28,4 +28,8 @@ default void setUpdateAttributes(Object input) {
   default boolean isListened() {
     return true;
   }
+
+  default ResourceType getResourceType() {
```
impolitepanda, line 39: I don't know if I like the idea of having a default implementation for this. Letting it be an interface method instead will force people to redefine it in every implementation of Base interface, thus making it so that we never forget, during coding or reviewing, to add and, most importantly, think about where does my resource fit within the RBAC system.

impolitepanda: I'd also like to see a comment explaining what this field is. resourceType is a very generic name, and to someone new to our codebase, it could be hard to see that it's linked to RBAC.

heditar: I will add some doc, for the default implementation I didn't want to force the class to have a n implementation, lot of the existing models will use the default implementation.

openbas-model/src/main/java/io/openbas/database/model/Agent.java
heditar: if we want to add @JsonIgnore that should be on the getter, I will add it

impolitepanda (approved):

PR #3752, opened 2025-08-01
[backend] Issue/375 RBAC Capa hierarchy 

### LINKED ISSUE #3722: RBAC: Chunk 1.2: Implement capa hierarchy in the backend
Refer to https://www.notion.so/filigran/RBAC-Implement-roles-capability-and-apply-them-to-grants-7ae145c285b64762beddfa30d07e7f99

heditar:
### Proposed changes

* Adding the concept "parent" in the capability enum so we can force the fact that when you add a "MANAGE" cap you will also get the "ACCESS" capa. It is already enforced in the front

### Testing Instructions

1. Step-by-step how to test

### Related issues

* Closes #3722 
* #375 

### Checklist

### DIFF
--- a/openbas-api/src/main/java/io/openbas/service/RoleService.java
+++ b/openbas-api/src/main/java/io/openbas/service/RoleService.java
@@ -2,6 +2,7 @@ package io.openbas.service;
 
 import static io.openbas.utils.pagination.PaginationUtils.buildPaginationJPA;
 
+import com.google.common.annotations.VisibleForTesting;
 import io.openbas.database.model.Capability;
 import io.openbas.database.model.Role;
 import io.openbas.database.repository.RoleRepository;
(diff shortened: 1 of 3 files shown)


openbas-api/src/main/java/io/openbas/service/RoleService.java
```diff
@@ -70,4 +71,24 @@ public void deleteRole(@NotBlank final String roleId) {
...
     roleRepository.deleteById(roleId);
   }
+
+  /**
+   * Get a set of capabilities as input and return a set containing the input + their parent
+   *
+   * @param capabilitiesInput
+   * @return
+   */
+  @VisibleForTesting
```
impolitepanda, line 81: Not a fan of this annotaiton at all. At the same time, the method needs to be tested and doens't need to be public, so.... Let's go with it none the less.

impolitepanda (approved):

PR #3762, opened 2025-08-06
[backend/frontend] Onboarding

RomuDeuxfois:
### Related issues

* Closes https://github.com/OpenBAS-Platform/openbas/issues/3513

### DIFF
--- a/openbas-front/src/admin/components/settings/Parameters.tsx
+++ b/openbas-front/src/admin/components/settings/Parameters.tsx
@@ -359,13 +359,13 @@ const Parameters = () => {
                 <ItemBoolean status={null} variant="large" neutralLabel={settings?.rabbitmq_version} />
               </ListItem>
               {settings.analytics_engine_type
-            && (
+                && (
+                  <ListItem divider>
+                    <ListItemText primary={t(settings.analytics_engine_type)} />
+                    <ItemBoolean status={null} variant="large" neutralLabel={settings?.analytics_engine_version} />
+                  </ListItem>
+                )}
               <ListItem divider>
-                <ListItemText primary={t(settings.analytics_engine_type)} />
-                <ItemBoolean status={null} variant="large" neutralLabel={settings?.analytics_engine_version} />
-              </ListItem>
-            )}
-          <ListItem divider>
                 <ListItemText primary={t('Telemetry manager')} />
                 <ItemBoolean status={settings?.telemetry_manager_enable} variant="large" label={settings?.telemetry_manager_enable ? t('Enable') : t('Disabled')} />
               </ListItem>


openbas-front/src/admin/components/onboarding/MenuItemOnboarding.tsx
```diff
@@ -0,0 +1,114 @@
...
+  };
+
+  return (
+    <Tooltip title={navOpen ? '' : `${progress}/${total} - ${t('onboarding_getting_started')}`} arrow disableHoverListener={navOpen}>
+      <Box
+        onClick={handleToggleSidebar}
+        sx={{
+          bgcolor: hexToRGB(theme.palette.primary.main, 0.24),
+          padding: 2,
+          marginBottom: 2,
```
camrrx, line 62: You should use theme.spacing for padding and marginBottom ?

RomuDeuxfois: Magic with MUI
padding: 2 ≃ padding: theme.spacing(2)

camrrx:
The first time I tested it, when a step was completed, the progress number was not updated. You had to click on “discover openbas” to synchronize it. Since it works well, I can't reproduce this use case.

camrrx:
The "enable onboarding" on the user form (profile) is not working, if you set a value, update it, then exit and go back on profile : the value is not saved

openbas-front/src/admin/components/profile/UserForm.tsx
```diff
@@ -0,0 +1,93 @@
...
+    mode: 'onTouched',
+    resolver: zodResolver(
+      zodImplement<UpdateProfileInput>().with({
+        user_email: z.string().email(t('Should be a valid email address')),
+        user_firstname: z.string().min(1, { message: t('Should not be empty') }),
+        user_lastname: z.string().min(1, { message: t('Should not be empty') }),
+        user_organization: z.string().optional(),
+        user_country: z.string().optional(),
+        user_theme: z.string().min(1, { message: t('Should not be empty') }),
+        user_lang: z.string().min(1, { message: t('Should not be empty') }),
```
camrrx, line 39: "user_onboarding_enable" is missing ?

openbas-api/src/main/java/io/openbas/rest/user/MeApi.java
```diff
@@ -118,6 +121,18 @@ public User updatePassword(@Valid @RequestBody UpdateMePasswordInput input)
     }
   }
 
+  @Secured(ROLE_USER)
+  @PutMapping(ME_URI + "/onboarding")
+  public User updateOnboarding(@Valid @RequestBody UpdateOnboardingInput input) {
+    User user =
```
savacano28: Maybe you should use a Service layer

openbas-front/src/admin/components/profile/Index.js
```diff
@@ -86,11 +79,13 @@ const Index = () => {
...
         </Typography>
-        <UserForm
-          organizations={R.values(organizationsMap)}
-          onSubmit={onUpdate}
-          initialValues={initialValues}
-        />
+        <UserForm onSubmit={onUpdate} initialValues={initialValues} />
+      </Paper>
+      <Paper>
+        <Typography variant="h1" style={{ marginBottom: 20 }}>
```
savacano28, line 85: theme? gutterBottom?

openbas-front/src/components/fields/CountryFieldController.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { Autocomplete as MuiAutocomplete, Box, TextField } from '@mui/material';
+import type { FunctionComponent } from 'react';
+import { Controller, useFormContext } from 'react-hook-form';
+import { makeStyles } from 'tss-react/mui';
+
+import { countryOptions } from '../../utils/Option';
+
+const useStyles = makeStyles()(() => ({
+  icon: {
+    paddingTop: 4,
```
savacano28: Maybe could we pass theme here?

openbas-front/src/components/fields/OrganizationFieldController.tsx
```diff
@@ -0,0 +1,140 @@
...
+  const { organizations }: { organizations: Organization[] } = useHelper((helper: OrganizationHelper) => ({ organizations: helper.getOrganizations() }));
+
+  useDataLoader(() => {
+    dispatch(fetchOrganizations());
+  });
+
+  const [open, setOpen] = useState(false);
+
+  const handleSubmit = async (data: OrganizationCreateInput) => {
+    const inputValues = R.pipe(
```
savacano28: Maybe we should avoid using ramda

openbas-front/src/constants/BaseUrls.ts
```diff
@@ -1,10 +1,27 @@
...
 export const SCENARIO_BASE_URL = '/admin/scenarios';
+export const SIMULATION_BASE_URL = '/admin/simulations';
+export const ATOMIC_BASE_URL = '/admin/atomic_testings';
+
+// Assets
 export const ENDPOINT_BASE_URL = '/admin/assets/endpoints';
-export const PAYLOAD_BASE_URL = '/admin/payloads';
 export const SECURITY_PLATFORM_BASE_URL = '/admin/assets/security_platforms';
+
+// Peoples
```
savacano28, line 13: peoples

openbas-front/src/utils/lang/en.json
```diff
@@ -1042,6 +1043,31 @@
...
+  "onboarding_collector_setup": "Set up a collector",
+  "onboarding_launch_scenario_get_started": "Launch a scenario and understand your results",
+  "onboarding_oaev_x_octi ": "OAEV x OCTI",
+  "onboarding_rbac_get_started": "RBAC",
+  "onboarding_display_contextual_buttons": "Display ‘How to..?’ Buttons",
+  "onboarding_contextual_buttons": "Contextual Help Buttons",
+  "onboarding_display_widget": "Display Onboarding Widget",
+  "onboarding_process": "Onboarding process",
+  "openbas_welcome": "Welcome",
+  "openbas_first_time": "1st Time on OpenBAS ? Let’s take a Look.",
```
savacano28: Could be: First time on OpenBA**S?** Let's take a look!

openbas-api/src/main/java/io/openbas/migration/V4_16__User_Onboarding_Enable.java
```diff
@@ -0,0 +1,35 @@
...
+
+  @Override
+  public void migrate(Context context) throws Exception {
+    Connection connection = context.getConnection();
+    try (Statement statement = connection.createStatement()) {
+      statement.execute(
+          """
+              ALTER TABLE users ADD COLUMN user_onboarding_widget_enable varchar(255) default 'DEFAULT';
+              ALTER TABLE users ADD COLUMN user_onboarding_contextual_help_enable varchar(255) default 'DEFAULT';
+              CREATE TABLE user_onboarding_progress (
```
savacano28: I think we should use: .._progresses ?

openbas-api/src/main/java/io/openbas/config/SessionHelper.java
```diff
@@ -4,6 +4,8 @@
 
 public class SessionHelper {
 
+  private SessionHelper() {}
```
antoinemzs, line 7: Lombok annotation `@NoArgsConstructor` does the same

openbas-api/src/main/java/io/openbas/api/onboarding/OnboardingApi.java
```diff
@@ -0,0 +1,40 @@
...
+
+  @GetMapping(ONBOARDING_URI + "/config")
+  public List<OnboardingConfig.OnboardingCategoryDTO> getOnboardingConfig() {
+    return OnboardingConfig.getOnboardingConfig();
+  }
+
+  @PutMapping(ONBOARDING_URI + "/skipped")
+  public UserOnboardingProgress skippedCategory(@RequestBody @NotNull final List<String> steps) {
+    return this.onboardingService.skippedSkipped(steps);
+  }
```
antoinemzs: Consider some OpenAPI documentation with annotations

openbas-model/src/main/java/io/openbas/utils/OnboardingConfig.java
```diff
@@ -0,0 +1,51 @@
...
+  public static final String TABLE_TOP_SETUP_ICON = "people";
+  public static final String GET_STARTED_ICON = "rocket";
+
+  public static final String VIDEO_URI = "https://app.storylane.io/demo/bxqijbtlfklz";
+
+  public static final String SCENARIO_BASE_URL = "/admin/scenarios";
+  public static final String ENDPOINT_BASE_URL = "/admin/assets/endpoints";
+  public static final String PLAYER_BASE_URL = "/admin/teams/players";
+  public static final String TEAM_BASE_URL = "/admin/teams/teams";
+  public static final String COLLECTOR_BASE_URL = "/admin/integrations/collectors";
```
antoinemzs: That's a bit of leaking the front into the back; these should be perhaps tokens rendered in the front to expand the actual URLs. What do you think?

openbas-api/src/main/java/io/openbas/migration/V4_16__User_Onboarding_Enable.java
```diff
@@ -0,0 +1,35 @@
...
+                onboarding_id UUID PRIMARY KEY,
+                user_id varchar(256) NOT NULL UNIQUE REFERENCES users(user_id) ON DELETE CASCADE,
+                onboarding_created_at TIMESTAMP WITH TIME ZONE DEFAULT now(),
+                onboarding_updated_at TIMESTAMP WITH TIME ZONE DEFAULT now()
+              );
+              CREATE TABLE user_onboarding_steps (
+                onboarding_id UUID NOT NULL REFERENCES user_onboarding_progress(onboarding_id) ON DELETE CASCADE,
+                step TEXT NOT NULL,
+                completed BOOLEAN NOT NULL DEFAULT FALSE,
+                skipped BOOLEAN NOT NULL DEFAULT FALSE
```
antoinemzs: Is it the case that we can have a step both completed and skipped?

antoinemzs: Column names don't follow the standard of `tablename_columnname`

openbas-front/src/admin/components/profile/UserForm.tsx
RomuDeuxfois: The enable onboarding field is not suppose to be here. I forgot to remove it.
Everything has been moved to the Help Settings section.

openbas-api/src/main/java/io/openbas/migration/V4_16__User_Onboarding_Enable.java
antoinemzs: Tagentially, and non blocking, why such a long name for the entity instead of just "onboarding"?

openbas-api/src/main/java/io/openbas/rest/agent/AgentApi.java
antoinemzs, line 1: :+1:

openbas-api/src/main/java/io/openbas/service/onboarding/OnboardingService.java
antoinemzs, line 1: Generally I'd rather have the service class methods take arguments instead of relying on an ambient session. The session should be relevant at the request layer, i.e. the API class or at the very least the AOP method. What do you think?

openbas-front/src/admin/components/profile/Index.js
RomuDeuxfois: I couldn’t refactor the entire file in this round.

openbas-front/src/components/fields/CountryFieldController.tsx
RomuDeuxfois: Good point

openbas-front/src/constants/BaseUrls.ts
RomuDeuxfois: ?

openbas-api/src/test/java/io/openbas/rest/onboarding/MeApiOnboardingTest.java
```diff
@@ -0,0 +1,76 @@
...
+            .getResponse()
+            .getContentAsString();
+
+    // -- ASSERT --
+    assertNotNull(response);
+    String widgetEnable = JsonPath.read(response, "$.user_onboarding_widget_enable");
+    assertEquals("DISABLED", widgetEnable);
+    String contextualHelpEnable =
+        JsonPath.read(response, "$.user_onboarding_contextual_help_enable");
+    assertEquals("DISABLED", contextualHelpEnable);
```
antoinemzs: nitpick: `assertThatJson` is a better tool to assert on json objects generally, it allows for more complete assertions on the objects. It's for future jobs, so let's not spend time changing this now.

savacano28:
@RomuDeuxfois : 
* 6/6 : ... Is almost ready to use.. : I have 6/6 but the message is: **Is almost ready to use**, maybe it should be: Ready to use?
* When I reset options in user side, it should have the default value from user (I think?), and in the case of default it takes Platforme value? I think I dont have these behaviors.

openbas-front/src/constants/BaseUrls.ts
savacano28: I'm not sure about the word: People**s** with **S**?

openbas-api/src/main/java/io/openbas/api/onboarding/OnboardingApi.java
antoinemzs, line 1: Shouldn't this be under `io.openbas.api.rest` ?

openbas-api/src/main/java/io/openbas/service/onboarding/OnboardingService.java
```diff
@@ -0,0 +1,74 @@
...
+      this.saveUserOnboardingProgress(progress);
+    }
+  }
+
+  public UserOnboardingProgress skippedSkipped(@NotNull final List<String> steps) {
+    User user = userService.user(currentUser().getId());
+    UserOnboardingProgress progress = getOrDefault(user);
+
+    List<UserOnboardingStepStatus> stepStatuses =
+        steps.stream().map(s -> progress.getProgressMap().get(s)).toList();
```
antoinemzs, line 53:
```suggestion
        steps.values().stream().toList();
```

savacano28:
@RomuDeuxfois Is it normal to have so many calls to /onboarding (admin/setting/parameteres): 
image

openbas-api/src/main/java/io/openbas/service/onboarding/OnboardingService.java
```diff
@@ -0,0 +1,74 @@
...
+    UserOnboardingProgress progress = getOrDefault(user);
+
+    UserOnboardingStepStatus stepStatus = progress.getProgressMap().get(step);
+    if (stepStatus != null && !stepStatus.isCompleted()) {
+      stepStatus.setCompleted(true);
+      this.saveUserOnboardingProgress(progress);
+    }
+  }
+
+  public UserOnboardingProgress skippedSkipped(@NotNull final List<String> steps) {
```
antoinemzs: I don't get the method's name: should this be `skipSteps` ?

openbas-api/src/main/java/io/openbas/service/onboarding/OnboardingService.java
```diff
@@ -0,0 +1,74 @@
...
+    User user = userService.user(currentUser().getId());
+    UserOnboardingProgress progress = getOrDefault(user);
+
+    List<UserOnboardingStepStatus> stepStatuses =
+        steps.stream().map(s -> progress.getProgressMap().get(s)).toList();
+    if (!stepStatuses.isEmpty()) {
+      stepStatuses.forEach(s -> s.setSkipped(true));
+      user.setOnboardingProgress(progress);
+      userService.updateUser(user);
+    }
```
antoinemzs: This would put a completed step in the possible state of both `completed == true && skipped == true`. Consider whether this is OK, or if skipping an already completed step should be disallowed.

openbas-api/src/test/java/io/openbas/rest/onboarding/OnboardingApiTest.java
```diff
@@ -0,0 +1,175 @@
...
+    // -- EXECUTE --
+    mvc.perform(
+            post(ENDPOINT_URI + "/agentless")
+                .content(asJsonString(endpointInput))
+                .contentType(MediaType.APPLICATION_JSON)
+                .accept(MediaType.APPLICATION_JSON))
+        .andExpect(status().is2xxSuccessful())
+        .andReturn()
+        .getResponse()
+        .getContentAsString();
```
antoinemzs: No need to call these if part of the test's setup

openbas-api/src/main/java/io/openbas/api/onboarding/OnboardingApi.java
```diff
@@ -0,0 +1,40 @@
...
+    return onboardingService.userOnboardingProgress();
+  }
+
+  @GetMapping(ONBOARDING_URI + "/config")
+  public List<OnboardingConfig.OnboardingCategoryDTO> getOnboardingConfig() {
+    return OnboardingConfig.getOnboardingConfig();
+  }
+
+  @PutMapping(ONBOARDING_URI + "/skipped")
+  public UserOnboardingProgress skippedCategory(@RequestBody @NotNull final List<String> steps) {
```
antoinemzs: I think we should have an input object here, that serialises the list of strings in a member. A bare list of strings makes it more difficult to expand functionality in a backwards compatible manner. Thoughts?

openbas-front/src/components/common/dialog/FiligranDialog.tsx
```diff
@@ -0,0 +1,58 @@
...
+  handleClose: () => void;
+  children: (() => ReactElement) | ReactElement | null;
+  title?: ReactNode;
+  maxWidth?: Breakpoint;
+  className?: string;
+  actions?: ReactElement | null;
+  showCloseIcon?: boolean;
+}
+
+const FiligranDialog: FunctionComponent<DialogProps> = ({
```
antoinemzs: I think I remember we wanted to not call components OpenBAS* or Filigran*

What was wrong with just "Dialog"?

openbas-front/src/admin/components/assets/endpoints/EndpointForm.tsx
```diff
@@ -35,7 +35,7 @@ const EndpointForm: FunctionComponent<Props> = ({
     endpoint_hostname: '',
     endpoint_ips: [],
     endpoint_mac_addresses: [],
-    endpoint_platform: undefined,
+    endpoint_platform: 'Linux',
```
antoinemzs, line 38: Excellent choice for a default if you ask me!

openbas-front/src/components/common/dialog/FiligranDialog.tsx
RomuDeuxfois: The name Dialog is already used by MUI, so it’s unclear which one the file is referencing.

openbas-api/src/main/java/io/openbas/api/onboarding/OnboardingApi.java
RomuDeuxfois: Good point, I will do it.

openbas-front/src/constants/BaseUrls.ts
RomuDeuxfois: This is the category name we have on the left menu.

openbas-api/src/main/java/io/openbas/config/SessionHelper.java
RomuDeuxfois: Yep but SonarLint don't handle lombok in this case :/
image

openbas-model/src/main/java/io/openbas/utils/OnboardingConfig.java
RomuDeuxfois: Could you share a small code example to illustrate your point? I’m not familiar with this mechanism

openbas-api/src/main/java/io/openbas/migration/V4_16__User_Onboarding_Enable.java
RomuDeuxfois: Yes, we can.
But completed is more powerfull than skipped.

RomuDeuxfois: I want to be specific in case we have more onboarding object to come.

openbas-api/src/main/java/io/openbas/service/onboarding/OnboardingService.java
RomuDeuxfois: I understand your point of view. Personally, I’d rather bind the service to the session if that’s the only viable way to prevent misuse of the service.

openbas-api/src/main/java/io/openbas/service/onboarding/OnboardingService.java
RomuDeuxfois: Not sure to understand what you want here.
Steps is a list and don't have a values() method

openbas-api/src/main/java/io/openbas/service/onboarding/OnboardingService.java
RomuDeuxfois: You can complete a skipped step.
Maybe is clearer to unskip it in this case ?

openbas-model/src/main/java/io/openbas/utils/OnboardingConfig.java
antoinemzs: I'm not speaking about a specific bit of code, rather the pattern: these are URLs that are theoretically meaningful in the UI, but not in the back.
What I'm suggesting specifically is that we should convey the _notion_ that there should be a destination URL to some location, but we shouldn't prescribe the actual URL that any frontend

openbas-api/src/main/java/io/openbas/service/onboarding/OnboardingService.java
antoinemzs: ofc i got confused nvm

openbas-front/src/components/common/dialog/FiligranDialog.tsx
antoinemzs: I think whenever names clash in imports, we should alias them locally rather than encoding a trademark in the class name?

openbas-model/src/main/java/io/openbas/utils/OnboardingConfig.java
RomuDeuxfois: This is why you are talking about tokens.
And have a map on front-end ?

openbas-front/src/components/common/dialog/FiligranDialog.tsx
RomuDeuxfois: I don't have a strong opinion on this. @guillaumejparis what do you think ?

openbas-api/src/main/java/io/openbas/service/onboarding/OnboardingService.java
antoinemzs: What kind of misuse ? If anything it precludes implementing hypothetical admin features (e.g. acting on a user's behalf) but at this point it's more of a talk of design purity. It also requires mocking a session to test it, although we have tooling for this.

openbas-api/src/main/java/io/openbas/service/onboarding/OnboardingService.java
antoinemzs: It really sounds to me that these are two mutually exclusive states. That one field overrides the other depends on the code processing the object, as opposed to when it is encoded directly in the data model (e.g. single field with backing state machine).

I don't expect this model to expand too much of course, but it looks a bit odd; also it's exposed to clients via the output object (on the GET operations), so it should probably at least be reconciled there.

openbas-api/src/main/java/io/openbas/service/onboarding/OnboardingService.java
RomuDeuxfois: After some reading, I understand your point, I made chnages.

openbas-api/src/main/java/io/openbas/service/onboarding/OnboardingService.java
RomuDeuxfois: You are right, as a user point of view it's a nonsense.
I made changes.

openbas-front/src/components/fields/CountryFieldController.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+const useStyles = makeStyles()(theme => ({
+  icon: {
+    paddingTop: theme.spacing(1),
+    display: 'inline-block',
+  },
+  text: {
+    display: 'inline-block',
+    flexGrow: 1,
+    marginLeft: 10,
```
savacano28: theme

openbas-front/src/components/fields/OrganizationFieldController.tsx
```diff
@@ -0,0 +1,140 @@
...
+import { addOrganization, fetchOrganizations } from '../../actions/Organization';
+import OrganizationForm from '../../admin/components/teams/organizations/OrganizationForm';
+import { useHelper } from '../../store';
+import { type Organization, type OrganizationCreateInput } from '../../utils/api-types';
+import { useAppDispatch } from '../../utils/hooks';
+import useDataLoader from '../../utils/hooks/useDataLoader';
+import { type Option } from '../../utils/Option';
+import { useFormatter } from '../i18n';
+
+const useStyles = makeStyles()(() => ({
```
savacano28: theme

RomuDeuxfois:
coucou

openbas-api/src/main/java/io/openbas/config/SessionHelper.java
antoinemzs: Ah I missed that the ctor was private, my bad, thanks for pointing it out.

openbas-model/src/main/java/io/openbas/utils/OnboardingConfig.java
antoinemzs: Theoretically yes. I guess a frontend navigation action is not really meaningful for an undefined API client. There is currently a single (recent) example of a backend-generated frontend URL.

So the choice to make here is one of the following:
1. The API does not expose URLs at all. Navigation is implemented in the frontend as a frontend-only behaviour, in a sense that fulfilling the Onboarding step requires to navigate when we are in the OBAS UI, but other clients may implement a different flow
2. The API exposes a hint only (not a URL) to help the client may implement its own URL pattern if applicable (what we are discussing here)
3. We accept to tie the OBAS API with the OBAS frontend and have some API objects expose specific frontend concepts

I'd be of a mind to implement 1., unless there is a need to have a URL in the API?
 (2. which I initially suggested looks like a middle of the road, less bad but also not very good compromise)

openbas-front/src/components/common/dialog/FiligranDialog.tsx
antoinemzs: In Guillaume's absence, he gave an opinion previously here: https://github.com/OpenBAS-Platform/openbas/pull/3499#discussion_r2179334049 if that helps moving the patch forward

PR #3765, opened 2025-08-06
[backend] RBAC adding create assesment capa Issue/375 

heditar:
### Proposed changes

* adding the capa "create assessment" to allow creation of scenarion/simulation

### Testing Instructions

1.  call the api to create a scenario with or without the capa

### Related issues

* #375 
*

### Checklist

### DIFF
--- a/openbas-api/src/main/java/io/openbas/service/PermissionService.java
+++ b/openbas-api/src/main/java/io/openbas/service/PermissionService.java
@@ -54,9 +54,7 @@ public class PermissionService {
     switch (action) {
       case READ:
         return grantService.hasReadGrant(resourceId, user);
-      case WRITE:
-        return grantService.hasWriteGrant(resourceId, user);
-      case DELETE:
+      case WRITE, DELETE:
         return grantService.hasWriteGrant(resourceId, user);
       case LAUNCH:
         return grantService.hasLaunchGrant(resourceId, user);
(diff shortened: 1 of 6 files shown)


openbas-model/src/main/java/io/openbas/database/model/Capability.java
```diff
@@ -14,6 +14,14 @@ public enum Capability {
   // Superuser
   BYPASS(null, pair(null, null)),
 
+  // Assesment
+  CREATE_ASSESSMENT(
+      null,
+      pair(ResourceType.SCENARIO, Action.CREATE),
```
RomuDeuxfois, line 20: Export & delete is handle by grant right ?

openbas-api/src/main/java/io/openbas/service/PermissionService.java
```diff
@@ -50,6 +56,8 @@ private boolean hasGrantPermission(
         return grantService.hasReadGrant(resourceId, user);
       case WRITE:
         return grantService.hasWriteGrant(resourceId, user);
+      case DELETE:
+        return grantService.hasWriteGrant(resourceId, user);
```
RomuDeuxfois:
```suggestion
      case WRITE, DELETE:
        return grantService.hasWriteGrant(resourceId, user);
```

openbas-model/src/main/java/io/openbas/database/model/Capability.java
heditar: yes, that should be linked to the READ action

PR #3790, opened 2025-08-13
[backend] RBAC new permission logic Issue/3748

### LINKED ISSUE #3748: RBAC: Chunk 4.1: Update permisison logic for taxonomies and injects
Taxonomies resources should be accessible by all the users

Article and Inject permission should be based on their parent permission

create/update/delete should be  available with only the manage_settings capa

heditar:
### Proposed changes

* open taxonomies apis
* check parent permission for injects
* add resourceType missing
* link new resource type to capa
*  -> to fix the login I had to open two apis /api/statictics and /api/settings but they should not be open so I created a bug to follow up https://github.com/OpenBAS-Platform/openbas/issues/3789

### Testing Instructions

1. create a user with no admin right and play with the api

### Related issues

* Closes #3748

### Checklist

### DIFF
--- a/openbas-api/src/main/java/io/openbas/service/PermissionService.java
+++ b/openbas-api/src/main/java/io/openbas/service/PermissionService.java
@@ -1,6 +1,7 @@
 package io.openbas.service;
 
 import io.openbas.database.model.*;
+import io.openbas.rest.inject.service.InjectService;
 import jakarta.validation.constraints.NotNull;
 import java.util.EnumSet;
 import java.util.Set;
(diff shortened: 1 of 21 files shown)


openbas-api/src/main/java/io/openbas/service/PermissionService.java
```diff
@@ -14,19 +15,28 @@ public class PermissionService {
 
   // TODO: today settings are necessary to login -> review that
   private static final EnumSet<ResourceType> RESOURCES_OPEN =
-      EnumSet.of(ResourceType.PLAYER, ResourceType.TEAM, ResourceType.PLATFORM_SETTING);
+      EnumSet.of(
+          ResourceType.PLAYER,
+          ResourceType.TEAM,
+          ResourceType.PLATFORM_SETTING,
```
impolitepanda, line 21: PLATFORM_SETTING in the open section is temporary as well, right? If yes, could you add a fixme comment ?

openbas-api/src/main/java/io/openbas/service/PermissionService.java
```diff
@@ -37,16 +47,28 @@ public boolean hasPermission(
       return user.isAdmin();
     }
 
+    // for inject/article the permission will be based on the parent's (scenario/simulation/test)
```
impolitepanda, line 50: Comment says that both injects and articles follow this. However, RESOURCES_USING_PARENT_PERMISSION only contains inject. Should it also contain articles?

openbas-api/src/main/java/io/openbas/service/PermissionService.java
```diff
@@ -85,12 +107,29 @@ private boolean hasCapaPermission(
...
+        && (Action.READ.equals(action) || Action.SEARCH.equals(action))) {
       return true;
     }
 
     Capability requiredCapability = Capability.of(resourceType, action).orElse(Capability.BYPASS);
 
     return userCapabilities.contains(requiredCapability);
   }
+
+  private Target resolveTarget(
```
impolitepanda, line 120: Following my previous comments, this function only works for injects, not articles. If articles are indeed treated like the injects, it should probably be reflected here?

openbas-dev/Frontent start.run.xml
```diff
@@ -1,6 +1,6 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="Frontent start" type="js.build_tools.npm">
-    <package-json value="$PROJECT_DIR$/core-engine/openbas-front/package.json" />
```
impolitepanda, line 3: My bad, sorry for that. It's because I'm on macos, and our project structure is different from the rest of the team.

openbas-api/src/main/java/io/openbas/service/PermissionService.java
heditar: yes it is temporary and I have a todo upper in this section + a github task https://github.com/OpenBAS-Platform/openbas/issues/3789

openbas-api/src/main/java/io/openbas/service/PermissionService.java
heditar: yes that's a miss, I forgot to implement that, I can do a follow up after this PR to add article

Issue #3531, opened 2025-07-03
Apply simply custom dashboards on home, scenarios and simulations


jborozco:
## Context

- Be able to apply custom dashboards on the homepage so you can customize your view based on your needs
- Be able to apply custom dashboards in scenarios
- Have a default dashboard preset in the platform so we still propose a default home page, scenario and simulations 
- Simplify the settings of the custom dashboard global parameters

PR #3793, opened 2025-08-14
[backend/frontend] Add scenario filter (#3531)

RomuDeuxfois:
### Proposed changes

- When creating or editing a **custom dashboard**, the user can now add a **new filter parameter** called `scenario`.
- This `scenario` filter works similarly to the existing simulation filter.
- The user can:
    - Choose the `scenario` filter in any widgets for the simulation dimension.
    - Bind it to any `scenario` param.
- However, when the dashboard is displayed **inside a running simulation**:
    - The `scenario` filter is **hidden** in the UI to prevent manual override.
    - Its value is set to the scenario link to the simulation or to null.

### DIFF
--- a/openbas-front/src/admin/components/workspaces/custom_dashboards/CustomDashboardForm.tsx
+++ b/openbas-front/src/admin/components/workspaces/custom_dashboards/CustomDashboardForm.tsx
@@ -6,6 +6,7 @@ import { type FunctionComponent, useMemo } from 'react';
 import { FormProvider, type SubmitHandler, useFieldArray, useForm } from 'react-hook-form';
 import { z } from 'zod';
 
+import SelectFieldController, { createItems, type Item } from '../../../../components/fields/SelectFieldController';
 import TextFieldController from '../../../../components/fields/TextFieldController';
 import { useFormatter } from '../../../../components/i18n';
 import { type CustomDashboardInput, type CustomDashboardParametersInput } from '../../../../utils/api-types';
@@ -35,7 +36,7 @@ const CustomDashboardForm: FunctionComponent<Props> = ({
   const parametersSchema = z.object({
     custom_dashboards_parameter_id: z.string().optional(),
     custom_dashboards_parameter_name: z.string().min(1, { message: t('Should not be empty') }),
-    custom_dashboards_parameter_type: z.literal('simulation'),
+    custom_dashboards_parameter_type: z.enum(['scenario', 'simulation']),
   });
 
   const validationSchema = useMemo(
@@ -64,11 +65,11 @@ const CustomDashboardForm: FunctionComponent<Props> = ({
     name: 'custom_dashboard_parameters',
   });
 
-  const items: CustomDashboardParametersInput['custom_dashboards_parameter_type'][] = ['simulation'];
+  const items: Item<CustomDashboardParametersInput['custom_dashboards_parameter_type']>[] = createItems(['scenario', 'simulation']);
   const handleAddParameter = (type: CustomDashboardParametersInput['custom_dashboards_parameter_type']) => {
     if (type) {
       append({
-        custom_dashboards_parameter_name: type,
+        custom_dashboards_parameter_name: '',
         custom_dashboards_parameter_type: type,
       });
     }
@@ -108,7 +109,7 @@ const CustomDashboardForm: FunctionComponent<Props> = ({
           <IconButton
             color="secondary"
             aria-label="Add"
-            onClick={() => handleAddParameter(items[0])} // For now, we handle just one type
+            onClick={() => handleAddParameter(items[0].value)}
             size="small"
           >
             <Add fontSize="small" />
@@ -130,13 +131,11 @@ const CustomDashboardForm: FunctionComponent<Props> = ({
               required
               noHelperText
             />
-            <TextFieldController
+            <SelectFieldController
               name={`custom_dashboard_parameters.${index}.custom_dashboards_parameter_type`}
               label={t('Parameter Type')}
-              variant="standard"
+              items={items}
               required
-              disabled
-              noHelperText
             />
             <Tooltip title={t('Delete')}>
               <IconButton color="error" onClick={() => remove(index)}>
--- a/openbas-front/src/admin/components/workspaces/custom_dashboards/CustomDashboardParameters.tsx
+++ b/openbas-front/src/admin/components/workspaces/custom_dashboards/CustomDashboardParameters.tsx
@@ -3,6 +3,7 @@ import { type FunctionComponent, useContext } from 'react';
 
 import ScenarioField from '../../../../components/fields/ScenarioField';
 import SimulationField from '../../../../components/fields/SimulationField';
+import { type CustomDashboardParameters as CustomDashboardParametersType } from '../../../../utils/api-types';
 import { CustomDashboardContext } from './CustomDashboardContext';
 
 const CustomDashboardParameters: FunctionComponent = () => {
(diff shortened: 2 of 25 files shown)


openbas-front/src/admin/components/workspaces/custom_dashboards/CustomDashboardForm.tsx
```diff
@@ -108,7 +109,7 @@ const CustomDashboardForm: FunctionComponent<Props> = ({
           <IconButton
             color="secondary"
             aria-label="Add"
-            onClick={() => handleAddParameter(items[0])} // For now, we handle just one type
+            onClick={() => handleAddParameter(items[0].value)}
```
antoinemzs, line 112: Can you elaborate on the desired behaviour? This will add a parameter with a default fixed type (there is more than one now). Is the user expected to change the type as needed before final validation?

Also mind whether taking the first item in the list risks changing the default type unexpectedly whenever the source list order is rearranged.

openbas-front/src/admin/components/workspaces/custom_dashboards/CustomDashboardParameters.tsx
```diff
@@ -19,11 +22,26 @@ const CustomDashboardParameters: FunctionComponent = () => {
...
+                onChange={(value: string | undefined) => handleParameters(p.custom_dashboards_parameter_id, value ?? '')}
+              />
+            </div>
+          );
+        } else if (p.custom_dashboards_parameter_type === 'simulation') {
+          return (
+            <div key={p.custom_dashboards_parameter_id}>
               <SimulationField
                 label={p.custom_dashboards_parameter_name}
                 value={getParameterValue(p.custom_dashboards_parameter_id)}
```
antoinemzs: style: use a switch case form

openbas-front/src/components/fields/ScenarioField.tsx
antoinemzs, line 1: It looks like this is an "Options select field" rather than strictly a "Scenario" field? Would it be possible to merge SimulationField and ScenarioField and pass the specific option retrieval behaviour (method) to the component?

antoinemzs: I realise now this is just a thin wrapper around AutocompleteField. There doesn't seem to be any typing issues between SimulationField and ScenarioField in the props, just a matter of passing the correct constants and parameterised methods to the AutocompleteField.

Could the latter be called directly from where SimulationField and ScenarioField are called instead?

openbas-front/src/utils/Option.ts
antoinemzs, line 1: Unused?

openbas-model/src/main/java/io/openbas/database/model/Scenario.java
```diff
@@ -250,8 +251,17 @@ public void setTags(Set<Tag> tags) {
       inverseJoinColumns = @JoinColumn(name = "exercise_id"))
   @JsonSerialize(using = MultiIdListDeserializer.class)
   @JsonProperty("scenario_exercises")
+  @Setter(NONE)
   private List<Exercise> exercises;
 
+  public void setExercises(List<Exercise> exercises) {
+    for (Exercise exercise : exercises) {
+      if (exercise != null) exercise.setUpdatedAt(now());
+    }
```
antoinemzs: missing null check on exercises?

openbas-model/src/main/java/io/openbas/database/model/Scenario.java
```diff
@@ -250,8 +251,17 @@ public void setTags(Set<Tag> tags) {
       inverseJoinColumns = @JoinColumn(name = "exercise_id"))
   @JsonSerialize(using = MultiIdListDeserializer.class)
   @JsonProperty("scenario_exercises")
+  @Setter(NONE)
   private List<Exercise> exercises;
 
+  public void setExercises(List<Exercise> exercises) {
+    for (Exercise exercise : exercises) {
+      if (exercise != null) exercise.setUpdatedAt(now());
```
antoinemzs: Incidentally, there's no point setting the createdAt and updatedAt attributes explicitly since the introduction of the audit annotations (`@CreationTimestamp`, `@UpdateTimestamp`) which force the value through a factory instantiated at boot time.

`exercise` needs to be saved somehow to persist the timestamps.

openbas-model/src/main/java/io/openbas/database/model/Exercise.java
```diff
@@ -161,10 +161,17 @@ public class Exercise implements Base {
...
-  @Schema(type = "string")
   @Queryable(filterable = true, dynamicValues = true)
+  @Schema(type = "string")
+  @Setter(NONE)
   private Scenario scenario;
 
+  public void setScenario(Scenario scenario) {
+    if (scenario != null) scenario.setUpdatedAt(now());
+    this.scenario = scenario;
+    this.setUpdatedAt(now());
```
antoinemzs, line 172: Likewise, there is no more point in setting the timestamps explicitly. There needs to be a save on all affected objects to persist new timestamps.

openbas-model/src/main/java/io/openbas/database/repository/ExerciseRepository.java
```diff
@@ -366,6 +366,7 @@ List<Object[]> findAllOptionByNameLinkedToFindingsWithContext(
               + "LEFT JOIN injects inj ON ex.exercise_id = inj.inject_exercise "
               + "LEFT JOIN injects_assets ia ON ia.inject_id = inj.inject_id "
               + "LEFT JOIN injects_asset_groups iag ON iag.inject_id = inj.inject_id "
+              + "LEFT JOIN scenarios_exercises se ON ex.exercise_id = se.exercise_id "
```
antoinemzs, line 369: weird db layout forcing use of aggregation method but that's how it is.

johanah29 (commented):
Testing NOK on my end

johanah29 (commented):
Testing OK

openbas-front/src/admin/components/workspaces/custom_dashboards/CustomDashboardForm.tsx
RomuDeuxfois: The idea is to provide an initial value so that it’s not empty, but this value can still be changed by the user.

openbas-front/src/components/fields/ScenarioField.tsx
RomuDeuxfois: The idea is to have a reusable component to search for scenarios or search for simulations.
Indeed, for now it doesn’t do much.

How does that cause an issue for you?

openbas-front/src/utils/Option.ts
RomuDeuxfois: I don't understand which line your are pointing.

openbas-model/src/main/java/io/openbas/database/model/Scenario.java
RomuDeuxfois: CreationTimestamp and UpdateTimestamp do not work on one-to-many, many-to-one, etc. relations.

openbas-model/src/main/java/io/openbas/database/repository/ExerciseRepository.java
RomuDeuxfois: "It’s historical." :D

RomuDeuxfois:
Thanks for the feedbacks
PR ready for a secound round.

openbas-front/src/utils/Option.ts
antoinemzs: ah yes, github does not show the diffs; nevermind.

openbas-front/src/components/fields/ScenarioField.tsx
antoinemzs: it's a feedback I've received in another change, so I figured this was some sort of team agreement. I don't have an issue with thin wrappers, it's mainly a matter of how many layers we want in a given code path.

but this is quite clear enough. Just keeping in mind that with this pattern we would have a thin wrapper per option retrieval type, but this is all refactorable at any time.

openbas-front/src/admin/components/workspaces/custom_dashboards/CustomDashboardParameters.tsx
```diff
@@ -19,23 +22,41 @@ const CustomDashboardParameters: FunctionComponent = () => {
...
-              <SimulationField
-                label={p.custom_dashboards_parameter_name}
-                value={getParameterValue(p.custom_dashboards_parameter_id)}
-                onChange={(value: string | undefined) => handleParameters(p.custom_dashboards_parameter_id, value ?? '')}
-              />
-            </div>
-          );
-        } else {
-          return (<></>);
+        switch (p.custom_dashboards_parameter_type) {
```
savacano28: Maybe you could extract this logic in a function?

openbas-front/src/admin/components/workspaces/custom_dashboards/CustomDashboardParameters.tsx
```diff
@@ -19,23 +22,41 @@ const CustomDashboardParameters: FunctionComponent = () => {
...
+                <SimulationField
+                  label={p.custom_dashboards_parameter_name}
+                  value={getParameterValue(p.custom_dashboards_parameter_id)}
+                  onChange={(value: string | undefined) =>
+                    handleParameters(p.custom_dashboards_parameter_id, value ?? '')}
+                />
+              </div>
+            );
+          default:
+            return (<></>);
```
savacano28: Maybe here you could return null?

RomuDeuxfois:
```suggestion
            return null;
```

openbas-front/src/admin/components/workspaces/custom_dashboards/CustomDashboardParameters.tsx
RomuDeuxfois: I don't see why. Can I have your point ?

savacano28: Improve readability, maybe in the future you will add more cases (scenario, smiulation..)?
Something like:
const renderDashbaoradtypes = (p) => {
**name** = p.custom_dashboard_name;
**switch** (p.type) ...
case 'sce':
default...
}

{params.map(0p} => (<div  key= {p.id}>{renderDashboardTypes(p)})...

openbas-model/src/main/java/io/openbas/database/model/Scenario.java
antoinemzs: We discussed we didn't have a better solution right now, in this timeframe

antoinemzs (approved):

PR #3805, opened 2025-08-19
[backend] Adding possibility to provide instance ID (#3644)

### LINKED ISSUE #3644: Make it possible to provide platform ID to be used at initial deployment or when cloning
## Context

OpenBAS has an internal identifier for each instance, which is automatically generated at bootstrap / initial deployment (UUIDv4).

For some usecases (cloning an existing instance, for example) we need to be able to change an existing ID, and for some others it would even be better to provide the ID as soon as bootstrapping the instance.

## Use case

[Similar to OCTI](https://github.com/OpenCTI-Platform/opencti/issues/11188), we have internal usecases.

Dimfacion:
### Proposed changes

* Adding possibility to provide instance ID
*

#### Use case 1
1. Delete your instance_id in the parameters database
2. Start openbas
3. The id should have been recreated with a generated UUID as well as the creation date

#### Use case 2
1. Add an UUID to the property openbas.instance-id in application.properties
2. Start openbas
3. The id should have been updated with the provided UUID as well as the creation date

### Related issues

* Closes #3644 
*

### Checklist

### DIFF
--- a/openbas-api/src/main/java/io/openbas/App.java
+++ b/openbas-api/src/main/java/io/openbas/App.java
@@ -1,14 +1,66 @@
 package io.openbas;
 
+import static io.openbas.database.model.SettingKeys.PLATFORM_INSTANCE;
+import static io.openbas.database.model.SettingKeys.PLATFORM_INSTANCE_CREATION;
+
+import io.openbas.config.OpenBASConfig;
+import io.openbas.database.model.Setting;
+import io.openbas.database.repository.SettingRepository;
 import io.openbas.tools.FlywayMigrationValidator;
+import jakarta.annotation.PostConstruct;
+import java.sql.Timestamp;
+import java.time.Instant;
+import java.util.Optional;
+import java.util.UUID;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.logging.log4j.util.Strings;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 
 @SpringBootApplication
+@Slf4j
+@RequiredArgsConstructor
 public class App {
+
+  private final SettingRepository settingRepository;
+  private final OpenBASConfig openBASConfig;
 
   public static void main(String[] args) {
     FlywayMigrationValidator.validateFlywayMigrationNames();
     SpringApplication.run(App.class, args);
+  }
+
+  @PostConstruct
+  public void init() {
+    log.info("Startup init");
+    // Get the platform instance id
+    Optional<Setting> instanceId = this.settingRepository.findByKey(PLATFORM_INSTANCE.key());
+    Setting instanceCreationDate =
+        this.settingRepository
+            .findByKey(PLATFORM_INSTANCE_CREATION.key())
+            .orElse(new Setting(PLATFORM_INSTANCE_CREATION.key(), ""));
+
+    // If we don't have a platform instance id or if it's been specified as another value than the
+    // one in the database
+    if (instanceId.isEmpty()
+        || (openBASConfig.getInstanceId() != null
+            && !instanceId.get().getValue().equals(openBASConfig.getInstanceId()))) {
+      log.info("Updating platform instance id");
+      // We update the platform instance id using a random UUID if the value does not exist in the
+      // database
+      Setting instanceIdSetting =
+          instanceId.orElse(new Setting(PLATFORM_INSTANCE.key(), UUID.randomUUID().toString()));
+
+      // If it's been specified as a specific id, we validate that it's a proper UUID and use it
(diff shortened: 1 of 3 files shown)


openbas-api/src/test/java/io/openbas/AppTest.java
```diff
@@ -0,0 +1,94 @@
...
+      context = SpringApplication.run(App.class);
+    } catch (Exception e) {
+      assertEquals("Invalid UUID string: pouet", e.getCause().getMessage());
+      return;
+    } finally {
+      if (context != null) {
+        context.close();
+      }
+    }
+    fail();
```
antoinemzs: tip: the assertj method `assertThatThrownBy` (as used in `InjectorContractApiTest`) captures this pattern well

openbas-framework/src/main/java/io/openbas/config/OpenBASConfig.java
```diff
@@ -61,6 +61,9 @@ public class OpenBASConfig {
   @JsonProperty("enabled_dev_features")
   private String enabledDevFeatures = "";
 
+  @JsonProperty("instance_id")
+  private String instanceId = null;
```
antoinemzs: nitpick: is it useful to explicitly initialise to null when it's the default for `String` type un java?

openbas-api/src/test/java/io/openbas/AppTest.java
```diff
@@ -0,0 +1,94 @@
...
+import static org.junit.jupiter.api.Assertions.*;
+
+import io.openbas.database.repository.SettingRepository;
+import java.util.UUID;
+import org.junit.jupiter.api.*;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.SpringApplication;
+import org.springframework.context.ConfigurableApplicationContext;
+
+class AppTest extends IntegrationTest {
```
antoinemzs, line 18: consider setting transactional to avoid keeping the default settings in the db

openbas-api/src/main/java/io/openbas/App.java
```diff
@@ -1,14 +1,66 @@
...
+  private final OpenBASConfig openBASConfig;
+
   public static void main(String[] args) {
     FlywayMigrationValidator.validateFlywayMigrationNames();
     SpringApplication.run(App.class, args);
   }
+
+  @PostConstruct
+  public void init() {
+    log.info("Startup init");
```
antoinemzs, line 36: suggestion: could we have an echo of the final instance UUID in the boot logs?

openbas-api/src/main/java/io/openbas/App.java
```diff
@@ -1,14 +1,66 @@
...
 
+  private final SettingRepository settingRepository;
+  private final OpenBASConfig openBASConfig;
+
   public static void main(String[] args) {
     FlywayMigrationValidator.validateFlywayMigrationNames();
     SpringApplication.run(App.class, args);
   }
+
+  @PostConstruct
```
RomuDeuxfois, line 34: We have some CommandLineRunner class to do things like this:
- InitAdminCommandLineRunner
- InitTagRuleCommandLineRunner

Do we want to homogenize our flows ?

Dimfacion: After some quick testing, the order of initialization is as follows :
- The init() method in App.java
- Then the OpenTelemetryConfig that uses the platform instance ID
- Then the CommandLineRunner implementations

Since we need the platform instance ID to be initialized before the OpenTelemetryConfig, I'm unsure there is another way. I'm happy to discuss this further tho if you have an idea.

RomuDeuxfois: Good to know and thanks for the tests !

openbas-framework/src/main/java/io/openbas/config/OpenBASConfig.java
Dimfacion: I just like being explicit but yeah, I'll remove it :p

openbas-api/src/test/java/io/openbas/AppTest.java
Dimfacion: Nice ! I'm changing this to use it, thanks !

openbas-api/src/test/java/io/openbas/AppTest.java
Dimfacion: I tried to do it but it ended up creating issues since we're starting a new SpringApplication inside an already existing SpringApplication leading to deadlocks and timeouts ... 
In the end, I moved to a solution involving directly instantiating the App class with mocks and calling the init method directly. This is way more elegant, faster and we're testing the same scope but do not impact the database. Do let me know if you have a better idea tho :)

antoinemzs: ah yes, good point. Your proposal makes sense, thanks!

Issue #3595, opened 2025-07-11
Add Stix2Scenario


Seb-MIGUEL:
Based on Julien feedbacks we need to generate scenario from opencti to openbas with stix format. 

We need to deploy it before integrate cve into opencti > openbas scenario

PR #3810, opened 2025-08-19
[backend] enable coverage bundle creation (#3595)

### LINKED ISSUE #3707: Push stix bundle from OpenBAS to OpenCTI

Yes / No (Help can be provided if you need assistance submitting a PR)

antoinemzs:
### Proposed changes

* Related to : https://github.com/OpenBAS-Platform/openbas/pull/3807
*

### Testing Instructions

1. From a scenario created via STIX bundle
2. Launch simulation
3. When all injects have executed: manually add a result to each expectation (note: currently, manually adding results to the higher order nodes (e.g. Asset Group) which then bubble down to the lower levels is a way to expedite the trigger for creating the stix response)
4. Wait a minute or so
5. (Only in debugging): inspect the generated stix response json object

### Related issues

* Closes #3707 
*

### Checklist

### Further comments


openbas-api/src/main/java/io/openbas/service/SecurityCoverageSendJobService.java
```diff
@@ -0,0 +1,90 @@
...
+import org.springframework.transaction.annotation.Transactional;
+
+@Service
+@RequiredArgsConstructor
+public class SecurityCoverageSendJobService {
+  private final SecurityCoverageSendJobRepository securityCoverageSendJobRepository;
+  private final ExerciseService exerciseService;
+  private final EntityManager entityManager;
+
+  public void createOrUpdateCoverageSendJobForSimulationsIfReady(List<Exercise> exercises) {
```
RomuDeuxfois, line 24: 👀  Security Coverage Job to push to OpenCTI

openbas-model/src/main/java/io/openbas/stix/types/Timestamp.java
```diff
@@ -1,9 +1,16 @@
...
 import java.time.Instant;
 
 public class Timestamp extends BaseType<Instant> {
   public Timestamp(Instant value) {
     super(value);
   }
+
+  @Override
+  public JsonNode toStix(ObjectMapper mapper) {
+    return mapper.valueToTree(this.getValue());
```
RomuDeuxfois: Can we move this method up, since it's often the same, in order to have a default implementation in the abstract class or the interface?

openbas-model/src/main/java/io/openbas/stix/parsing/ObjectFactory.java
```diff
@@ -0,0 +1,33 @@
...
+import io.openbas.stix.objects.ObjectBase;
+import io.openbas.stix.objects.RelationshipObject;
+import io.openbas.stix.objects.constants.CommonProperties;
+import io.openbas.stix.objects.constants.ObjectTypes;
+import io.openbas.stix.types.BaseType;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Function;
+
+public class ObjectFactory {
```
RomuDeuxfois, line 13: praise: I really like the concept

openbas-model/src/main/java/io/openbas/stix/objects/constants/ObjectTypes.java
```diff
@@ -0,0 +1,29 @@
+package io.openbas.stix.objects.constants;
+
+import jakarta.validation.constraints.NotBlank;
+
+public enum ObjectTypes {
+  RELATIONSHIP("relationship"),
+  SIGHTING("sighting"),
+  DEFAULT("__default__");
```
RomuDeuxfois, line 10: That’s not supposed to happen, right ?

openbas-model/src/main/java/io/openbas/database/repository/SecurityCoverageSendJobRepository.java
```diff
@@ -0,0 +1,47 @@
...
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.data.repository.CrudRepository;
+import org.springframework.stereotype.Repository;
+
+@Repository
+public interface SecurityCoverageSendJobRepository
+    extends CrudRepository<SecurityCoverageSendJob, String>,
+        JpaSpecificationExecutor<SecurityCoverageSendJob> {
+
+  String findByStatusAndUpdatedAtBeforeStatement =
```
RomuDeuxfois: todo: Field 'findByStatusAndUpdatedAtBeforeStatement' is never used

openbas-model/src/main/java/io/openbas/database/repository/InjectExpectationRepository.java
```diff
@@ -233,6 +234,11 @@ List<RawInjectExpectation> rawForComputeGlobalByInjectIds(
   List<RawInjectExpectation> rawForComputeGlobalByExerciseIds(
       @Param("exerciseIds") Set<String> exerciseIds);
 
+  @Query(
+      value =
+          "select i from InjectExpectation i where i.inject.id in :injectIds and i.agent is null and i.user is null")
+  List<InjectExpectation> findAllForGlobalScoreByInjects(@Param("injectIds") Set<String> injectIds);
```
RomuDeuxfois, line 240: I'm not sure I understand the naming of this method in relation to what we are doing.

openbas-api/src/main/java/io/openbas/rest/attack_pattern/service/AttackPatternService.java
```diff
@@ -235,18 +236,16 @@ public List<String> searchAttackPatternWithTTPAIWebservice(
...
+   * Resolves external AttackPattern references from a {@link SecurityCoverage} into internal {@link
+   * AttackPattern} entities using the {@code attackPatternService}.
    *
-   * @param attackPatternRefs list attackPatternRefs to resolve with internal attackPattern refs
+   * @param stixRefs list of tuples linking an atatck pattern ext ID with a stix ID
    * @return list of resolved internal AttackPattern entities
    */
-  public Map<String, AttackPattern> fetchInternalAttackPatternIds(
-      List<StixRefToExternalRef> attackPatternRefs) {
+  public Map<String, AttackPattern> fetchInternalAttackPatternIdsFromSecurityCoverage(
```
RomuDeuxfois: nitpick: I am not a big fan of add securty coverage concept inside the method name. Can you rename it to fetchInternalAttackPatternIds ?

openbas-api/src/main/java/io/openbas/rest/exercise/service/ExerciseService.java
```diff
@@ -199,6 +202,28 @@ public List<Document> getExercisePlayerDocuments(Exercise exercise) {
     return documentService.getPlayerDocuments(articles, injects);
   }
 
+  public boolean hasPendingResults(Exercise exercise) {
```
RomuDeuxfois, line 206: next step: should be merged with end of simulation mechanism

openbas-model/src/main/java/io/openbas/database/repository/ExerciseRepository.java
```diff
@@ -88,6 +88,15 @@ List<Object[]> userCountGroupByWeek(
       nativeQuery = true)
   List<Exercise> thatMustBeFinished();
 
+  @Query(
+      """
+    SELECT e FROM Exercise e, Scenario s
+    WHERE e.scenario.id = s.id AND s.id = :#{#exercise.scenario.id}
```
RomuDeuxfois, line 87: question: I’m not sure I understand what you’re trying to achieve here. Wouldn’t it be simpler to just do WHERE e.id = :exercise.id AND e.scenario IS NOT NULL ?

openbas-api/src/main/java/io/openbas/rest/exercise/service/ExerciseService.java
```diff
@@ -199,6 +202,28 @@ public List<Document> getExercisePlayerDocuments(Exercise exercise) {
...
+  public Optional<Instant> getLatestValidityDate(Exercise exercise) {
+    Optional<Exercise> follower = this.getFollowingSimulation(exercise);
+    if (follower.isPresent()) {
+      return follower.get().getStart();
+    }
+
+    return exercise.getStart().isPresent() && exercise.getScenario() != null
+        ? cronService.getNextExecutionFromInstant(
+            exercise.getStart().get(), ZoneId.of("UTC"), exercise.getScenario().getRecurrence())
+        : Optional.empty();
```
RomuDeuxfois, line 225: question: In this case, you’re specifying a time in the future — why is that?

openbas-api/src/main/java/io/openbas/scheduler/PlatformJobDefinitions.java
```diff
@@ -45,4 +42,12 @@ public JobDetail getEngineSyncExecution() {
         .withIdentity(jobKey("ElasticSyncExecutionJob"))
         .build();
   }
+
+  @Bean
+  public JobDetail getSecurityCoverageJobExecution() {
```
RomuDeuxfois, line 47: next step: check for an event publisher

openbas-api/src/main/java/io/openbas/scheduler/jobs/ScenarioExecutionJob.java
```diff
@@ -114,16 +109,8 @@ private boolean isScenarioOutdated(@NotNull final Scenario scenario) {
...
-    Cron cron = parser.parse(cronExpression);
-    ExecutionTime executionTime = ExecutionTime.forCron(cron);
-
-    Duration timeToNextExecution =
-        executionTime
-            .timeToNextExecution(ZonedDateTime.now(ZoneId.of("UTC")))
-            .orElse(Duration.ZERO);
-
-    return Instant.now().plus(timeToNextExecution);
+    return cronService
```
RomuDeuxfois, line 112: praise: clearer

openbas-api/src/main/java/io/openbas/service/SecurityCoverageSendJobService.java
```diff
@@ -0,0 +1,90 @@
...
+    } else {
+      SecurityCoverageSendJob newJob = new SecurityCoverageSendJob();
+      newJob.setSimulation(exercise);
+      newJob.setUpdatedAt(Instant.now());
+      return Optional.of(newJob);
+    }
+  }
+
+  public List<SecurityCoverageSendJob> getPendingSecurityCoverageSendJobs() {
+    return securityCoverageSendJobRepository.findByStatusAndUpdatedAtBeforeNoLock(
```
RomuDeuxfois, line 54: question: not sure to understand the NoLock name of this method

openbas-api/src/main/java/io/openbas/service/SecurityCoverageService.java
```diff
@@ -185,9 +206,167 @@ private void updatePropertiesFromSecurityCoverage(Scenario scenario, SecurityCov
...
+
+  private List<ObjectBase> getCoverageForSimulation(Exercise exercise)
+      throws ParsingException, JsonProcessingException {
+    List<ObjectBase> objects = new ArrayList<>();
+
+    // create the main coverage object
+    SecurityCoverage assessment = exercise.getSecurityCoverage();
+    DomainObject coverage = (DomainObject) stixParser.parseObject(assessment.getContent());
+    coverage.setProperty(CommonProperties.MODIFIED.toString(), new Timestamp(Instant.now()));
+    coverage.setProperty("coverage", getOverallCoverage(exercise));
```
RomuDeuxfois: nitpick: "coverage' should be a const

openbas-api/src/main/java/io/openbas/service/SecurityCoverageService.java
```diff
@@ -185,9 +206,167 @@ private void updatePropertiesFromSecurityCoverage(Scenario scenario, SecurityCov
...
+      boolean covered = !((Map<String, BaseType<?>>) attackPatternCoverage.getValue()).isEmpty();
+      RelationshipObject sro =
+          new RelationshipObject(
+              new HashMap<>(
+                  Map.of(
+                      CommonProperties.ID.toString(),
+                      new Identifier(ObjectTypes.RELATIONSHIP + "--" + exercise.getId()),
+                      CommonProperties.TYPE.toString(),
+                      new StixString(ObjectTypes.RELATIONSHIP.toString()),
+                      "relationship_type",
```
RomuDeuxfois: nitpick: "relationship_type" should be a const

openbas-api/src/main/java/io/openbas/service/SecurityCoverageService.java
```diff
@@ -185,9 +206,167 @@ private void updatePropertiesFromSecurityCoverage(Scenario scenario, SecurityCov
...
+                      new StixString("has-assessed"),
+                      RelationshipObject.Properties.SOURCE_REF.toString(),
+                      coverage.getId(),
+                      RelationshipObject.Properties.TARGET_REF.toString(),
+                      new Identifier(stixRef.getStixRef()),
+                      "covered",
```
RomuDeuxfois: nitpick: "covered" should be a const

openbas-api/src/main/java/io/openbas/service/SecurityCoverageService.java
```diff
@@ -185,9 +206,167 @@ private void updatePropertiesFromSecurityCoverage(Scenario scenario, SecurityCov
...
+                      new io.openbas.stix.types.Boolean(covered))));
+      sroStartTime.ifPresent(instant -> sro.setProperty("start_time", instant));
```
RomuDeuxfois: Properties.START_TIME

openbas-api/src/main/java/io/openbas/service/SecurityCoverageService.java
```diff
@@ -185,9 +206,167 @@ private void updatePropertiesFromSecurityCoverage(Scenario scenario, SecurityCov
...
+      sroStopTime.ifPresent(instant -> sro.setProperty("stop_time", instant));
```
RomuDeuxfois: Properties.STOP_TIME

openbas-api/src/main/java/io/openbas/rest/attack_pattern/service/AttackPatternService.java
savacano28: I thought we changed this :)

openbas-api/src/main/java/io/openbas/rest/inject/form/InjectAssistantInput.java
```diff
@@ -34,5 +34,5 @@ public class InjectAssistantInput {
   @NotNull
   @JsonProperty("inject_by_ttp_number")
   @Schema(description = "Number of injects to generate for each TTP")
-  private int injectByTTPNumber = 1;
+  private Integer injectByTTPNumber = 1;
```
savacano28, line 37: thought: I had changed it because I had set a default value

openbas-api/src/main/java/io/openbas/service/SecurityCoverageService.java
```diff
@@ -185,9 +206,167 @@ private void updatePropertiesFromSecurityCoverage(Scenario scenario, SecurityCov
...
+    }
+
+    for (SecurityPlatform securityPlatform : assetService.securityPlatforms()) {
+      DomainObject platformIdentity = securityPlatform.toStixDomainObject();
+      objects.add(platformIdentity);
+
+      BaseType<?> platformCoverage = getOverallCoveragePerPlatform(exercise, securityPlatform);
+      boolean covered = !((Map<String, BaseType<?>>) platformCoverage.getValue()).isEmpty();
+      RelationshipObject sro =
+          new RelationshipObject(
```
RomuDeuxfois, line 283: question: Why not using a real class instead of a property map ?

openbas-api/src/main/java/io/openbas/service/SecurityCoverageService.java
```diff
@@ -185,9 +206,167 @@ private void updatePropertiesFromSecurityCoverage(Scenario scenario, SecurityCov
...
+
+    // get all injects involved in attack pattern
+    List<Inject> injects =
+        exercise.getInjects().stream()
+            .filter(
+                i ->
+                    i.getInjectorContract().isPresent()
+                        && i.getInjectorContract().get().getAttackPatterns().stream()
+                            .anyMatch(
+                                attackPattern -> attackPattern.getId().equals(ap.get().getId())))
```
RomuDeuxfois, line 337: question: why just using the first attack pattern here ?

openbas-api/src/main/java/io/openbas/utils/ResultUtils.java
```diff
@@ -41,6 +37,34 @@ public List<ExpectationResultsByType> computeGlobalExpectationResults(Set<String
...
 
+  public List<ExpectationResultsByType> computeGlobalExpectationResultsForPlatform(
+      Set<String> injectIds, SecurityPlatform securityPlatform) {
+
+    if (injectIds == null || injectIds.isEmpty()) {
+      return emptyList();
+    }
+
+    List<InjectExpectation> expectations =
+        injectExpectationRepository.findAllForGlobalScoreByInjects(injectIds).stream()
```
RomuDeuxfois, line 48: I’m not sure I understand the reasoning behind the naming of this method findAllForGlobalScoreByInjects

openbas-api/src/main/java/io/openbas/utils/ResultUtils.java
```diff
@@ -41,6 +37,34 @@ public List<ExpectationResultsByType> computeGlobalExpectationResults(Set<String
...
+          exp.setResults(
+              exp.getResults().stream()
+                  .filter(r -> r.getSourceId().equals(securityPlatform.getId()))
+                  .toList());
+
+          exp.setScore(
+              exp.getResults().stream()
+                  .max(Comparator.comparing(InjectExpectationResult::getScore))
+                  .map(InjectExpectationResult::getScore)
+                  .orElse(0.0));
```
RomuDeuxfois: I’m not sure this is what we want. If the score is null, it should be in a PENDING state, not FAILED.

openbas-api/src/main/java/io/openbas/service/cron/CronService.java
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  public String getCronExpression(ScheduleFrequency scheduling, Instant seed) {
+    ZonedDateTime zdt = seed.atZone(ZoneId.of("UTC"));
+    int minute = zdt.getMinute();
+    int hour = zdt.getHour();
+    int dayOfMonth = zdt.getDayOfMonth();
+    int dayOfWeek = zdt.getDayOfWeek().getValue();
+
+    return switch (scheduling) {
```
RomuDeuxfois, line 36: This could be a SecuryFrenquency (enum class) method

openbas-model/src/main/java/io/openbas/cron/ScheduleFrequencyConverter.java
```diff
@@ -0,0 +1,18 @@
+package io.openbas.cron;
+
+import jakarta.persistence.AttributeConverter;
+import jakarta.persistence.Converter;
+
+@Converter
+public class ScheduleFrequencyConverter implements AttributeConverter<ScheduleFrequency, String> {
```
RomuDeuxfois, line 7: You can avoid this by using @Enumerated(EnumType.STRING) on private ScheduleFrequency scheduling property

openbas-model/src/main/java/io/openbas/database/model/InjectExpectation.java
```diff
@@ -27,7 +28,18 @@
 @Entity
 @Table(name = "injects_expectations")
 @EntityListeners(ModelBaseListener.class)
-public class InjectExpectation implements Base {
+public class InjectExpectation implements Base, Cloneable {
```
RomuDeuxfois, line 31: question: Why is Cloneable needed here?

openbas-model/src/main/java/io/openbas/database/model/Scenario.java
```diff
@@ -56,6 +56,14 @@ public enum SEVERITY {
     critical,
   }
 
+  public static final String MAIN_FOCUS_INCIDENT_RESPONSE = "incident-response";
```
RomuDeuxfois, line 60: praise: glad to see it on back-end side

openbas-model/src/main/java/io/openbas/database/repository/ExerciseRepository.java
antoinemzs: This finds whether there is a "next" simulation of the same scenario; hence I must pass the scenario ID to the query.

openbas-api/src/main/java/io/openbas/rest/exercise/service/ExerciseService.java
antoinemzs: This sets the time after which the simulation is expected to be superseded (by a successor simulation of the same scenario). This event may occur in the future in this code branch. In other situations, it may be set to the time when the successor effectively started running, and in some other situtations again it cannot be determined (i.e. this is the latest simulation of a given, non-scheduled scenario, or an orphaned simulation).

openbas-api/src/main/java/io/openbas/service/SecurityCoverageSendJobService.java
antoinemzs: This repository method is decorated with the hibernate "readonly" annotation, making it will not hold any exclusive locks on the rows it will select (I've noticed the ORM take exclusive locks on `injects_expectations` prior to using this annotation).
This allows for selecting such rows early in the transaction and still allow other threads to modify the rows while the original transaction is still ongoing (e.g. it takes a long time to process the stix response).

Actual locks are explictly acquired later (not this method) for the shortest time possible.

openbas-api/src/main/java/io/openbas/service/SecurityCoverageService.java
antoinemzs: `Map<String, BaseType<?>>` is a real class :grin: 
There is no real point for a custom class IMO, and I actually rely on some `Map` methods to implement the functionality, e.g.
https://github.com/OpenAEV-Platform/openaev/blob/c4774546efbb6a563c906e54e6fc2f220bbc4bfc/openbas-model/src/main/java/io/openbas/stix/objects/ObjectBase.java#L45C3-L47C4

If anything, I think we should probably improve the API of `ObjectBase`, `DomainObject`, `RelationshipObject` to not need to instantiate the Map outside of the ctor, to hide the internals. This parser is a PoC that has devoled into production code so some things are still a bit rough.

openbas-api/src/main/java/io/openbas/service/SecurityCoverageService.java
antoinemzs: In this method, we are considering a single attack pattern. However I'm using service methods that select multiple rows, so maybe this makes it a bit confusing; selecting a list of attack pattern where ID matches a single value, it's expected to get a single or no results, hence then calling `getFirst()` if the method succeeds -- to get a single object.

openbas-model/src/main/java/io/openbas/cron/ScheduleFrequencyConverter.java
antoinemzs: This was my initial thinking, but there appeared to be a need to map arbitrary string values to enum values (e.g. `DAILY` is `d`, `WEEKLY` is `w`, etc...); `@Enumerated` just maps the transliteration of the enum name to the column value.

openbas-model/src/main/java/io/openbas/database/model/InjectExpectation.java
antoinemzs: I need to obtain a shallow copy of the expectations to draw the `SecurityPlatform` coverage. From the copies I strip the reuslts that are not sourced from the considered specific security platform. Can't do that with the original instances since they are bound to the transaction context and modifying them risks miscalculating scores later.

openbas-model/src/main/java/io/openbas/database/model/Scenario.java
antoinemzs: Glad you like it, but it's dead code after I rebased after the STIX import patch was merged. It seems the same constant was moved away to another place, so that's something I need to clean up.

openbas-model/src/main/java/io/openbas/stix/types/Timestamp.java
antoinemzs: Looks like a lot of the overrides reimplement the base class method verbatim. Good catch thanks

openbas-model/src/main/java/io/openbas/stix/objects/constants/ObjectTypes.java
antoinemzs: So that's a placeholder for when the `fromString` method does not find a matching value -- and if an object in the stix input has the type "`__default__`", well it's going to follow the correct route in the factory.
Arguably this is not the most clever implementation, but it's a way to internalise the exception catching away from consumers; who get one of few enum values.

openbas-api/src/main/java/io/openbas/rest/attack_pattern/service/AttackPatternService.java
antoinemzs: Its's a merge issue, will fix

openbas-api/src/main/java/io/openbas/service/cron/CronService.java
antoinemzs: I might disagree. The enum just exposes option values, but logic should perhaps live in a service; moreover, IMO the concepts of ScheduleFrequency (generic) and CronExpression (implementation) are somewhat only loosely related, the former being a parameter of the latter.

openbas-api/src/main/java/io/openbas/service/SecurityCoverageService.java
antoinemzs: added constants

openbas-api/src/main/java/io/openbas/service/SecurityCoverageService.java
antoinemzs: added constants

openbas-api/src/main/java/io/openbas/service/SecurityCoverageService.java
antoinemzs: added constants

openbas-api/src/main/java/io/openbas/service/SecurityCoverageService.java
antoinemzs: added constants

openbas-api/src/main/java/io/openbas/utils/ResultUtils.java
antoinemzs: Find all InjectExpectations for computing the Global Score, by Injects (pass list of inject ids).

The idea is to exclude Player and Agent expectations in this query. This replicates the logic already in production dealing with raw objects (`rawForComputeGlobalByInjectIds`)

openbas-api/src/main/java/io/openbas/utils/ResultUtils.java
antoinemzs: Good thinking, I wasn't sure either. Note that none of this is persisted, this is purely for computing a success rate per security platform; in this situation a PENDING or a FAILED have the same impact on the _success_ rate.

openbas-model/src/main/java/io/openbas/database/repository/InjectExpectationRepository.java
antoinemzs: It's a specialised method used when computing global scores, so that is mentioned in the method's name. The actual behaviour is to exclude expectations ascribed to Players and Agents.

openbas-api/src/main/java/io/openbas/scheduler/PlatformTriggers.java
```diff
@@ -57,4 +57,14 @@ public Trigger elasticSyncExecutionTrigger() {
         .withSchedule(_15_seconds.withMisfireHandlingInstructionNextWithRemainingCount())
         .build();
   }
+
+  @Bean
```
Dimfacion, line 63: **question :** Just like yesterday when we talked about deactivating the elastic sync job, should we deactivate it here too ?

openbas-model/src/main/java/io/openbas/database/repository/InjectExpectationRepository.java
RomuDeuxfois: Yeah, that’s exactly what’s bothering me. The query excludes inject expectations but at the same time we say it’s retrieving stuff for the global score calculation. So the name should probably reflect the actual logic — something like findAllFromTeamAndAssetOrAssetGroup.

I don’t really have a less verbose idea..

openbas-api/src/main/java/io/openbas/service/SecurityCoverageSendJobService.java
RomuDeuxfois: Thanks for the explanations, I’ve learned something again!

openbas-api/src/main/java/io/openbas/service/cron/CronService.java
RomuDeuxfois: It’s true that linking them would make little sense, I agree.

openbas-model/src/main/java/io/openbas/cron/ScheduleFrequencyConverter.java
RomuDeuxfois: True, I see why a converter is needed in this case.

openbas-model/src/main/java/io/openbas/stix/types/Hashes.java
```diff
@@ -11,6 +12,11 @@ public Hashes(Map<HashingAlgorithms, java.lang.String> value) {
     super(value);
   }
 
+  @Override
+  public JsonNode toStix(ObjectMapper mapper) {
+    return mapper.valueToTree(this.getValue());
+  }
```
RomuDeuxfois: **nitpick:** This override can be removed due to implementation in BaseType class.

openbas-model/src/main/java/io/openbas/stix/objects/constants/CustomProperties.java
```diff
@@ -0,0 +1,29 @@
+package io.openbas.stix.objects.constants;
+
+import jakarta.validation.constraints.NotBlank;
+
+public enum CustomProperties {
```
RomuDeuxfois: **Though**: did you talk of ‘extending the STIX standard’? If so, maybe ExtendedProperties would be a more meaningful name

openbas-model/src/main/java/io/openbas/stix/types/Hashes.java
antoinemzs: Oops forgot this one sorry

openbas-model/src/main/java/io/openbas/stix/objects/constants/CustomProperties.java
antoinemzs: sounds good

openbas-api/src/main/java/io/openbas/scheduler/PlatformTriggers.java
antoinemzs: Yes good thinking. I'm testing the job functionality directly and I don't want it to be consumed unexpectedly and cause a problem in CI.

PR #3818, opened 2025-08-20
[backend/frontend] Add UI for Grant on Atomic Testing/Payload

RomuDeuxfois:
### Related issues

* Closes https://github.com/OpenBAS-Platform/openbas/issues/3804

### DIFF
--- a/openbas-api/src/main/java/io/openbas/rest/group/GroupApi.java
+++ b/openbas-api/src/main/java/io/openbas/rest/group/GroupApi.java
@@ -3,7 +3,6 @@ package io.openbas.rest.group;
 import static io.openbas.database.model.User.ROLE_ADMIN;
 import static io.openbas.database.model.User.ROLE_USER;
 import static io.openbas.utils.pagination.PaginationUtils.buildPaginationJPA;
-import static java.time.Instant.now;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.StreamSupport.stream;
 
@@ -206,7 +205,7 @@ public class GroupApi extends RestBehavior {
       actionPerformed = Action.WRITE,
       resourceType = ResourceType.USER_GROUP)
   @Transactional(rollbackOn = Exception.class)
-  public Grant groupGrant(@PathVariable String groupId, @Valid @RequestBody GroupGrantInput input) {
+  public Group groupGrant(@PathVariable String groupId, @Valid @RequestBody GroupGrantInput input) {
     if (input.getExerciseId() == null && input.getScenarioId() == null) {
       throw new IllegalArgumentException("At least one of exercise or scenario should be present");
     }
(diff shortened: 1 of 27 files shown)


openbas-api/src/main/java/io/openbas/rest/group/GroupApi.java
```diff
@@ -226,31 +225,25 @@ public Grant groupGrant(@PathVariable String groupId, @Valid @RequestBody GroupG
...
-      scenario.getGrants().add(savedGrant);
-      scenario.setUpdatedAt(now());
-      scenarioRepository.save(scenario);
-    }
+    group.getGrants().add(grant);
+    return groupRepository.save(group);
+  }
 
-    return savedGrant;
+  @Secured(ROLE_ADMIN)
```
savacano28: Do we need still secured with RBAC?

RomuDeuxfois: This will be override by the RBAC team work
@heditar can you confirm ?

heditar: yes, it's coming in Camille's PR

impolitepanda: I can confirm. @ Secured and @ preAuthorize will all disapear (work being done by Camille)

heditar: but if it's a new api you can just let the RBAC annotation and remove this one

openbas-api/src/main/java/io/openbas/rest/group/GroupApi.java
```diff
@@ -226,31 +225,25 @@ public Grant groupGrant(@PathVariable String groupId, @Valid @RequestBody GroupG
...
+  @PostMapping("/api/groups/{groupId}/grants/{grantId}")
```
impolitepanda: Post or Put?

openbas-api/src/main/java/io/openbas/rest/group/GroupApi.java
```diff
@@ -226,31 +225,25 @@ public Grant groupGrant(@PathVariable String groupId, @Valid @RequestBody GroupG
...
+  @RBAC(
+      resourceId = "#grantId",
+      actionPerformed = Action.WRITE,
```
heditar, line 162: if the resource type is Group the resource id should be the group id

openbas-api/src/main/java/io/openbas/rest/group/GroupApi.java
```diff
@@ -226,31 +225,25 @@ public Grant groupGrant(@PathVariable String groupId, @Valid @RequestBody GroupG
...
-    }
+    group.getGrants().add(grant);
+    return groupRepository.save(group);
+  }
 
-    return savedGrant;
+  @Secured(ROLE_ADMIN)
+  @PostMapping("/api/groups/{groupId}/grants/{grantId}")
+  @RBAC(
+      resourceId = "#grantId",
```
impolitepanda: Should probably be #groupId, I think

openbas-api/src/main/java/io/openbas/rest/group/GroupApi.java
```diff
@@ -273,7 +266,7 @@ public Group groupOrganization(
   }
 
   @Secured(ROLE_ADMIN)
-  @DeleteMapping("/api/groups/{groupId}/organizations/{organizationId}")
+  @PostMapping("/api/groups/{groupId}/organizations/{organizationId}")
```
impolitepanda: Post or Put?

heditar: I think we use PUT for similar actions on other apis

openbas-api/src/main/java/io/openbas/rest/group/GroupApi.java
RomuDeuxfois: I read again the Restfull standards and it seems that creating or removing ressource on an entity should be a POST
Updating a simple property should be PUT.
But I can just follow what is already implemented.
Let me know :)

openbas-api/src/main/java/io/openbas/rest/group/GroupApi.java
RomuDeuxfois: I can stick with PUT if you confirm this.

heditar: Sorry I tough it was to add an organization, so use POST to delete with the path /api/groups/{groupId}/organizations/{organizationId} ? Then that should be DELETE, why are we changing that ?

camrrx:
We don't have access anymore on the group popover because of the side of the header. The display is good on a larger screen.

Capture d’écran 2025-08-27 à 08 59 59

impolitepanda:
> We don't have access anymore on the group popover because of the side of the header. The display is good on a larger screen.
> Capture d’écran 2025-08-27 à 08 59 59

We will let this like that for now. Default grant will be removed in the future. For now, the table is horizontally scrollable. Not good, but good enough ^^'

heditar (approved):

PR #3831, opened 2025-08-22
[backend/frontend] RBAC add  assessment capa Issue/375

heditar:
### Proposed changes

* add assesment capa 
* -> doesn't incule handling capa in Bulk actions, that will be done in a separate PR

### Testing Instructions

1. Step-by-step how to test

### Related issues

* Closes #ISSUE-NUMBER
*

### Checklist


openbas-api/src/main/java/io/openbas/rest/exercise/service/ExerciseService.java
```diff
@@ -363,9 +365,11 @@ private void getObjectives(Exercise duplicatedExercise, Exercise originalExercis
 
   // -- EXERCISES --
   public List<ExerciseSimple> exercises() {
-    // We get the exercises depending on whether or not we are granted
+    // We get the exercises depending on whether or not we are granted or have the capa
+    User currentUser = userService.currentUser();
     List<RawExerciseSimple> exercises =
-        currentUser().isAdmin()
+        currentUser.isAdminOrBypass()
```
RomuDeuxfois, line 371: Would be great to have a helper who wrapp the isAdminOrBypass() inside.
To just call user.hasCapa("ACCESS_ASSESSMENT")
And the helper check is bypass or admin or has capa

What do you think ?

openbas-model/src/main/java/io/openbas/database/model/User.java
```diff
@@ -345,6 +345,12 @@ public boolean isOnlyPlayer() {
     return !isAdmin() && !isManager();
   }
 
+  @JsonProperty("user_is_admin_or_bypass")
```
RomuDeuxfois, line 348: Usefull on front-end ?

openbas-front/src/admin/components/settings/roles/capabilities.json
```diff
@@ -1 +1 @@
...
          {
            "name": "Manage platform settings",
            "value": "MANAGE_PLATFORM_SETTINGS",
            "checkable": true
          }
        ]
      }
    ]
  }
]
```
RomuDeuxfois, line 1: Does this file should be removed ?

openbas-model/src/main/java/io/openbas/database/model/User.java
heditar: right not is it not used but I was thinking that it could be used, it is not adding any information as we have the list of capa + isAdmin but now we decided that it should have the same behavior, I was thinking that it can be useful

openbas-front/src/admin/components/settings/roles/capabilities.json
heditar: no today it is used to display the capa screen in the front, we didn't work on a way to get that dynamically yet

RomuDeuxfois (approved):

openbas-api/src/main/java/io/openbas/rest/exercise/service/ExerciseService.java
heditar: I talked to @RomuDeuxfois, we will let it like it is to not add any more logic in the user model

Issue #3489, opened 2025-06-30
RBAC: Update the Simulation screen so the action button are hide/shown based on the permissisons+6


heditar:
## Context
The permission logic has been completely refactored as part of the RBAC task. The Simulation screen currently displays some buttons that are not consistent with the user's permissions and could trigger a 401 error.

## Goal
In the Simulation screen, only the actions allowed by the user's permissions are shown.

PR #3871, opened 2025-08-28
[frontend] Update the Simulation screen so only the allowed button are visible in the front (#3489)

camrrx:
### Proposed changes

* Display or not actions related to simulation depending on capability and grant

### Testing Instructions

1. Create 2 roles with and without capability "create assessment", and give an "access" grant to the user for one simulation, a "manage" grant to another simulation and a "launch" grant to another one.
2. Try to do different actions on simulation pages on the different simulation granted

### Related issues

* Closes #3861 
*

### Checklist

### Further comments

### DIFF
--- a/openbas-api/src/main/java/io/openbas/migration/V4_23__Force_es_reindex.java
+++ b/openbas-api/src/main/java/io/openbas/migration/V4_23__Force_es_reindex.java
@@ -6,7 +6,7 @@ import org.flywaydb.core.api.migration.Context;
 import org.springframework.stereotype.Component;
 
 @Component
-public class V4_21__Force_es_reindex extends BaseJavaMigration {
+public class V4_23__Force_es_reindex extends BaseJavaMigration {
 
   @Override
   public void migrate(Context context) throws Exception {
--- a/openbas-front/src/admin/components/simulations/SimulationList.tsx
+++ b/openbas-front/src/admin/components/simulations/SimulationList.tsx
@@ -194,8 +194,8 @@ const SimulationList: FunctionComponent<Props> = ({
               >
                 <ListItemButton
                   classes={{ root: classes.item }}
-                  component={ability.can(ACTIONS.ACCESS, SUBJECTS.RESOURCE, exercise.exercise_id) ? Link : 'div'}
-                  to={ability.can(ACTIONS.ACCESS, SUBJECTS.RESOURCE, exercise.exercise_id) ? `/admin/simulations/${exercise.exercise_id}` : undefined}
+                  component={(ability.can(ACTIONS.ACCESS, SUBJECTS.RESOURCE, exercise.exercise_id) || ability.can(ACTIONS.ACCESS, SUBJECTS.ASSESSMENT)) ? Link : 'div'}
+                  to={(ability.can(ACTIONS.ACCESS, SUBJECTS.RESOURCE, exercise.exercise_id) || ability.can(ACTIONS.ACCESS, SUBJECTS.ASSESSMENT)) ? `/admin/simulations/${exercise.exercise_id}` : undefined}
                 >
                   <ListItemIcon>
                     <HubOutlined color="primary" />


openbas-front/src/admin/components/common/injects/InjectPopover.tsx
```diff
@@ -242,13 +242,13 @@ const InjectPopover: FunctionComponent<Props> = ({
...
     disabled: isDisabled,
-    userRight: true,
+    userRight: permissions.canManage,
   });
   if (inject.inject_type !== 'openbas_manual' && canTriggerNow && onUpdateInjectTrigger) entries.push({
     label: t('Trigger now'),
     action: () => handleOpenTrigger(),
     disabled: isDisabled || !permissions.isRunning,
-    userRight: true,
+    userRight: permissions.canManage,
```
MarineLeM: why is not permission.canLaunch ?

openbas-front/src/admin/components/injects/InjectTestPopover.tsx
```diff
@@ -75,31 +65,20 @@ const InjectTestPopover: FunctionComponent<Props> = ({
...
-        <MenuItem
-          onClick={handleOpenTest}
-        >
-          {t('Replay test')}
-        </MenuItem>
-        <MenuItem onClick={handleOpenDelete}>
-          {t('Delete test')}
-        </MenuItem>
-      </Menu>
+      <ButtonPopover entries={entries} variant="icon" />
```
MarineLeM, line 81: 👍

openbas-front/src/admin/components/common/injects/InjectPopover.tsx
camrrx: You're right, i'll update it, thanks !

openbas-front/src/admin/components/lessons/scenarios/Lessons.tsx
```diff
@@ -34,6 +35,7 @@ interface Props {
   lessonsQuestions: LessonsQuestion[];
   lessonsAnswers?: LessonsAnswer[];
   lessonsTemplates: LessonsTemplate[];
+  usersMap: Record<string, User>;
```
MarineLeM: userMap is not use, you should remove it

openbas-front/src/admin/components/scenarios/scenario/lessons/ScenarioLessons.tsx
```diff
@@ -141,6 +143,7 @@ const ScenarioLessons = () => {
         lessonsCategories={lessonsCategories}
         lessonsQuestions={lessonsQuestions}
         lessonsTemplates={lessonsTemplates}
+        usersMap={usersMap}
```
MarineLeM: Should also be removed, because is not used in the Lessions component

openbas-front/src/utils/permissions/simulationPermissions.ts
```diff
@@ -0,0 +1,61 @@
+import * as R from 'ramda';
```
MarineLeM: avoid to use ramda please

openbas-front/src/utils/permissions/simulationPermissions.ts
```diff
@@ -0,0 +1,61 @@
...
+    canDelete,
+    readOnly,
+    isLoggedIn: !R.isEmpty(logged),
+    isRunning,
+  };
+};
+
+export default useSimulationPermissions;
+
+export const secondsFromToNow = (date: Date) => {
```
MarineLeM: I don't think this function should be in this field, may be in the Time.ts file ?

openbas-front/src/utils/permissions/simulationPermissions.ts
MarineLeM: Most of our hook files follow the pattern "useNameHook.js", but this one is just NameHook.js. We should maintain consistency across the entire project. May be it's something we can discuss with all the team ?

MarineLeM:
As a user who can only access the simulation, when I go to the Injects tab and open the interactive view, I’m able to move the inject box and therefore have the impresssion that i am able to update the start time
image

MarineLeM:
As a user who can only access the simulation, I should not be able to click on this button
image

openbas-front/src/admin/components/components/teams/TeamPlayers.tsx
```diff
@@ -323,7 +325,7 @@ const TeamPlayers: FunctionComponent<Props> = ({ teamId, handleClose }) => {
                     <div
                       className={classes.bodyItem}
                       style={inlineStyles.user_enabled}
```
MarineLeM: The user_enabled class has cursor: pointer even for disabled buttons. Non-clickable elements shouldn't show a pointer cursor - it's misleading UX.

PR #3939, opened 2025-09-08
[backend] Rework ES deletions (#3858)

### LINKED ISSUE #3858: Add filter simulation/scenario on all dimensions
## Context

To provide a better custom dashboard usage, we should be able to bind simulation/scenario param on all dimensions/widgets.

damgouj:
### Proposed changes

* Add missing objects in EsBase
* Remove base_dependencies to not "delete cascade" each object linked
* Add an API call at the end of the ES delete to remove properly the ids in the documents

### Testing Instructions

1. Create dashboards and widgets
2. Play with OpenAEV and check the data are consistent (add, update and delete data)

### Related issues

* Closes #3858  

### Checklist

### DIFF
--- a/openbas-model/src/main/java/io/openbas/engine/model/EsBase.java
+++ b/openbas-model/src/main/java/io/openbas/engine/model/EsBase.java
@@ -3,15 +3,17 @@ package io.openbas.engine.model;
 import io.openbas.annotation.EsQueryable;
 import io.openbas.annotation.Indexable;
 import io.openbas.annotation.Queryable;
-import io.openbas.database.model.*;
+import io.openbas.engine.model.assetgroup.EsAssetGroup;
 import io.openbas.engine.model.attackpattern.EsAttackPattern;
 import io.openbas.engine.model.endpoint.EsEndpoint;
 import io.openbas.engine.model.finding.EsFinding;
 import io.openbas.engine.model.inject.EsInject;
 import io.openbas.engine.model.injectexpectation.EsInjectExpectation;
 import io.openbas.engine.model.scenario.EsScenario;
+import io.openbas.engine.model.securityplatform.EsSecurityPlatform;
 import io.openbas.engine.model.simulation.EsSimulation;
 import io.openbas.engine.model.tag.EsTag;
+import io.openbas.engine.model.team.EsTeam;
 import io.openbas.engine.model.vulnerableendpoint.EsVulnerableEndpoint;
 import io.swagger.v3.oas.annotations.media.DiscriminatorMapping;
 import io.swagger.v3.oas.annotations.media.Schema;
@@ -35,6 +37,9 @@ import lombok.Setter;
       EsSimulation.class,
       EsTag.class,
       EsVulnerableEndpoint.class,
+      EsTeam.class,
+      EsAssetGroup.class,
+      EsSecurityPlatform.class,
     },
     discriminatorMapping = {
       @DiscriminatorMapping(value = "attack-pattern", schema = EsAttackPattern.class),
@@ -46,6 +51,9 @@ import lombok.Setter;
       @DiscriminatorMapping(value = "scenario", schema = EsScenario.class),
       @DiscriminatorMapping(value = "tag", schema = EsTag.class),
       @DiscriminatorMapping(value = "vulnerable-endpoint", schema = EsVulnerableEndpoint.class),
+      @DiscriminatorMapping(value = "team", schema = EsTeam.class),
+      @DiscriminatorMapping(value = "security-platform", schema = EsSecurityPlatform.class),
+      @DiscriminatorMapping(value = "asset-group", schema = EsAssetGroup.class),
     })
 public class EsBase {
 
@@ -68,7 +76,10 @@ public class EsBase {
   // -- Base for ACL --
   private List<String> base_restrictions;
 
-  // To support logical side deletions
+  // To support logical side deletions, means "DELETE CASCADE", to set only if you want to delete
+  // the linked object itself
+  // Example : I delete the inject Id-A, all objects which contain in their base_dependencies the
+  // Id-A will be entirely deleted
   // https://github.com/rieske/postgres-cdc could be an alternative.
   private List<String> base_dependencies = new ArrayList<>();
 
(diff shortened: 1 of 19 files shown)


openbas-model/src/main/java/io/openbas/engine/model/inject/InjectHandler.java
```diff
@@ -80,27 +79,21 @@ public List<EsInject> fetch(Instant from) {
                     inject.getAttack_patterns_children());
               }
               if (!isEmpty(inject.getInject_kill_chain_phases())) {
-                dependencies.addAll(inject.getInject_kill_chain_phases());
                 esInject.setBase_kill_chain_phases_side(inject.getInject_kill_chain_phases());
               }
               if (hasText(inject.getInject_injector_contract())) {
-                dependencies.add(inject.getInject_injector_contract());
                 esInject.setBase_inject_contract_side(inject.getInject_injector_contract());
```
MarineLeM, line 98: I think if we remove the injectorContract we also need to remove the inject.

openbas-model/src/main/java/io/openbas/engine/model/simulation/SimulationHandler.java
```diff
@@ -42,23 +42,18 @@ public List<EsSimulation> fetch(Instant from) {
...
-                dependencies.addAll(simulation.getExercise_asset_groups());
                 esSimulation.setBase_asset_groups_side(simulation.getExercise_asset_groups());
               }
               if (!isEmpty(simulation.getExercise_teams())) {
-                dependencies.addAll(simulation.getExercise_teams());
                 esSimulation.setBase_teams_side(simulation.getExercise_teams());
               }
               if (hasText(simulation.getScenario_id())) {
-                dependencies.add(simulation.getScenario_id());
                 esSimulation.setBase_scenario_side(simulation.getScenario_id());
```
MarineLeM, line 64: I think if we remove a scenario, we should also remove all its associated simulations ?

openbas-model/src/main/java/io/openbas/engine/model/EsBase.java
```diff
@@ -68,7 +76,10 @@ public class EsBase {
   // -- Base for ACL --
   private List<String> base_restrictions;
 
-  // To support logical side deletions
+  // To support logical side deletions, means "DELETE CASCADE", to set only if you want to delete
+  // the linked object itself
+  // Example : I delete the inject Id-A, all objects which contain in their base_dependencies the
+  // Id-A will be entirely deleted
```
MarineLeM, line 82: 👍

openbas-model/src/main/java/io/openbas/engine/model/injectexpectation/InjectExpectationHandler.java
```diff
@@ -94,12 +94,10 @@ public List<EsInjectExpectation> fetch(Instant from) {
                 esInjectExpectation.setBase_asset_group_side(injectExpectation.getAsset_group_id());
               }
               if (!isEmpty(injectExpectation.getAttack_pattern_ids())) {
-                dependencies.addAll(injectExpectation.getAttack_pattern_ids());
                 esInjectExpectation.setBase_attack_patterns_side(
                     injectExpectation.getAttack_pattern_ids());
               }
               if (!isEmpty(injectExpectation.getSecurity_platform_ids())) {
-                dependencies.addAll(injectExpectation.getSecurity_platform_ids());
                 esInjectExpectation.setBase_security_platforms_side(
```
MarineLeM, line 120: I think we should remove the expectation when we remove the security platform ?
Btw i set a manual security platform on an inject and nothing appears in the security_platforms_side, is it normal ?

openbas-model/src/main/java/io/openbas/engine/model/simulation/SimulationHandler.java
guillaumejparis: If we do that in elastic we need to do that in postgres ? is this really waht we want ?

openbas-model/src/main/java/io/openbas/engine/model/inject/InjectHandler.java
damgouj: I did the same as in Postgres, you can remove a Payload, it will set "null" on the injector contract attribute of the inject table ;)

openbas-model/src/main/java/io/openbas/engine/model/simulation/SimulationHandler.java
damgouj: I did the same as in Postgres, if you remove a scenario, you don't remove all its associated simulations ;)

openbas-model/src/main/java/io/openbas/engine/model/injectexpectation/InjectExpectationHandler.java
damgouj: - I did the same as in Postgres, if you remove a security platform, you don't remove the expectation ;)
- Well seen, I have to fix the part for the manual security platform, thanks !

openbas-model/src/main/java/io/openbas/engine/model/simulation/SimulationHandler.java
MarineLeM: ok my bad i thought we delete the simulation also

openbas-model/src/main/java/io/openbas/engine/model/EsBase.java
```diff
@@ -46,6 +51,9 @@
       @DiscriminatorMapping(value = "scenario", schema = EsScenario.class),
       @DiscriminatorMapping(value = "tag", schema = EsTag.class),
       @DiscriminatorMapping(value = "vulnerable-endpoint", schema = EsVulnerableEndpoint.class),
+      @DiscriminatorMapping(value = "team", schema = EsTeam.class),
```
savacano28, line 54: Praise!

openbas-model/src/main/java/io/openbas/engine/model/finding/FindingHandler.java
```diff
@@ -40,25 +40,37 @@ public List<EsFinding> fetch(Instant from) {
...
+                esFinding.setBase_scenario_side(null);
               }
               if (hasText(finding.getAsset_id())) {
                 dependencies.add(finding.getAsset_id());
                 esFinding.setBase_endpoint_side(finding.getAsset_id());
+              } else {
+                esFinding.setBase_endpoint_side(null);
+              }
+              if (dependencies.isEmpty()) {
+                esFinding.setBase_dependencies(List.of());
```
MarineLeM: This condition is redundant. You already initialized dependencies as an empty ArrayList, so checking if it's empty to set another empty list is unnecessary. Just use esFinding.setBase_dependencies(dependencies) directly - it works whether the list is empty or has elements.

MarineLeM (approved):

PR #3945, opened 2025-09-09
[frontend] RBAC - Fix select inject

camrrx:
### Proposed changes

* Fetch asset groups from scenario or exercise if we do not have asset right

### Testing Instructions

1. Log in with user granted on some scenario/simulations (manage) but without asset capa
2. Go on simulation / inject 
3. Check the checkbox of one inject
4. You shouldn't have 401

### Checklist

### Further comments

### DIFF
--- a/openbas-front/src/admin/components/common/ToolBar.js
+++ b/openbas-front/src/admin/components/common/ToolBar.js
@@ -32,7 +32,6 @@ import { Component, forwardRef } from 'react';
 import { connect } from 'react-redux';
 import { withStyles } from 'tss-react/mui';
 
-import { fetchAssetGroups } from '../../../actions/asset_groups/assetgroup-action';
 import { fetchEndpoints } from '../../../actions/assets/endpoint-actions';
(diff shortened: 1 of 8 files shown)


openbas-front/src/admin/components/common/ToolBar.js
```diff
@@ -644,22 +643,24 @@ class ToolBar extends Component {
...
-                  onClick={this.handleOpenBulkTest.bind(this)}
-                  color="primary"
-                  size="small"
-                >
-                  <ForwardToInbox fontSize="small" />
-                </IconButton>
-              </span>
-            </Tooltip>
+            {canLaunch ? (
+              <Tooltip title={t('Test')}>
```
heditar: This question is still open: do we need launch permission to test, given that we’re not actually launching?
@jborozco, could we get your thoughts on this?

heditar: JB confirmed that just Manage is enough

Issue #3334, opened 2025-06-10
Add nuclei CVE through Injector


Seb-MIGUEL:
The goal of that feature is to add all the 3000 CVES to be readable from OpenBAS, we need to upgrade the nuclei injector with a new kind of folder which  will contain all the CVE from: https://github.com/projectdiscovery/nuclei-templates/blob/main/cves.json

RomuDeuxfois:
Can we tackle this issue in this one and add vulnerable expectation by default in a nuclei injector contract ?
https://github.com/OpenBAS-Platform/openbas/issues/3243

Seb-MIGUEL:
Yes make sense

PR #3976, opened 2025-09-11
[backend] add possibility to link contract to vuln by vuln ext id (#3334)

antoinemzs:
### Proposed changes

* add possibility to link contract to vuln by vuln ext id

### Testing Instructions

1. API only: fill the `contract_vulnerability_external_ids` in the input json on `POST /api/injector_contracts`
2. 404 if some external ids are not foudn, otherwise 200 and vulns are linked in the association table

### Related issues

* Contributes to #3334

### Checklist

### Further comments

### DIFF
--- a/openbas-api/src/main/java/io/openbas/rest/cve/service/CveService.java
+++ b/openbas-api/src/main/java/io/openbas/rest/cve/service/CveService.java
@@ -67,7 +67,7 @@ public class CveService {
 
     // Batch fetch existing CVEs
     Map<String, Cve> existingCvesMap =
-        cveRepository.findAllByExternalIdIn(externalIds.stream().toList()).stream()
+        cveRepository.findAllByExternalIdInIgnoreCase(externalIds.stream().toList()).stream()
             .collect(Collectors.toMap(Cve::getExternalId, Function.identity()));
 
     // Process with pre-fetched data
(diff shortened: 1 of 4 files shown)


openbas-api/src/main/java/io/openbas/rest/cve/service/CveService.java
```diff
@@ -152,6 +145,27 @@ public Cve findByExternalId(String externalId) {
...
+    return vulns;
+  }
+
+  private void throwIfMissing(
+      Set<String> requiredIds,
+      List<Cve> fetchedVulnerabilities,
+      Function<? super Cve, String> getId) {
+    List<String> missingIds =
+        requiredIds.stream()
+            .filter(id -> !fetchedVulnerabilities.stream().map(getId).toList().contains(id))
```
savacano28: Questions:
* Could we have some problems with the case: Cve - cve?
* Why are we throwing an exception? wouldn't it be better to just log? (maybe I forget something :))

antoinemzs: On your second point:
> Why are we throwing an exception? wouldn't it be better to just log? (maybe I forget something :))

The idea here is to not silentily fail. There is a similar functionality for attack patterns; if the request asked for something that didn't exist, silently failing risks skewing the expectations of the client, whereas with an error response, the client has the opportunity to fix the request to a working state.

PR #3994, opened 2025-09-12
[backend] Upsert injects with contract related to vulnerabilities (#3508)

### LINKED ISSUE #3508: Parse Vulnerabilities from StixBundle
As we mapped the TTPs based on payloads attack patterns we will mapped the CVE used by a threat actor to nuclei CVE available into the platform and them used them into the scenario which will be generated

savacano28:
### Proposed changes

Chunk 3: Upsert injects related to vulenrabilities

* Add a method to upsert injects related to vulnerabilities(Import Stix):
* * Compare the current asset group configuration (tag rules) and existing vulnerabilities in the Stix bundle with injects already created in the same scenario.
* * Verify that an inject exists for every vulnerability and for the assets list in the requiremnt
* * Injects with a vulnerability not present in the bundle has to be deleted

/!\ In the next PR, we update the result in security coverage including the response from vulnerability expectations

### Testing Instructions

1. Import stix bundle (you can re sue bundle from pr)
2. Check creation Sceanrio, creation injects
3. Update your asset group (s) related to tag rule
4. Check Injects (upserts)

### Related issues

* Closes #3508 

### Checklist


openbas-api/src/main/java/io/openbas/rest/inject/service/InjectAssistantService.java
```diff
@@ -380,32 +377,20 @@ private Set<Inject> buildInjectsWithTargetsByVulnerability(
...
+    String targetField = extractTargetField(injectorContract);
+
+    if (CONTRACT_ELEMENT_CONTENT_TYPE_ASSET_GROUP.equals(targetField)) {
+      // Priority: asset-group exists because We use tag rules to fetch asset groups
+      inject.setAssetGroups(new ArrayList<>(assetGroupListMap.keySet()));
+    } else if (CONTRACT_ELEMENT_CONTENT_TYPE_ASSET.equals(targetField)) {
+      // Only compute flattened endpoints if asset exists
+      Set<Endpoint> flatEndpointsFromMap =
+          assetGroupListMap.values().stream().flatMap(List::stream).collect(Collectors.toSet());
+      inject.setAssets(new ArrayList<>(flatEndpointsFromMap));
```
antoinemzs: IMO this is breaking encapsulation; see it this way: on pne hand we have asset groups, and on the other we have an inject. We don't know whether we can assign asset groups directly or whether we have to expand the asset groups, deduplicate the assets and assign them instead.

The message is: `(AssetGroup | Assets) => Inject`

Therefore the inject (or rather the inject _service_ according to our architecture) should be dealing with this. It's a fairly generic situation: the expansion should be transparent from the caller's POV.
The inject service is responsible for prodding the associated injector contract (via service) for what it supports

openbas-api/src/test/java/io/openbas/api/stix_process/StixApiTest.java
```diff
@@ -563,6 +603,95 @@ void shouldCreateScenarioWithOneInjectWhenNoAssetGroupsExist() throws Exception
...
+
+      scenario = scenarioRepository.findById(scenarioId).orElseThrow();
+      injects = injectRepository.findByScenarioId(scenario.getId());
+      assertThat(injects).hasSize(1);
+      assertThat(
+              injects.stream()
+                  .filter(updated -> updated.getId().equals(inject.getId()))
+                  .map(Inject::getAssets)
+                  .toList())
+          .hasSize(1);
```
antoinemzs: Why I recommend doing actual asserts rather than just counting elements: here we are just counting how many injects have a collection of assets (no matter how many assets in each collection), which is always strictly equal to the number of injects (= each inject has a collection of assets, empty or not).

If we want to count how many assets in the filtered inject, we need to flatten before counting:
```suggestion
      assertThat(
              injects.stream()
                  .filter(updated -> updated.getId().equals(inject.getId()))
                  .flatMap(i -> i.getAssets().stream())
                  .toList())
          .hasSize(1);
```

The test passes by accident because the we expected 1 assets and counted 1 inject, 1 == 1.

antoinemzs (approved):

PR #4000, opened 2025-09-15
[backend] Fix npe in injector callback

### LINKED ISSUE #3550: My scenario can't run because injects stay in pending but it's executed

savacano28:
### Proposed changes

* Add condition for agent null (case of injectors)

### Testing Instructions

1. Create inject with contract nuclei
2. Execute
3. Inject should be correctly executed

### Related issues

* Closes #3550
*

### Checklist

### Further comments


openbas-api/src/main/java/io/openbas/rest/inject/service/InjectStatusService.java
```diff
@@ -127,11 +127,19 @@ public ExecutionTrace createExecutionTrace(
...
+    boolean noTraces = injectStatus.getTraces().isEmpty();
+    boolean noDuration = input.getDuration() == 0;
+    String agentId = (agent == null) ? null : agent.getId();
+
+    if (noTraces || noDuration) {
+      traceCreationTime = Instant.now();
+    } else {
+      traceCreationTime =
+          getExecutionTimeFromStartTraceTimeAndDurationByAgentId(
+              injectStatus, agentId, input.getDuration());
```
impolitepanda, line 141: We will still have an NPE here. If angent is null, agentId is also null (line 134) and in getExecutionTimeFromStartTraceTimeAndDurationByAgentId, line 115, we do agentId.equals.

impolitepanda: Maybe we should do noTraces || noDuration || agent == null, and remove line 134 ?

PR #4015, opened 2025-09-16
[frontend] Add skeleton on list loading

gabriel-peze:
### Proposed changes

Loading can take time on some modals, adding a skeleton on the display of :
- Add Team dialog on inject
- Add Asset Group dialog on inject
- Add Endpoint dialog on inject
- Team update
- User Group management

### Related issues

* Closes [Add teams → time take time to fetch, it can be good to have a loader here](https://www.notion.so/filigran/Add-teams-time-take-time-to-fetch-it-can-be-good-to-have-a-loader-here-26a8fce17f2a80c295d7e4f9020f76d0)


openbas-front/src/components/common/queryable/pagination/PaginationComponentV2.tsx
```diff
@@ -82,6 +84,12 @@ const PaginationComponentV2 = <T extends object>({
   const [properties, setProperties] = useState<PropertySchemaDTO[]>([]);
   const [options, setOptions] = useState<OptionPropertySchema[]>([]);
 
+  const setParentLoading = (loading: boolean) => {
+    if (typeof setLoading === 'function') {
+      setLoading(loading);
+    }
+  };
```
MarineLeM: You don’t need to wrap setLoading inside a separate setParentLoading function. Since setLoading might be undefined, you can directly use the optional chaining operator:
`setLoading?.(loading);`

Issue #4019, opened 2025-09-16
Fail gracefully at startup if username or password is missing


RomuDeuxfois:
## Context

In order to facilitate a user in his set-up:
- Fail gracefully at startup if username or password is missing

PR #4020, opened 2025-09-16
[backend] Fail gracefully at startup if username or password is missing (#4019)

RomuDeuxfois:
### Testing Instructions

1. Remove username and password from application.properties and change log level to logging.level.io.openbas=error

### Related issues

* Closes https://github.com/OpenAEV-Platform/openaev/issues/4019

### Breaking changes

* The default spring.mail configuration is now based on gmail

### DIFF
--- a/openbas-api/src/main/resources/application.properties
+++ b/openbas-api/src/main/resources/application.properties
@@ -16,9 +16,9 @@ openbas.cookie-duration=P1D
 openbas.unsecured-certificate=false
 openbas.with-proxy=false
 openbas.extra-trusted-certs-dir=
-#openbas.admin.email=admin@openbas.io
-#openbas.admin.password=ChangeMe
-#openbas.admin.token=ChangeMe
+#openbas.admin.email=admin@openbas.io # mandatory
+#openbas.admin.password=ChangeMe # mandatory
+#openbas.admin.token=ChangeMe # mandatory
 
 ########################
 # RUNTIME DEPENDENCIES #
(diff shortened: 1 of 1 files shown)


openbas-api/src/main/resources/application.properties
```diff
@@ -238,20 +238,20 @@ executor.openbas.binaries.version=@project.version@
...
 
-### /!\ MAKE SURE TO SETUP RABBITMQ (see RabbitMQ section above)
+### /!\ MAKE SURE TO SET UP RABBITMQ (see RabbitMQ section above)
 
 # Mail sending config (Always available, mandatory)
 openbas.default-mailer=no-reply@openbas.io
 openbas.default-reply-to=contact@openbas.io
-spring.mail.host=smtp.mail.com
-spring.mail.port=465
+spring.mail.host=smtp.gmail.com
```
antoinemzs, line 246: If this is supposed to be an example, let's not make it possible to leak credentials by mistake to google ;)
```suggestion
spring.mail.host=smtp.mailprovider.example
```
as per https://www.rfc-editor.org/rfc/rfc2606#page-2

RomuDeuxfois: I’m not sure I understand.
The idea is to set the Gmail configuration as the default. It’s just the Gmail SMTP server name.

RomuDeuxfois: Oh my mistake, could be clearer with this issue https://github.com/OpenAEV-Platform/docker/issues/67

antoinemzs: Ok the debate has happened and we are good to go

PR #4023, opened 2025-09-17
[frontend] Fix inject form

camrrx:
### Proposed changes

* Add the teams as default values 
* Make the difference between create mode and update mode :
    -  **create mode** we reset the default values as much as we click on injector contract : it has to fill the field in the inject form.
    - **update mode** the default values are reset when the component is mounted. 
* When we create a team in the inject form, it's automatically assigned to the simulation you are working on

### Testing Instructions

1. Go on simulation / inject tab 
2. Try to create a new inject : should be 🆗 
3. Go on media inject 
4. Try to update a team : should be 🆗 
5. Try to create a team : should appears in the definition tabs
6. Try to update the email subject : should be 🆗 
7. Play with the different sections where inject are

### Checklist

### Further comments

### DIFF
--- a/openbas-front/src/admin/components/common/injects/form/teams/InjectAddTeams.tsx
+++ b/openbas-front/src/admin/components/common/injects/form/teams/InjectAddTeams.tsx
@@ -46,7 +46,7 @@ const InjectAddTeams: FunctionComponent<Props> = ({
   // Standard hooks
   const { t } = useFormatter();
   const { classes } = useStyles();
-  const { searchTeams } = useContext(TeamContext);
+  const { searchTeams, onReplaceTeam } = useContext(TeamContext);
 
   const [teamValues, setTeamValues] = useState<TeamOutput[]>([]);
   const [selectedTeamValues, setSelectedTeamValues] = useState<TeamOutput[]>([]);
@@ -56,7 +56,6 @@ const InjectAddTeams: FunctionComponent<Props> = ({
 
   const handleClose = () => {
     setOpen(false);
-    setSelectedTeamValues([]);
   };
 
   const submitAddTeams = () => {
@@ -150,6 +149,8 @@ const InjectAddTeams: FunctionComponent<Props> = ({
                     onCreate={(team) => {
                       setTeamValues([...teamValues, team as TeamOutput]);
                       setSelectedTeamValues([...selectedTeamValues, team as TeamOutput]);
+                      // If a team is created, it has to be linked to the simulation/scenario
+                      onReplaceTeam?.([...selectedTeamValues, team as TeamOutput].map(v => v.team_id));
                     }}
                   />
                 </Can>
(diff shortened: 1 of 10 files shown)


openbas-front/src/admin/components/common/injects/form/teams/InjectTeamsList.tsx
```diff
@@ -53,14 +53,30 @@ const InjectTeamsList: FunctionComponent<Props> = ({ readOnly = false, hideEnabl
...
     const updatedTeamIds = injectTeamIds.filter((id: string) => id !== teamId);
     setValue('inject_teams', updatedTeamIds, { shouldValidate: true });
+    if (updatedTeamIds.length > 0) {
+      setTeams(prev => prev.filter(team => team.team_id !== teamId));
+    } else {
+      setTeams([]);
+    }
   };
 
   const teamListItem = (team: TeamOutput, userEnabled: number) => (
```
MarineLeM: Actually I tested without the changes from this file and it's working, so the problem really comes from the useEffect in the InjectForm file.

The older version has less duplication and keeps the logic in only one place. Moreover, in the older version we have separation of concerns, meaning the onTeamsChange and onRemoveTeam only handle form state updates.

So is that possible to remove the change from this file ? 
What do you think ?

MarineLeM (changes requested):

openbas-front/src/admin/components/common/injects/form/teams/InjectTeamsList.tsx
```diff
@@ -56,7 +56,7 @@ const InjectTeamsList: FunctionComponent<Props> = ({ readOnly = false, hideEnabl
   }, [injectTeamIds]);
 
   // -- ACTIONS --
-  const onTeamsChange = (teamIds: string[]) => setValue('inject_teams', teamIds, { shouldValidate: true });
+  const onTeamsChange = async (teamIds: string[]) => setValue('inject_teams', teamIds, { shouldValidate: true });
```
MarineLeM: We don't need the async here, it can be removed

openbas-front/src/admin/components/common/injects/form/teams/InjectAddTeams.tsx
```diff
@@ -150,6 +149,8 @@ const InjectAddTeams: FunctionComponent<Props> = ({
                     onCreate={(team) => {
                       setTeamValues([...teamValues, team as TeamOutput]);
                       setSelectedTeamValues([...selectedTeamValues, team as TeamOutput]);
+                      // If a team is created, it has to be linked to the simulation/scenario
+                      onReplaceTeam?.([...selectedTeamValues, team as TeamOutput].map(v => v.team_id));
```
MarineLeM, line 153: I think the replace is called too fast. Sample : When creating a new team without updating the inject form, should this team be automatically added to the simulation?

MarineLeM (approved):

PR #4024, opened 2025-09-17
[backend] RBAC - Fix unit tests

### LINKED ISSUE #3495: RBAC: Chunk 10: Integration tests
Refer to https://www.notion.so/filigran/RBAC-Implement-roles-capability-and-apply-them-to-grants-7ae145c285b64762beddfa30d07e7f99

impolitepanda:
### Proposed changes

* Re-enabled tests that were deactivated during the develoment of the RBAC features
* Rescoped test users that were switched to admin during the develoment of the RBAC features
* Removed old test annotation to manage Security Context user
* Replaced it with a global @ WithMockUser annotation taking the parameters wished for
* Implemented utility method to add grants to test users if needed
* Fixed issues revealed by rescope tests

### Testing Instructions

1. Launch unit tests

### Related issues

* Closes #3495 
* Related to issue #375 

### Checklist


openbas-api/src/main/java/io/openbas/service/PermissionService.java
```diff
@@ -77,10 +77,15 @@ public boolean hasPermission(
     // for inject/article the permission will be based on the parent's (scenario/simulation/test)
     // permission
     if (RESOURCES_USING_PARENT_PERMISSION.contains(resourceType)) {
-      Target parentTarget = resolveTarget(resourceId, resourceType, action);
-      resourceId = parentTarget.resourceId;
-      resourceType = parentTarget.resourceType;
-      action = parentTarget.action;
+      // Inject import fall back to MANAGE_ASSESSMENT
```
heditar: for the inject creation are we not using the permision on the parent resource ?

impolitepanda: So yes and no. When creating an inject in a standard way, you're absolutely right. And in those case, the RBAC is actually WRITE, not CREATE, and on the resource type of the parent (scneario, simulation, or atomic testing).
But there is one specific case : /api/injects/imports. It's the only case where we have CREATE + INJECT and it's a route that can import on several parents at the same time, so the previous logic can't apply, hence this fix.

impolitepanda: Rollbacked this change, and split the inject import endpoint into import for simulations and import for scenarios, thus falling back to RBAC for these resources instead of injects.

Issue #3172, opened 2025-05-14
Dashboard scope


jborozco:
## Context

be able to reduce scope of your dashboard on an time frame and on scope of scenario, inject etc..

PR #4030, opened 2025-09-17
[frontend] Fix custom dashboard form (#3172)

guillaumejparis:
Fix  : When creating a new dashboard and selecting “default home” it is not set as the home dashboard
Fix : Dashboard creation fails when starting from Parameters tab

### DIFF
--- a/openbas-front/src/admin/components/common/injects/form/InjectForm.tsx
+++ b/openbas-front/src/admin/components/common/injects/form/InjectForm.tsx
@@ -219,11 +219,12 @@ const InjectForm = ({
     // defaultValues: defaultValues,
   });
 
-  const { handleSubmit, reset, subscribe, getValues, clearErrors, trigger, formState: { isSubmitting } } = methods;
+  const { handleSubmit, reset, subscribe, getValues, setError, clearErrors, trigger, formState: { isSubmitting } } = methods;
 
   const onSubmit: SubmitHandler<InjectInputForm> = async (data) => {
     // we cannot save, even in draft, without title
     if (!data.inject_title?.length) {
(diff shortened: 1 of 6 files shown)


openbas-front/src/admin/components/common/injects/form/InjectForm.tsx
```diff
@@ -219,11 +219,12 @@ const InjectForm = ({
     // defaultValues: defaultValues,
   });
 
-  const { handleSubmit, reset, subscribe, getValues, clearErrors, trigger, formState: { isSubmitting } } = methods;
+  const { handleSubmit, reset, subscribe, getValues, setError, clearErrors, trigger, formState: { isSubmitting } } = methods;
 
   const onSubmit: SubmitHandler<InjectInputForm> = async (data) => {
     // we cannot save, even in draft, without title
     if (!data.inject_title?.length) {
+      setError('inject_title', { message: t('This field is required.') });
```
MarineLeM, line 227: Good catch

openbas-front/src/admin/components/workspaces/custom_dashboards/CustomDashboardForm.tsx
```diff
@@ -139,7 +115,7 @@ const CustomDashboardForm: FunctionComponent<Props> = ({
           minHeight: '100%',
           gap: theme.spacing(2),
         }}
-        onSubmit={handleSubmitWithoutDefault}
+        onSubmit={handleSubmit(onSubmit, handleSubmitWithErrors)}
```
MarineLeM, line 118: Nice one

Issue #3421, opened 2025-06-21
Payload creation tabs is missing border bottom


SamuelHassine:
## Description

Payload creation tabs is missing border bottom

![Image](https://github.com/user-attachments/assets/21bea16c-c7a1-435f-99de-2076a64c5ad7)

vs

![Image](https://github.com/user-attachments/assets/7babfbdc-ea45-47b4-9661-f62fcab8b7a6)

PR #4031, opened 2025-09-18
[frontend] Homogenize tabs style (#3421)

RomuDeuxfois:
### Related issues

* Closes https://github.com/OpenAEV-Platform/openaev/issues/3421

### DIFF
--- a/openbas-front/src/components/common/tabs/Tabs.tsx
+++ b/openbas-front/src/components/common/tabs/Tabs.tsx
@@ -1,0 +1,33 @@
+import { Box, Tab, Tabs as MUITabs } from '@mui/material';
+import { type FunctionComponent, type ReactNode, useCallback } from 'react';
+
+export interface TabsEntry {
+  key: string;
+  label: ReactNode;
+}
+
+const Tabs: FunctionComponent<{
+  entries: TabsEntry[];
+  currentTab: string;
+  onChange: (newValue: string) => void;
+}> = ({ entries = [], currentTab, onChange }) => {
+  const handleChange = useCallback((_e: unknown, newValue: string) => {
+    onChange(newValue);
+  }, [onChange]);
+
+  return (
+    <Box sx={{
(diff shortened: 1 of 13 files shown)


openbas-front/src/components/common/tabs/Tabs.tsx
```diff
@@ -0,0 +1,33 @@
+import { Box, Tab, Tabs as MUITabs } from '@mui/material';
+import { type FunctionComponent, type ReactNode, useCallback } from 'react';
+
+export interface TabsEntry {
+  key: string;
+  label: ReactNode;
+}
+
+const Tabs: FunctionComponent<{
```
MarineLeM, line 9: It’s confusing to name your own component the same as a Material UI component. When you use it in code, it becomes very easy to mix them up

guillaumejparis: Yep, I think we need to choose what to do here because we will have the same problem when integrating filigran-ui with potentially three Tabs component

RomuDeuxfois: We discussed this a few months ago and agreed on keeping a simple name without any prefix or suffix.

That said, happy to revisit the topic if it no longer makes sense.
Possible alternatives could be MuiTabs, FiligranTabs (from filigran-ui), or CustomTabs (as in OpenAEV before the proper design system?).

guillaumejparis: Mmmm maybe the good thing to do for the future is Tabs is the openaev component MuiTabs is the mui one & FiligranTabs the filigran-ui one 
Ofc we will need to modify a lof of mui import but it seems to be the good idea to avoid confusion

RomuDeuxfois: Ok for me. I’ll leave it like that for now, right ?

PR #4040, opened 2025-09-18
[backend] Stability improvement

### LINKED ISSUE #4034: [Tooling] - Equip the application with load/performance tests
## Context

OpenBAS lacks load testing, making it hard to validate performance and stability.

## Use case

Run load tests to detect bottlenecks and ensure production readiness.

## Current Workaround

Use external tools manually (Locust, JMeter) without integration.

## Proposed Solution

Provide built-in support or official tooling/docs for load testing OpenBAS.

## Additional Information

Could also be used in CI/CD to catch regressions.

Dimfacion:
### Proposed changes

* Disabling concurrent job for the simulation creation
* Better copy of variable into a newly created simulation as it seems to be taking a lot of CPU usage

### Testing Instructions

1. Test recurring jobs with variable to check if they are created and have all their variables

### Related issues

* Closes #4034
*

### Checklist

### DIFF
--- a/openbas-api/src/main/java/io/openbas/scheduler/jobs/ScenarioExecutionJob.java
+++ b/openbas-api/src/main/java/io/openbas/scheduler/jobs/ScenarioExecutionJob.java
@@ -16,6 +16,7 @@ import java.time.temporal.ChronoUnit;
 import java.util.List;
 import lombok.RequiredArgsConstructor;
 import org.jetbrains.annotations.NotNull;
+import org.quartz.DisallowConcurrentExecution;
 import org.quartz.Job;
 import org.quartz.JobExecutionContext;
 import org.quartz.JobExecutionException;
@@ -24,6 +25,7 @@ import org.springframework.transaction.annotation.Transactional;
 
 @Component
 @RequiredArgsConstructor
+@DisallowConcurrentExecution
 public class ScenarioExecutionJob implements Job {
 
   private final ScenarioService scenarioService;
--- a/openbas-api/src/main/java/io/openbas/rest/stream/StreamApi.java
+++ b/openbas-api/src/main/java/io/openbas/rest/stream/StreamApi.java
@@ -23,6 +23,7 @@ import org.springframework.http.HttpHeaders;
 import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
 import org.springframework.http.codec.ServerSentEvent;
+import org.springframework.scheduling.annotation.Async;
 import org.springframework.transaction.event.TransactionalEventListener;
 import org.springframework.web.bind.annotation.GetMapping;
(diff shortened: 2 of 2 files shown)


openbas-api/src/main/java/io/openbas/scheduler/jobs/ScenarioExecutionJob.java
```diff
@@ -24,6 +25,7 @@
 
 @Component
 @RequiredArgsConstructor
+@DisallowConcurrentExecution
```
RomuDeuxfois, line 28: Do we need to add this annotation on every job ?

openbas-api/src/main/java/io/openbas/service/ScenarioToExerciseService.java
```diff
@@ -318,21 +318,8 @@ public Exercise toExercise(
...
-                  exerciseVariable.setKey(scenarioVariable.getKey());
-                  exerciseVariable.setValue(scenarioVariable.getValue());
-                  exerciseVariable.setDescription(scenarioVariable.getDescription());
-                  exerciseVariable.setType(scenarioVariable.getType());
-                  exerciseVariable.setExercise(exerciseSaved);
-                  return exerciseVariable;
-                })
-            .toList();
-    this.variableService.createVariables(exerciseVariables);
+    this.variableService.copyVariableFromScenarioForSimulation(
```
RomuDeuxfois: I’d prefer that we keep it as it was for now, to make sure we properly fix the job concurrency issue. We can improve the queries later on. What do you think?

openbas-api/src/main/java/io/openbas/scheduler/jobs/ScenarioExecutionJob.java
Dimfacion: Not on each jobs in my opinion but we should study those that might become a problem if they collide (like they take a longer time than they periodicity). I think it is out of scope for this issue but would you be ok if we add it to our tech pains list ?

RomuDeuxfois: Ok, up to you :)

openbas-api/src/main/java/io/openbas/service/ScenarioToExerciseService.java
Dimfacion: I can move it to another PR but honestly, I would rather have that merged asap. It's abnormal that 13% of our CPU time in the last 3 hours has been on this little part of the code ...

Dimfacion: Here's the new PR : https://github.com/OpenAEV-Platform/openaev/pull/4046 :)

PR #4050, opened 2025-09-19
[backend/frontend] feat: add XTM Hub connectivity check

### LINKED ISSUE #4022: [Chunk 8] Add connectivity check

Kroustille:
### Proposed changes

* add settings to store connectivity check date and should send email
* add connector executed periocally to check connectivity
* add XTM Hub client which sends a request to get connectivity status
* add method in XTM Hub service to check connectivity and send email
* add endpoint in API to let front request a connectivity refresh
* refresh connectivity when experience page is loaded

### Testing Instructions

1. Step-by-step how to test

### Related issues

* Closes #4022

### Checklist

### Further comments

### DIFF
--- a/openbas-api/src/main/java/io/openbas/xtmhub/XtmHubEmailService.java
+++ b/openbas-api/src/main/java/io/openbas/xtmhub/XtmHubEmailService.java
@@ -60,7 +60,7 @@ public class XtmHubEmailService {
         """
       <p>We wanted to inform you that the connectivity between OpenAEV and the XTM Hub has been lost.
       As a result, the integration is currently inactive.</p>
-      <p>To restore functionality, please navigate to the <strong>Settings</strong> section and
+      <p>To restore the functionality, please navigate to the <strong>Settings</strong> section and
       re-initiate the registration process for the OpenAEV platform. This will re-establish the
       connection and allow continued use of the integrated features.</p>
       <p>If you need assistance during the process, don't hesitate to reach out.</p>


openbas-api/src/main/resources/application.properties
```diff
@@ -273,6 +273,10 @@ openbas.xtm.opencti.token=<opencti-token>
 openbas.xtm.hub.enable=true
 openbas.xtm.hub.url=https://hub.filigran.io
 #openbas.xtm.hub.override-api-url=<override-hub-api-url>
+openbas.xtm.hub.collector.enable=false
+openbas.xtm.hub.collector.id=b402f1f5-29ba-4ee3-b366-f0467754cf4e
+#openbas.xtm.hub.collector.connectivity-check-interval=3600000
```
Kroustille, line 278: I will update OpenAEV documentation once this PR is approved, to be sure I have the right parameter names 😉

openbas-api/src/main/java/io/openbas/xtmhub/XtmHubClient.java
```diff
@@ -0,0 +1,77 @@
...
+          httpPost,
+          classicHttpResponse -> {
+            if (classicHttpResponse.getCode() == HttpStatus.SC_OK) {
+              return XtmHubConnectivityStatus.ACTIVE;
+            } else {
+              return XtmHubConnectivityStatus.INACTIVE;
+            }
+          });
+    } catch (IOException e) {
+      log.warn("XTM Hub is unreachable on {}", config.getApiUrl(), e);
```
impolitepanda: log.warn("XTM Hub is unreachable on {}: {}", config.getApiUrl(), e.getMessage(), e);

While this ins't necessary per say, it's a bit easier for us when reading grafana if the exception message is directly in the log message instead if the trace.

impolitepanda: Also, are you ok with a warn only level here ? I don't believe we monitor them at all

openbas-api/src/main/java/io/openbas/xtmhub/XtmHubClient.java
```diff
@@ -0,0 +1,77 @@
...
+      httpPost.addHeader("Accept", "application/json");
+
+      StringEntity httpBody = buildMutationBody(platformId, token, platformVersion);
+      httpPost.setEntity(httpBody);
+      return httpClient.execute(
+          httpPost,
+          classicHttpResponse -> {
+            if (classicHttpResponse.getCode() == HttpStatus.SC_OK) {
+              return XtmHubConnectivityStatus.ACTIVE;
+            } else {
```
impolitepanda: Else is unnecessary
if (classicHttpResponse.getCode() == HttpStatus.SC_OK) {
  return XtmHubConnectivityStatus.ACTIVE;
}
return XtmHubConnectivityStatus.INACTIVE;

openbas-api/src/main/java/io/openbas/xtmhub/XtmHubClient.java
```diff
@@ -0,0 +1,77 @@
...
+          });
+    } catch (IOException e) {
+      log.warn("XTM Hub is unreachable on {}", config.getApiUrl(), e);
+
+      return XtmHubConnectivityStatus.INACTIVE;
+    }
+  }
+
+  @NotNull
+  private static StringEntity buildMutationBody(
```
impolitepanda: Is there a reason it's a static method ? As far as I can see, it's only called in line 31, where this.buildMutationBody would work fine

openbas-api/src/main/java/io/openbas/xtmhub/XtmHubEmailService.java
```diff
@@ -0,0 +1,90 @@
...
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.core.io.ResourceLoader;
+import org.springframework.stereotype.Service;
+
+@Service
+@RequiredArgsConstructor
+@Slf4j
+public class XtmHubEmailService {
+  private static final String EMAIL_SUBJECT =
```
impolitepanda, line 22: Hmmm... I don't know how to do it, but there is an i18n issue, here. Need to be discussed with staff engineers. Problem we haven't encountered yet, I believe (or that was put under the rug)

openbas-api/src/main/java/io/openbas/xtmhub/XtmHubEmailService.java
```diff
@@ -0,0 +1,90 @@
...
+
+  private final UserRepository userRepository;
+  private final MailingService mailingService;
+  private final PlatformSettingsService platformSettingsService;
+  private final ResourceLoader resourceLoader;
+
+  public void sendLostConnectivityEmail() {
+    List<User> administrators = findPlatformAdministrators();
+    if (administrators.isEmpty()) {
+      log.warn("No administrators found to send XTM Hub lost connectivity email");
```
impolitepanda: I believe this should be an error log. I don't think we should ever have a platform without any admins

openbas-api/src/main/java/io/openbas/xtmhub/XtmHubEmailService.java
```diff
@@ -0,0 +1,90 @@
...
+      return;
```
impolitepanda: In the same vein as previous comment, shouldn't we throw an exception here ? It's a case that, in my mind, should never occur and should probably trigger a 500

openbas-api/src/main/java/io/openbas/xtmhub/XtmHubEmailService.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    try {
+      String emailBody = buildEmailBody(administrators.getFirst());
+      mailingService.sendEmail(EMAIL_SUBJECT, emailBody, administrators);
+      log.info("XTM Hub lost connectivity email sent to {} administrators", administrators.size());
+    } catch (Exception e) {
+      log.error("Failed to send lost connectivity email", e);
```
impolitepanda: log.error("Failed to send lost connectivity email: {}", e.getMessage(), e);

openbas-api/src/main/java/io/openbas/xtmhub/XtmHubEmailService.java
```diff
@@ -0,0 +1,90 @@
...
+
+  public void sendLostConnectivityEmail() {
+    List<User> administrators = findPlatformAdministrators();
+    if (administrators.isEmpty()) {
+      log.warn("No administrators found to send XTM Hub lost connectivity email");
+      return;
+    }
+
+    try {
+      String emailBody = buildEmailBody(administrators.getFirst());
```
impolitepanda: I'm not sure i follow the getFirst logic, here. Can you explain why the executionContext only takes the first admin user ?

openbas-model/src/main/java/io/openbas/database/repository/UserRepository.java
```diff
@@ -143,6 +143,21 @@ List<RawPlayer> rawPlayersAccessibleFromOrganizations(
...
+  @Query(
+      value =
+          "SELECT users.*\n"
+              + "FROM users\n"
+              + "LEFT JOIN users_groups ON users_groups.user_id = users.user_id\n"
+              + "LEFT JOIN groups ON groups.group_id = users_groups.group_id\n"
+              + "LEFT JOIN groups_roles ON groups_roles.group_id = groups.group_id\n"
+              + "LEFT JOIN roles ON roles.role_id = groups_roles.role_id\n"
+              + "LEFT JOIN roles_capabilities ON roles_capabilities.role_id = roles.role_id\n"
+              + "WHERE capability = :capability\n"
```
impolitepanda: Capability BYPASS should also count.

openbas-model/src/main/java/io/openbas/database/repository/UserRepository.java
```diff
@@ -143,6 +143,21 @@ List<RawPlayer> rawPlayersAccessibleFromOrganizations(
       nativeQuery = true)
   Set<RawUser> rawUserByIds(@Param("ids") Set<String> ids);
 
+  @Query(
+      value =
+          "SELECT users.*\n"
```
impolitepanda: \n are not necessary

openbas-model/src/main/java/io/openbas/database/repository/UserRepository.java
```diff
@@ -143,6 +143,21 @@ List<RawPlayer> rawPlayersAccessibleFromOrganizations(
...
+              + "FROM users\n"
+              + "LEFT JOIN users_groups ON users_groups.user_id = users.user_id\n"
+              + "LEFT JOIN groups ON groups.group_id = users_groups.group_id\n"
+              + "LEFT JOIN groups_roles ON groups_roles.group_id = groups.group_id\n"
+              + "LEFT JOIN roles ON roles.role_id = groups_roles.role_id\n"
+              + "LEFT JOIN roles_capabilities ON roles_capabilities.role_id = roles.role_id\n"
+              + "WHERE capability = :capability\n"
+              + "OR users.user_admin = true\n"
+              + "GROUP BY users.user_id;",
+      nativeQuery = true)
```
impolitepanda: It wasn't possible to use a JPA query instead of a native one ?

openbas-model/src/main/java/io/openbas/database/repository/UserRepository.java
```diff
@@ -143,6 +143,21 @@ List<RawPlayer> rawPlayersAccessibleFromOrganizations(
...
+  List<User> usersHavingCapability(@Param("capability") String capability);
```
impolitepanda: Rename to document the fact it returns user with capa or admin or bypass

openbas-api/src/main/java/io/openbas/xtmhub/XtmHubEmailService.java
```diff
@@ -0,0 +1,90 @@
...
+      <p>
+        <a href="%s">Access OpenAEV</a><br />
+        Best,<br />
+        Filigran Team<br />
+      </p>
+      """,
+        baseUrl);
+  }
+
+  private List<User> findPlatformAdministrators() {
```
impolitepanda: So this is a wrong name in regards to what the function does. You do not get only admins here, but any user with write capability on platform settings.

openbas-api/src/main/java/io/openbas/xtmhub/XtmHubEmailService.java
```diff
@@ -0,0 +1,90 @@
...
+    return userRepository.usersHavingCapability(Capability.MANAGE_PLATFORM_SETTINGS.toString());
+  }
+
+  private String getTemplate() {
+    String templatePath = String.format(TEMPLATE_PATH_FORMAT, "en");
+    try (InputStream inputStream = resourceLoader.getResource(templatePath).getInputStream()) {
+      return new String(inputStream.readAllBytes());
+    } catch (IOException e) {
+      throw new RuntimeException("Failed to read template", e);
```
impolitepanda, line 92: Can you add an error log here ?

openbas-api/src/main/java/io/openbas/xtmhub/XtmHubService.java
```diff
@@ -24,11 +30,102 @@ public PlatformSettings register(@NotBlank final String token) {
         LocalDateTime.now(),
         XtmHubRegistrationStatus.REGISTERED,
         currentUser.getId(),
-        currentUser.getName());
+        currentUser.getName(),
```
impolitepanda: As there is nothing done on the currentUser in this class/method, shouldn't we deffer the call to User currentUser = userService.currentUser(); into updateXTMHubRegistration ?

openbas-api/src/main/java/io/openbas/xtmhub/XtmHubService.java
```diff
@@ -24,11 +30,102 @@ public PlatformSettings register(@NotBlank final String token) {
...
+
+  private boolean shouldSendConnectivityLossEmail(
+      PlatformSettings settings, ConnectivityCheckResult checkResult) {
+
+    return !checkResult.isActive()
+        && hasConnectivityBeenLostForTooLong(checkResult.lastCheck())
+        && isEmailNotificationEnabled(settings);
+  }
+
+  private boolean hasConnectivityBeenLostForTooLong(LocalDateTime lastCheck) {
```
impolitepanda, line 90: Simplify with somethin like this :
    return lastCheck.isBefore(LocalDateTime.now().minusHours(CONNECTIVITY_EMAIL_THRESHOLD_HOURS));

openbas-api/src/main/java/io/openbas/xtmhub/collector/XtmHubCollectorConfig.java
```diff
@@ -0,0 +1,18 @@
...
+import lombok.Getter;
+import lombok.Setter;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.stereotype.Component;
+
+@Setter
+@Component
+@ConfigurationProperties(prefix = "openbas.xtm.hub.collector")
+public class XtmHubCollectorConfig {
+  @Getter private boolean enable;
```
impolitepanda: Seeing as all fields have a @ Getter, set it at class level instead?

openbas-api/src/main/java/io/openbas/xtmhub/collector/XtmHubCollectorConfig.java
```diff
@@ -0,0 +1,18 @@
...
+
+  @Getter @NotBlank private String id;
+
+  @Getter @NotBlank private Integer connectivityCheckInterval = 60 * 60 * 1000;
```
impolitepanda: This feels like magic numbers. Can you comment on what this really represents (i mean, it's pretty obvious but none the less, but just a comment like "default to 1h")

openbas-api/src/main/java/io/openbas/xtmhub/collector/XtmHubCollector.java
```diff
@@ -0,0 +1,28 @@
...
+@RequiredArgsConstructor
+@Service
+public class XtmHubCollector {
+
+  private final XtmHubCollectorConfig config;
+  private final ThreadPoolTaskScheduler taskScheduler;
+  private final XtmHubService xtmHubService;
+
+  @PostConstruct
+  public void init() {
```
impolitepanda, line 19: After reviewing and discussions with staffs, there might be a need to rework this scheduler here. Will keep you updated asap.

impolitepanda (changes requested):

openbas-api/src/main/java/io/openbas/xtmhub/XtmHubService.java
Kroustille: There's a case where we want to update connectivity status with a collector. In this case, `currentUser` can't be found. 

I could propose another `PlatformSettingsService` methods which updates only necessary settings for the connectivity check, instead of passing every XTM Hub setting. 

What do you think @impolitepanda ?

openbas-api/src/main/java/io/openbas/xtmhub/collector/XtmHubCollector.java
impolitepanda: Decision made : keep it like this :)

openbas-api/src/main/java/io/openbas/xtmhub/XtmHubService.java
impolitepanda: You're absolutely correct, then. Could you send the whole user instead of two params instead, then ?

openbas-api/src/main/java/io/openbas/xtmhub/XtmHubClient.java
Kroustille: I think we did the same on OpenCTI because when XTM Hub is deployed, it becomes unreachable and we could potentially log errors but don't have anything to do to fix it.

I'm thinking that if we log an error instead of a warning, at least we could see how many times it occurs and decide if we need to something about it.

What do you think @impolitepanda ?

openbas-api/src/main/java/io/openbas/xtmhub/XtmHubEmailService.java
Kroustille: I agree that if there's no administrators on the platform, we have big problems 😅 

I can throw an `Exception` but I can't remember how to it properly. Should I throw a custom Exception here ? Or an Exception with a custom message do the trick ? 

Should I specify in parent methods that it throws an Exception ?

Do you have a best practice documentation about that somewhere ?

openbas-api/src/main/java/io/openbas/xtmhub/XtmHubEmailService.java
Kroustille: Yeah I wasn't really sure about what I was doing here, but ExecutionContext takes only one user and I didn't know how to handle it with a list of users, or if I should handle it or not.

By taking the first administrators, I was sure to have only one user in the `ExecutionContext` 😝 

Do you have a suggestion to improve it ? 🙏

openbas-api/src/main/java/io/openbas/xtmhub/XtmHubClient.java
impolitepanda: I would agree. It's not abnormal to have connection issues logged as error from time to time. As you say it's more the number of occurrences that matters. I would go for an error level log, I think.

openbas-api/src/main/java/io/openbas/xtmhub/XtmHubEmailService.java
impolitepanda: Confirmed, no i18n in the backend for now. I'll resolve this conv, then.

openbas-api/src/main/java/io/openbas/xtmhub/XtmHubEmailService.java
impolitepanda: Unfortunately, we know that our exception management is... far from perfect, let's say. So no guidelines, no best practices. It will come, one day :)

Apart from that, the best way to throw an exception in a case like this is to throw a RuntimeException. This way, it will be converted to a 500 on client side and you won't have to update the function and its parents' signatures to indcate they can throw.

openbas-api/src/main/java/io/openbas/xtmhub/XtmHubEmailService.java
```diff
@@ -0,0 +1,97 @@
...
+        Best,<br />
+        Filigran Team<br />
+      </p>
+      """,
+        baseUrl);
+  }
+
+  private List<User> findUsersAbleToManageSettings() {
+    List<String> capabilities =
+        Arrays.asList(Capability.MANAGE_PLATFORM_SETTINGS.toString(), Capability.BYPASS.toString());
```
impolitepanda: If you're building immutable lists, like this one, it's preferrable to use List.of instead of Arrays.asList. Arrays.asList instantiate both an array and a list, and is not fully immutable, it's just fixed length.

openbas-api/src/main/java/io/openbas/xtmhub/XtmHubService.java
Kroustille: I needed to use a record because I don't have the user when I'm checking connectivity, only id and name coming from PlatformSettings.

openbas-api/src/main/java/io/openbas/xtmhub/XtmHubEmailService.java
Kroustille: I created a new method which only takes a data `Map` to build the content from template because we don't explicitly needs the execution context 🙂

impolitepanda (approved):

openbas-api/src/main/resources/application.properties
Kroustille: Documentation PR available [here](https://github.com/OpenAEV-Platform/docs/pull/201)

PR #4053, opened 2025-09-22
[frontend/backend] add scenario name in list widget

MarineLeM:
### Proposed changes

* Add scenario name list widget


openbas-model/src/main/java/io/openbas/engine/model/scenario/EsScenario.java
```diff
@@ -17,7 +17,7 @@ public class EsScenario extends EsBase {
   /* Every attribute must be uniq, so prefixed with the entity type! */
   /* Except relationships, they should have same name on every model! */
 
-  @Queryable(label = "scenario name")
```
damgouj, line 20: I think you forgot to put the annotation "EsQueryable(keyword = true)" because if you sort or beakdown by name, it doesn't work ;)

PR #4058, opened 2025-09-22
[frontend/backend] fix access right on actionability

MarineLeM:
### Proposed changes

* Be able to see the actionability list even if you are a read only user


openbas-api/src/main/java/io/openbas/rest/settings/PlatformSettingsApi.java
```diff
@@ -152,6 +154,13 @@ public List<EsBase> homeDashboardEntities(
     return customDashboardService.homeDashboardEntities(widgetId, parameters);
   }
 
+  @PostMapping("/home_dashboard/entities-runtime/{widgetId}")
```
damgouj: Weird pattern here: _ and - in the URL

PR #4069, opened 2025-09-23
[frontend]Fix on dashboards

johanah29:


openbas-front/src/admin/components/workspaces/custom_dashboards/CustomDashboardParameters.tsx
```diff
@@ -87,14 +87,19 @@ const CustomDashboardParameters: FunctionComponent = () => {
...
-      <TimeRangeFilters
-        timeRangeValue={getParameter(dateParameters.get('timeRange'))?.value ?? LAST_QUARTER_TIME_RANGE}
-        handleTimeRange={data => handleParametersValue(dateParameters.get('timeRange'), data)}
-        startDateValue={getParameter(dateParameters.get('startDate'))?.value}
-        handleStartDate={data => handleParametersValue(dateParameters.get('startDate'), data)}
-        endDateValue={getParameter(dateParameters.get('endDate'))?.value}
-        handleEndDate={data => handleParametersValue(dateParameters.get('endDate'), data)}
-      />
+      {
+        !contextId && (
```
damgouj: How does it work ? You remove the time range for scenario and simulation but what is the default value ? Example -> I have a scenario which runs a simulation each day since 1 year -> if the default value is 3 months, we won't have all the data in the analysis tab.

johanah29: What do you suggest ?

damgouj: Check with the product team what they really want ?

PR #4085, opened 2025-09-25
[frontend]Fix layout of custom dashboard parameters

johanah29:
Former layout
image

New layout
image


RomuDeuxfois:
I am not a big fan of the new screen
image

We can just put the scenario and simulation filter at first
image

johanah29:
> I am not a big fan of the new screen image
> 
> We can just put the scenario and simulation filter at first image


New layout:
image

RomuDeuxfois:
@johanah29 please test the different use cases before requesting another review. We're losing time otherwise, thanks.

image

RomuDeuxfois (approved):

PR #4104, opened 2025-09-26
[backen/frontend] Add time range comparison based on date filter for number widget

### LINKED ISSUE #4071: Add time range comparison based on date filter
Until now before the elastic stack we had the following screen

Image

We can see  the variation on time and that's something we would like to have into our new dashboards. The idea is to use the custom time filter from the checkbox as a period and duplicate this period to compare the datas and show it through the widgets:

IE: below the today's view:

Image

We would like to use the same time period, here "last 3 months" and compare the results to the 3 months before that period then show the comparison.

impolitepanda:
### Proposed changes

* Updated count function for widgets to return an object containing count for current interval, previous interval, and difference between the two
* Updated the number widget to display the difference
* Updated widgets to have a tooltip with a message explaining in details what the number is

### Testing Instructions

1. On a dashboard with a number widget, play with the dashboard or widget time range to see the difference being displayed correctly.

### Related issues

* Closes #4071 

### Checklist

### Further comments

### DIFF
--- a/openbas-front/src/admin/components/workspaces/custom_dashboards/widgets/WidgetViz.tsx
+++ b/openbas-front/src/admin/components/workspaces/custom_dashboards/widgets/WidgetViz.tsx
@@ -87,40 +87,35 @@ const WidgetViz = ({ widget, fullscreen, setFullscreen, setTooltipMessage }: Wid
       Icon = ArrowDownwardOutlined;
       color = 'error.main';
     }
-    const labelKey = `${getTimeRangeItem(widgetTimeRange, t)?.label_key}_progression`;
+    const labelKey = `${getTimeRangeItem(widgetTimeRange)?.label_key}_progression`;
     return (
       <Box display="flex" alignItems="center">
         <Icon sx={{ color }} fontSize="small" />
-        <Box
-          display="flex"
-          alignItems="center"
+        <Typography
+          variant="body2"
+          component="span"
         >
-          <Typography
-            variant="body2"
-            component="span"
+          <Box
+            component="strong"
+            sx={{ color }}
           >
-            <Box
-              component="strong"
-              sx={{ color }}
-            >
-              {formattedDiff}
-            </Box>
+            {formattedDiff}
+          </Box>
+          {' '}
+          <strong>
+            {resourceName}
+          </strong>
+          {' '}
+          <span>
+            {t(labelKey)}
             {' '}
+          </span>
+          {widgetTimeRange !== ALL_TIME_TIME_RANGE && (
             <strong>
-              {resourceName}
+              {t('was previously', { previous_number: previous_interval_count })}
             </strong>
-            {' '}
-            <span>
-              {t(labelKey)}
-              {' '}
-            </span>
-            {widgetTimeRange !== ALL_TIME_TIME_RANGE && (
-              <strong>
-                {t('was previously', { previous_number: previous_interval_count })}
-              </strong>
-            )}
-          </Typography>
-        </Box>
+          )}
+        </Typography>
       </Box>
     );
   };
--- a/openbas-front/src/admin/components/workspaces/custom_dashboards/TimeRangeFilters.tsx
+++ b/openbas-front/src/admin/components/workspaces/custom_dashboards/TimeRangeFilters.tsx
@@ -20,7 +20,7 @@ const TimeRangeFilters: FunctionComponent<Props> = ({ handleTimeRange, handleSta
   const { t } = useFormatter();
   const theme = useTheme();
 
-  const timeRangeItems = getTimeRangeItems(t);
+  const timeRangeItems = getTimeRangeItems();
 
   return (
     <div style={{
@@ -46,7 +46,7 @@ const TimeRangeFilters: FunctionComponent<Props> = ({ handleTimeRange, handleSta
         >
           {timeRangeItems.map(item => (
             <MenuItem key={item.value} value={item.value}>
-              {item.label}
+              {t(item.label_key)}
             </MenuItem>
           ))}
         </Select>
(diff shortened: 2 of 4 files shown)


openbas-api/src/main/java/io/openbas/rest/dashboard/DashboardService.java
```diff
@@ -41,9 +42,10 @@ public class DashboardService {
    *
    * @param widgetId the id from the {@link Widget} defining the type and configuration
    * @param parameters parameters passed at runtime (e.g. filters, date ranges)
-   * @return long representing the count result
+   * @return EsCountInterval a count object, including the previous interval count and the
```
damgouj: current interval too ;)

openbas-model/src/main/java/io/openbas/engine/EngineService.java
```diff
@@ -44,9 +45,10 @@ public interface EngineService {
    *
    * @param user the user to use
    * @param runtime the count runtime to use
-   * @return a count
+   * @return a count object, including the previous interval count and the difference between the
```
damgouj: current interval too ;)

openbas-front/src/admin/components/workspaces/custom_dashboards/widgets/WidgetViz.tsx
```diff
@@ -27,26 +37,103 @@ export type SerieData = {
...
+    }
+    // Compute the widget time range to get the correct sentence
+    let widgetTimeRange;
+    if (widget.widget_config.time_range == 'DEFAULT') {
+      widgetTimeRange = getTimeRangeFromDashboard(customDashboard, customDashboardParameters);
+    } else {
+      widgetTimeRange = widget.widget_config.time_range;
+    }
+    const formattedDiff
+        = !difference_count || difference_count > 0 ? `+${difference_count}` : `${difference_count}`;
```
damgouj, line 79: Maybe you forgot "-" for the second value ?

MarineLeM:
We have the time range comparaison tooltip over all the widget, is that normal ? 
Additionally, from a UX perspective, it would be more intuitive for the time range explanation tooltip to appear when hovering over the counter itself, rather than requiring users to hover over the information icon. What do you think ?
image

damgouj (changes requested):
I am sorry but could you explain how to test the number in the "I" please ? I have always 0.
image
and it doesn't refresh the range in it when I update the time range filter.

My white and green numbers are the same, could we check together please ?
image

openbas-front/src/admin/components/workspaces/custom_dashboards/widgets/WidgetViz.tsx
impolitepanda: Nope, the number is already negative

openbas-front/src/components/ItemNumberDifference.js
```diff
@@ -8,9 +8,9 @@ import inject18n from './i18n';
 
 const styles = theme => ({
   diff: {
-    float: 'left',
-    margin: '23px 0 0 10px',
-    padding: '2px 5px 2px 5px',
+    display: 'flex',
+    alignItems: 'center',
+    padding: '2px 5px',
```
MarineLeM: Since you're refactoring this section - it would be great if you could use the theme for set the padding here

impolitepanda:
Added missing translation, thanks!

openbas-model/src/main/java/io/openbas/engine/EngineService.java
impolitepanda: Fixed

openbas-api/src/main/java/io/openbas/rest/dashboard/DashboardService.java
impolitepanda: Fixed

openbas-front/src/utils/lang/fr.json
```diff
@@ -77,6 +77,7 @@
   "All teams": "Toutes les équipes",
   "All teams value": "Valeur de toutes les équipes",
   "All time": "Tous les temps",
+  "All time_progression": "depuis le début",
```
MarineLeM, line 80: It's just a preference, but I find it odd that the translation file keys use spaces and underscore like 'All time_progression'. What do you think to use only underscores like : 'all_time_progression ?

MarineLeM (approved):

openbas-front/src/admin/components/workspaces/custom_dashboards/widgets/WidgetViz.tsx
```diff
@@ -27,25 +37,103 @@ export type SerieData = {
...
+      color = 'success.main';
+    } else if (difference_count && difference_count < 0) {
+      Icon = ArrowDownwardOutlined;
+      color = 'error.main';
+    }
+    const labelKey = `${getTimeRangeItem(widgetTimeRange, t)?.label_key}_progression`;
+    return (
+      <Box display="flex" alignItems="center">
+        <Icon sx={{ color }} fontSize="small" />
+        <Box
```
MarineLeM: I think you can remove this tag.

impolitepanda: Absolutely correct. Is was a relic from all the tries I made to have a good behavior on my text. Thanks for the spot!

openbas-front/src/utils/lang/fr.json
impolitepanda: I do agree with you. However, I'm a bit frisky about changing the keys as I don't know where they are used and could create a regression this way. I'll let it like this for now.

openbas-front/src/components/ItemNumberDifference.js
impolitepanda: Done ! Thanks

PR #4118, opened 2025-09-30
[backend] Fix import existing dashboard

RomuDeuxfois:


openbas-api/src/main/java/io/openbas/rest/custom_dashboard/CustomDashboardService.java
```diff
@@ -67,6 +67,20 @@ public CustomDashboard createCustomDashboard(@NotNull final CustomDashboard cust
     return this.customDashboardRepository.save(customDashboardWithDefaultParams);
   }
 
+  public static CustomDashboard sanityCheck(@NotNull final CustomDashboard customDashboard) {
+    CustomDashboard customDashboardWithDefaultParams = customDashboard;
+    if (customDashboard.getParameters().stream().noneMatch(p -> p.getType().equals(timeRange))) {
```
damgouj, line 71: Could we have the case where customDashboard.getParameters == null ? Maybe it's better to do it to avoid a NPE.

openbas-api/src/main/java/io/openbas/runner/InitStarterPackCommandLineRunner.java
```diff
@@ -159,7 +159,7 @@ private void importDashboardsFromResources() {
               try {
                 JsonApiDocument<ResourceObject> dashboard =
                     this.zipJsonService.handleImport(
-                        resourceToAdd.getContentAsByteArray(), "custom_dashboard_name", null);
+                        resourceToAdd.getContentAsByteArray(), "custom_dashboard_name", null, null);
```
damgouj: To think, maybe tomorrow we'll need the sanity check for the starter pack if the dashboards continue to evolve

openbas-model/src/main/java/io/openbas/jsonapi/GenericJsonApiImporter.java
```diff
@@ -70,7 +74,12 @@ public T handleImportEntity(JsonApiDocument<ResourceObject> doc, IncludeOptions
     entityManager.flush();
 
     // Persist root entity
-    return entityManager.merge(entity);
+    if (sanityCheck != null) {
+      T sanitized = sanityCheck.apply(entity);
+      return entityManager.merge(sanitized);
+    } else {
+      return entityManager.merge(entity);
+    }
```
gabriel-peze: I would do something like this, less code for the same result, and i think more readable
```
T entityToMerge = Optional.ofNullable(sanityCheck).map(check -> check.apply(entity)).orElse(entity);
return entityManager.merge(entity);
```

RomuDeuxfois:
```suggestion
T entityToMerge = Optional.ofNullable(sanityCheck).map(check -> check.apply(entity)).orElse(entity);
return entityManager.merge(entity);
```

RomuDeuxfois: I like it.

openbas-api/src/main/java/io/openbas/rest/custom_dashboard/CustomDashboardService.java
```diff
@@ -67,6 +67,20 @@ public CustomDashboard createCustomDashboard(@NotNull final CustomDashboard cust
...
+      customDashboardWithDefaultParams = customDashboard.addParameter("Time range", timeRange);
+    }
```
RomuDeuxfois, line 74:
```suggestion
    if(customDashboard.getParameters() == null) {
       return customDashboard;
    }
    if (customDashboard.getParameters().stream().noneMatch(p -> p.getType().equals(timeRange))) {
      customDashboardWithDefaultParams = customDashboard.addParameter("Time range", timeRange);
    }
```

openbas-api/src/main/java/io/openbas/rest/custom_dashboard/CustomDashboardService.java
RomuDeuxfois: Good point.

openbas-api/src/main/java/io/openbas/runner/InitStarterPackCommandLineRunner.java
RomuDeuxfois: Really good point.

gabriel-peze (approved):

damgouj (approved):

PR #4133, opened 2025-10-01
[frontend/backend] set last simulation ended as default scenario para…

### LINKED ISSUE #4070: Add last simulation into scenario's MITRE context
The MITRE widget does not have a “scenario” parameter since it displays the results of a simulation.
The goal is that, when a dashboard is used in a scenario context, the MITRE widget should be automatically pre-filled:
If the widget is configured with a simulation filter,
and the dashboard is opened in a scenario context,
then the simulation filter should default to the last ending simulation executed for that scenario.

MarineLeM:
### Proposed changes

* If the widget is configured with a simulation filter,
and the dashboard is opened in a scenario context,
then the simulation filter should default to the last ending simulation executed for that scenario.

### Related issues

* Closes #4070


openaev-api/src/main/java/io/openaev/utils/pagination/SortField.java
```diff
@@ -2,6 +2,8 @@
 
 import javax.annotation.Nullable;
 import lombok.Builder;
+import org.springframework.data.domain.Sort;
 
 @Builder
-public record SortField(String property, @Nullable String direction) {}
+public record SortField(
```
impolitepanda, line 8: Is there a need to keep both utils.SortField and utils.pagination.SortField ? They seem to be identical

impolitepanda (approved):
Test OK

openaev-api/src/test/java/io/openaev/rest/scenario/ScenarioSimulationApiTest.java
```diff
@@ -0,0 +1,238 @@
...
+  @Autowired private ScenarioRepository scenarioRepository;
+  @Autowired private ExerciseRepository exerciseRepository;
+  @Autowired private TagRepository tagRepository;
+
+  private static Scenario scenario;
+  private static Exercise exercise1FromScenario;
+  private static Exercise exercise2FromScenario;
+  private static Exercise exercise3NotFromScenario;
+
+  @BeforeAll
```
antoinemzs: `@BeforeAll` is done outside of the tests transaction, perhaps you wanted to do a `@BeforeEach` (which is done within the transaction, so is rolled back at the end of each test)?

MarineLeM: not necessary in this class I only test the search

antoinemzs: yes that's fine, but as it is, the objects persisted in this method will remain in the database (not rolled back), so it's maybe better to switch to a BeforeEach, so that they are effectively rolled back :)

PR #4143, opened 2025-10-03
[backend] Add starterpack flag on scenarios and remove Import label

### LINKED ISSUE #3527: Integrate a data starter pack by default on OBAS installation
## Context

- One technical scenario
- ready to use Nuclei injects
- ready to use nmap injects
- One technical dashboard
- All filigran collectors

gabriel-peze:
### Proposed changes

* Flag imported scenarios by starterpack to find them for the getting started page
* Remove "(Import)" label from items imported by the starter pack

### Testing Instructions

1. Launch the project with scenarios and dashboards in your starterpack ressources folder.
2. Check the getting started page and imported names

### Related issues

* Closes #3527


openaev-api/src/main/java/io/openaev/service/ImportService.java
```diff
@@ -53,13 +53,14 @@ private void handleDataImport(
       InputStream inputStream,
       Map<String, ImportEntry> docReferences,
       Exercise exercise,
-      Scenario scenario) {
+      Scenario scenario,
+      boolean isFromStarterPack) {
```
antoinemzs, line 59: Could we imagine having a String parameter instead to represent the suffix, and overloading the method so that the "suffix parameter"-less version defaults to the current known constant?
It would be simpler down the line to append the suffix (or no suffix) rather than if-elseing :)

openaev-api/src/main/java/io/openaev/importer/V1_DataImporter.java
```diff
@@ -365,6 +370,7 @@ private Scenario importScenario(JsonNode importNode, Map<String, Base> baseIds)
             .map(baseIds::get)
             .map(Tag.class::cast)
             .collect(Collectors.toSet()));
+    scenario.setFromStarterPack(isFromStarterPack);
```
antoinemzs, line 367: perhaps this should already be a serialised attribute of the imported material, not a special-case in the importer.
What do you think ?

gabriel-peze: pain point have been created, it will be manage later :)

PR #4144, opened 2025-10-03
[backend] Performance stream

camrrx:
### Proposed changes

* Add a ressourcetype blacklist to remove cve and payload from the stream

### Testing Instructions

1. Add a breakpoint in the if which return nothing in listenDatabaseUpdate of streamAPI
2. Update CVE : the breakpoint should be activate 
3. Update something else : should be ok

### Checklist

### DIFF
--- a/openaev-api/src/main/java/io/openaev/rest/stream/StreamApi.java
+++ b/openaev-api/src/main/java/io/openaev/rest/stream/StreamApi.java
@@ -11,6 +11,7 @@ import io.openaev.aop.RBAC;
 import io.openaev.config.OpenAEVPrincipal;
 import io.openaev.database.audit.BaseEvent;
 import io.openaev.database.model.Action;
+import io.openaev.database.model.ResourceType;
 import io.openaev.database.model.User;
 import io.openaev.rest.helper.RestBehavior;
 import io.openaev.service.PermissionService;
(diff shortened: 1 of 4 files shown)


openaev-api/src/main/java/io/openaev/rest/stream/StreamApi.java
```diff
@@ -64,10 +66,16 @@ private void sendStreamEvent(FluxSink<Object> flux, BaseEvent event) {
...
   }
 
+  private static final EnumSet<ResourceType> RESOURCES_STREAM_BLACKLIST =
+      EnumSet.of(ResourceType.CVE, ResourceType.PAYLOAD);
+
   @Async
   @Transactional
   @TransactionalEventListener
   public void listenDatabaseUpdate(BaseEvent event) {
+    if (RESOURCES_STREAM_BLACKLIST.contains(event.getInstance().getResourceType())) {
```
Dimfacion:
```suggestion
    if (!event.isListened()) {
```
**Question :** Wouldn't it be better to rely on the isListened and override it in the Java POJO to avoid having multiple ways of disabling listening of events ?

heditar: yes that's probably better, as we don't have more logic than check the resource type

heditar (approved):

PR #4147, opened 2025-10-03
[backend] Compare player and user for the update

camrrx:
### Proposed changes

* Compare the user and the player to know if we do the update or not

### Testing Instructions

1. Call upsertPlayer with identical input → no update
2. Call with modified fields → update

### Checklist


heditar:
Not sure why, but always insert when there is tags in the list

heditar (approved):

PR #4162, opened 2025-10-08
[backend] StarterPack - Fix empty Asset/AssetGroup on imported injects and OCTI tag rule management (#3527)

### LINKED ISSUE #3527: Integrate a data starter pack by default on OBAS installation

gabriel-peze:
### Proposed changes

* Add honey.scan.me asset or All endpoints AssetGroup by default on created injects (AssetGroup by default if option exist)
* Create OpenCti rule when it doesn't exist on StarterPack runner process

### Testing Instructions

1. Relaunch OpenAEV whitout the StarterPack scenarios, dashboards, asset, asset group, tags and tag rule on database, and all the payloads and injector contracts from NVD Nist CVE, NMAP, Nuclei existing on database.
2. OpenCTI tag rule should have the "All endpoints" asset group, and all imported technical scenarios should have injects with honey.scan.me asset or All endpoints if asset group field exist.

### Related issues

* Closes #3527


openaev-api/src/main/java/io/openaev/rest/injector_contract/InjectorContractContentUtils.java
```diff
@@ -102,4 +104,31 @@ public static ObjectNode getDynamicInjectorContractFieldsForInject(
...
+  public boolean hasField(InjectorContract injectorContract, String field) {
+    if (injectorContract == null || injectorContract.getContent() == null) {
+      return false;
+    }
+
+    try {
+      ObjectMapper mapper = new ObjectMapper();
+      ObjectNode objectNode = (ObjectNode) mapper.readTree(injectorContract.getContent());
+
+      return objectNode.get("fields") != null
```
damgouj, line 124: I am sorry but I am not sure to understand this method, could you explain please ?

openaev-api/src/main/java/io/openaev/runner/InitStarterPackCommandLineRunner.java
```diff
@@ -110,9 +107,12 @@ public void run(String... args) {
     try {
       Tag tagVulnerability = this.createTag(Tags.VULNERABILITY);
       Tag tagCisco = this.createTag(Tags.CISCO);
-      this.createHoneyScanMeAgentlessEndpoint(List.of(tagVulnerability.getId(), tagCisco.getId()));
-      this.createAllEndpointsAssetGroup();
-      this.importScenariosFromResources();
+      Tag openCti = this.createTag(Tags.OPENCTI);
```
damgouj, line 110: The OpenCTI tag is not created in another CommandLineRunner class ?

openaev-api/src/test/java/io/openaev/runner/InitStarterPackCommandLineRunnerTest.java
```diff
@@ -259,6 +280,158 @@ public void shouldInitStarterPack() {
...
+    this.verifyParameterExist();
+    this.verifyDefaultHomeDashboardParameterExist();
+    this.verifyDefaultScenarioDashboardParameterExist();
+    this.verifyDefaultSimulationDashboardParameterExist();
+    this.verifyTagRuleExist();
+  }
+
+  @Test
+  @DisplayName("Should init StarterPack with honey.scan.me asset")
+  public void shouldInitStarterPackWithDefaultAssets() throws JsonProcessingException {
```
damgouj, line 326: Cool :) !

openaev-api/src/main/java/io/openaev/rest/injector_contract/InjectorContractContentUtils.java
gabriel-peze: Sometimes we have only the Content on the injector contract, and not directly the ConvertedContract that's why i use the content, because this value is always set.
And i need to know if in the content (data who define the form on the frontend) there is a field asset or asset group, to be able to set it for the imported injects.
I wan't to check this, because i need to set an asset or an asset group only if it's necessary on the injector contract, for example mail sender injects don't need an asset or an asset group.

openaev-api/src/main/java/io/openaev/runner/InitStarterPackCommandLineRunner.java
gabriel-peze: Yes, InittagRuleCommandLineRunner, but the execution order can change for the runners, and on docker stack, on the StarterPack runner, the tag rule doesn't exist yet when we execute the code, so i create this security in case the starter pack run before the tag rule runner.

Issue #3705, opened 2025-07-24
Add OpenBAS connector into OpenCTI


Seb-MIGUEL:
## If the feature request is approved, would you be willing to submit a PR?

Yes / No (Help can be provided if you need assistance submitting a PR)

PR #4164, opened 2025-10-08
[backend] actually push stix over graphql (#3705)

antoinemzs:
### Proposed changes

*
*

### Testing Instructions

1. Step-by-step how to test

### Related issues

* Closes #ISSUE-NUMBER
*

### Checklist

### Further comments


openaev-api/src/main/java/io/openaev/opencti/client/mutations/PushStixBundle.java
```diff
@@ -0,0 +1,43 @@
...
+          stixBundlePush(connectorId: $connectorId, bundle: $bundle, work_id: $work_id)
+      }
+      """;
+  }
+
+  @Override
+  public JsonNode getVariables() throws JsonProcessingException {
+    ObjectMapper mapper = new ObjectMapper();
+    ObjectNode node = mapper.createObjectNode();
+    node.set("connectorId", mapper.valueToTree(connector.getId()));
```
impolitepanda, line 32: nitpick: we mix up snake case and camel case here (connectorId vs work_id). Can we do something about using the same case ?

openaev-api/src/main/java/io/openaev/opencti/connectors/service/OpenCTIConnectorService.java
```diff
@@ -47,4 +50,21 @@ public void registerOrPingAllConnectors() {
...
     }
   }
+
+  public void pushSecurityCoverageStixBundle(Bundle bundle) throws ConnectorError, IOException {
+    // don't examine the bundle
+    // pick the first occurrence of the correct connector type
+    // it's not supported yet to have more than one active connector of each type
+    Optional<ConnectorBase> connector =
+        connectors.stream()
+            .filter(c -> c instanceof SecurityCoverageConnector && c.shouldRegister())
```
impolitepanda, line 53: nitpick: instanceof here is fine as it's the single check we do of this kind, but maybe in the future we will need to differentiate types a lot more times. What do you think about creating a ConnectorType enum (name is an example) and having an abstract getter in the parent class, defined in the subclasses?

openaev-api/src/main/java/io/openaev/opencti/client/mutations/PushStixBundle.java
antoinemzs: agreed, unfortunately, it's the graphql contract on the OpenCTI side

openaev-api/src/main/java/io/openaev/opencti/connectors/service/OpenCTIConnectorService.java
antoinemzs: I'm not certain this is how this is meant to evolve. Currently, we make the assumption that there is a single instance of a given collector (it's aso a current limitation of the implementation), hence a single active SecurityCoverageConnector tied to a single OpenCTI instance; so we do not examine the sender of the input bundle, the output bundle will always be blindly sent to the same OpenCTI instance connected via that single active connector.

Later, I would expect to be discriminating the sender of the input bundle and make a decision of a destination for the output bundle based on that, not the reflected type of a connector.

Note that this object is only useful for providing URL and credentials for a standard POST request.

tl;dr: I don't expect this pattern to scale at all beyond the current state, and future evolutions to use a real attribute (albeit not a "type"), possibly from the input bundle.

openaev-api/src/main/java/io/openaev/service/stix/SecurityCoverageService.java
```diff
@@ -75,6 +72,9 @@ public class SecurityCoverageService {
 
   private final PreviewFeatureService previewFeatureService;
 
+  @Value("${openaev.xtm.opencti.connector.security-coverage.url}")
+  private String openctiUrl;
```
antoinemzs: Possibly safer to obtain this value via the connector itself, which is instanciated in the spring context. CI broke because the key is not defined, and no default is set here. The connector instanciation does some sanitisation too, which would be beneficial here since we expect to either use part or all of the key's value.

antoinemzs: Currently the connector appends "/graphql" at the end of the configured URL value, but perhaps it is useful to do something similar to the classic OpenCTI configuration and have both base and API url getters

openaev-api/src/main/java/io/openaev/service/stix/SecurityCoverageService.java
```diff
@@ -103,6 +103,10 @@ public SecurityCoverage buildSecurityCoverageFromStix(String stixJson)
     String name = stixCoverageObj.getRequiredProperty(STIX_NAME);
     securityCoverage.setName(name);
 
+    String configuredStripped = StringUtils.stripEnd(openctiUrl, "/");
+    String coveredRef = stixCoverageObj.getRequiredProperty(STIX_COVERED_REF);
+    securityCoverage.setExternalUrl(configuredStripped + "/dashboard/id/" + coveredRef);
```
antoinemzs: Would it make sense that a remote URL be sent as a whole in the input as an attribute to the STIX object? Or do you expect abuse?

openaev-api/src/main/java/io/openaev/service/stix/SecurityCoverageService.java
```diff
@@ -204,18 +207,26 @@ private void updatePropertiesFromSecurityCoverage(Scenario scenario, SecurityCov
...
+      scenario.setRecurrence(cron);
+    }
 
-    scenario.setTags(tagService.fetchTagsFromLabels(sa.getLabels()));
+    Set<Tag> tags = tagService.fetchTagsFromLabels(sa.getLabels());
+    TagCreateInput octiLabel = new TagCreateInput();
+    octiLabel.setName("opencti");
+    octiLabel.setColor("#001bda");
+    tags.add(this.tagService.upsertTag(octiLabel));
+    scenario.setTags(tags);
```
antoinemzs: AFAIK a tag with name `opencti` is already being created by default in all OpenAEV instances 
image
https://github.com/OpenAEV-Platform/openaev/blob/bc5e0c4a2cbc84c9988185a041e71c608999e660/openaev-api/src/main/java/io/openaev/runner/InitTagRuleCommandLineRunner.java#L37

antoinemzs: Here we risk changing the colour, no huge deal but for product consistency, there ought to be a factory for well known tags.
In our case here I think it's sufficient to expect the tag to exist already as part of the platform initalisation.

openaev-api/src/main/java/io/openaev/service/stix/CoverageResult.java
```diff
@@ -0,0 +1,17 @@
+package io.openaev.service.stix;
+
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class CoverageResult {
```
antoinemzs, line 8: Thanks, good call

openaev-api/src/main/java/io/openaev/service/stix/SecurityCoverageService.java
antoinemzs: postponed: use the configured base URL (`openaev.xtm.opencti.connector.security-coverage.url`) to construct the remote URL

openaev-api/src/main/java/io/openaev/service/stix/SecurityCoverageService.java
antoinemzs: Fixed.

openaev-api/src/main/java/io/openaev/service/stix/SecurityCoverageService.java
antoinemzs: Will do in another PR

PR #4166, opened 2025-10-08
[backend] null check on SecurityCoverageConnector.getUrl() (#3705)

antoinemzs:
### Proposed changes

* missing null check on getUrl for when a connector is not configured

### Testing Instructions

1. Unconfigured connectors should not trigger NPE in background job

### Related issues

* Contributes to #3705

### Checklist

### DIFF
--- a/openaev-api/src/test/resources/application.properties
+++ b/openaev-api/src/test/resources/application.properties
@@ -115,12 +115,6 @@ openaev.xtm.opencti.enable=false
 openaev.xtm.opencti.url=<opencti-url>
 openaev.xtm.opencti.token=<opencti-token>
 
-
-# security coverage connector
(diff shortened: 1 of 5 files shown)


openaev-api/src/test/resources/application.properties
```diff
@@ -115,12 +115,6 @@ openaev.xtm.opencti.enable=false
 openaev.xtm.opencti.url=<opencti-url>
 openaev.xtm.opencti.token=<opencti-token>
 
-
-# security coverage connector
```
savacano28, line 119: Suggestion: We could add these properties in application.properties as well?

savacano28 (approved):

openaev-api/src/test/resources/application.properties
antoinemzs: done

PR #4176, opened 2025-10-09
[backend ]import contract Issue/3527 

### LINKED ISSUE #3527: Integrate a data starter pack by default on OBAS installation

heditar:
### Proposed changes

* When the import is for the starter pack we will create the injector contract, these contract should come from the injector but as the starter pack is populated we try to import it and create some mock injector so the user can see the injects
* Once the injector register it will override the injector contract and delete the mock injector

### Testing Instructions

1. Start your service with a new db and no injector
2.  You should see the the starter pack scenarios with injects and "mock" injector in the injetor page
3. once you start your injectors you should see the regular injectors and no more mock
4. If you start a scenario before your injects are registered, it should be executed correctly after the registration (if it is done before the scenario is marked as maybe prevented)

### Related issues

* Closes #3527
*

### Checklist

### DIFF
--- a/openaev-api/src/main/java/io/openaev/service/InjectorService.java
+++ b/openaev-api/src/main/java/io/openaev/service/InjectorService.java
@@ -74,9 +74,9 @@ public class InjectorService {
   }
 
   /**
-   * This method will check if the injector type is a mock if yes it will remove the mock suffix if
-   * no it will return the parameter It is used to send the execution to the correct injector even
-   * if the current one is just a mock
+   * This method will check if the injector type is a dummy if yes it will remove the dummy suffix
+   * if no it will return the parameter It is used to send the execution to the correct injector
+   * even if the current one is just a dummy injector
    *
    * @param injectorType
    * @return
@@ -150,7 +150,7 @@ public class InjectorService {
                 .toList();
         injectorContractRepository.saveAll(injectorContracts);
 
-        // delete the mock injector if it was created when importing the starter pack
+        // delete the dummy injector if it was created when importing the starter pack
         deleteDummyInjectorIfItExists(input.getType());
       }
       InjectorConnection conn =


openaev-api/src/main/java/io/openaev/executors/Executor.java
```diff
@@ -40,13 +42,17 @@ public class Executor {
   private final ExecutionExecutorService executionExecutorService;
   private final InjectStatusService injectStatusService;
 
+  @Qualifier("coreInjectorService")
```
impolitepanda, line 45: Do we have several services called InjectorService? If that's the case, we should probably either combine them or rename one.
Qualifiers like this are a bit dangerous and could lead to confusing errors when developing.
I would avoid the use of qualifier as much as possible

openaev-api/src/main/java/io/openaev/importer/V1_DataImporter.java
```diff
@@ -1158,13 +1183,15 @@ private void importInjects(
 
           // Define default AssetsGroup or Assets
           Optional<Inject> injectOpt = injectRepository.findById(injectId);
-          if (injectorContract.isPresent() && injectOpt.isPresent()) {
+          if (injectOpt.isPresent() && injectOpt.get().getInjectorContract().isPresent()) {
             Inject inject = injectOpt.get();
             if (assetGroup != null
-                && injectorContractContentUtils.hasField(injectorContract.get(), "asset_groups")) {
+                && injectorContractContentUtils.hasField(
```
impolitepanda, line 1189: Should we log a warn or do something to inform the user that he is trying to add an assetGroup to an inject whose contract doesn't support them ?
Here, this case is just silently ignored

openaev-api/src/main/java/io/openaev/importer/V1_DataImporter.java
```diff
@@ -1194,10 +1221,53 @@ private void importInjects(
...
   }
 
+  /**
+   * Used to create an injector to be able to import injector contract from the starterpack before
+   * thye real contract is created by the real injector
+   *
+   * @param importNode contract node
+   * @return
+   */
+  private Injector createMockInjector(JsonNode importNode) {
```
impolitepanda: nitpick: the term mock is more often use in the context of tests. Could be call it something different ? Like Temporary, maybe ?

openaev-api/src/main/java/io/openaev/service/InjectorService.java
```diff
@@ -0,0 +1,255 @@
...
+
+  /**
+   * Create a mock injector, that is used when importing the starter pack before the real injector
+   * are registered
+   *
+   * @param injectorType
+   * @param injectorName
+   * @return
+   */
+  public Injector createMockInjector(
```
impolitepanda: nitpick: cf other comment about the use of the term mock

openaev-api/src/main/java/io/openaev/service/InjectorService.java
```diff
@@ -0,0 +1,255 @@
...
+    factory.setHost(rabbitmqConfig.getHostname());
+    factory.setPort(rabbitmqConfig.getPort());
+    factory.setUsername(rabbitmqConfig.getUser());
+    factory.setPassword(rabbitmqConfig.getPass());
+    factory.setVirtualHost(rabbitmqConfig.getVhost());
+    // Declare queueing
+    Connection connection = null;
+    try {
+      // Upload icon
+      if (file.isPresent() && "image/png".equals(file.get().getContentType())) {
```
impolitepanda, line 103: Could you use the MedaiType constant here instead of a string? (like MediaType.IMAGE_PNG_VALUE)

openaev-api/src/main/java/io/openaev/service/InjectorService.java
```diff
@@ -0,0 +1,255 @@
...
+        fileService.uploadFile(
+            FileService.INJECTORS_IMAGES_BASE_PATH + input.getType() + ".png", file.get());
+      }
+      connection = factory.newConnection();
```
impolitepanda, line 107: Is there a way to isolate the whole queue connection/configuration in a dedicated method to make this code a little bit more readable and reusable ?

openaev-api/src/main/java/io/openaev/service/InjectorService.java
```diff
@@ -0,0 +1,255 @@
...
+            .filter(c -> !existing.contains(c.getId()))
+            .map(in -> convertInjectorFromInput(in, injector))
+            .toList();
+    injectorContractRepository.deleteAllById(toDeletes);
+    injectorContractRepository.saveAll(toCreates);
+    return injectorRepository.save(injector);
+  }
+
+  // TODO JRI => REFACTOR TO RELY ON INJECTOR SERVICE
+  private InjectorContract convertInjectorFromInput(InjectorContractInput in, Injector injector) {
```
impolitepanda: I feel like this function would be better placed in the InjectorContractService instead. What do you think ?

openaev-api/src/main/java/io/openaev/executors/Executor.java
heditar: I will rename the existing one

openaev-api/src/main/java/io/openaev/service/InjectorService.java
heditar: Won't fix this comment here as it was exiting code

openaev-api/src/main/java/io/openaev/service/InjectorService.java
heditar: Will update with "dummy"

openaev-api/src/main/java/io/openaev/service/InjectorService.java
heditar: Won't fix this comment here as it was exiting code

openaev-api/src/main/java/io/openaev/service/InjectorService.java
heditar: Won't fix this comment here as it was exiting code

openaev-api/src/main/java/io/openaev/importer/V1_DataImporter.java
heditar: will update with "dummy"

openaev-api/src/main/java/io/openaev/executors/Executor.java
heditar: I added a TODO to merge the two. I’ll handle it separately during the cooldown week since I need a bit more understanding, and I want to ship this feature first.

PR #4178, opened 2025-10-10
[backend/frontend] Fix too many teams in bulk inject updates

impolitepanda:
### Proposed changes

* [frontend] fixed call to fetch teams that was unscoped in simulation for the distribution component
* [frontend] re-scoped getTeams from store that was too global for scenario and simulation inject tabs
* [backend] changed left outer join to inner join on team fetching to avoid having unrelated rows returned

### Testing Instructions

Replicating the bug before this PR :
1. Open a simulation with one team in it's definition
2. Open the inject tab and bulk update any injects to add a team. You should see 1 team, the one defined in the definition tab
3. Open a scenario with one different team in it's definition
4. Open the inject tab and bulk update any injects to add a team. You will see 2 teams, the one defined in the definition tab for the scenario and the one for the simulation you opened before

This PR fixes that.

### Related issues

* None

### Checklist


openaev-dev/Frontent start (MacOS).run.xml
```diff
@@ -0,0 +1,10 @@
+<component name="ProjectRunConfigurationManager">
```
damgouj: No need to commit this file I think ?

impolitepanda: Removed, thanks !

PR #4183, opened 2025-10-13
[frontend] Fix update asset groups for default asset rules

damgouj:
### Proposed changes

* Fix update asset groups for default asset rules

### Testing Instructions

1. Go to settings -> customization
2. Update a rule and check if everything is OK

### Checklist

### DIFF
--- a/openaev-front/src/admin/components/assets/asset_groups/AssetGroupsList.tsx
+++ b/openaev-front/src/admin/components/assets/asset_groups/AssetGroupsList.tsx
@@ -69,17 +69,11 @@ const AssetGroupsList: FunctionComponent<Props> = ({
 
     if (missingIds.length > 0) {
       // Can't check if the EndpointContext exists so check the URL to know which method used
-      if (location.pathname.includes(ASSET_RULES_BASE_URL)) {
-        findAssetGroups(missingIds).then((result) => {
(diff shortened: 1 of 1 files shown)


openaev-front/src/admin/components/assets/asset_groups/AssetGroupsList.tsx
```diff
@@ -64,10 +68,18 @@ const AssetGroupsList: FunctionComponent<Props> = ({
...
+        findAssetGroups(missingIds).then((result) => {
+          setAssetGroupValues([...result.data, ...assetGroups]);
+          setLoading(false);
+        });
+      } else {
+        fetchAssetGroupsByIds(missingIds).then((result) => {
+          setAssetGroupValues([...result.data, ...assetGroups]);
+          setLoading(false);
+        });
+      }
```
gabriel-peze: I think it should be possible to do something like that : 
```suggestion
      const assetGroupPromise = location.pathname.includes(ASSET_RULES_BASE_URL) ? findAssetGroups(missingIds) : fetchAssetGroupsByIds(missingIds);
      assetGroupPromise.then((result) => {
          setAssetGroupValues([...result.data, ...assetGroups]);
          setLoading(false);
        });
```

PR #4277, opened 2025-10-17
[frontend/backend] feat: use healthchecks to manage inject readyness (#4263)

### LINKED ISSUE #4263: Improve inject readiness with healthchecks
## Context
Improve the inject readyness process by centralizing it and by using healthchecks.

gabriel-peze:
### Proposed changes

* Use healthchecks to evaluate inject readyness and add more details into inject missing content state 
* Improve code by removing complex methods from DTOs and Entities

### Testing Instructions

1. Check for inject readyness state and details into scenarios injects, should not be changed but should have more details into the scenario readyness tooltip.

### Related issues

* Closes #4263

### DIFF
--- a/openaev-api/src/main/java/io/openaev/service/InjectImportService.java
+++ b/openaev-api/src/main/java/io/openaev/service/InjectImportService.java
@@ -370,6 +370,7 @@ public class InjectImportService {
 
       ZoneOffset zoneOffset = ZoneOffset.ofTotalSeconds(timezoneOffset * 60);
 
+      AtomicInteger count = new AtomicInteger(0);
       // For each rows of the selected sheet
       selectedSheet
           .rowIterator()


openaev-api/src/main/java/io/openaev/healthcheck/utils/HealthCheckUtils.java
```diff
@@ -157,6 +171,34 @@ public List<HealthCheck> runInjectorCheck(
...
+      List<HealthCheck> injectsHealthChecks) {
+    List<HealthCheck> result = new ArrayList<>();
+
+    if (injectsHealthChecks.stream()
+        .anyMatch(
+            healthCheck ->
+                Objects.equals(type, healthCheck.getType())
+                    && Objects.equals(detail, healthCheck.getDetail())
+                    && Objects.equals(status, healthCheck.getStatus()))) {
+      result.add(new HealthCheck(type, detail, status, now()));
```
RomuDeuxfois, line 192: **nitpick**: now() can be set as a default value, so we could remove it from here.

openaev-api/src/main/java/io/openaev/service/InjectImportService.java
```diff
@@ -178,9 +178,13 @@ public ImportTestSummary importInjectIntoFromXLS(
...
             "At least one of exercise or scenario should be present");
       }
-      importTestSummary.setInjects(new ArrayList<>());
+      importTestSummary.setInjectOutputs(new ArrayList<>());
     } else {
-      importTestSummary.setInjects(importTestSummary.getInjects().stream().limit(5).toList());
+      importTestSummary.setInjectOutputs(
+          importTestSummary.getInjects().stream()
+              .limit(5)
+              .map(inject -> injectMapper.toInjectOutput(inject, injectService.runChecks(inject)))
```
RomuDeuxfois: **question:** I am not sure we need to run an health check in this method call

openaev-api/src/main/java/io/openaev/rest/inject/output/InjectOutput.java
```diff
@@ -33,110 +40,179 @@ public class InjectOutput {
   @Schema(description = "Title of the inject")
   private String title;
 
+  @JsonProperty("inject_description")
```
RomuDeuxfois: **question**: I don’t quite understand this change — the DTO seems to be turning into something that looks like the original entity.
This could lead to performance issues if we end up returning the full inject object, which contains a large number of properties.
What’s the reason behind this change?

openaev-api/src/main/java/io/openaev/rest/inject/output/InjectOutput.java
```diff
@@ -1,18 +1,25 @@
 package io.openaev.rest.inject.output;
 
+import static java.time.Instant.now;
```
RomuDeuxfois: **question:** This isn’t directly related to the PR, but I don’t quite understand the purpose of `runInjectsChecksFor`.
From what I can tell, it seems to act as a kind of garbage collector for health checks that weren’t handled earlier — but why weren’t they?
It makes the code feel quite repetitive. Why do we need to handle it this way?

RomuDeuxfois:
Haven’t tested that part yet — just doing some code review to understand the mechanisms in place.

openaev-api/src/main/java/io/openaev/rest/inject/output/InjectOutput.java
gabriel-peze: The new healthchecks are the ones who creates the missing content failure, we want to filter them because the cause of the missing content is not linked to the scenario. Only the missing content information are important for the scenario.

For other checks (like SMTP, IMAP, Collector, etc), the only check to do is to verify if, into the injects healthcheck, the failure exist, and because into the injects healthcheck all checks are not necessary related to the scenario (Missing body, subject, asset, etc) we can't simply add all the injects checks to the scenario, and instead, verify the match into the injects checks.

openaev-api/src/main/java/io/openaev/rest/inject/output/InjectOutput.java
gabriel-peze: We shouldn't add the healthchecks informations to the Inject entity because there is no healthcheck information into the database, so we have to use a DTO to return the informations with healthchecks.
To avoid any regression, the InjectOutput have to hold the same informations as before, but the process didn't change.
All attributes exist into the InjectOutput, but it doesn't mean every time an InjectOutput is returned to the frontend all attributes are set.
For example, for the inject search, only a few informations are set to the InjectOutput object, the ones needed to display a light list of inject. And the get one Inject API still return the full informations as before.

RomuDeuxfois:
@gabriel-peze can I re read this one ?

openaev-api/src/main/java/io/openaev/service/InjectSearchService.java
```diff
@@ -184,23 +186,90 @@ private void selectForInject(
...
+              inject.setEnabled(tuple.get("inject_enabled", Boolean.class));
+              inject.setContent(tuple.get("inject_content", ObjectNode.class));
+              inject.setAllTeams(tuple.get("inject_all_teams", Boolean.class));
+              inject.setExercise(
+                  ofNullable(tuple.get("inject_exercise", String.class))
+                      .map(
+                          id -> {
+                            Exercise exercise = new Exercise();
+                            exercise.setId(id);
+                            return exercise;
```
RomuDeuxfois, line 213: **Question**: Why are we doing this? I’m not sure I understand the idea here.
I’m concerned that if someone uses the execInject method directly without looking at its implementation, they might think that it returns a full exercise.

gabriel-peze: Sorry for the late response.

Here in that case, and all the others, the final inject object need a full object, and can't work only with a String, and i didn't want to load the full Exercise, Scenario, Tags, Contracts, etc objects. To do it it would create joins on the request to get the informations from the ids provided by the inject table, and it will increase the execution time a lot.
But in the check process i need thoose ids, and because the inject need an Exercise object, and not only a String to store the provided id, i need to save him in an Exercise object, to be managed by the check process.

gabriel-peze:
To retest before merging, and validate with product before too.

PR #4287, opened 2025-10-20
[backend] fix response time for atomic test relaunch

impolitepanda:
### Proposed changes

* Replace deleteById hibernate call by a native one for Atomic Testing relaunch. Cascade delete are still handled correctly by cascade in the foreign key declaration in DB. Orphan removal should not be a problem as this is only used on relaunch.

### Testing Instructions

1. Relaunch an AT. 
2. In local : An AT used to take 1.2-1.4s to relaunch, now 200ms

### Related issues

* 

### Checklist

### DIFF
--- a/openaev-model/src/main/java/io/openaev/database/repository/InjectRepository.java
+++ b/openaev-model/src/main/java/io/openaev/database/repository/InjectRepository.java
@@ -385,7 +385,6 @@ public interface InjectRepository
   void deleteAllByScenarioIdAndInjectorContract(String injectorContract, String scenarioId);
 
   @Modifying
-  @Transactional
   @Query(value = "DELETE FROM injects WHERE inject_id = :id", nativeQuery = true)
-  int deleteByIdNative(@Param("id") String id);
+  void deleteByIdNative(@Param("id") String id);
 }


openaev-model/src/main/java/io/openaev/database/repository/InjectRepository.java
```diff
@@ -383,4 +383,9 @@ void deleteAllInjectsWithAttackPatternContractsByScenarioId(
           "DELETE FROM injects i WHERE i.inject_injector_contract = :injectorContract AND i.inject_scenario = :scenarioId",
       nativeQuery = true)
   void deleteAllByScenarioIdAndInjectorContract(String injectorContract, String scenarioId);
+
+  @Modifying
+  @Transactional
```
savacano28: Maybe we could remove this Transactional and return a void?

savacano28:
Tested OK: The relaunch time has been significantly reduced!

openaev-model/src/main/java/io/openaev/database/repository/InjectRepository.java
impolitepanda: Absolutely correct, thanks for spoting that

PR #4290, opened 2025-10-21
[frontend] Translation user's files from javascript to typescript

camrrx:
### Proposed changes

* Translate user's files from javascript to typescript

### Testing Instructions

1. Go on the user page and try all the functionalities, it should works as usual

### DIFF
--- a/openaev-front/src/admin/components/settings/users/CreateUser.tsx
+++ b/openaev-front/src/admin/components/settings/users/CreateUser.tsx
@@ -38,11 +38,7 @@ const CreateUser = ({ onCreate }: CreateUserProps) => {
     };
 
     return dispatch(addUser(inputValues)).then((result: UserResult) => {
-      if (!result?.entities?.users) {
-        return result;
-      }
-
-      if (onCreate) {
+      if (result?.entities?.users && onCreate) {
         const userCreated = result.entities.users[result.result];
 
         const orgId = userCreated.user_organization;
--- a/openaev-front/src/admin/components/settings/users/UserPopover.tsx
+++ b/openaev-front/src/admin/components/settings/users/UserPopover.tsx
@@ -42,9 +42,7 @@ const UserPopover = ({ user, onUpdate, onDelete }: UserPopoverProps) => {
     },
   );
 
-  const handleOpenEdit = () => {
-    setOpenEdit(true);
-  };
+  const handleOpenEdit = () => setOpenEdit(true);
 
   const handleCloseEdit = () => setOpenEdit(false);
 
@@ -56,11 +54,7 @@ const UserPopover = ({ user, onUpdate, onDelete }: UserPopoverProps) => {
     };
 
     return dispatch(updateUser(user.user_id, inputValues)).then((result: UserResult) => {
-      if (!result?.entities?.users) {
-        return result;
-      }
-
-      if (onUpdate) {
+      if (result?.entities?.users && onUpdate) {
         const userUpdated = result.entities.users[result.result];
 
         const orgId = userUpdated.user_organization;
@@ -78,19 +72,13 @@ const UserPopover = ({ user, onUpdate, onDelete }: UserPopoverProps) => {
     });
   };
 
-  const handleOpenEditPassword = () => {
-    setOpenEditPassword(true);
-  };
+  const handleOpenEditPassword = () => setOpenEditPassword(true);
 
   const handleCloseEditPassword = () => setOpenEditPassword(false);
 
-  const onSubmitEditPassword = (data: ChangePasswordInput) => {
-    dispatch(updateUserPassword(user.user_id, data)).then(() => handleCloseEditPassword());
-  };
+  const onSubmitEditPassword = (data: ChangePasswordInput) => dispatch(updateUserPassword(user.user_id, data)).then(() => handleCloseEditPassword());
 
-  const handleOpenDelete = () => {
-    setOpenDelete(true);
-  };
+  const handleOpenDelete = () => setOpenDelete(true);
 
   const handleCloseDelete = () => setOpenDelete(false);
 


openaev-front/src/admin/components/settings/users/UserPopover.tsx
```diff
@@ -0,0 +1,169 @@
...
+      user_organization: data.user_organization?.id,
+      user_tags: data.user_tags?.map((tag: Option) => tag.id),
+    };
+
+    return dispatch(updateUser(user.user_id, inputValues)).then((result: {
+      result: string;
+      entities: { users: Record<string, User> };
+    }) => {
+      if (onUpdate) {
+        const userUpdated = result.entities.users[result.result];
```
gabriel-peze, line 58: `result.entities?.users[result.result]`

In the case where there is an issue in the form (for example the phone miss an +) we go throw this method.
In this case it fail because result.entities is undefined.
image

But should we catch the 400 instead, and display a better message available into the result on the form, instead of only a toast ?
image

openaev-front/src/admin/components/settings/users/CreateUser.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const onSubmit = (data: UserInputForm) => {
+    const inputValues = {
+      ...data,
+      user_organization: data.user_organization?.id,
+      user_tags: data.user_tags?.map((tag: Option) => tag.id),
+    };
+
+    return dispatch(addUser(inputValues)).then((result: UserResult) => {
+      if (onCreate) {
+        const userCreated = result.entities.users[result.result];
```
gabriel-peze, line 42: I didn't succeed to have an issue here, but should we catch the issues to avoid getting a result without entities like in the update ?

openaev-front/src/admin/components/settings/users/UserPopover.tsx
camrrx: Good point, I’ve added the phone number format check directly in the form validation.
Now, if the number is filled but not in the expected format (+ followed by digits), the error is raised at the form level before the request is sent, so we won’t rely on the 400 response anymore for this case

openaev-front/src/admin/components/settings/users/CreateUser.tsx
```diff
@@ -0,0 +1,83 @@
...
+      user_organization: data.user_organization?.id,
+      user_tags: data.user_tags?.map((tag: Option) => tag.id),
+    };
+
+    return dispatch(addUser(inputValues)).then((result: UserResult) => {
+      if (!result?.entities?.users) {
+        return result;
+      }
+
+      if (onCreate) {
```
gabriel-peze:
```
      if (!result?.entities?.users) {
        return result;
      }

      if (onCreate) {
```

Can be replaced by 

```
if (result?.entities?.users && onCreate) {
```

openaev-front/src/admin/components/settings/users/UserPopover.tsx
```diff
@@ -0,0 +1,170 @@
...
+      return {
+        organizationsMap: helper.getOrganizationsMap(),
+        tagsMap: helper.getTagsMap(),
+      };
+    },
+  );
+
+  const handleOpenEdit = () => {
+    setOpenEdit(true);
+  };
```
gabriel-peze:
```
const handleOpenEdit = () => {
    setOpenEdit(true);
  };
```

Can be replaced by 

`const handleOpenEdit = () => setOpenEdit(true);`

openaev-front/src/admin/components/settings/users/UserPopover.tsx
```diff
@@ -0,0 +1,170 @@
...
+      user_organization: data.user_organization?.id,
+      user_tags: data.user_tags?.map((tag: Option) => tag.id),
+    };
+
+    return dispatch(updateUser(user.user_id, inputValues)).then((result: UserResult) => {
+      if (!result?.entities?.users) {
+        return result;
+      }
+
+      if (onUpdate) {
```
gabriel-peze:
```
if (!result?.entities?.users) {
        return result;
      }

      if (onUpdate) {
```

Can be replaced by 

`if (result?.entities?.users && onUpdate) {`

openaev-front/src/admin/components/settings/users/UserPopover.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+        onUpdate(userToUpdate);
+      }
+      return result.result ? handleCloseEdit() : result;
+    });
+  };
+
+  const handleOpenEditPassword = () => {
+    setOpenEditPassword(true);
+  };
```
gabriel-peze:
```
const handleOpenEditPassword = () => {
    setOpenEditPassword(true);
  };
```

Can be replaced by 

`const handleOpenEditPassword = () => setOpenEditPassword(true);`

openaev-front/src/admin/components/settings/users/UserPopover.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+  const handleCloseEditPassword = () => setOpenEditPassword(false);
+
+  const onSubmitEditPassword = (data: ChangePasswordInput) => {
+    dispatch(updateUserPassword(user.user_id, data)).then(() => handleCloseEditPassword());
+  };
```
gabriel-peze:
```
const onSubmitEditPassword = (data: ChangePasswordInput) => {
    dispatch(updateUserPassword(user.user_id, data)).then(() => handleCloseEditPassword());
  };
```
  
  Can be replaced by 

`const onSubmitEditPassword = (data: ChangePasswordInput) => dispatch(updateUserPassword(user.user_id, data)).then(() => handleCloseEditPassword());`

openaev-front/src/admin/components/settings/users/UserPopover.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+  const handleOpenDelete = () => {
+    setOpenDelete(true);
+  };
```
gabriel-peze:
```
const handleOpenDelete = () => {
    setOpenDelete(true);
  };
```
  
  Can be replaced by 

`const handleOpenDelete = () => setOpenDelete(true);`

gabriel-peze (approved):

PR #4291, opened 2025-10-21
[frontend/backend] fix expiration time inject expectations

MarineLeM:
### Proposed changes

* Fix in UI the expiration time
* Fix query to get inject expectation in the security manager job

### DIFF
--- a/openaev-api/src/main/java/io/openaev/service/InjectExpectationService.java
+++ b/openaev-api/src/main/java/io/openaev/service/InjectExpectationService.java
@@ -321,7 +321,7 @@ public class InjectExpectationService {
       InjectExpectation injectExpectation,
       Collector collector,
       InjectExpectationUpdateInput input,
-      Boolean shouldPropagateLastInjectExpectationResult) {
+      boolean shouldPropagateLastInjectExpectationResult) {
     // Update inject expectation at agent level
     injectExpectation =
         this.computeInjectExpectationForAgentOrAssetAgentless(injectExpectation, input, collector);
--- a/openaev-api/src/main/java/io/openaev/service/InjectExpectationUtils.java
+++ b/openaev-api/src/main/java/io/openaev/service/InjectExpectationUtils.java
@@ -23,7 +23,8 @@ import java.util.function.Function;
 import java.util.function.Predicate;
 
 public class InjectExpectationUtils {
-  public InjectExpectationUtils() {}
+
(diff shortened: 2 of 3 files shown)


openaev-api/src/main/java/io/openaev/service/InjectExpectationService.java
```diff
@@ -309,19 +313,25 @@ public void bulkUpdateInjectExpectation(
...
+      computeTechnicalExpectation(injectExpectation, collector, input, false);
     }
   }
 
   public void computeTechnicalExpectation(
       InjectExpectation injectExpectation,
       Collector collector,
-      InjectExpectationUpdateInput input) {
+      InjectExpectationUpdateInput input,
+      Boolean shouldPropagateLastInjectExpectationResult) {
```
damgouj: boolean ?

openaev-api/src/main/java/io/openaev/service/InjectExpectationUtils.java
```diff
@@ -16,12 +18,12 @@
...
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.function.Function;
 import java.util.function.Predicate;
 
 public class InjectExpectationUtils {
-
-  private InjectExpectationUtils() {}
+  public InjectExpectationUtils() {}
```
damgouj: rollback with private ?

openaev-api/src/main/java/io/openaev/service/InjectExpectationUtils.java
```diff
@@ -159,7 +161,22 @@ public static void computeScores(
...
+                    .filter(result -> newResultToAdd.getSourceId().equals(result.getSourceId()))
+                    .findFirst();
+            existingResult.ifPresent(
+                injectExpectationResult ->
+                    expectation.getResults().remove(injectExpectationResult));
+            expectation.getResults().add(newResultToAdd);
+
+            // IF RESULT TO ADD IS EXPIRATION MANAGER => SO I EXPIRE ALL the inject expectation with
+            // no result to expired
+            ExpectationsExpirationManagerConfig conf = new ExpectationsExpirationManagerConfig();
```
damgouj: Each time you go here, you initialize your config. I am not sure of the perf here ?
As you only need the collector id, maybe you could update the id by a "public static final" ?

PR #4295, opened 2025-10-21
[frontend/backend] Remove use of fromStarterPack for Dependencies (#4294)

### LINKED ISSUE #4294: Improve process by avoiding using fromStarterPack boolean
Use instead of fromStarterPack boolean a List of dependencies saved in database.

gabriel-peze:
### Proposed changes

* Use scenario dependencies instead of fromStarterPack value for more reusable code
* Add the dependencies value into the scenarios zip files, to get the starterpack information hold by the scenario instead of all the global import process

### Testing Instructions

1. All starterpack scenarios should implement as usual, with dummy injectors or real ones if implemented

### Related issues

* Closes #4294

### DIFF
--- a/openaev-api/src/main/java/io/openaev/importer/V1_DataImporter.java
+++ b/openaev-api/src/main/java/io/openaev/importer/V1_DataImporter.java
@@ -1055,9 +1055,10 @@ public class V1_DataImporter implements Importer {
           }
 
           if (injectorContractId == null) {
-            if (Arrays.asList(scenario.getDependencies())
-                .contains(Scenario.Dependency.STARTERPACK)) {
-              // if the we are importing starter pack, we will create the injector contract so the
+            if (scenario.getDependencies() != null
+                && Arrays.asList(scenario.getDependencies())
+                    .contains(Scenario.Dependency.STARTERPACK)) {
+              // if we are importing the starter pack, we will create the injector contract so the
               // injects are created before the injector registered
               // once the injector register the contract will be overriden and will be the one
               // provided by the injector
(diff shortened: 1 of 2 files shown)


openaev-api/src/main/java/io/openaev/importer/V1_DataImporter.java
```diff
@@ -1061,7 +1055,8 @@ private void importInjects(
           }
 
           if (injectorContractId == null) {
-            if (isFromStarterPack) {
+            if (Arrays.asList(scenario.getDependencies())
```
damgouj: Check null in this if or no need ?

openaev-api/src/main/java/io/openaev/service/ScenarioService.java
```diff
@@ -777,6 +777,7 @@ private Scenario copyScenario(Scenario scenario) {
     scenarioDuplicate.setLessonsAnonymized(scenario.isLessonsAnonymized());
     scenarioDuplicate.setDocuments(new ArrayList<>(scenario.getDocuments()));
     scenarioDuplicate.setGrants(new ArrayList<>(scenario.getGrants()));
+    scenarioDuplicate.setDependencies(scenario.getDependencies());
```
damgouj: If you duplicate a "starter pack" scenario, it goes to the "starter page" too ? There will be two same scenarios if I am not wrong ?

openaev-api/src/main/resources/starterpack/scenarios/scenario-CVE_TTPs_validation.zip
damgouj: Just to be sure, normal to delete this one ?

openaev-api/src/main/java/io/openaev/importer/V1_DataImporter.java
gabriel-peze: you're right, Arrays.asList is not null safe and will throw an error if a null is pass to him.

openaev-api/src/main/java/io/openaev/service/ScenarioService.java
gabriel-peze: You're right too, duplicated scenario should not be set as StarterPack, i'll fix it.

PR #4323, opened 2025-10-28
[frontend/backend] Rbac filter params #4342

camrrx:
### Proposed changes

* Check the contextId in the request params to allow the filter or not depending on the user permissions

### Related issues

* Closes: #4342 

### Testing Instructions

1. User granted manage on one simulation/scenario
2. Go on the inject tabs 
3. Add filters and check if there is not permission error (403)

### DIFF
--- a/openaev-api/src/main/java/io/openaev/service/PermissionService.java
+++ b/openaev-api/src/main/java/io/openaev/service/PermissionService.java
@@ -111,13 +111,13 @@ public class PermissionService {
     if (httpMappingInfo.isEmpty()) {
       return false;
     }
-    RBACAspect.HttpMappingInfo mappingIno = httpMappingInfo.get();
+    RBACAspect.HttpMappingInfo mappingInfo = httpMappingInfo.get();
     boolean endsWithOptions =
-        Arrays.stream(mappingIno.paths()).anyMatch(path -> path.endsWith("/options"));
+        Arrays.stream(mappingInfo.paths()).anyMatch(path -> path.endsWith("/options"));
     if (endsWithOptions) {
       // Retrieve the request param to check if a source ID is provided
-      if (mappingIno.args().containsKey("sourceId")) {
-        String sourceId = mappingIno.args().get("sourceId").toString();
+      if (mappingInfo.args().containsKey("sourceId")) {
+        String sourceId = mappingInfo.args().get("sourceId").toString();
         return hasGrantPermission(user, sourceId, resourceType, action);
       }
     }


openaev-api/src/main/java/io/openaev/service/PermissionService.java
```diff
@@ -97,6 +104,23 @@ public boolean hasPermission(
...
       return hasGrantPermission(user, resourceId, resourceType, action);
     }
+
+    // Specific case: /options endpoints are used to filter tables.
+    // In the context of a grantable resource, they  should be accessible if the sourceId associated
+    // with the request is a resource on which the user is granted
+    if (httpMappingInfo.isEmpty()) {
+      return false;
+    }
+    RBACAspect.HttpMappingInfo mappingIno = httpMappingInfo.get();
```
damgouj: mappingIno -> mappingInfo

openaev-api/src/main/java/io/openaev/aop/RBACAspect.java
```diff
@@ -91,4 +107,27 @@ public void methodRBACVerification(JoinPoint joinPoint, RBAC rbac)
...
     }
   }
+
+  public record HttpMappingInfo(
+      RequestMethod httpMethod, String[] paths, Map<String, Object> args) {}
+
+  private Optional<HttpMappingInfo> getHttpMappingInfo(Method method, Map<String, Object> args) {
+    if (method.isAnnotationPresent(GetMapping.class)) {
+      GetMapping ann = method.getAnnotation(GetMapping.class);
+      return Optional.of(new HttpMappingInfo(RequestMethod.GET, ann.value(), args));
```
damgouj, line 117: Maybe I miss something but I prefer to ask: why do you need the GET/POST/... ? It seems in the methods after, you just check the URL and parameters ?

impolitepanda: It's just future proofing. As of this moment, it's useless indeed, but it doesn't cost more to have it in case one day we need to have specific case that would need it.

openaev-api/src/main/java/io/openaev/service/PermissionService.java
impolitepanda: Fixed, thanks !

impolitepanda:
> Tests OK for me but just to be sure, the access grant is enough to have filters, right ?

Yes, exactly. Filters should be open relative to data you're allowed to see. And when you're granted on a scenario, atomic testing or a simulation, you are granted on all sub-resources as well. This include filter data

PR #4344, opened 2025-10-30
[backend/frontend] refactor: use Include Parameter Pattern on Team API (#1)

RomuDeuxfois:
### Issues

* We have some new perf issues in a specific endpoint

### Proposed changes

**Commit 1 - Include Paramater Pattern**
Instead of relying on implicit behavior or hidden persistence, the method now explicitly receives what needs to be included, improving readability, testability, and developer intent.

**Commit 2 - Layered architecture**

* The API controller delegates to a Service
* The Service builds the database query using a Query Object / Criteria Builder
* The DB result is mapped into a DTO before being returned
* The Service maps the DB DTO to the Service DTO
* The API returns this new DTO, and the front handles a clean DTO

Used by
Common in backend architectures at **Netflix**, **Spotify**, **Amazon**, and most **Spring Boot** enterprise systems.


RomuDeuxfois:
18/11/2025 -> Next steps: POC GraphQL

PR #4349, opened 2025-10-31
[frontend] test: add  e2e tests (#1)

MarineLeM:
### Proposed changes

Add e2e test
- payload form
- add team into scenario context

### DIFF
--- a/openaev-front/src/admin/components/common/injects/form/InjectForm.tsx
+++ b/openaev-front/src/admin/components/common/injects/form/InjectForm.tsx
@@ -523,6 +523,7 @@ const InjectForm = ({
           <Button
             variant="contained"
             color="secondary"
+            data-testid="inject-form-submit-button"
             onClick={() => {
               onSubmit(getValues());
             }}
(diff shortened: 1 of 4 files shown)


EmilieFo17 (commented):
```
  3 failed
    [chromium] › tests_e2e/tests/scenario/scenario-teams.spec.ts:58:5 › Scenario - Teams management › Player Management › should be able to activate and deactivate player 
    [chromium] › tests_e2e/tests/scenario/scenario-teams.spec.ts:88:5 › Scenario - Teams management › Teams in Injects › should only show scenario teams in inject form 
    [chromium] › tests_e2e/tests/scenario/scenario-teams.spec.ts:176:5 › Scenario - Teams management › Teams in Injects › should be able to add all the scenario teams
```



You must also add the .auth folder, otherwise the first test will fail.

The three tests fail when I run the command yarn playwright test --headed, but when I run them individually, they are OK.

openaev-front/playwright.config.ts
savacano28: Maybe we could add more browsers? https://playwright.dev/docs/best-practices#test-across-all-browsers

EmilieFo17: Discussing with @guillaumejparis, we just added Safari and Chrome so as not to overload the CI too much

guillaumejparis: Yep I'm thinking about a nightly workflow with tests on all browsers and maybe more e2e tests -> to discuss

guillaumejparis (approved):

PR #4358, opened 2025-11-10
[tools] chore(ci): add warning for test coverage (#1)

savacano28:
### Proposed changes

In this pull request, we introduced a warning message in step api-tests that triggers when test coverage falls below 50% for lines and 30% for branches in the openai-api project (Ci will not fail with this message) and also a check codecov for threshold 50%.

Currently, oav-api is at 49.5% line coverage, so the immediate goal is to increase coverage to at least 50% to avoid the warning in the step api-test.

The next step is to remove the warning message and instead make the CI fail when coverage drops below the required levels.

image

image

image

### Testing Instructions

1. You can modify goal coverage to check the message warning/success.

### Checklist

### Further comments

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -10,7 +10,3 @@ coverage:
         target: 50%
         threshold: 0
         informational: true
-comment:
-  layout: "diff"
-  comment: |
-    #### TEST COMMENT


.drone.yml
```diff
@@ -66,6 +68,7 @@ steps:
       - cd ..
       - mvn install -DskipTests=true
       - java -jar openaev-api/target/openaev-api.jar
+      - sleep 20
```
guillaumejparis: just to be sure it is really needed ?

codecov.yml
```diff
@@ -0,0 +1,10 @@
+coverage:
+  status:
+    project:
+      default:
+        target: 90%
```
Dimfacion: **question :** Shouldn't we put back 50% now that we finished our tests ?

codecov.yml
```diff
@@ -0,0 +1,10 @@
...
+        threshold: 0
+    patch:
+      default:
+        target: 90%
```
Dimfacion: **question :** same as https://github.com/OpenAEV-Platform/openaev/pull/4358/files#r2542400948

Issue #4370, opened 2025-11-14
NPE on /api/simulations/:id when id does not exist


antoinemzs:
## Description
```
Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception [Request processing failed: java.lang.RuntimeException: java.lang.NullPointerException: Cannot invoke "io.openaev.database.raw.RawSimulation.getInject_ids()" because "rawSimulation" is null] with root cause
```

## Environment

1. OS (where OpenAEV server runs): { e.g. Mac OS 10, Windows 10, Ubuntu 16.4, etc. }
2. OpenAEV version: { e.g. OpenAEV 1.0.2 }
3. OpenAEV client: { e.g. frontend or python }
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

PR #4371, opened 2025-11-14
[backend/frontend] fix(simulation): go to 404 page when simulation id is not found (#4370)

RomuDeuxfois:


openaev-api/src/main/java/io/openaev/rest/exercise/ExerciseApi.java
```diff
@@ -527,6 +527,9 @@ public void deleteExercise(@PathVariable String exerciseId) {
   public SimulationDetails exercise(@PathVariable String exerciseId) {
     // We get the raw exercise
     RawSimulation rawSimulation = exerciseRepository.rawDetailsById(exerciseId);
+    if (rawSimulation == null) {
+      throw new ElementNotFoundException("Simulation not found");
+    }
```
antoinemzs: Do you think we could improve this just so that it goes through a service layer and deals with NOT FOUND there ? This would be more in line with our practice (I understand the present code was already like this before)

antoinemzs (approved):
Approved with the caveat in comments: it would be great if it could just be slightly refactored before merging. Thanks!

PR #4376, opened 2025-11-17
[backend] fix(import): remap IDs in non entity content (#4364)

### LINKED ISSUE #4364: The dashboard isn’t working properly when applied to a simulation
## Description

When importing a dashboard that contains widgets linked to simulation parameters, all widget-to-parameter links are lost.
As a result, after the dashboard is applied to a simulation, the widgets no longer point to any simulation parameter, leaving the dashboard empty.

This happens on the Tabletop Exercice simulation from the starter pack.

## Environment

1. OpenAEV version: 2.0.0

## Reproducible Steps

1. Run a fresh OpenAEV instance.
2. Go to the Tabletop Exercice simulation dashboard.
3. Widgets are not linked to the simulation parameters.

## Expected Output

Imported dashboards should preserve widget → simulation parameter links.

Widgets should automatically bind to the referenced simulation parameters, and the dashboard should be populated accordingly.

## Actual Output

All widget → parameter links are lost during import.
Widgets show no associated parameter.
The dashboard remains empty even though widgets were correctly configured prior to export.

Image

## Important

Should work for the new instance and for the existing one.
Check every dashboard on the starter pack.

antoinemzs:
### Proposed changes

* Provide scheme to generically remap IDs in non entity content

### Testing Instructions

1. Establish custom dashboard with widgets linked to dashboard parameter (ex: dashboard-bound simulation param)
2. Export dashboard
3. Reimport
4. Imported dashboard widgets should have links correctly remapped to new dashboard

### Related issues

* Closes #4364

### Checklist

### Further comments

### DIFF
--- a/openaev-front/src/utils/api-types.d.ts
+++ b/openaev-front/src/utils/api-types.d.ts
@@ -580,6 +580,14 @@ type BasePayloadPayloadTypeMapping<Key, Type> = {
   payload_type: Key;
 } & Type;
 
+interface BaseWidgetConfiguration {
+  widget_configuration_type: string;
+}
+
+type BaseWidgetConfigurationWidgetConfigurationTypeMapping<Key, Type> = {
+  widget_configuration_type: Key;
+} & Type;
+
 export interface CVEBulkInsertInput {
   cves: CveCreateInput[];
   initial_dataset_completed?: boolean;
@@ -1206,7 +1214,7 @@ export interface CweOutput {
 
 export type DateHistogramWidget = UtilRequiredKeys<
   WidgetConfiguration,
-  "series" | "widget_configuration_type" | "time_range" | "date_attribute"
+  "widget_configuration_type"
 > & {
   display_legend?: boolean;
   interval: "year" | "month" | "week" | "day" | "hour" | "quarter";
@@ -2466,7 +2474,7 @@ export interface FlagInput {
 
 export type FlatConfiguration = UtilRequiredKeys<
   WidgetConfiguration,
-  "series" | "widget_configuration_type" | "time_range" | "date_attribute"
+  "widget_configuration_type"
 >;
 
 export interface FullTextSearchCountResult {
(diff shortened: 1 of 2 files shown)


openaev-model/src/main/java/io/openaev/engine/api/ListConfiguration.java
```diff
@@ -35,4 +36,24 @@ public static class ListPerspective {
   public ListConfiguration() {
     super(WidgetConfigurationType.LIST);
   }
+
+  @Override
+  public void remap(Map<String, String> map) {
+    if (this.perspective != null) {
```
RomuDeuxfois, line 42: **question:** series is always null on this ListConfiguration ?

openaev-model/src/main/java/io/openaev/jsonapi/GenericJsonApiImporter.java
```diff
@@ -56,7 +58,12 @@ public T handleImportEntity(
     // Cache keyed by id, with a boolean indicating whether the entity should be persisted or not at
     // the end
     Map<String, Pair<T, Boolean>> entityCache = new HashMap<>();
-    T entity = buildEntity(doc.data(), includedMap, entityCache, includeOptions, true);
+    List<CanRemapWeakRelationships> entitiesNeedingRemapping = new ArrayList<>();
```
RomuDeuxfois, line 61: **question:** Seems possible to use entityCache to store which relationship need a remaping. Or I am missing something ?

openaev-model/src/main/java/io/openaev/engine/api/ListConfiguration.java
antoinemzs: Yes since a recent refactor that moved `List<Series> series` to the WidgetConfiguration super class. This was intentionally not the case before since a ListConfiguration always has a single perspective by design.

openaev-model/src/main/java/io/openaev/jsonapi/GenericJsonApiImporter.java
antoinemzs: This is the main thing I'd like to refactor. The `entityCache` and `entitiesNeedingRemapping` collections don't contain the same set of objects; some in the latter are just attributes of the entities (i.e. the "non entities" in the hibernate sense).
So for the purpose of this change, the answer is no. In the long term I'd like a better scheme that doesn't require so much tracking.

PR #4379, opened 2025-11-18
[backend/frontend] feat(catalog): add catalog connector ingestion data (#4311)

### LINKED ISSUE #4311: [EE] Implement the catalog (UI+Backend)
## Context

Add the catalog logic to OBAS

camrrx:
### Proposed changes

* Ingestion json connectors file
* Display catalog page 

### Testing Instructions

1. Go on catalog page 
2. You should have access to all the connectors

### Related issues

* Closes #4311

### Checklist

### Further comments

### DIFF
--- a/openaev-api/src/main/java/io/openaev/runner/CatalogConnectorIngestionRunner.java
+++ b/openaev-api/src/main/java/io/openaev/runner/CatalogConnectorIngestionRunner.java
@@ -8,11 +8,13 @@ import java.io.InputStream;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.boot.CommandLineRunner;
+import org.springframework.context.annotation.Profile;
 import org.springframework.stereotype.Component;
 
 @Slf4j
 @RequiredArgsConstructor
 @Component
+@Profile("!test")
 public class CatalogConnectorIngestionRunner implements CommandLineRunner {
   private static final ObjectMapper mapper = new ObjectMapper();
   private static final String resourcePath = "/catalog/catalog-integrators.json";
--- a/openaev-api/src/test/java/io/openaev/service/CatalogConnectorIngestionServiceTest.java
+++ b/openaev-api/src/test/java/io/openaev/service/CatalogConnectorIngestionServiceTest.java
@@ -10,7 +10,6 @@ import io.openaev.database.repository.ConnectorInstanceConfigurationRepository;
 import io.openaev.rest.connector_instance.service.ConnectorInstanceService;
 import io.openaev.service.catalog_connectors.CatalogConnectorIngestionService;
 import io.openaev.service.catalog_connectors.CatalogConnectorService;
-import io.openaev.utilstest.RabbitMQTestListener;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
@@ -18,13 +17,9 @@ import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.context.SpringBootTest;
-import org.springframework.test.context.TestExecutionListeners;
 import org.springframework.transaction.annotation.Transactional;
 
 @SpringBootTest
-@TestExecutionListeners(
-    value = {RabbitMQTestListener.class},
-    mergeMode = TestExecutionListeners.MergeMode.MERGE_WITH_DEFAULTS)
 @DisplayName("Catalog connectors process tests")
 @Transactional
 public class CatalogConnectorIngestionServiceTest {


openaev-api/src/main/java/io/openaev/runner/CatalogConnectorIngestionRunner.java
```diff
@@ -0,0 +1,215 @@
...
+@Component
+public class CatalogConnectorIngestionRunner implements CommandLineRunner {
+  private final CatalogConnectorService catalogConnectorService;
+  private static final ObjectMapper mapper = new ObjectMapper();
+  private final CatalogConnectorConfigurationService catalogConnectorConfigurationService;
+  private final FileService fileService;
+
+  @Override
+  public void run(String... args) {
+    String resourcePath = "/catalog/catalog-integrators.json";
```
MarineLeM: Should this be an environment variable? What do you think?

openaev-api/src/main/java/io/openaev/runner/CatalogConnectorIngestionRunner.java
```diff
@@ -0,0 +1,215 @@
...
+        throw new IOException("File not found : " + resourcePath);
+      }
+
+      JsonNode rootNode = mapper.readTree(is);
+
+      List<CatalogConnector> catalog = extractCatalog(rootNode);
+
+      System.out.println("Build catalog:");
+      for (CatalogConnector connector : catalog) {
+        System.out.println("  • " + connector.getTitle());
```
MarineLeM: Please remove System.out.println and use @Slf4j logging instead. What is the purpose of this 'for' loop ? It only log connector titles?

openaev-api/src/main/java/io/openaev/runner/CatalogConnectorIngestionRunner.java
```diff
@@ -0,0 +1,215 @@
...
+import org.springframework.boot.CommandLineRunner;
+import org.springframework.stereotype.Component;
+
+@Slf4j
+@RequiredArgsConstructor
+@Component
+public class CatalogConnectorIngestionRunner implements CommandLineRunner {
+  private final CatalogConnectorService catalogConnectorService;
+  private static final ObjectMapper mapper = new ObjectMapper();
+  private final CatalogConnectorConfigurationService catalogConnectorConfigurationService;
```
MarineLeM: This service is not used in this class

openaev-api/src/main/java/io/openaev/rest/catalog_connector/CatalogConnectorApi.java
```diff
@@ -0,0 +1,62 @@
...
+@RestController
+@RequiredArgsConstructor
+public class CatalogConnectorApi extends RestBehavior {
+  public static final String CATALOG_CONNECTOR_URI = "/api/connectors";
+  private final CatalogConnectorService catalogConnectorService;
+  private final FileService fileService;
+
+  @GetMapping(CATALOG_CONNECTOR_URI)
+  @RBAC(actionPerformed = Action.READ, resourceType = ResourceType.CATALOG)
+  public List<CatalogConnector> getCatalogConnectors() {
```
MarineLeM: You should create a DTO to avoid fetching unnecessary CatalogConnector informations when retrieving the catalog list

openaev-front/src/admin/components/integrations/catalog_connectors/ConnectorDetails.tsx
```diff
@@ -0,0 +1,128 @@
...
+    flexShrink: 0,
+    padding: theme.spacing(2),
+    alignSelf: 'flex-start',
+    textTransform: 'uppercase',
+    borderRadius: 4,
+  },
+  paperConnector: {
+    display: 'flex',
+    flexDirection: 'column',
+    borderRadius: 4,
```
MarineLeM: You can remove
`borderRadius: 4,`
because this css property is already include on the MUI Paper component

openaev-front/src/admin/components/integrations/catalog_connectors/ConnectorDetails.tsx
```diff
@@ -0,0 +1,128 @@
...
+        />
+      </div>
+      <div className={classes.content}>
+        <Typography variant="h4">{t('Overview')}</Typography>
+        <Typography variant="h4">{t('Basic Information')}</Typography>
+
+        <Paper variant="outlined" className={classes.paperConnector}>
+          {connector.catalog_connector_description}
+        </Paper>
+        <Paper variant="outlined" className={classes.paperConnector}>
```
MarineLeM:
```suggestion
        <Paper variant="outlined" className={`paper ${classes.paperConnector}`}>
```

The paper CSS class is already defined in index.css with standard padding that's used consistently across Paper components in our application. By using this existing class, you can remove the redundant padding definition on line 43 (padding: theme.spacing(2),).

openaev-api/src/main/java/io/openaev/service/CatalogConnectorService.java
```diff
@@ -4,16 +4,46 @@
...
+  public Optional<CatalogConnector> findBySlug(String slug) {
+    return catalogConnectorRepository.findBySlugWithConfigurations(slug);
+  }
+
+  public Optional<CatalogConnector> findById(String id) {
+    return catalogConnectorRepository.findById(id);
+  }
+
+  @Transactional
+  public List<CatalogConnector> upsertAll(List<CatalogConnector> connectors) {
```
MarineLeM: It seems that this function is not used ? Should we delete it ?

openaev-front/src/utils/api-types.d.ts
```diff
@@ -590,6 +590,90 @@ export interface CVEBulkInsertInput {
...
+   * Connector configuration enum
+   * @uniqueItems true
+   */
+  connector_configuration_enum?: string[];
+  /** Connector configuration format */
+  connector_configuration_format?: string;
+  /** Connector ID */
+  connector_configuration_id?: string;
+  /** Connector configuration key */
+  connector_configuration_key?: string;
```
MarineLeM: connector_configuration_key should be mandatory as connector_configuration_type so we need to update the CatalogConnectorConfiguration.java class

openaev-front/src/utils/api-types.d.ts
```diff
@@ -590,6 +590,90 @@ export interface CVEBulkInsertInput {
...
+  /** Connector configuration required */
+  connector_configuration_required?: boolean;
+  /** Connector configuration type */
+  connector_configuration_type?: string;
```
MarineLeM: Btw the connector_configuration_type should be an enum

openaev-api/src/main/java/io/openaev/runner/CatalogConnectorIngestionRunner.java
```diff
@@ -0,0 +1,237 @@
...
+      InputStream dataStream = new ByteArrayInputStream(imageBytes);
+
+      String fileName = connectorSlug + "-logo.png";
+
+      fileService.uploadStream(FileService.CONNECTORS_LOGO_PATH, fileName, dataStream);
+
+      return fileName;
+    } catch (Exception e) {
+      log.error("Error upload image MinIO", e);
+      return "img/icon-connector-default.png";
```
camrrx: We don't have default image yet

openaev-api/src/main/java/io/openaev/rest/catalog_connector/CatalogConnectorApi.java
```diff
@@ -0,0 +1,62 @@
...
+import org.apache.commons.io.IOUtils;
+import org.springframework.http.CacheControl;
+import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.*;
+
+@RestController
+@RequiredArgsConstructor
+public class CatalogConnectorApi extends RestBehavior {
+  public static final String CATALOG_CONNECTOR_URI = "/api/connectors";
```
MarineLeM: should be /api/catalog-connector what do you think ?

openaev-front/src/admin/components/integrations/catalog_connectors/Index.tsx
```diff
@@ -0,0 +1,62 @@
+import { lazy } from 'react';
```
MarineLeM: Suggestion: Since this file contains only a single route, consider moving it directly into integrations.Index file to reduce complexity. We typically benefit from separate route files when we have multiple sub-routes.

MarineLeM: Or for better organization consider moving the catalog list routes into this files
so you will have in this file something like

```
<Routes>
  <Route path="" element={errorWrapper(Catalog)()} />
  <Route path=":connectorId" element={errorWrapper(ConnectorDetails)()} />
  <Route path="*" element={<NotFound />} />
</Routes>
```

and in the integrations.Index file
`<Route path="catalog/*" element={errorWrapper(IndexCatalog)()} />`

openaev-api/src/main/java/io/openaev/rest/catalog_connector/dto/CatalogConnectorOutput.java
```diff
@@ -0,0 +1,51 @@
...
+  @JsonProperty("catalog_connector_subscription_link")
+  private String subscriptionLink;
+
+  @JsonProperty("catalog_connector_source_code")
+  private String sourceCode;
+
+  @Enumerated(EnumType.STRING)
+  @JdbcTypeCode(SqlTypes.NAMED_ENUM)
+  @JsonProperty("catalog_connector_type")
+  private CatalogConnector.CONNECTOR_TYPE containerType;
```
MarineLeM, line 50: This field should be mandatory, isn't it ?

openaev-api/src/main/java/io/openaev/service/catalog_connectors/CatalogConnectorIngestionService.java
```diff
@@ -0,0 +1,307 @@
...
+    public CatalogConnectorIngestionService(
+            CatalogConnectorService catalogConnectorService,
+            FileService fileService,
+            ConnectorInstanceService connectorInstanceService,
+            ConnectorInstanceConfigurationRepository connectorInstanceConfigurationRepository) {
+        this.catalogConnectorService = catalogConnectorService;
+        this.fileService = fileService;
+        this.connectorInstanceService = connectorInstanceService;
+        this.connectorInstanceConfigurationRepository = connectorInstanceConfigurationRepository;
+    }
```
antoinemzs, line 37: Suggestion: to keep source code cleaner (or more "magical"), Lombok provides class annotations to hide constructors such as this one which only assigns instance fields.

`@RequiredArgsConstructor` on the class allows you to delete this constructor (Lombok will autogenerate it).

```suggestion
@Slf4j
@Service
@Transactional
@RequiredArgsConstructor
public class CatalogConnectorIngestionService {
    private static final Set<String> PROTECTED_KEYS = Set.of(
            "COLLECTOR_ID",
            "INJECTOR_ID",
            "EXECUTOR_ID"
    );
    private final CatalogConnectorService catalogConnectorService;
    private final FileService fileService;
    private final ConnectorInstanceService connectorInstanceService;
    private final ConnectorInstanceConfigurationRepository connectorInstanceConfigurationRepository;
```

openaev-api/src/main/java/io/openaev/service/catalog_connectors/CatalogConnectorIngestionService.java
```diff
@@ -0,0 +1,307 @@
...
+        if (contracts == null) {
+            throw new IllegalArgumentException("contracts is null");
+        }
+
+        List<CatalogConnector> catalogConnectorList = new ArrayList<>();
+
+        for (JsonNode contract : contracts) {
+            CatalogConnector catalogConnector = buildCatalogConnector(contract);
+            catalogConnectorList.add(catalogConnector);
+        }
```
antoinemzs: Note: no error handling here. A lot of things can go wrong in deserialisation situations, so are we prepared for an all-or-nothing ingestion (because the process is transactional) ?
I'm asking because functionally, what are we expecting to see if there is, for example, a single contract throwing an error?

openaev-api/src/main/java/io/openaev/service/catalog_connectors/CatalogConnectorIngestionService.java
```diff
@@ -0,0 +1,307 @@
...
+                String[] parts = base64Image.split(",");
+                if (parts.length == 2) {
+                    base64Data = parts[1];
+                }
+            }
+
+            byte[] imageBytes = Base64.getDecoder().decode(base64Data);
+            InputStream dataStream = new ByteArrayInputStream(imageBytes);
+
+            String fileName = connectorSlug + "-logo.png";
```
antoinemzs: it seems like this would have the potential for collisions, but not terribly dangerous apart if we accidentally upload the same path elsewhere in the codebase.

Suggestion: using the DB-generated ID for the persisted catalog connector object as one of the filename fragments.

This would have you remove the use of `.saveAll(List<CatalogConnector>)` in favour of a loop with `.save(CatalogConnector)` in `extractCatalog`; but that's OK because we are already within a transaction scope (the overhead of using a loop+save is to create a new transaction scope for each iteration)

```java
@Override
@Transactional
public <S extends T> List<S> saveAll(Iterable<S> entities) {

	Assert.notNull(entities, ENTITIES_MUST_NOT_BE_NULL);

	List<S> result = new ArrayList<>();

	for (S entity : entities) {
		result.add(save(entity));
	}

	return result;
}
```

openaev-api/src/main/java/io/openaev/service/catalog_connectors/CatalogConnectorIngestionService.java
```diff
@@ -0,0 +1,307 @@
...
+
+        for (JsonNode contract : contracts) {
+            CatalogConnector catalogConnector = buildCatalogConnector(contract);
+            catalogConnectorList.add(catalogConnector);
+        }
+
+        List<CatalogConnector> saved = catalogConnectorService.saveAll(catalogConnectorList);
+
+        for (CatalogConnector connector : saved) {
+            cleanupInstanceConfigurations(connector);
```
antoinemzs: good thinking

openaev-api/src/main/java/io/openaev/service/catalog_connectors/CatalogConnectorIngestionService.java
```diff
@@ -0,0 +1,307 @@
...
+                // - Protected system keys must NEVER be deleted
+                boolean keyRemovedFromSchema =
+                        schemaConf == null && !PROTECTED_KEYS.contains(instConf.getKey().toUpperCase());
+
+                // Configuration changed to "PASSWORD" format but the stored value is not encrypted:
+                // - schemaConf exists
+                // - format is PASSWORD
+                // - must be deleted for security
+                boolean mustDeleteBecausePassword =
+                        schemaConf != null && isFormastPasswordButNotEncrypted(schemaConf, instConf);
```
antoinemzs: It looks like we should make an effort to encrypt the currently plaintext value with the known XTM Composer public key instead of nuking the config line altogether. What are your thoughts?

openaev-api/src/test/java/io/openaev/service/CatalogConnectorIngestionServiceTest.java
```diff
@@ -0,0 +1,349 @@
...
+    CatalogConnectorIngestionService ingestion =
+        new CatalogConnectorIngestionService(
+            catalogConnectorService,
+            fileService,
+            connectorInstanceService,
+            connectorInstanceConfigurationRepository);
+
+    CatalogConnector connector = ingestion.buildCatalogConnector(contract);
+
+    assertThat(connector.getContainerType()).isNull();
```
antoinemzs, line 229: Is this correct? This looks like this could cause issues down the line. What are the impacts of this occurrence?

openaev-api/src/main/java/io/openaev/service/catalog_connectors/CatalogConnectorIngestionService.java
```diff
@@ -0,0 +1,307 @@
...
+            // required
+            boolean isRequired = false;
+            if (required != null && required.isArray()) {
+                for (JsonNode req : required) {
+                    if (req.asText().equals(key)) {
+                        isRequired = true;
+                        break;
+                    }
+                }
+            }
```
antoinemzs: should we be concerned that perhaps there is an entry in [required] that doesn't match any listed properties? Thoughts?

Note this would meant the contract is malformed. Consider adding some logging.

openaev-api/src/main/java/io/openaev/service/catalog_connectors/CatalogConnectorIngestionService.java
camrrx: To confirm with @jborozco !

openaev-api/src/main/java/io/openaev/service/catalog_connectors/CatalogConnectorIngestionService.java
camrrx: Good suggestion, as discuss we'll not do it now.

openaev-api/src/main/java/io/openaev/service/catalog_connectors/CatalogConnectorIngestionService.java
camrrx: Todo : another PR when the method to encrypt will be merged

openaev-api/src/test/java/io/openaev/service/CatalogConnectorIngestionServiceTest.java
camrrx: We keep it like this as discussed

Issue #4283, opened 2025-10-20
[SCV] Display and Filter Payloads by Security Domains


DeniseMonreale:
As a security analyst

I want to filter Security Domains

So that I can quickly identify Payloads/Inject/Contracts that contain specific domains and prioritize comprehensive tests that provide maximum coverage across my security infrastructure

PR #4414, opened 2025-11-25
[frontend/backend] feat(payloads): filter payloads/contracts by security domains (#4283)

johanah29:
### Testing Instructions

1. Create some payloads with various domains
2. Filter the list on domains 
In a simulation/scenario injects tab
3. Filter the list on domains
4. Try to add an inject, filter the payload list on domains

Available operators for the filter must be contains, not contains, empty and not empty.

### DIFF
--- a/openaev-front/src/utils/lang/en.json
+++ b/openaev-front/src/utils/lang/en.json
@@ -901,6 +901,7 @@
   "inject-singular": "inject",
   "InjectExpectation": "Inject expectations",
   "injector_contract_arch": "Architecture",
+  "injector_contract_domains": "Domains",
   "injector_contract_injector": "Injector",
   "injector_contract_kill_chain_phases": "Kill chain phase",
   "injector_contract_labels": "Label",
--- a/openaev-front/src/utils/lang/es.json
+++ b/openaev-front/src/utils/lang/es.json
@@ -901,6 +901,7 @@
   "inject-singular": "inyección",
   "InjectExpectation": "Inyectar expectativas",
   "injector_contract_arch": "Arquitectura",
+  "injector_contract_domains": "Dominios",
   "injector_contract_injector": "Inyector",
   "injector_contract_kill_chain_phases": "fase de la cadena de ataque",
   "injector_contract_labels": "Etiqueta",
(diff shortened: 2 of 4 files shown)


openaev-front/src/admin/components/common/injects/CreateInject.tsx
```diff
@@ -127,7 +127,7 @@ const CreateInject: FunctionComponent<Props> = ({
     },
     {
       field: 'injector_contract_domains',
-      label: t('payload_domains'),
+      label: 'Domains',
```
gabriel-peze: Is it normal to stop using a translation here ?

openaev-front/src/utils/lang/en.json
```diff
@@ -869,6 +869,7 @@
   "inject_asset_groups": "Asset groups",
   "inject_assets": "Assets",
   "inject_attack_patterns": "Attack pattern",
+  "inject_contract_domains": "Domains",
```
gabriel-peze, line 874: Is it possible to use only one trad "Domains" and delete "inject_contract_domains", "injector_contract_domains" and "payload_domains" ?

openaev-front/src/admin/components/common/injects/Injects.tsx
```diff
@@ -130,22 +130,14 @@ const Injects: FunctionComponent<Props> = ({
...
-                  = inject.inject_injector_contract?.injector_contract_domains;
-
-        const domains
-                  = (payloadDomains && payloadDomains.length > 0 && payloadDomains)
-                    || (contractDomains && contractDomains.length > 0 && contractDomains);
-
-        return domains
-          ? <ItemDomains domains={domains} variant="reduced-view" />
+      value: (inject: InjectOutputType, _: InjectorContractConverted['convertedContent']) => {
+        return inject.inject_contract_domains && inject.inject_contract_domains.length > 0
```
gabriel-peze: Could be replaced by `return inject.inject_contract_domains?.length`

openaev-front/src/admin/components/common/injects/CreateInject.tsx
johanah29: The type used for the headers automatically translates the label

openaev-front/src/utils/lang/en.json
johanah29: Was unified to Paload domains

gabriel-peze:
I found one issue with labels :
The filter work but there is a missing key for Domains on inject creation label
image

PR #4466, opened 2025-11-28
[backend] docs(queryable): add javadoc on queryable annotation (#1)

RomuDeuxfois:


openaev-model/src/main/java/io/openaev/annotation/Queryable.java
```diff
@@ -6,21 +6,51 @@
...
   boolean sortable() default false;
 
+  /** Human-readable label used for documentation or UI generation. */
   String label() default "";
 
+  /**
+   * Defines the JPA path used to query this property. Follows Spring Data Specification conventions
+   * (e.g. {@code "organization.id"}).
+   */
   String path() default "";
```
RomuDeuxfois: Can lead to UnsupportableOperation. Path should en by a primitiv value.

PR #4467, opened 2025-11-28
test

RomuDeuxfois:
### Proposed changes

* Display which type is required to match the regex.
* Show a warning without failing the job. Otherwise, it can be confusing when reviewing a PR: the PR can be merged, but at first glance it appears red and looks broken.


RomuDeuxfois:
@savacano28, @guillaumejparis and @Dimfacion happy to discuss the failure point.

Issue #4266, opened 2025-10-15
[SCV] Categorize existing payloads into domains


DeniseMonreale:
**As a** Security Analyst
I want all existing payloads to be categorized into predefined Security Domains
So that I can easily browse, filter, and analyze payloads by domain, ensuring better organization and visibility across attack surfaces.

DeniseMonreale:
Test 01

To FIX

+ On the Payload page, when I export to CSV, the domains are not included

+ On the Payload page, clicking the export button next to the Payload label does not trigger any action

PR #4481, opened 2025-12-02
[backend] feat(scv): add domain categorization (#4266)

gabriel-peze:
### Proposed changes

* Start project as usual, payloads and contracts should have a domain set

### Testing Instructions

1. Start the application (and injectors and/or collectors, but is optionnal)
2. Check for payloads and domains on the platform, domains should be se

### DIFF
--- a/openaev-api/src/main/java/io/openaev/rest/domain/enums/PresetDomain.java
+++ b/openaev-api/src/main/java/io/openaev/rest/domain/enums/PresetDomain.java
@@ -23,9 +23,9 @@ public class PresetDomain {
 
   private static final Map<Domain, List<String>> domainKeywordsMap =
       Map.of(
-          NETWORK, List.of("network", "ftp", "smb", "llmnr"),
+          NETWORK, List.of("network", "ftp", "smb", "llmnr", "nmap"),
           WEB_APP, List.of("web"),
-          EMAIL_INFILTRATION, List.of("email", "phishing"),
+          EMAIL_INFILTRATION, List.of("mail", "phishing"),
           DATA_EXFILTRATION, List.of("exfiltrat"),
           URL_FILTERING, List.of("bitsadmin"),
           CLOUD, List.of("aws", "azure", "gcp"));


openaev-api/src/main/java/io/openaev/rest/domain/enums/DefaultDomain.java
```diff
@@ -0,0 +1,24 @@
+package io.openaev.rest.domain.enums;
+
+import io.openaev.database.model.Domain;
+import java.time.Instant;
+import lombok.Getter;
+
+@Getter
+public enum DefaultDomain {
```
antoinemzs: Suggestion: name this `PresetDomain` instead; it's not a default domain per se before it's been selected as such by an algorithm

openaev-api/src/main/java/io/openaev/migration/V4_52__Implement_Domains_notion.java
```diff
@@ -78,16 +78,18 @@ REFERENCES domains(domain_id)
...
 
       stmt.execute(
           """
-insert into injectors_contracts_domains (injector_contract_id, domain_id)
-select ic.injector_contract_id, d.domain_id from injectors_contracts ic
-inner join domains d on d.domain_name = 'To classify'
-where ic.injector_contract_payload is null;""");
+                insert into injectors_contracts_domains (injector_contract_id, domain_id)
+                select ic.injector_contract_id, d.domain_id from injectors_contracts ic
+                inner join domains d on d.domain_name = 'To classify'
```
antoinemzs: TODO: try to use constants in this sort of situation; case sensitivity counts for example, or the cost of typos, etc...

openaev-api/src/main/java/io/openaev/rest/domain/enums/DefaultDomain.java
antoinemzs: It looks like it'd be easier to use this structure if it was a POJO with a number of `public static Domain` members.

instead of calling `DefaultDomain.ENDPOINT.getDomain()` it'd just be `DefaultDomain.ENDPOINT`.

Example:
```java
public class DefaultDomain {
  public static Domain ENDPOINT = new Domain(null, "Endpoint", "#389CFF", Instant.now(), null);
  ...
}
```
Callsite:
```java
if (findInKeywords(DomainKeyWords.ENDPOINT, name)
    || findInKeywords(DomainKeyWords.ENDPOINT, description)) {   
  domains.add(DefaultDomain.ENDPOINT);
}
```

openaev-api/src/main/java/io/openaev/rest/domain/DomainService.java
```diff
@@ -78,7 +94,64 @@ public Domain upsert(final String name, final String color) {
...
+      domains.add(DefaultDomain.DATA_EXFILTRATION.getDomain());
+    }
+    if (findInKeywords(DomainKeyWords.URL_FILTERING, name)
+        || findInKeywords(DomainKeyWords.URL_FILTERING, description)) {
+      domains.add(DefaultDomain.URL_FILTERING.getDomain());
+    }
+    if (findInKeywords(DomainKeyWords.CLOUD, name)
+        || findInKeywords(DomainKeyWords.CLOUD, description)) {
+      domains.add(DefaultDomain.CLOUD.getDomain());
+    }
```
antoinemzs: Looks like you could statically map wordlists with preset domains already, and not go through this long `if` stack.

pseudo code
```java
public class DefaultDomain {
  public List<Domain> getRelevantDomainsFromKeywords(String input) {
    return domainWordsMap.stream().filter(pair -> pair.matchWords(input)).map(pair -> pair.getDomain()).toList();
  }
}
```
Callsite:
```java
private final DefaultDomain defaultDomain;
List<Domain> relevantDomains = defaultDomain.getRelevantDomainsFromKeywords(name + description);
```

openaev-api/src/main/java/io/openaev/rest/injector_contract/InjectorContractService.java
```diff
@@ -216,12 +218,15 @@ public InjectorContract createNewInjectorContract(InjectorContractAddInput input
...
-
     setVulnerabilitiesFromExternalOrInternalIds(
         input.getVulnerabilityExternalIds(), input.getVulnerabilityIds(), injectorContract);
 
     injectorContract.setInjector(
         updateRelation(input.getInjectorId(), injectorContract.getInjector(), injectorRepository));
+    injectorContract.setDomains(
+        !injectorContract.getInjector().isPayloads()
+            ? this.domainService.upserts(input.getDomains())
+            : Set.of());
```
antoinemzs: TODO: immutable lists in Hibernate entities like this can cause issues during persistence as Hibernate sometimes tries to mutate a collection.

Prefer `new HashSet<>()`;

openaev-api/src/test/java/io/openaev/rest/inject/InjectApiTest.java
```diff
@@ -162,6 +163,8 @@ void beforeAll() {
     Agent agent = AgentFixture.createDefaultAgentService();
     agent.setAsset(endpointSaved);
     AGENT = agentRepository.save(agent);
+
+    DOMAINS = domainComposer.forDefaultToClassifyDomain().persist().getSet();
```
antoinemzs, line 168: praise: looks like a "forDefault...()" method on composers is really useful.

openaev-api/src/main/java/io/openaev/migration/V4_52__Implement_Domains_notion.java
antoinemzs: It's not too late for making this change as we haven't released the code yet.

openaev-api/src/test/java/io/openaev/rest/domain/service/DomainServiceTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+  @Test
+  @DisplayName("Set should not be merged, because existing is to classify")
+  void setShouldNotBeMergedBecauseExistingIsToClassify() {
+    Set<Domain> domainsA = Set.of(DefaultDomain.TOCLASSIFY.getDomain());
+    Set<Domain> domainsB = Set.of(DefaultDomain.ENDPOINT.getDomain());
+
+    Set<Domain> domains = this.domainService.mergeDomains(domainsA, domainsB);
+
+    assertFalse(domains.isEmpty());
```
antoinemzs: suggestion for later: look at `org.assertj.core.api.Assertions.assertThat` for richer assertions, e.g.
```java
assertThat(domains).containsExactly(DefaultDomain.ENDPOINT.getDomain());
// the following two are not needed since the above assertion takes care of them
// assertThat(domains.size()).isEqualTo(1);
// assertThat(domains).isEmpty();
```
Test failures are more explicit with this lib.

openaev-api/src/main/java/io/openaev/rest/domain/DomainService.java
```diff
@@ -78,7 +94,64 @@ public Domain upsert(final String name, final String color) {
...
+
+    if (domains.isEmpty()) {
+      domains.add(DefaultDomain.ENDPOINT.getDomain());
+    }
```
antoinemzs: Suggestion: the "default" domain should be exposed as a special member of the `DefaultDomain` class. We are mixing concerns here.

openaev-api/src/main/java/io/openaev/rest/domain/DomainService.java
```diff
@@ -78,7 +94,64 @@ public Domain upsert(final String name, final String color) {
...
+        || findInKeywords(DomainKeyWords.URL_FILTERING, description)) {
+      domains.add(DefaultDomain.URL_FILTERING.getDomain());
+    }
+    if (findInKeywords(DomainKeyWords.CLOUD, name)
+        || findInKeywords(DomainKeyWords.CLOUD, description)) {
+      domains.add(DefaultDomain.CLOUD.getDomain());
+    }
+
+    if (domains.isEmpty()) {
+      domains.add(DefaultDomain.ENDPOINT.getDomain());
```
antoinemzs: Inconsistency with the migration file: why is the migration assigning "to classify" as a default domain, but here we are using "Endpoint"? It sounds like we are considering two defaults.

openaev-api/src/main/java/io/openaev/rest/injector_contract/form/InjectorContractInput.java
```diff
@@ -37,4 +36,7 @@ public class InjectorContractInput {
 
   @JsonProperty("contract_platforms")
   private PLATFORM_TYPE[] platforms = new PLATFORM_TYPE[0];
+
+  @JsonProperty("contract_domains")
+  private Set<Domain> domains;
 }
```
antoinemzs, line 42: Breaking change: the API breaks if we don't pass a Set<Domain> (seralised in JSON); we can already see this because you needed to change a lot of setups in the `InjectorContractApiTest` class

TODO: make this non breaking and assign a default empty set (`new HashSet<>()`)

make appropriate changes in code to accomodate this.

antoinemzs (changes requested):
Good stuff overall; I'm mostly concerned by a breaking change in the API which appears to be easy to revert

openaev-api/src/main/java/io/openaev/rest/domain/DomainService.java
gabriel-peze: To classify is set by default at migration because domains are mandatory in database, and we don't want to set a false domain on existing payloads or contracts, so to classify is set to tell the user that this payload or domain is to classify.
Here, endpoint is the default domain for caldera payloads/contracts, here we have to set the domain endpoint, and calculate the others domains, so by default, all caldera payloads/contracts are Endpoints domains, and maybe more.

PR #4499, opened 2025-12-04
[frontend/backend] feat: deploy with xtmcomposer (#4312)

### LINKED ISSUE #4312: Deploy collector and injector from the catalog
## Context

Be able to pdeploy injector and collectors from the catalog

MarineLeM:
### Proposed changes
Add collectors/injectors/executors deploiement with xtmComposer

### Related issues

* Closes #4312 

### Linked PRs:
XtmComposer : https://github.com/FiligranHQ/xtm-composer/pull/42

### DIFF
--- a/openaev-api/src/main/java/io/openaev/service/connector_instances/ConnectorInstanceLogService.java
+++ b/openaev-api/src/main/java/io/openaev/service/connector_instances/ConnectorInstanceLogService.java
@@ -15,7 +15,7 @@ import org.springframework.stereotype.Service;
 @Service
 @RequiredArgsConstructor
 public class ConnectorInstanceLogService {
-  public static final long LOG_SIZE_LIMIT = 3L;
+  public static final long LOG_SIZE_LIMIT = 1000L;
   private final ConnectorInstanceLogRepository connectorInstanceLogRepository;
 
   private void cleanupExcessLogs(String connectorInstanceId) {


openaev-api/src/main/java/io/openaev/service/connector_instances/EncryptionFactory.java
```diff
@@ -0,0 +1,28 @@
...
+   *
+   * @param instance the connector instance
+   * @return the encryption strategy, or null if no encryption needed
+   */
+  public EncryptionService getEncryptionService(ConnectorInstance instance) {
+    if (instance.getCatalogConnector().isManagerSupported()) {
+      return xtmComposerEncryptionService;
+    }
+    log.warn("Built-in encryption not yet implemented for instance: {}", instance.getId());
+    return null; // TODO Later
```
camrrx: You can add the Antoine's issue number

openaev-api/src/main/java/io/openaev/api/xtm_composer/XtmComposerApi.java
```diff
@@ -0,0 +1,134 @@
...
+
+  @PutMapping(value = XTMCOMPOSER_URI + "/{xtmComposerId}/refresh-connectivity")
+  @Operation(
+      summary = "Refresh connectivity with XTM composer",
+      description = "Refresh last check connectivity in settings and version in XTM Composer")
+  @RBAC(actionPerformed = Action.WRITE, resourceType = ResourceType.CATALOG)
+  @ApiResponses(value = {@ApiResponse(responseCode = "200", description = "Successful refresh")})
+  @Transactional(rollbackFor = Exception.class)
+  public XtmComposerOutput refreshConnectivity(@PathVariable @NotBlank final String xtmComposerId) {
+    return xtmComposerService.refreshConnectivity(xtmComposerId, LocalDateTime.now());
```
Dimfacion: **nitpick :** Given that LocalDateTime does not have an explicit timezone while Instant does, I think it would be better to usually use Instant for technical dates (although in this use case, it's kind of the same, I guess)

openaev-api/src/main/java/io/openaev/service/connectors/XtmComposerService.java
```diff
@@ -0,0 +1,208 @@
...
+   */
+  public void validateXtmComposerId(String xtmComposerId) {
+    Map<String, Setting> xtmComposerInformation = this.getXtmComposerSettings();
+    if (!xtmComposerId.equals(xtmComposerInformation.get(XTM_COMPOSER_ID.key()).getValue())) {
+      throw new BadRequestException("Invalid xtm-composer identifier");
+    }
+  }
+
+  /** Validate XTM Composer reachability */
+  public void validateXtmComposerReachability() {
```
Dimfacion: **nitpick :** Maybe it would make sense to return a boolean instead of returning nothing and relying on the caller expecting an exception that is not in the signature of the method ?

MarineLeM: Actually the endpoint already return a boolean
image

The validateXtmComposerReachability function it's also called when creating an instance or updating the requested status of an instance

Dimfacion: > Actually the endpoint already return a boolean


That's what confuses me. Why don't you return a boolean in validateXtmComposerReachability instead of catching a generic exception in the caller function ? You could then just return the result of validateXtmComposerReachability in isXtmComposerReachable. 
Even if that bother you to return a boolean, we should at least have a signature that explicitly says that we throw an exception so that it's explicit what exception can happen.
```suggestion
  public void validateXtmComposerReachability() throws BadRequestException {
```

openaev-api/src/main/java/io/openaev/api/xtm_composer/dto/XtmComposerInstanceOutput.java
```diff
@@ -0,0 +1,69 @@
...
+    @Schema(description = "Configuration key")
+    @JsonProperty("configuration_key")
+    @NotBlank
+    private String key;
+
+    @Schema(description = "Configuration value")
+    @JsonProperty("configuration_value")
+    private String value;
+
+    @Schema(description = "Configuration value")
```
Dimfacion:
```suggestion
    @Schema(description = "Configuration is encrypted")
```

openaev-api/src/main/java/io/openaev/api/xtm_composer/dto/XtmComposerInstanceOutput.java
```diff
@@ -0,0 +1,69 @@
...
+
+  @Schema(description = "Connector Instance requested status")
+  @JsonProperty("connector_instance_requested_status")
+  @NotBlank
+  private ConnectorInstance.REQUESTED_STATUS_TYPE requestedStatus;
+
+  @Schema(description = "Connector Instance configuration")
+  @JsonProperty("connector_instance_configuration")
+  @NotBlank
+  private List<Configuration> configuration;
```
Dimfacion: **nitpick :**
```suggestion
  private List<Configuration> configurations;
```

openaev-api/src/main/java/io/openaev/migration/V4_55__Update_catalog_connector_configuration_type.java
```diff
@@ -12,26 +12,36 @@ public void migrate(Context context) throws Exception {
...
       select.execute(
           """
-                                  ALTER TABLE connector_instance_logs
-                                  RENAME COLUMN connector_configuration_created_at TO connector_instance_log_created_at;
-                            """);
+          ALTER TABLE connector_instance_logs
+          RENAME COLUMN connector_configuration_created_at TO connector_instance_log_created_at;
+        """);
+
+      select.execute(
```
Dimfacion: **question :** Why are we modifying an already deployed migration on prerelease ? Shouldn't we do a new one ?

MarineLeM: I don't understand we don't have this migration yet on prerelease

openaev-api/src/main/java/io/openaev/rest/collector/service/CollectorService.java
```diff
@@ -38,12 +98,39 @@ public Collector collector(String id) {
         .orElseThrow(() -> new ElementNotFoundException("Collector not found with id: " + id));
   }
 
+  public Iterable<Collector> collectors() {
```
Dimfacion, line 105: **nitpick :** I think we should have javadoc on public methods of our services

openaev-api/src/main/java/io/openaev/rest/injector/form/InjectorOutput.java
```diff
@@ -0,0 +1,40 @@
...
+  private String name;
+
+  @JsonProperty("injector_type")
+  @NotBlank
+  private String type;
+
+  @JsonProperty("injector_external")
+  private boolean external = false;
+
+  // TODO check need category, customContracts, executorCommands, executorClearCommands, payloads...
```
Dimfacion: **question :** Still TODO or is it done ?

openaev-api/src/main/java/io/openaev/service/InjectorService.java
```diff
@@ -90,15 +152,35 @@ public String getOriginInjectorType(@NotBlank final String injectorType) {
...
+   * Retrieve all injectors.
+   *
+   * @param isIncludeNext Include pending injectors.
+   * @return List of injector output
+   */
+  public Iterable<InjectorOutput> injectorsOutput(boolean isIncludeNext) {
+    return getConnectorsOutput(isIncludeNext);
+  }
+
+  public Optional<Injector> injectorByType(@NotBlank final String injectorType) {
```
Dimfacion, line 178: **nitpick :** Sorry, I'm gonna be annoying again with javadoc ;)

openaev-api/src/main/java/io/openaev/service/catalog_connectors/CatalogConnectorService.java
```diff
@@ -34,4 +58,23 @@ public Optional<CatalogConnector> findBySlug(String slug) {
   public Optional<CatalogConnector> findById(String id) {
     return catalogConnectorRepository.findById(id);
   }
+
+  public Set<CatalogConnectorConfiguration> getCatalogConnectorConfigurations(
+      String catalogConnectorId) {
+
+    Set<String> EXCLUDED_CONFIG_KEYS = Set.of("OPENAEV_TOKEN");
```
Dimfacion: **suggestion :** Maybe that could be a constant in this class ?

openaev-api/src/main/java/io/openaev/service/connector_instances/ConnectorInstanceLogService.java
```diff
@@ -0,0 +1,76 @@
...
+import org.springframework.stereotype.Service;
+
+@Slf4j
+@Service
+@RequiredArgsConstructor
+public class ConnectorInstanceLogService {
+  @Getter private final Integer LOG_SIZE_LIMIT = 10000;
+  private final ConnectorInstanceLogRepository connectorInstanceLogRepository;
+
+  public String transformRawLogsLineToLog(Set<String> rawLogLines) {
```
Dimfacion: **nitpick :** javadoc on public method of services

openaev-api/src/main/java/io/openaev/service/connector_instances/ConnectorInstanceService.java
```diff
@@ -0,0 +1,269 @@
...
+  }
+
+  private ConnectorInstanceConfiguration createConfigurationFromInput(
+      CreateConnectorInstanceInput.ConfigurationInput input,
+      CatalogConnectorConfiguration definition,
+      ConnectorInstance instance,
+      EncryptionService encryptionService) {
+
+    boolean isEncrypted =
+        CatalogConnectorConfiguration.CONNECTOR_CONFIGURATION_FORMAT.PASSWORD.equals(
```
Dimfacion, line 224: **thought :** Maybe that would be useful to have a list of encrypted configuration as maybe later, we won't have only the passwords that are encrypted ?

MarineLeM: Actually for now in the catalog all the sensitives datas are in password format, I propose to update this piece of code if the catalog evolve

openaev-api/src/main/java/io/openaev/service/connectors/ConnectorOrchestrationService.java
```diff
@@ -0,0 +1,230 @@
...
+    }
+  }
+
+  private void throwIfConnectorAlreadyExist(
+      String catalogConnectorSlug, CatalogConnector.CONNECTOR_TYPE catalogConnectorType)
+      throws IllegalArgumentException {
+    BaseConnectorEntity connector;
+    if (CatalogConnector.CONNECTOR_TYPE.COLLECTOR.equals(catalogConnectorType)) {
+      connector =
+          collectorService.findCollectorByType("openaev_" + catalogConnectorSlug).orElse(null);
```
Dimfacion: **suggestion :** Maybe have a constant called prefix for "openaev_"

openaev-api/src/main/java/io/openaev/service/connectors/XtmComposerService.java
```diff
@@ -0,0 +1,216 @@
...
+  /**
+   * Check if the last connectivity check is older than 1 day
+   *
+   * @param lastConnectivityCheckValue Last connectivity check value as string
+   * @return true if the last connectivity check is older than 1 day, false otherwise
+   */
+  public boolean isLastConnectivityCheckTooOld(String lastConnectivityCheckValue) {
+    try {
+      Instant lastCheck = Instant.parse(lastConnectivityCheckValue);
+      Instant oneDayAgo = Instant.now().minus(1, ChronoUnit.DAYS);
```
Dimfacion: **suggestion :** Maybe we should put this in properties (with minutes instead of days) ?

MarineLeM: Yes, one day was arbitrary. Should we set it to 2 hours instead? Or another duration?

openaev-api/src/main/java/io/openaev/rest/connector_instance/ConnectorInstanceApi.java
```diff
@@ -0,0 +1,136 @@
...
+  public ConnectorInstance updateRequestedStatus(
+      @PathVariable @NotBlank final String connectorInstanceId,
+      @Valid @RequestBody UpdateConnectorInstanceRequestedStatus input) {
+    return orchestrationService.updateRequestedStatus(
+        connectorInstanceId, input.getRequestedStatus());
+  }
+
+  @DeleteMapping(value = CONNECTOR_INSTANCE_URI + "/{connectorInstanceId}")
+  @Operation(summary = "Delete connector instance")
+  @RBAC(actionPerformed = Action.DELETE, resourceType = ResourceType.CATALOG)
```
impolitepanda, line 146: There doesn't seem to be any RBAC on any method allowing the creation of catalog (I couldn't find any @ RBAC(actionPerformed = Action.CREATE, resourceType = ResourceType.CATALOG) . Is that normal ?

impolitepanda: Also, IMO, this should probably not be a DELETE but a WRITE (it would mean this method would require modification permission on a catalog in order to remove a connector instance from it). Or are we considering that a connector instance is a catalog ?

openaev-api/src/main/java/io/openaev/rest/connector_instance/dto/UpdateConnectorInstanceRequestedStatus.java
```diff
@@ -0,0 +1,17 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.openaev.database.model.ConnectorInstance;
+import io.swagger.v3.oas.annotations.media.Schema;
+import jakarta.validation.constraints.NotNull;
+import lombok.Getter;
+
+@Getter
+public class UpdateConnectorInstanceRequestedStatus {
+  @NotNull(message = MANDATORY_MESSAGE)
```
impolitepanda, line 13: Not sure if this notnull is necessary, the class having only getter methods.

openaev-api/src/main/java/io/openaev/rest/executor/ExecutorApi.java
```diff
@@ -59,6 +64,7 @@ public class ExecutorApi extends RestBehavior {
   private TokenRepository tokenRepository;
 
   @Resource protected ObjectMapper mapper;
+  @Autowired private ExecutorService executorService;
```
impolitepanda: Could we use the fact that we are modifying this class to remove all @ Autowired fields and methods and replace them by a constuctor (@ RequiredArgsConstructor) + private final fields to fit the newer coding style ?

openaev-api/src/main/java/io/openaev/service/catalog_connectors/CatalogConnectorIngestionService.java
```diff
@@ -192,6 +192,9 @@ public Set<CatalogConnectorConfiguration> buildConnectorConfigurations(
 
     for (Iterator<String> it = properties.fieldNames(); it.hasNext(); ) {
       String key = it.next();
+      if ("OPENAEV_URL".equals(key)) {
```
impolitepanda: There is no constant for this string?

openaev-api/src/main/java/io/openaev/service/connector_instances/ConnectorInstanceLogService.java
```diff
@@ -0,0 +1,82 @@
...
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Service;
+
+@Slf4j
+@Service
+@RequiredArgsConstructor
+public class ConnectorInstanceLogService {
+  @Getter private final Integer LOG_SIZE_LIMIT = 10000;
```
impolitepanda: No need for a getter, set it to public instead, as it's a constant.

openaev-api/src/main/java/io/openaev/service/connector_instances/ConnectorInstanceLogService.java
```diff
@@ -0,0 +1,82 @@
...
+  private final ConnectorInstanceLogRepository connectorInstanceLogRepository;
+
+  private void cleanupExcessLogs(String connectorInstanceId) {
+    long currentCount =
+        connectorInstanceLogRepository.countByConnectorInstanceId(connectorInstanceId);
+
+    if (currentCount > LOG_SIZE_LIMIT) {
+      int excessCount = (int) (currentCount - LOG_SIZE_LIMIT);
```
impolitepanda: I think using long instead of int would be more accurate, as it's the superior type in this calculus

openaev-api/src/main/java/io/openaev/service/connector_instances/ConnectorInstanceLogService.java
```diff
@@ -0,0 +1,82 @@
...
+   * @param connectorInstance the connector instance to log for
+   * @param rawLog the log content to store
+   * @return the saved log entry
+   * @throws IllegalArgumentException if rawLog is empty
+   */
+  @Transactional
+  public ConnectorInstanceLog pushLogByConnectorInstance(
+      ConnectorInstance connectorInstance, String rawLog) throws IllegalArgumentException {
+    if (rawLog.isEmpty()) {
+      throw new IllegalArgumentException("Instance ID and log cannot be empty");
```
impolitepanda: Should we really fail with an exception here ? Maybe just doing nothing would be better ?

openaev-api/src/main/java/io/openaev/service/connector_instances/ConnectorInstanceLogService.java
```diff
@@ -0,0 +1,82 @@
...
+    return saved;
+  }
+
+  /**
+   * Retrieves all logs for a specific connector instance.
+   *
+   * @param connectorInstanceId the connector instance identifier
+   * @return list of logs for the connector instance, empty if none found
+   */
+  public List<ConnectorInstanceLog> findLogsByConnectorInstanceId(String connectorInstanceId) {
```
impolitepanda, line 78: This can return quite a large list (LOG_SIZE_LIMIT max elements, and a log can be up to 65kB long). Could we put a pagination system in place ? (maybe with a subsequent tasks if it's too big to include now)

openaev-api/src/main/java/io/openaev/service/connector_instances/ConnectorInstanceService.java
```diff
@@ -0,0 +1,377 @@
...
+            definition.getConnectorConfigurationFormat());
+    JsonNode value = input.getValue();
+
+    if (isEncrypted) {
```
impolitepanda: I believe this encryption is done too late. It should be done almost immediately when receiving the data to make sure it never transit unencrypted within our stack.

I haven't seen it yet so it may not be needed, but it should be the same for decryption, it should be done at the last possible moment before returning data to the client

This prevents accidental logging of unencrypted data and reduce the risk of leakage

openaev-api/src/main/java/io/openaev/service/connector_instances/ConnectorInstanceService.java
```diff
@@ -0,0 +1,377 @@
...
+        tokenRepository.findAll(fromUser(currentUser().getId())).stream()
+            .findFirst()
+            .orElseThrow(() -> new IllegalArgumentException("No token found for current user"));
+    return createConfiguration(
+        "OPENAEV_TOKEN", objectMapper.getNodeFactory().textNode(token.getValue()), false, instance);
+  }
+
+  private ConnectorInstanceConfiguration createContainerIdConfiguration(
+      ConnectorInstance instance, CatalogConnector.CONNECTOR_TYPE type) {
+    return createConfiguration(
```
impolitepanda, line 306: Building a string key like this is kind of dangerous for future developers. Could you instead add a constant in each CatalogConnector type ?

impolitepanda: Also, it it's linked to the hardcoded values from getConfigurationKey() functions, then it might be a good idea to use the CatalogConnector type there as well

openaev-api/src/main/java/io/openaev/service/connectors/ConnectorOrchestrationService.java
```diff
@@ -0,0 +1,233 @@
...
+    throwIfEnterpriseLicenseNotActive();
+
+    ConnectorInstance instance =
+        connectorInstanceService.connectorInstanceById(connectorInstanceId);
+    throwIfXtmComposerDownAndNeeded(instance.getCatalogConnector());
+
+    return connectorInstanceService.updateRequestedStatus(instance, requestedStatus);
+  }
+
+  private void throwIfConnectorInstanceAlreadyExist(String catalogId) throws BadRequestException {
```
impolitepanda: Should we maybe throw a DataIntegrityViolationException instead of a bad request ? It would translate to an HTTP response code for CONFLICT, which makes a bit more sense, maybe ?

openaev-api/src/main/java/io/openaev/service/connectors/ConnectorOrchestrationService.java
```diff
@@ -0,0 +1,233 @@
...
+        connectorInstanceService.findAllByCatalogConnectorId(catalogId);
+    if (!existingInstances.isEmpty()) {
+      throw new BadRequestException(
+          "ConnectorInstance with CatalogConnector id " + catalogId + " already exists");
+    }
+  }
+
+  private void throwIfConnectorAlreadyExist(
+      String catalogConnectorSlug, CatalogConnector.CONNECTOR_TYPE catalogConnectorType)
+      throws BadRequestException {
```
impolitepanda: same comment as above

openaev-api/src/main/java/io/openaev/service/connectors/ConnectorOrchestrationService.java
```diff
@@ -0,0 +1,233 @@
...
+      throw new BadRequestException(
+          "Connector with slug " + catalogConnectorSlug + " already exists");
+    }
+  }
+
+  private void throwIfInstanceOrConnectorAlreadyExist(
+      String catalogConnectorId,
+      String catalogConnectorSlug,
+      CatalogConnector.CONNECTOR_TYPE catalogConnectorType)
+      throws BadRequestException {
```
impolitepanda: Same comment as above

openaev-api/src/main/java/io/openaev/service/connector_instances/ConnectorInstanceService.java
MarineLeM: Just for information, the decryption part for this PR is done in XtmComposer

openaev-api/src/main/java/io/openaev/service/connector_instances/ConnectorInstanceLogService.java
MarineLeM: You completely right i will create an issue to improve this part

MarineLeM: https://github.com/OpenAEV-Platform/openaev/issues/4595

impolitepanda (approved):
Tests ok!

PR #4527, opened 2025-12-09
[backend] fix(securitycoverage): lock Stix updates and skip obsolete/retry bundles (#4521)

### LINKED ISSUE #4521: Lockdown postgres related to Update security coverage and retry
## Description

Currently, we are experiencing multiple system lock-downs related to updates/ retry error send back for oaev of the security coverage. The issue originates from an instruction set on the Octi side that was recently modified in bulk, causing it to trigger updates every few minutes and also several retries. Each update/retry generates a call to the OAEV STIX API, which in turn initiates the security-coverage update process.

This update process is computationally heavy in OAEV, as it requires creating and deleting scenarios, injects, and asset groups. When these operations are triggered repeatedly, the accumulated workload results in lock-downs on the OAEV database (DDB).

Issue related OCTI side: https://github.com/OpenCTI-Platform/opencti/issues/13633

Image

Image

## Environment

1. Openaev Demo interne 2.0.5

savacano28:
### Proposed changes

* Add lock stixApi @antoinemzs 
* Add a condition to prevent re-saving a security coverage that has already been saved (compute md5 on stix json)
* Add logs for security coverage update
* Clean alerts sonar

Closes #4521 

### Checklist

### Further comments

### DIFF
--- a/openaev-api/src/main/java/io/openaev/service/stix/StixService.java
+++ b/openaev-api/src/main/java/io/openaev/service/stix/StixService.java
@@ -36,10 +36,11 @@ public class StixService {
           securityCoverageService.processAndBuildStixToSecurityCoverage(stixJson);
 
       // Update Scenario using the last SecurityCoverage
-      Scenario  scenario = securityCoverageService.buildScenarioFromSecurityCoverage(securityCoverage);
+      Scenario scenario =
+          securityCoverageService.buildScenarioFromSecurityCoverage(securityCoverage);
       return scenario;
     } catch (BadRequestException | ParsingException e) {
-      log.error("Error while processing STIX bundle: {}", e.getMessage());
+      log.error(String.format("Error while processing STIX bundle: %s", e.getMessage()), e);
       throw e;
     }
   }


openaev-api/src/main/java/io/openaev/api/stix_process/StixApi.java
```diff
@@ -72,12 +73,11 @@ public ResponseEntity<?> processBundle(@RequestBody String ctiEvent) {
...
       BundleImportReport importReport = new BundleImportReport(scenario.getId(), summary);
       return ResponseEntity.ok(importReport);
-    } catch (ParsingException | IOException e) {
+    } catch (BadRequestException | ParsingException | IOException e) {
       log.error(
           String.format(
               "Parsing error while processing STIX bundle (workId=%s). ctiEvent=%s. Error: %s",
-              workId, ctiEvent, e.getMessage()),
-          e);
+              workId, ctiEvent, e.getMessage()));
```
antoinemzs: Why removing the `Throwable` argument from the log? It's generally useful to pass it to the log function so that the trace gets printed out.
See https://slf4j.org/apidocs/org/slf4j/Logger.html#error(java.lang.String,java.lang.Throwable)

openaev-api/src/main/java/io/openaev/utils/SecurityCoverageUtils.java
```diff
@@ -83,4 +86,14 @@ public static Set<String> getExternalIds(Set<StixRefToExternalRef> objectRefs) {
...
+
+  public static String computeMd5(String input) {
+    try {
+      MessageDigest md = MessageDigest.getInstance("MD5");
+      byte[] digest = md.digest(input.getBytes(StandardCharsets.UTF_8));
+      return DatatypeConverter.printHexBinary(digest).toLowerCase();
+    } catch (Exception e) {
+      throw new RuntimeException("Failed to compute MD5", e);
+    }
+  }
```
antoinemzs: Is the choice of MD5 based on speed requirements (which would be applicable to rejecting duplicate messages)?
However note that for any sensitive application, MD5 should not be used.

openaev-api/src/main/java/io/openaev/migration/V4_52__Add_coverage_content_hash_column.java
```diff
@@ -0,0 +1,40 @@
...
+      statement.execute(
+          """
+                ALTER TABLE security_coverages
+                ADD COLUMN security_coverage_content_hash VARCHAR(64);
+                """);
+
+      statement.execute(
+          """
+                UPDATE security_coverages
+                SET security_coverage_content_hash = md5(security_coverage_content::text)
```
antoinemzs: small nitpick, but make sure the hex digest is lower case like in the utility function

openaev-api/src/main/java/io/openaev/service/stix/SecurityCoverageService.java
```diff
@@ -84,23 +85,35 @@ public class SecurityCoverageService {
    * SecurityCoverage} domain object, and saves it. It also extracts referenced attack patterns and
    * sets optional fields like description and scheduling.
    *
-   * @param stixJson STIX-formatted JSON string representing a security coverage
+   * @param stixCoverageObj STIX-formatted JSON string representing a security coverage
```
antoinemzs: Incorrect, it is now a parsed ObjectBase from the STIX parsing subsystem

openaev-api/src/main/java/io/openaev/migration/V4_52__Add_coverage_content_hash_column.java
```diff
@@ -0,0 +1,40 @@
...
+@Component
+public class V4_52__Add_coverage_content_hash_column extends BaseJavaMigration {
+
+  @Override
+  public void migrate(Context context) throws Exception {
+    try (Statement statement = context.getConnection().createStatement()) {
+      statement.execute(
+          """
+                ALTER TABLE security_coverages
+                ADD COLUMN security_coverage_content_hash VARCHAR(64);
```
antoinemzs: Note that we store the hash of the entire submitted bundle, not just the security coverage, so this should be reflected in the column name (and related class attributes)

openaev-api/src/main/java/io/openaev/service/stix/StixService.java
```diff
@@ -24,12 +36,26 @@ public class StixService {
...
+
+      JsonNode root = objectMapper.readTree(stixJson);
+      String stixJsonHash = computeMd5(stixJson);
+      Bundle bundle = stixParser.parseBundle(root.toString());
+      ObjectBase stixCoverageObj = extractAndValidateCoverage(bundle);
+      String externalId = stixCoverageObj.getRequiredProperty(CommonProperties.ID.toString());
+
+      SecurityCoverage securityCoverage =
+          securityCoverageService.buildSecurityCoverageFromStix(
+              stixCoverageObj, bundle, externalId, stixJsonHash);
```
antoinemzs: IMO all this should be done inside the existing method `SecurityCoverageService.buildSecurityCoverageFromStix(String)`, because right now it makes the signature complex and difficult to understand; and we must fetch the existing security coverage object from the database anyway to do the hash check, so we don't benefit from doing it oustide of this function.

openaev-api/src/main/java/io/openaev/service/stix/SecurityCoverageService.java
```diff
@@ -84,23 +85,35 @@ public class SecurityCoverageService {
...
+   * @param externalId external reference
    * @return the saved {@link SecurityCoverage} object
    * @throws IOException if the input cannot be parsed into JSON
    * @throws ParsingException if the STIX bundle is malformed
    */
-  public SecurityCoverage buildSecurityCoverageFromStix(String stixJson)
-      throws IOException, ParsingException {
+  @Lock(type = LockResourceType.SECURITY_COVERAGE, key = "#externalId")
+  public SecurityCoverage buildSecurityCoverageFromStix(
+      ObjectBase stixCoverageObj, Bundle bundle, String externalId, String stixJsonHash)
```
antoinemzs, line 120: Suggestion: consider moving the code inside a private method and put the lock on that one. Leave the public method lock free and doing the preprocessing.

We should be mindful to keep the public interfaces clean and intelligible.

openaev-model/src/main/java/io/openaev/stix/objects/constants/ExtendedProperties.java
```diff
@@ -6,7 +6,8 @@ public enum ExtendedProperties {
   COVERED("covered"),
   COVERAGE("coverage"),
   MITRE_EXTENSION_DEFINITION("extension-definition--322b8f77-262a-4cb8-a915-1e441e00329b"),
-  OPENCTI_EXTENSION_DEFINITION("extension-definition--ea279b3e-5c71-4632-ac08-831c66a786ba");
+  OPENCTI_EXTENSION_DEFINITION("extension-definition--ea279b3e-5c71-4632-ac08-831c66a786ba"),
+  OPENCTI_MODIFIED_AT("x_opencti_modifed_at");
```
antoinemzs: Does not look used anywhere

openaev-model/src/main/java/io/openaev/database/model/SecurityCoverage.java
```diff
@@ -78,6 +78,10 @@ public class SecurityCoverage implements Base {
   @JsonProperty("security_coverage_content")
   private String content;
 
+  @Column(name = "security_coverage_content_hash", nullable = false, unique = true, length = 64)
```
antoinemzs: Re: length, IMO this might not be too useful to put this constraint down, as it is only relevant to MD5 (other hash algos will have different lengths).
If you wish to keep it, please comment that this constraint is chosen because the only hashing algo in place right now for this functionality is MD5.

openaev-api/src/main/java/io/openaev/utils/SecurityCoverageUtils.java
antoinemzs: Did you check that there wasn't already an implementation for producing an MD5 hex digest? https://github.com/OpenAEV-Platform/openaev/blob/bc5e0c4a2cbc84c9988185a041e71c608999e660/openaev-model/src/main/java/io/openaev/helper/CryptoHelper.java

openaev-api/src/main/java/io/openaev/service/stix/StixService.java
```diff
@@ -24,12 +29,19 @@ public class StixService {
    */
   @Transactional(rollbackFor = Exception.class)
   public Scenario processBundle(String stixJson) throws IOException, ParsingException {
-    // Update securityCoverage with the last bundle
-    SecurityCoverage securityCoverage =
-        securityCoverageService.buildSecurityCoverageFromStix(stixJson);
-    // Update Scenario using the last SecurityCoverage
-    Scenario scenario = securityCoverageService.buildScenarioFromSecurityCoverage(securityCoverage);
-    return scenario;
+    Scenario scenario = new Scenario();
```
antoinemzs: maybe not useful to preintantiate the scenario as this instance gets either overwritten or discarded

openaev-api/src/main/java/io/openaev/service/stix/StixService.java
```diff
@@ -24,12 +29,19 @@ public class StixService {
...
+    try {
+      // Update securityCoverage with the last bundle
+      SecurityCoverage securityCoverage =
+          securityCoverageService.processAndBuildStixToSecurityCoverage(stixJson);
+
+      // Update Scenario using the last SecurityCoverage
+      scenario = securityCoverageService.buildScenarioFromSecurityCoverage(securityCoverage);
+      return scenario;
+    } catch (BadRequestException | ParsingException e) {
+      log.error("Error while processing STIX bundle: {}", e.getMessage());
```
antoinemzs: prefer using the version of the logger method that takes a Throwable as a last argument to get full context in the case of an exception

PR #4528, opened 2025-12-09
[backend/frontend] fix(scenario): fix direct start when a scenario is scheduled (#4329)

### LINKED ISSUE #4329: Handle direct start for scenario + schedule creation
## Use case

Currently, when a scenario is created and a schedule is added, the system either does not execute immediately or waits for the next scheduled run.

We need to implement a way to manage a direct start, allowing the scenario to begin execution immediately after creation, without waiting for the schedule trigger.

Behavior expected:
* When creating a new scenario with a schedule, there should be an option (or default behavior, UI?) to start the first run right away.
* Subsequent runs should follow the defined schedule as normal.

## Current Workaround

Set Start date as the same day.

## Proposed Solution

* Add configuration or flag for immediate start on creation.
* Update scheduling logic to support this mode.

## Additional Information

Check Scenarios created from Octi and AI assistant.

Yes

damgouj:
### Proposed changes

* Fix direct start for a simulation when a scenario is scheduled from UI and from OCTI

### Testing Instructions

1. Create a report in OCTI and then create a security coverage for OpenAEV
2. Check if the scenario is scheduled and simulation started in OpenAEV
3. Create a scheduled scenario in OpenAEV
4. Check if the simulation is started

### Related issues

* Closes #4329 

### Checklist

### DIFF
--- a/openaev-api/src/main/java/io/openaev/service/stix/StixService.java
+++ b/openaev-api/src/main/java/io/openaev/service/stix/StixService.java
@@ -2,8 +2,6 @@ package io.openaev.service.stix;
 
 import io.openaev.database.model.Scenario;
 import io.openaev.database.model.SecurityCoverage;
-import io.openaev.service.ScenarioService;
-import io.openaev.service.cron.CronService;
 import io.openaev.stix.parsing.ParsingException;
 import java.io.IOException;
 import lombok.RequiredArgsConstructor;
@@ -17,8 +15,6 @@ import org.springframework.transaction.annotation.Transactional;
 public class StixService {
 
   private final SecurityCoverageService securityCoverageService;
-  private final CronService cronService;
-  private final ScenarioService scenarioService;
 
   /**
    * Generate or update a Scenario from Stix bundle


openaev-api/src/main/java/io/openaev/api/stix_process/StixApi.java
```diff
@@ -64,9 +65,9 @@ public ResponseEntity<?> processBundle(@RequestBody String ctiEvent) {
       openCTIService.acknowledgeReceivedOfCoverage(
           workId, "OpenAEV ready to process the operation");
       // Create scenario from stix bundle
-      Scenario scenario = stixService.processBundle(stixJson);
-      // TODO Schedule or not, start directly on execution after create/update
+      ScenarioSecurityCoverage scenarioSecurityCoverage = stixService.processBundle(stixJson);
       // If no simulation for this scenario is in progress, start an execution right away
+      Scenario scenario = stixService.setRecurrence(scenarioSecurityCoverage);
```
savacano28: Question: I think you could set the recurrence in processBundle -> buildScenarioFromSecurityCoverage?

antoinemzs (approved):

PR #4568, opened 2025-12-14
[backend] fix(chaining): bulk delete injects with dependencies (#3944)

### LINKED ISSUE #3944: When I try delete some chaining injects in scenario/simulation i have: ObjectOptimisticLockingFailureException: Row was updated or deleted by another transaction
## Description

When I try delete some chaining injects I have error:

  **"message": "Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception [Request processing failed: org.springframework.orm.ObjectOptimisticLockingFailureException: Row was updated or deleted by another transaction (or unsaved-value mapping was incorrect) :** 

## Environment

1. OpenAEV 1.18.18

## Reproducible Steps

1. Create an scenario/simulation
2. Add some injects
3. Create some dependencies between injects (chining injects)
4. Select the chaining injects
5. Delete bulk

## Workaround
Remove the chaining conditional before

## Expected Output

No error (delete all selected injects)
 
## Actual Output

Error

## Screenshots (optional)

Image

Image

corinnekrych:
### Proposed changes

* Delete dependencies between injects child/parent
* Remove cascade delete
* Delete injects

### Testing Instructions

1. Step-by-step how to test

### Related issues

* Closes [#3944](https://github.com/OpenAEV-Platform/openaev/issues/3944)

### Checklist

### Further comments


task.md
```diff
@@ -0,0 +1,14 @@
+Avec le fichier openaev-de/docker-compose fait un reverse engineering
```
savacano28: Nitpick : I think these files were useful for context, but we probably shouldn’t commit it.

guillaumejparis:
> > The failing check is linked to [.github/workflows/pr-title-check-worker.yml](https://github.com/OpenAEV-Platform/openaev/blob/master/.github/workflows/pr-title-check-worker.yml?rgh-link-date=2025-12-15T07%3A27%3A09Z#L20-L21) as I am contributing from a fork repo, I do not have access to those secrets.
> 
> Hi @corinnekrych! thanks for the report :). We will address this shortly @guillaumejparis.

Yep we need to running this action in a "reduced way" on forks

openaev-api/src/main/java/io/openaev/rest/inject/service/InjectService.java
```diff
@@ -265,6 +266,22 @@ public void deleteAllByIds(List<String> injectIds) {
   @Transactional(rollbackOn = Exception.class)
   public void deleteAll(List<Inject> injects) {
     if (!CollectionUtils.isEmpty(injects)) {
+      List<String> injectIds = injects.stream().map(Inject::getId).toList();
```
savacano28, line 269: Todo: Maybe we should check whether these IDs are already handled by the controller (or other methods using this).

corinnekrych:
@savacano28 Oops sorry i've cleaned up the commits removing the one that brought files that should not have been committed. As a result, i did a force push to clean my branch!

openaev-api/src/main/java/io/openaev/rest/inject/service/InjectService.java
corinnekrych: Not sure what you mean here. Do you want to move part of this logic a layer up in the controller.
Here is a suggestion: I’ve just noticed there is also a InjectService.deleteAllByIds method (just above), do you suggest:
* We create a InjectService.deleteInjectDependencies that take a list of injectIds and delete dependencies
* From InjectApi.bulkDelete we call InjectService.deleteInjectDependencies + InjectService.deleteAllByIds

savacano28: Sorry, what I mean is that in the current implementation, the controller receives a list of IDs and then fetches the Injects, but in the service we extract the IDs again. This means the controller is fetching Injects unnecessarily.
I think it would be cleaner to move this logic entirely into the service. For example, we could have a method like deleteAllInjects in the service, which internally calls (as you propose :)) :
- deleteInjectDependencies(injectIds) to remove related dependencies
- deleteAllByIds(injectIds) to delete the Injects

We might also need to check the method in SecurityCoverageInjectService that uses deleteAll.

Thank you for the pr :)

guillaumejparis:
> > > The failing check is linked to [.github/workflows/pr-title-check-worker.yml](https://github.com/OpenAEV-Platform/openaev/blob/master/.github/workflows/pr-title-check-worker.yml?rgh-link-date=2025-12-15T12%3A53%3A20Z#L20-L21) as I am contributing from a fork repo, I do not have access to those secrets.
> > 
> > 
> > Hi @corinnekrych! thanks for the report :). We will address this shortly @guillaumejparis.
> 
> Yep we need to running this action in a "reduced way" on forks

Ok it's fixed, normally error will disappear after a rebase 👍

RomuDeuxfois:
I close it for now, don't hesitate to reopen it if needed

PR #4581, opened 2025-12-16
[backend] fix(inject): inject expectations with existing results are overwritten when expiration_time is set (#4532)

### LINKED ISSUE #4532: Inject expectations with existing results are overwritten when expiration_time is set
## Description

When multiple inject expectations exist for the same asset (some already detected / prevented, others undetected / unprevented), valid detected/prevented expectations are incorrectly reverted to undetected/unprevented after 45 minutes (collector delta).

Example:

Inject A is detected/prevented → marked detected/prevented 

Inject B is never detected → stays failed (expected)

But when the expectation reaches expiration:

Inject A, which was already detected/prevented, is overwritten and becomes failed again.

This happens only when at least one inject remains undetected.

Image

Image

## Environment

1. OpenAEV 2.0.5

## Reproducible Steps

1. Create a simulation with 2 injects: one inject is correctly marked as prevented/detected, while the other is not (because Defender never actually caught it — which is expected).
2. Wait for expectationExpirationManager and check results of expectations

## Expected Output

Expected behavior: previously set results (green) should not be overwritten at expiration.

camrrx:
### Proposed changes

* Prevent inject expectations with existing results from being reprocessed when expiration_time is set.
* Centralize logic to apply both not expired and no result checks

### Testing Instructions

1. Create a simulation with 2 injects: one inject is correctly marked as prevented/detected, while the other is not (because Defender never actually caught it — which is expected).
2. Verify expectations with existing results are not updated

### Related issues

* Closes #4532 
*

### Checklist

### Further comments

### DIFF
--- a/openaev-api/src/main/java/io/openaev/rest/expectation/ExpectationApi.java
+++ b/openaev-api/src/main/java/io/openaev/rest/expectation/ExpectationApi.java
@@ -54,6 +54,15 @@ public class ExpectationApi extends RestBehavior {
   @RBAC(actionPerformed = Action.READ, resourceType = ResourceType.SIMULATION)
   public List<InjectExpectation> getInjectExpectationsNotFilledAndNotExpired(
       @RequestParam(required = false, name = "expiration_time") final Integer expirationTime) {
+    if (expirationTime == null) {
+      return Stream.of(
+              injectExpectationService.manualExpectationsNotFill(),
+              injectExpectationService.preventionExpectationsNotFill(),
+              injectExpectationService.detectionExpectationsNotFill())
+          .flatMap(List::stream)
+          .toList();
+    }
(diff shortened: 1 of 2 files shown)


openaev-api/src/main/java/io/openaev/rest/expectation/ExpectationApi.java
```diff
@@ -52,7 +52,7 @@ public InjectExpectation deleteInjectExpectationResult(
           "Retrieves inject expectations of agents installed on an asset. If an expiration time is provided, it will return all expectations not expired within this timeframe independently of their results. Otherwise, it will return all expectations without any result.")
```
guillaumejparis, line 52: We may need to change this comment ?

PR #4585, opened 2025-12-16
[backend] feat: adding spring actuator to collect metrics (#4438)

### LINKED ISSUE #4438: [Performance tests] Add Latency/Cpu/Memory metrics to OpenAev
Add metrics to OpenAev
- add spring metrics to OpenAev, 

Depends on https://github.com/FiligranHQ/platform/issues/1672 will require some discussion with platform before starting

heditar:
### Proposed changes

* Adding spring actuator to OpenAev, disabled by default but will be used in the test instances to collect metrics during performance tests

### Testing Instructions

1. Uncomment the configuration and call localhost:8080/actuator 

### Related issues

* Closes #4438 

### Checklist

- [NA] I wrote test cases for the relevant uses case
- [NA] Where necessary I refactored code to improve the overall quality


heditar:
> > **suggestion :** I think we should add commented code in AppSecurityConfig.java that allows calls to /actuator/ so that it's easier for people to understand what they need to activate if they want the metrics working. Wdyt ?
> 
> That is working for me as it is without any modification to AppSecurityConfig

I actually added the code in AppSecurityConfig to only allow metrics api to authenticated users

PR #4631, opened 2025-12-19
[frontend] chore: improve frontend bundle (#4641)

guillaumejparis:
Using vite to make prod bundle : smaller js chunks 

For the login screen 
BEFORE : 
image
AFTER:
image

For a simple dashboard
BEFORE:
image
AFTER:
image

There is still room for improvment : we can replace deprecated lib like moment, do more lazy loading, ..

### DIFF
--- a/openaev-front/package.json
+++ b/openaev-front/package.json
@@ -116,7 +116,7 @@
     "swagger-typescript-api": "13.2.16",
     "typescript": "5.9.3",
     "typescript-eslint": "8.48.0",
-    "vite": "npm:rolldown-vite@7.2.4",
+    "vite": "npm:rolldown-vite@7.3.1",
     "vitest": "3.2.4"
   },
   "resolutions": {
--- a/openaev-front/yarn.lock
+++ b/openaev-front/yarn.lock
@@ -2746,14 +2746,14 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@napi-rs/wasm-runtime@npm:^1.0.7":
-  version: 1.1.0
-  resolution: "@napi-rs/wasm-runtime@npm:1.1.0"
+"@napi-rs/wasm-runtime@npm:^1.1.0":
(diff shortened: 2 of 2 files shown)


openaev-front/src/components/errorWrapper.tsx
```diff
@@ -0,0 +1,17 @@
+import { type ComponentProps, type ComponentType } from 'react';
+
+import { ErrorBoundary, SimpleError } from './Error';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const errorWrapper = (Component: ComponentType<any>) => {
```
RomuDeuxfois: You should remove the export here and use the export from the export default at line 17

openaev-api/src/main/java/io/openaev/rest/HomeApi.java
```diff
@@ -48,8 +51,8 @@ public ResponseEntity<String> home() {
             .replaceAll(
                 "%APP_DESCRIPTION%",
                 "OpenAEV is an open source platform allowing organizations to plan, schedule and conduct adversary simulation campaigns and cyber crisis exercises.")
-            .replaceAll("%APP_FAVICON%", basePath + "/static/ext/favicon.png")
-            .replaceAll("%APP_MANIFEST%", basePath + "/static/ext/manifest.json")
+            .replaceAll("%APP_FAVICON%", basePath + "/static/favicon.png")
```
RomuDeuxfois, line 54: I don't understand the full changes from static/ext/ to static/

guillaumejparis: With vite doing the build we do not have the same "freedom" to set the public dir and to get the same uri withthe ext I needed to add another layer to an already weird static folder in the frontend : 
image

Do you thinbk it can causes retro compatibility problems ?

PR #4643, opened 2025-12-23
[frontend] fix(payload): error recovery when creating a payload (#4582)

### LINKED ISSUE #4582: CREATE CTA does not guide user to remaining mandatory fields
Description

When creating a Payload, once all mandatory fields in the “General” tab are filled. However, clicking create CTA it does nothing on the first attempt

Only after clicking CREATE a second time does the user realize that another mandatory field is still missing in the Command tab

Steps to Reproduce

+ Go to Create Payload

Fill all mandatory fields in the General section

+ Click CREATE

+ Observe: nothing happens

+ Click CREATE again

+ User is then informed that a mandatory field is missing in the Command tab

Expected Behavior

The CREATE button should guide the user to the next missing mandatory field

If required fields exist in another tab, the UI should:

Automatically navigate to the relevant tab

Actual Behavior

First click on CREATE has no visible effect

Validation feedback only appears after a second click

EmilieFo17:
### Related issues

* Closes #4582

### DIFF
--- a/openaev-front/src/admin/components/payloads/PayloadForm.tsx
+++ b/openaev-front/src/admin/components/payloads/PayloadForm.tsx
@@ -159,7 +159,7 @@ const PayloadForm = ({
   const {
     handleSubmit,
     control,
-    formState: { errors, isDirty, isSubmitting, defaultValues, dirtyFields },
+    formState: { isDirty, isSubmitting, defaultValues, dirtyFields },
   } = methods;
 
   const getTabForField = (fieldName: string): string | undefined => {


openaev-front/src/admin/components/payloads/PayloadForm.tsx
```diff
@@ -202,8 +202,9 @@ const PayloadForm = ({
     e.preventDefault();
     const isValid = await methods.trigger();
     if (!isValid) {
-      const firstErrorField = Object.keys(errors)[0];
+      const firstErrorField = Object.keys(methods.formState.errors)[0];
```
MarineLeM, line 205: It's exactly the same code. I tested it and the problem still persists.

MarineLeM (approved):

PR #4666, opened 2026-01-02
[backend] feat(stix): add stix management for domain names (#3509)

### LINKED ISSUE #3509: Parse DNS resolution from StixBundle
When we generate scenario from OpenCTI we would like to generate a payload which will run a DNS resolution on all the IOC type FQDN from OpenCTI. We can think about use IP observables too

To avoid creating one payload/fqdn we could think about a payload in list format but let you guys to find the best way.

gabriel-peze:
### Proposed changes

* Add STIX management for domain name indicators

### Testing Instructions

1. Create a report and Security Coverage on OpenCTI, linked to OpenAEV
2. Scenario should be created into OpenAEV using STIX calls, and runnable, with coverage shown on OpenCTI from OpenAEV.

### Related issues

* Closes #3509

### DIFF
--- a/openaev-api/src/main/java/io/openaev/utils/SecurityCoverageUtils.java
+++ b/openaev-api/src/main/java/io/openaev/utils/SecurityCoverageUtils.java
@@ -92,7 +92,7 @@ public class SecurityCoverageUtils {
       }
 
       if (obj.hasProperty(STIX_NAME) && StringUtils.isBlank(refId) && !isIndicator) {
-          refId = (String) obj.getProperty(STIX_NAME).getValue();
+        refId = (String) obj.getProperty(STIX_NAME).getValue();
       }
 
       if (!StringUtils.isBlank(refId)) {


openaev-api/src/main/java/io/openaev/rest/payload/service/PayloadService.java
```diff
@@ -307,4 +318,58 @@ public Page<Payload> searchPayloads(@NotNull final SearchPaginationInput searchP
...
+  private DnsResolution createDynamicDnsResolutionPayload() {
+    DnsResolution dynamicDnsResolutionPayload = new DnsResolution();
+    dynamicDnsResolutionPayload.setId(DYNAMIC_DNS_RESOLUTION_UUID);
+    dynamicDnsResolutionPayload.setHostname(DYNAMIC_DNS_RESOLUTION_HOSTNAME_VARIABLE);
+    dynamicDnsResolutionPayload.setName("Dynamic DNS Resolution");
+    dynamicDnsResolutionPayload.setDescription("Dynamic DNS Resolution by argument");
+    dynamicDnsResolutionPayload.setStatus(Payload.PAYLOAD_STATUS.VERIFIED);
+    dynamicDnsResolutionPayload.setSource(Payload.PAYLOAD_SOURCE.FILIGRAN);
+    dynamicDnsResolutionPayload.setType(DnsResolution.DNS_RESOLUTION_TYPE);
+    dynamicDnsResolutionPayload.setPlatforms(ALL_PLATFORMS);
```
savacano28, line 353: TODO:
Add dynamicDnsResolutionPayload.setExecutionArch(Payload.EXECUTION_ARCHITECTURE.ALL);

openaev-api/src/main/java/io/openaev/rest/payload/service/PayloadService.java
```diff
@@ -307,4 +318,58 @@ public Page<Payload> searchPayloads(@NotNull final SearchPaginationInput searchP
...
         Payload.class);
   }
+
+  /**
+   * Upsert for the Dynamic DNS Resolution payload, who run DNS Resolution by domain name given by
+   * argument
+   *
+   * @return the Dynamic DNS Resolution payload
+   */
+  public DnsResolution getDynamicDnsResolutionPayload() {
```
savacano28, line 330: Praise! 
A clean approach!

openaev-api/src/main/java/io/openaev/service/stix/SecurityCoverageInjectService.java
```diff
@@ -29,15 +35,22 @@
 public class SecurityCoverageInjectService {
 
   public static final int TARGET_NUMBER_OF_INJECTS = 1;
+  public static final Endpoint.PLATFORM_TYPE[] ALL_PLATFORMS =
```
savacano28, line 38: Praise!

openaev-api/src/main/java/io/openaev/rest/inject/service/InjectService.java
```diff
@@ -1180,6 +1180,10 @@ public Specification<Inject> hasGrantAccessForInject(Grant.GRANT_TYPE grantType)
       extractCombinationAttackPatternPlatformArchitecture(Scenario scenario) {
 
     return scenario.getInjects().stream()
+        .filter(
+            inject ->
+                inject.getPayload().isEmpty()
```
savacano28, line 1185: Question:
Can you check what happens in a scenario where an inject does not have a contract?
It could throw a NPE?

openaev-api/src/main/java/io/openaev/service/stix/SecurityCoverageInjectService.java
```diff
@@ -85,6 +101,75 @@ private void cleanInjectPlaceholders(String scenarioId) {
...
+
+          // 6. Create an inject, linked to the scenario for each contract
+          createInjectsByInjectorContracts(
+              indicator.getHostname(), dynamicDnsResolutionPayload, scenario);
+        });
+
+    // 7. Delete all previous injects non existing anymore on the OpenCTI report
+    previousExistingInject.stream()
+        .filter(inject -> !managedInjectsIds.contains(inject.getId()))
+        .forEach(injectRepository::delete);
```
savacano28: Suggestion:
Maybe you can try: deleteAllByIdInBatch and check perfomance with both cases?

openaev-api/src/main/java/io/openaev/service/stix/SecurityCoverageInjectService.java
```diff
@@ -512,4 +597,85 @@ private Map<AssetGroup, List<Endpoint>> computeMissingAssetGroups(
...
+   * Retrieve an existing inject on the given scenario by Payload hostname
+   *
+   * @param scenario to retrieve injects
+   * @param hostname to find
+   * @return founded inject id, null if not
+   */
+  private String findExistingInjectIdByHostname(Scenario scenario, String hostname) {
+    return scenario.getInjects().stream()
+        .filter(inject -> hasDnsResolutionFor(inject, hostname))
+        .findFirst()
```
savacano28: Question:
Why findFirst? it could be findAny()? => at least one (best case?)

openaev-api/src/main/java/io/openaev/service/stix/SecurityCoverageInjectService.java
```diff
@@ -512,4 +597,85 @@ private Map<AssetGroup, List<Endpoint>> computeMissingAssetGroups(
...
+    }
+
+    Object payload = inject.getInjectorContract().get().getPayload();
+    if (!(payload instanceof DnsResolution dns)) {
+      return false;
+    }
+
+    return inject.getContent() != null
+        && inject.getContent().has(DYNAMIC_DNS_RESOLUTION_HOSTNAME_KEY)
+        && hostname.equals(inject.getContent().get(DYNAMIC_DNS_RESOLUTION_HOSTNAME_KEY).textValue())
```
savacano28: Todo:
Could hostname be null? If so, you might want to invert the comparison to avoid a NPE

openaev-api/src/main/java/io/openaev/service/stix/SecurityCoverageService.java
```diff
@@ -416,6 +426,15 @@ private List<ObjectBase> getCoverageForSimulation(Exercise exercise)
           objects);
     }
 
+    processCoverageRefs(
```
savacano28: Suggestion:
Maybe you could replace the word exercise for simulation :) ?
Ty!

openaev-api/src/main/java/io/openaev/service/stix/SecurityCoverageInjectService.java
savacano28: Question:
hostname can be null? in that case you could make this a safer comparison?

openaev-api/src/test/java/io/openaev/utils/fixtures/composers/PayloadComposer.java
```diff
@@ -12,6 +13,7 @@
 public class PayloadComposer extends ComposerBase<Payload> {
 
   @Autowired PayloadRepository payloadRepository;
+  @Autowired EntityManager entityManager;
```
savacano28: Todo:
EM appears to be unused -> to  remove

openaev-model/src/main/java/io/openaev/stix/objects/constants/ExtendedProperties.java
```diff
@@ -6,7 +6,8 @@ public enum ExtendedProperties {
   COVERED("covered"),
   COVERAGE("coverage"),
   MITRE_EXTENSION_DEFINITION("extension-definition--322b8f77-262a-4cb8-a915-1e441e00329b"),
-  OPENCTI_EXTENSION_DEFINITION("extension-definition--ea279b3e-5c71-4632-ac08-831c66a786ba");
+  OPENCTI_EXTENSION_DEFINITION("extension-definition--ea279b3e-5c71-4632-ac08-831c66a786ba"),
+  DOMAIN_NAME("Domain-Name");
```
savacano28: Question:
Could you check if Octi send always this kebab case: Domain-Name?

openaev-api/src/main/java/io/openaev/rest/inject/service/InjectService.java
gabriel-peze: In that case, getPayload always return an Optional with a Payload, or null, so inject.getPayload().isEmpty() can't return an NPE issue.
The instanceof played with the content of the Optional is null safe, so it won't create an NPE.
The only potential NPE can occur into the getPayload method, when this.getInjectorContract().isPresent() is called, but fonctionnaly, inject should always have a contract, so it shouldn't be possible to create an NPE.

openaev-model/src/main/java/io/openaev/stix/objects/constants/ExtendedProperties.java
gabriel-peze: Yes, all the Observables types are in this format, i found them into the OCTI project.
https://github.com/OpenCTI-Platform/opencti/blob/master/client-python/pycti/utils/constants.py

antoinemzs: This is a value, not a property name, therefore should not be in this enum. It should be a constant, though.

openaev-model/src/main/java/io/openaev/database/repository/InjectRepository.java
```diff
@@ -425,6 +425,19 @@ List<Object[]> findAllByTitleLinkedToFindingsWithContext(
...
   boolean existsByIdAndScenarioIsNullAndExerciseIsNull(String id);
 
+  @Modifying
+  @Query(
+      value =
+          "DELETE FROM injects i "
+              + "USING injectors_contracts ic, payloads p "
+              + "WHERE i.inject_injector_contract = ic.injector_contract_id "
+              + "AND ic.injector_contract_payload = p.payload_id "
+              + "AND p.payload_type = 'DnsResolution' "
```
antoinemzs: Should use a constant for the payload_type value passed here

openaev-model/src/main/java/io/openaev/database/model/StixRefToExternalRef.java
```diff
@@ -17,6 +17,39 @@ public class StixRefToExternalRef {
   @JsonProperty("external_ref")
   private String externalRef;
 
+  @JsonProperty("name")
+  private String name;
+
+  @JsonProperty("description")
+  private String description;
```
antoinemzs: `name` and `description` are unused, so perhaps remove them

openaev-api/src/main/java/io/openaev/service/stix/SecurityCoverageService.java
```diff
@@ -493,38 +515,73 @@ private void processCoverageRefs(
...
+                                    .getContent()
+                                    .get(DYNAMIC_DNS_RESOLUTION_HOSTNAME_KEY)
+                                    .textValue()))
+                .collect(Collectors.toList()),
+        inject ->
+            Optional.ofNullable(inject)
+                .map(Collections::singletonList)
+                .orElse(Collections.emptyList()),
+        Inject::getId);
+  }
```
antoinemzs: You aren't using the `externalRef` attribute here, why did you need to create a dedicated `hostname` attribute in its stead?

openaev-api/src/main/java/io/openaev/rest/payload/service/PayloadService.java
```diff
@@ -307,4 +318,59 @@ public Page<Payload> searchPayloads(@NotNull final SearchPaginationInput searchP
...
+    return payloadRepository
+        .findById(DYNAMIC_DNS_RESOLUTION_UUID)
+        .map(payload -> (DnsResolution) payload)
+        .orElseGet(this::createDynamicDnsResolutionPayload);
```
antoinemzs, line 334: Possible thread safety issue? There could theoretically be a race condition on the retrieval of the payload, which would lead to a clash with creating the static id payload in the `orElseGet` path.

Possible solutions:
* Check out https://www.baeldung.com/java-synchronized for details on the synchronized keyword. It would work here since PayloadService is a singleton, hence there would only be a signle instance so the synchronized keyword would create a global mutex situation.
* Use our own `@Lock` annotation with locking on the static ID. I would favour this option because it is consistent with our standard.

openaev-api/src/main/java/io/openaev/service/stix/SecurityCoverageService.java
gabriel-peze: I understand.
Previously on devs i had to use the names and descriptions from OCTI, and i saved directly the hostname too instead of saving it on externalRefs.
But the concept change a little bit during the development process, and name and description become useless, and hostname became the only id.
So, i forgot to change it on ExternalRef object. I'll change it to use the externalRef attribute instead, and remove the hostname attribute on StixRefToExternalRef before merging.

gabriel-peze:
@antoinemzs i let you review the changes i made after your feedbacks :)

PR #4674, opened 2026-01-05
[backend] fix(securitycoverage): add stix modified column (#4521)

savacano28:
### Proposed changes

* Extract Stix modified in a column to check if last stix bundle was modified directly and no reuse updated column from security coverage.

### Testing Instructions

1. Check column has value from octi Stix bundle

### Related issues

* #4521 

### Checklist

### Further comments

### DIFF
--- a/openaev-api/src/test/java/io/openaev/api/stix_process/StixApiTest.java
+++ b/openaev-api/src/test/java/io/openaev/api/stix_process/StixApiTest.java
@@ -924,14 +924,14 @@ class StixApiTest extends IntegrationTest {
       return rootNode;
     }
 
-    ObjectNode firstObject = (ObjectNode) objectsNode.get(index);
+    ObjectNode objectNode = (ObjectNode) objectsNode.get(index);
 
     if (newValues != null && !newValues.isEmpty()) {
       ArrayNode arrayNode = objectMapper.createArrayNode();
       newValues.forEach(arrayNode::add);
-      firstObject.set(fieldName, arrayNode);
+      objectNode.set(fieldName, arrayNode);
     } else if (newValue != null) {
-      firstObject.put(fieldName, newValue);
+      objectNode.put(fieldName, newValue);
     }
 
     ((ObjectNode) rootNode.path("event"))


openaev-api/src/main/java/io/openaev/migration/V4_53__Add_stix_modified_column.java
```diff
@@ -0,0 +1,26 @@
...
+    try (Statement statement = context.getConnection().createStatement()) {
+      statement.execute(
+          """
+              ALTER TABLE security_coverages ADD COLUMN security_coverage_stix_modified TIMESTAMPTZ;
+                """);
+
+      statement.execute(
+          """
+                UPDATE security_coverages SET security_coverage_stix_modified = security_coverage_updated_at
+                WHERE security_coverage_stix_modified IS NULL;
```
antoinemzs: I thought we'd dig into the stored, raw json to find the actual value? We can still fall back to `updated_at` if the value cannot be found (via `COALESCE`).

Have a look at https://www.postgresql.org/docs/17/functions-json.html for starters to safely query json structures using a postgres-native API.

openaev-api/src/test/java/io/openaev/api/stix_process/StixApiTest.java
antoinemzs, line 1: praise: good thinking for loading the stix documents as json nodes rather than plain text 🙏

openaev-api/src/test/java/io/openaev/api/stix_process/StixApiTest.java
```diff
@@ -464,129 +517,149 @@ void shouldCreateScenario() throws Exception {
...
-      mvc.perform(
-              post(STIX_URI + "/process-bundle")
-                  .contentType(MediaType.APPLICATION_JSON)
-                  .content(stixSecurityCoverage))
-          .andExpect(status().isBadRequest());
+      JsonNode updated =
+          updateStixObjectField(
+              stixSecurityCoverage,
+              CommonProperties.MODIFIED.toString(),
+              Instant.now().toString(),
```
antoinemzs, line 627: nitpick: it works here because time flows only one way (or at least, as we experience it), and we are trying to set a date *later* that the original reference (which we don't see here but is set as a past date in the source json).
Note that in general it's best practice in tests to use absolutely static time references so that a test is always guaranteed to work forever.

savacano28: Some people blame entropy :D

PR #4675, opened 2026-01-05
[backend] fix: update stix test api (#4521)

GaetanSantucci:
We use the field modified from Stix in order to know if this object has changed.
We will refact these test in other pr.

### DIFF
--- a/openaev-api/src/test/java/io/openaev/api/stix_process/StixApiTest.java
+++ b/openaev-api/src/test/java/io/openaev/api/stix_process/StixApiTest.java
@@ -40,7 +40,6 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Set;
 import org.apache.commons.io.IOUtils;
-import org.h2.schema.Domain;
 import org.junit.jupiter.api.*;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.MediaType;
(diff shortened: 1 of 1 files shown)


savacano28 (approved):

openaev-api/src/test/java/io/openaev/api/stix_process/StixApiTest.java
```diff
@@ -455,7 +456,7 @@ void shouldUpdateScenarioAndKeepSameNumberInjectsWhenUpdatedStixHasSameAttacks()
       entityManager.clear();
 
       String modifiedSecurityCoverage =
-          stixSecurityCoverage.replace("2025-08-04T14:00:00Z", "2025-12-20T14:00:00Z");
+          stixSecurityCoverage.replace("2025-12-31T14:00:00Z", Instant.now().toString());
```
Dimfacion, line 459: **nitpick :** I guess it's too late now but it would have been more robust to rely on the path of the property to update rather than on an string that might be changed later on.

GaetanSantucci: Yes, we opted for a quick fix to unblock the CI tests, but we've opened a PR (#4674) to improve the tests.

PR #4681, opened 2026-01-06
[backend] feat: convert built-in injectors (#4606)

### LINKED ISSUE #4606: [EE] Built-in integrations: convert built-in injectors to new scheme
## Context

Several built-in injectors are still pending conversion to be activated by the new system:
Image

## Use case

All integrations should go through the built-in manager if possible

## Current Workaround

THey work as they are, but cannot be part of the Catalog

## Proposed Solution

Convert injectors with help of internal documentation.

Some case-by-case discernment will be needed to handle edge cases (design choices that may make it harder to implement an Integration) e.g. with the Email injector.

Note some injectors should be set to autostart, not needing any configurations; they are the easiest wins.

MarineLeM:
### Proposed changes

* Convert built in injector to new scheme for catalog

### Testing Instructions

1. Step-by-step how to test

### Related issues

* Closes #4606

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -58,8 +58,9 @@ steps:
       SPRING_PROFILES_ACTIVE: ci
       OPENBAS_RABBITMQ_HOSTNAME: rabbitmq-e2e
     commands:
-      - apt update && apt install -y gnupg nodejs git npm
+      - apt update && apt install -y gnupg git npm
       - curl -fsSL https://deb.nodesource.com/setup_22.x | bash -
+      - apt update && apt install -y nodejs
       - npm install --global yarn
       - cd openaev-front
       - yarn install


openaev-api/src/main/java/io/openaev/executors/InjectorContext.java
```diff
@@ -0,0 +1,28 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.openaev.config.OpenAEVConfig;
+import io.openaev.database.repository.DocumentRepository;
+import io.openaev.service.FileService;
+import lombok.Getter;
+import org.springframework.stereotype.Component;
+
+@Component
+@Getter
+public class InjectorContext {
```
RomuDeuxfois, line 12: **nitpick**: I’m not a big fan of this approach on the backend. It aggregates services to make them easier to access, but without adding any additional logic. From experience, it’s better to keep explicit and clean dependency injection where the services are actually used.

openaev-api/src/main/java/io/openaev/injectors/caldera/CalderaExecutor.java
```diff
@@ -37,13 +38,9 @@
...
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
-import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
-import org.springframework.stereotype.Component;
 import org.springframework.transaction.annotation.Transactional;
 
-@Component(CalderaContract.TYPE)
-@RequiredArgsConstructor
 @Slf4j
```
RomuDeuxfois: @Component(CalderaContract.TYPE) is gone, one click to remove it. So nice!

openaev-api/src/main/java/io/openaev/injectors/caldera/CalderaInjector.java
RomuDeuxfois, line 13: @Component(CalderaContract.TYPE) can be removed to try a tentative removal without fully removing everything.

openaev-api/src/main/java/io/openaev/integration/impl/injectors/challenge/ChallengeInjectorIntegration.java
```diff
@@ -0,0 +1,73 @@
...
+import io.openaev.injectors.email.service.EmailService;
+import io.openaev.integration.ComponentRequestEngine;
+import io.openaev.integration.Integration;
+import io.openaev.integration.QualifiedComponent;
+import io.openaev.integrations.InjectorService;
+import io.openaev.service.InjectExpectationService;
+import io.openaev.service.connector_instances.ConnectorInstanceService;
+import java.util.List;
+
+public class ChallengeInjectorIntegration extends Integration {
```
RomuDeuxfois: It’s a bit unfortunate that the integration injectors are not in the same package. We now have integration/impl/injectors and injectors.

openaev-api/src/main/java/io/openaev/injectors/lade/LadeExecutor.java
```diff
@@ -10,19 +10,21 @@
...
 import io.openaev.injectors.lade.service.LadeService;
 import io.openaev.model.ExecutionProcess;
 import jakarta.validation.constraints.NotNull;
 import java.util.List;
-import lombok.RequiredArgsConstructor;
-import org.springframework.stereotype.Component;
 
-@Component(LadeContract.TYPE)
-@RequiredArgsConstructor
 public class LadeExecutor extends Injector {
```
RomuDeuxfois: Are you going to remove this one as well?

openaev-api/src/main/java/io/openaev/injectors/mastodon/MastodonExecutor.java
```diff
@@ -6,20 +6,22 @@
...
-import lombok.RequiredArgsConstructor;
-import org.springframework.stereotype.Component;
 
-@Component(MastodonContract.TYPE)
-@RequiredArgsConstructor
 public class MastodonExecutor extends Injector {
 
   private final MastodonService mastodonService;
 
+  public MastodonExecutor(InjectorContext context, final MastodonService mastodonService) {
```
RomuDeuxfois: This one has been used recently by the community by the way.

openaev-api/src/main/java/io/openaev/integration/impl/executors/caldera/CalderaExecutorIntegration.java
```diff
@@ -63,13 +65,19 @@ public CalderaExecutorIntegration(
...
     this.taskScheduler = taskScheduler;
     this.executorService = executorService;
+    this.connectorInstanceService = connectorInstanceService;
+    this.connectorInstance = connectorInstance;
   }
 
   @Override
   protected void innerStart() throws Exception {
+    String executorId =
+        connectorInstanceService.getConnectorInstanceConfigurationsByIdAndKey(
```
RomuDeuxfois, line 93: What happens if the migration is executed after the inner start here?

openaev-api/src/main/java/io/openaev/integration/impl/executors/tanium/TaniumExecutorIntegration.java
```diff
@@ -73,13 +75,19 @@ public TaniumExecutorIntegration(
     this.licenseCacheManager = licenseCacheManager;
     this.executorService = executorService;
     this.taskScheduler = taskScheduler;
+    this.connectorInstanceService = connectorInstanceService;
+    this.connectorInstance = connectorInstance;
   }
 
   @Override
   protected void innerStart() throws Exception {
+    String executorId =
```
RomuDeuxfois, line 103: EXECUTOR_ID can be a constant aswell as other one relativ to integration config key

openaev-api/src/main/java/io/openaev/integration/impl/injectors/challenge/ChallengeInjectorIntegrationFactory.java
```diff
@@ -0,0 +1,97 @@
...
+    this.challengeContract = challengeContract;
+    this.injectorContext = injectorContext;
+    this.emailService = emailService;
+    this.injectorService = injectorService;
+    this.injectExpectationService = injectExpectationService;
+    this.challengeRepository = challengeRepository;
+  }
+
+  @Override
+  protected final String getClassName() {
```
RomuDeuxfois: Since this implementation never changes, it should live directly in the IntegrationFactory abstract class.

openaev-api/src/main/java/io/openaev/integration/impl/injectors/opencti/OpenctiInjectorIntegrationFactory.java
```diff
@@ -0,0 +1,115 @@
...
+
+  @Override
+  protected void insertCatalogEntry() throws Exception {
+    String logoFilename = "%s-logo.png".formatted(getClassName());
+    fileService.uploadStream(
+        FileService.CONNECTORS_LOGO_PATH,
+        logoFilename,
+        getClass().getResourceAsStream("/img/icon-opencti.png"));
+    CatalogConnector connector = new CatalogConnector();
+    connector.setTitle("OpenCTI");
```
RomuDeuxfois: Can be replaced by OPENCTI_INJECTOR_NAME

openaev-api/src/main/java/io/openaev/integration/impl/injectors/ovh/OvhInjectorIntegration.java
```diff
@@ -0,0 +1,69 @@
...
+        false,
+        List.of());
+    OvhSmsService ovhSmsService = new OvhSmsService(this.config);
+    this.ovhSmsExecutor =
+        new OvhSmsExecutor(injectorContext, ovhSmsService, injectExpectationService);
+  }
+
+  @Override
+  protected void innerStop() {
+    // TODO
```
RomuDeuxfois, line 104: TODO ? or noop ?

openaev-api/src/main/java/io/openaev/integration/ManagerFactory.java
```diff
@@ -19,8 +19,8 @@ public class ManagerFactory {
   public Manager getManager() throws Exception {
     if (manager == null) {
       this.manager = new Manager(factories);
-      this.manager.monitorIntegrations();
     }
+    this.manager.monitorIntegrations();
```
RomuDeuxfois: This is already fixed in the current release by the code in the ManagerIntegrationsSyncJob class

  public void execute(JobExecutionContext jobExecutionContext) throws JobExecutionException {
    try {
      managerFactory.getManager().monitorIntegrations();
    } catch (Exception e) {
      throw new JobExecutionException(e);
    }
  }

openaev-api/src/main/java/io/openaev/service/connector_instances/ConnectorInstanceService.java
```diff
@@ -121,6 +121,21 @@ public Set<ConnectorInstanceConfiguration> getConnectorInstanceConfigurations(St
...
   }
 
+  /**
+   * Retrieve the value of a specific connector instance configuration by instance ID and key.
+   *
+   * @param instanceId the connector instance ID to search for the configuration
+   * @param key the configuration key to retrieve
+   * @return the configuration value as a String
+   */
+  public String getConnectorInstanceConfigurationsByIdAndKey(String instanceId, String key) {
```
RomuDeuxfois, line 184: Shouldn’t we throw an exception if the key is not found? Otherwise, if it ends up being null at the registration method level, we’ll throw later without clearly understanding why.

openaev-framework/src/main/java/io/openaev/integrations/InjectorService.java
```diff
@@ -155,10 +148,6 @@ public void register(
     log.info("Successfully registered injector '{}' (type: {})", name, contractor.getType());
   }
 
-  private void handleNonExposedInjector(String id) {
```
RomuDeuxfois, line 158: Oh, this is really interesting! It's what we were discussing earlier… so it looks like we can currently remove injectors on the fly.

openaev-model/src/main/java/io/openaev/healthcheck/enums/ExternalServiceDependency.java
```diff
@@ -56,26 +56,37 @@ public String getValue() {
...
+      throw new IllegalArgumentException("Injector type cannot be null or blank");
     }
+
+    // Special cases
+    if ("openaev_email".equalsIgnoreCase(injectorType)) {
+      return new ExternalServiceDependency[] {SMTP, IMAP};
+    }
+    if ("openaev_implant".equalsIgnoreCase(injectorType)) {
+      return new ExternalServiceDependency[] {};
+    }
```
RomuDeuxfois, line 85: I’m not a big fan of duplicated configuration here. We could use some introspection to discover which dependencies are related to a given type, but as a first step, we can simply add a constant on EmailInjectorIntegration for the dependencies and reuse it here.
We should also use the openaev_email constant.

RomuDeuxfois:
## Tests

* The `_dummy` injectors are still present after a from scratch start-up. It seems we never go through `deleteDummyInjectorIfItExists`, which is called in `service.InjectorService.registerInjector` (itself called by the API), but not in `integrations.InjectorService.register`. There is also quite a bit of code duplication here.
* The **SMS contract** and **OpenCTI** are missing in the injectors section.
Oh, after a second round, everyone is here…
* I see the **OpenAEV executor** in the catalog and an OpenAEV Agent on the executor part, why is that? I don’t see the other possible executors.
 Oh, after a second round, everyone is here…
* I launch **Nuclei** and it works, but on the UI side the button keeps loading instead of refreshing and indicating that the injector is up and running.
* On the logs screen, would it be possible to automatically scroll to the bottom instead of doing it manually? Alternatively, having a “scroll to bottom” button like in Drone would be helpful.
* I stopped my Nuclei injector, when I launch an Atomic Testing would it be possible to check whether the injector is enabled instead of waiting for the 10-minute garbage collector? That would greatly improve the user experience.

## UI

* We now have an additional panel
{575951B8-913A-4338-AF73-043794713338}
compared to before.
{B403F6D1-E288-43BF-A970-16BCE970C2DE}
  Is this expected?

* Breadcrumbs should be capitalized: **Injectors**, **Collectors**, **Executors**.
{A2B6D3C0-867B-4BDB-B617-A2A4B0E5A3E2}

* The default sort for integrations no longer seems to be alphabetical, is that intended?

* Search no longer works.

* There is no longer a color distinction between **built-in** and **external** tags like before.
{4190879D-0698-48AB-9C50-4D9B51937823}

* When I deploy an injector but don’t start it, it’s hard to see the difference. Will the badge address this?

PR #4685, opened 2026-01-06
[backend/frontend] feat(stix): introduce platform affinity (#4517)

### LINKED ISSUE #4517: [SC☂️] Platform-Aware Payload and Asset Selection => platform_affinity logic
OAEV must select eligible assets for simulation execution based on the platform_affinity list sent by OCTI for each TTP.
The current label-based mechanism must be removed/replaced

antoinemzs:
### Proposed changes

* Match default asset rules with provided platform affinity
* Default to [windows, linux, macos] when no affinity is set
* Deliver configuration to openaev instance for creating defaut asset groups and associated rules

### Testing Instructions

1. Create security coverage in OCTI with various combinations of Platform Affinity
2. OAEV scenario should create injects based on chosen platforms with non  mixed asset groups
3. There should be 3 new default asset rules based on 3 new tags: "security coverage: %s" where %s is one of windows, linux, macos
4. There should be non mixed asset groups assigned to these rules with matching platform (both arm64 and x86_65 archs should be assigne to the same default asset rule)

### Related issues

* Closes #4517

### Checklist

### Further comments

### DIFF
--- a/openaev-api/src/main/java/io/openaev/service/TagRuleService.java
+++ b/openaev-api/src/main/java/io/openaev/service/TagRuleService.java
@@ -63,7 +63,7 @@ public class TagRuleService {
     // we block creation of tag rules for reserved tags
     if (TagRule.RESERVED_TAG_NAMES.contains(tag.getName()) && !allowCreatingReserved) {
       throw new ForbiddenException(
-          "Creation of a rule is not allowed for the tag " + tag.getName());
+          "Creating a rule for the reserved tag '%s' is not permitted.".formatted(tag.getName()));
     }
 
     // if the tag  or one of the asset group doesn't exist we exist throw a ElementNotFoundException


openaev-model/src/main/java/io/openaev/database/model/AssetGroup.java
```diff
@@ -37,10 +37,25 @@
 })
 public class AssetGroup implements Base {
 
+  public Map<Endpoint.PLATFORM_TYPE, Map<Endpoint.PLATFORM_ARCH, String>> wellKnown() {
```
RomuDeuxfois: Seems unused.

openaev-model/src/main/java/io/openaev/database/model/TagRule.java
```diff
@@ -31,22 +40,36 @@ public class TagRule implements Base {
...
-  @Queryable(filterable = true, path = "assetGroups.name")
+  @Queryable(filterable = true, sortable = true, path = "assetGroups.name")
   @JsonProperty("tag_rule_asset_groups")
   private List<AssetGroup> assetGroups = new ArrayList<>();
 
   @Getter(onMethod_ = @JsonIgnore)
   @Transient
   private final ResourceType resourceType = ResourceType.TAG_RULE;
 
+  public void setTag(Tag tag) {
```
RomuDeuxfois, line 64: It feels a bit odd to block a parent entity based on a child entity. Just a personal impression.

openaev-model/src/main/java/io/openaev/database/model/TagRule.java
```diff
@@ -18,6 +21,12 @@
 @Table(name = "tag_rules")
 @EntityListeners(ModelBaseListener.class)
 public class TagRule implements Base {
+  public static Set<String> RESERVED_TAG_NAMES =
```
RomuDeuxfois, line 24: Relative to the Tag, so it should be placed on the Tag entity, right?

openaev-api/src/main/java/io/openaev/datapack/packs/V20260101_Starter_pack.java
```diff
@@ -1,38 +1,48 @@
...
-import org.springframework.transaction.annotation.Transactional;
 
-/** Command line runner that initializes the starter pack on first application start. */
-@Slf4j
 @Component
-@Transactional()
-@RequiredArgsConstructor
-public class InitStarterPackCommandLineRunner implements CommandLineRunner {
+@Slf4j
+public class V20260101_Starter_pack extends DataPack {
```
RomuDeuxfois, line 22: This is our new way to run starter pack feature right ?

openaev-api/src/main/java/io/openaev/datapack/packs/V20260101_Starter_pack.java
```diff
@@ -1,38 +1,48 @@
...
+      TagService tagService,
+      EndpointService endpointService,
+      AssetGroupService assetGroupService,
+      TagRuleService tagRuleService,
+      ImportService importService,
+      ZipJsonService<CustomDashboard> zipJsonService,
+      ResourcePatternResolver resolver) {
+    super(dataPackService);
+    this.settingRepository = settingRepository;
+    this.tagRuleRepository = tagRuleRepository;
```
RomuDeuxfois: Seems unused.

openaev-api/src/main/java/io/openaev/service/stix/SecurityCoverageService.java
```diff
@@ -149,10 +146,19 @@ private SecurityCoverage buildSecurityCoverageFromStix(
         labels.add(stixString.getValue());
       }
     }
-    // Force opencti tag
-    labels.add(OPENCTI_TAG_NAME);
```
RomuDeuxfois, line 153: Should we keep OpenCTI as the source for security coverage ?

openaev-api/src/main/java/io/openaev/datapack/packs/V20260107_Tags_and_tagrules_and_assetgroups.java
```diff
@@ -0,0 +1,94 @@
...
+
+        Filters.FilterGroup filterGroup = new Filters.FilterGroup();
+        filterGroup.setMode(Filters.FilterMode.and);
+        filterGroup.setFilters(List.of(filterArch, filterPlatform));
+
+        AssetGroup assetGroup = new AssetGroup();
+        assetGroup.setName("All %s %s".formatted(platform.toString(), arch.toString()));
+        assetGroup.setDynamicFilter(filterGroup);
+
+        AssetGroup saved = this.assetGroupService.createAssetGroup(assetGroup);
```
RomuDeuxfois, line 84: Seems unused.

openaev-model/src/main/java/io/openaev/database/model/TagRule.java
antoinemzs: It just incidentally bijects to a subset of well known tag names, but that isn't necessarily the case by design.

openaev-api/src/main/java/io/openaev/datapack/packs/V20260101_Starter_pack.java
antoinemzs: Yes,
the migration has been taken care of: I retained the initial guards in the code, so that instances that have already imported the original starter pack don't import it again by accident with the new scheme. This datapack will run, do nothing if the guard is activated, and mark itself as run in the db, as per the design.

openaev-model/src/main/java/io/openaev/database/model/TagRule.java
antoinemzs: All I have done here is make a guard unavoidable compared to the prior implementation. Some tag rules are meant to avoid being deleted, only modified in which asset groups are assigned to them. The code was riddled with various guards that prevented such update based on a hardcoded "opencti" tag, and I've recentralised the behaviour here, removed the duplicated hardcoding and made the protected status refer to a set of tags, not just one single hardcoded one.

openaev-api/src/main/java/io/openaev/service/stix/SecurityCoverageService.java
antoinemzs: That wasn't what we have discussed with Sam and Denise, so I guess not

openaev-api/src/main/java/io/openaev/service/TagRuleService.java
```diff
@@ -45,47 +46,84 @@ public List<TagRule> findAll() {
...
+
+  public TagRule createTagRule(
+      @NotBlank final Tag tag,
+      final List<String> assetGroupIds,
+      final boolean allowCreatingReserved) {
+    // we block creation of tag rules for reserved tags
+    if (TagRule.RESERVED_TAG_NAMES.contains(tag.getName()) && !allowCreatingReserved) {
       throw new ForbiddenException(
-          "Creation of a rule is not allowed for the tag " + OPENCTI_TAG_NAME);
+          "Creation of a rule is not allowed for the tag " + tag.getName());
```
RomuDeuxfois: **nitpick**: The front-end explanation doesn’t really make sense. When I try to create a tag rule, I can’t understand why it’s not possible without additional context. It would be much clearer to display a message like: “A tag rule with this tag already exists.”
Alternatively, the front-end could simply prevent offering the option to create a tag rule with an existing tag, to better guide the user.

RomuDeuxfois:
Tests:
- Creating a new tag rule with an already protected tag is not allowed 🟢
- Updating a tag rule with an existing protected tag results in duplicate tag rules 🟢

Minor front-end details to improve. Everything else looks good to me.

PR #4687, opened 2026-01-06
[backend] feat(catalog): add a connector instance hash (#4604)

### LINKED ISSUE #4604: [EE] Built-in integrations: monitor ConnectorInstance hash for reconfiguring running integration
## Context

The configuration of a running Integration may be modified at any time. The Integrations must check the computed hash of their internal ConnectorInstance to reconfigure so that new configuration values are taken into account.

## Use case

Deployed instances from Catalog entries may be reconfigured at any time via API and via GUI. New values should be taken into account as soon as possible.

## Current Workaround

Manually stop and then start the integration again

## Proposed Solution

Use the same system as for the XTM Composer integration (external) using a hash of the ConnectorInstance (with config) to trigger the reconfiguration of an Integration object.

## Additional Information

Be mindful that reconfiguring may break running integrations if they are doing work based on the previous version of the configuration (e.g. a hostname changes mid-course in the execution of an inject).

We could make a copy of the configuration prior to starting a work, so that the main copy can be changed or swapped without side effects.

camrrx:
### Proposed changes

*  Compute a hash from an identifier 
* Update built-in integration lifecycle to automatically stop and restart a running integration when the computer hash changes.

### Testing Instructions

1. Update a connector instance configuration from a built-in connector 
2. The hash will change and the connector will stop and start. 

### Related issues

* Closes #4604 

### Checklist

### Further comments


openaev-model/src/main/java/io/openaev/database/model/ConnectorInstanceInMemory.java
```diff
@@ -19,4 +19,11 @@ public class ConnectorInstanceInMemory extends ConnectorInstance {
   private String className;
 
   private String id;
+
+  private CatalogConnector catalogConnector;
```
RomuDeuxfois: question: What's used for ? I can't find any instanciation of this property.

openaev-api/src/main/java/io/openaev/integration/Integration.java
```diff
@@ -36,6 +39,8 @@ private void start() throws Exception {
     if (ConnectorInstancePersisted.CURRENT_STATUS_TYPE.stopped.equals(this.currentStatus)) {
       this.innerStart();
       this.currentStatus = ConnectorInstance.CURRENT_STATUS_TYPE.started;
+      this.appliedHash = ConnectorInstanceHashHelper.computeInstanceHash(this.connectorInstance);
+      ;
```
RomuDeuxfois:
```suggestion
```

openaev-api/src/main/java/io/openaev/integration/Integration.java
```diff
@@ -58,6 +63,25 @@ public void initialise() throws Exception {
         this.stop();
         return;
       }
+
+      String instanceHash = ConnectorInstanceHashHelper.computeInstanceHash(this.connectorInstance);
+      if (this.appliedHash == null) {
+        this.appliedHash = instanceHash;
+      }
```
RomuDeuxfois: Question: It's not necessary I think. appliedHash is computed if the instance is running, and if not we just start it at ligne 89.

openaev-model/src/main/java/io/openaev/database/model/ConnectorInstanceInMemory.java
camrrx: I added this property mainly because ConnectorInstance now declares the abstract method getCatalogConnector().
Without it, ConnectorInstanceInMemory fails to compile. Open to suggestions if you see a cleaner alternative.

RomuDeuxfois: I have a few ideas, but since I don’t yet fully understand the overall architecture that has been put in place, I might be mistaken.

From my point of view, the issue comes from the use of the abstract class. The idea of enforcing the use of this abstract class across all mandatory methods forces the declaration of the private CatalogConnector catalogConnector property, which leads to the issue I’m describing.

At what point do we not have a ConnectorInstancePersisted anymore, but instead a ConnectorInstanceInMemory?

camrrx: We end up with a ConnectorInstanceInMemory during the factory sync (Manager -> OpenAEVExecutorIntegrationFactory.findRelatedInstances -> return ConnectorInstanceInMemory) : factories can return in-memory instances before the integration is refreshed and bound to a persisted ConnectorInstancePersisted from the database. After that refresh, the integration usually works with the persisted instance.

RomuDeuxfois: Ok, so if I understand correctly, the catalog connector is not available until we have a `ConnectorInstancePersisted`.
In that case, would it make sense to update the code like this, assuming it aligns with your intended code structure?

```
@Override
public CatalogConnector getCatalogConnector() {
  return null; // Not persisted yet, so not available
}
```

camrrx: We discussed this with @antoinemzs and there was indeed some confusion on my side.
To clarify the model : 
- _ConnectorInstance_ is the abstract base class
- _ConnectorInstancePersisted_ is the Hibernate entity backed by the database
- _ConnectorInstanceInMemory_ is a plain Java object, not persisted, used to satisfy auto-start integrations

An integration is created with either an inMemory or a persisted one, and keeps it for its entire lifecycle, we never convert an inMemory into a persisted one. 

Following the discussion, I’ll change the approach.
Instead, I’ll add a getHashIdentity() method on ConnectorInstance, implemented by :
ConnectorInstancePersisted will get its identity from the catalog (container image/version or className).
ConnectorInstanceInMemory will provide its own identity.

openaev-api/src/main/java/io/openaev/integration/Integration.java
```diff
@@ -58,6 +62,23 @@ public void initialise() throws Exception {
...
+              this.connectorInstance.getRequestedStatus());
+
+      if (isRunning
+          && !isStoppingRequested
+          && this.appliedHash != null
+          && !instanceHash.equals(this.appliedHash)) {
+
+        this.stop();
+        this.start();
+      }
```
antoinemzs: I'm sure we could refactor the contents of this method to avoid some repetition and clearly delineate the various states and actions.

for example, if we assign a local variable like so:
```java
boolean isRunning =
          ConnectorInstancePersisted.CURRENT_STATUS_TYPE.started.equals(this.currentStatus);
```
we should be using it throughout.

at the very least, the method should guarantee that only a single action can be taken through the course of the method; stacked `if` blocks don't scale well and are harder to understand.

openaev-api/src/main/java/io/openaev/helper/ConnectorInstanceHashHelper.java
```diff
@@ -0,0 +1,64 @@
...
+
+    return configurations.stream()
+        .filter(c -> c != null && c.getKey() != null && c.getValue() != null)
+        .sorted(Comparator.comparing(ConnectorInstanceConfiguration::getKey))
+        .map(c -> String.format("%s=%s", c.getKey(), c.getValue()))
+        .collect(Collectors.joining(";"));
+  }
+
+  // Hashing
+  private static String hashWithSHA256(String text) {
```
savacano28: There is already a hash generation method in another class (CryptoHelper.XXAlgoXXHex and hex). Maybe we could consider a small refactor to mutualize the implementation (and update class name)?

openaev-model/src/main/java/io/openaev/helper/CryptoHelper.java
```diff
@@ -21,4 +23,25 @@ public static String md5Hex(String message) {
     }
     return null;
   }
+
+  public static String hashWithSHA256(String text) {
```
savacano28, line 55: I think we could refact this funtion with md5Hex() and hex()

openaev-api/src/main/java/io/openaev/integration/Integration.java
```diff
@@ -58,18 +62,40 @@ public void initialise() throws Exception {
...
         this.stop();
+        return;
+      }
+      if (isRunning && hasHashChanged) {
+        this.stop();
+        this.start();
+        return;
+      }
+      if (isStartingRequested && isStopped) {
+        this.start();
```
RomuDeuxfois, line 96: **praise**: clear !

openaev-api/src/test/java/io/openaev/integration/IntegrationTest.java
```diff
@@ -0,0 +1,102 @@
...
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.TestExecutionListeners;
+import org.springframework.transaction.annotation.Transactional;
+
+@SpringBootTest
+@Transactional
+@TestExecutionListeners(
+    value = {RabbitMQTestListener.class},
+    mergeMode = TestExecutionListeners.MergeMode.MERGE_WITH_DEFAULTS)
+public class IntegrationTest {
```
RomuDeuxfois: **nitpick:** Can you just change the name of this test ? IntegrationTest is really confuse to me

RomuDeuxfois (approved):

PR #4689, opened 2026-01-06
[backend] feat(catalog): encryption of pwd values in db (#4605)

### LINKED ISSUE #4605: [EE] Built-in integrations: introduce encryption of sensitive values in the database
## Context

Sensitive values (API keys, secrets...) may be part of Integration configuration sets.

## Use case

We need to encrypt these values, in the same way sensitive values are encrypted for the XTM Composer.

## Current Workaround

None

## Proposed Solution

Leverage the work done for encrypting sensitive values for the XTM Composer. Use a platform-wide symmetric cipher key.
Consider storing it in an env var or application.properties (must be persisted; transient cipher keys mean encrypted values can never be decrypted once the platform restarts).

Dimfacion:
### Proposed changes

* Encryption/decryption with an AES 256 bit with GCM key of sensitive data in database
* Small fix to update the config of the executors when stopping/starting the executor

### Testing Instructions

1. In your application.properties (or application-dev.properties), fill the openaev.admin.encryption_key and openaev.admin.encryption_salt properties
2. (Optional) If your catalog_connectors table is not empty, empty it
3. Start your backend
4. Check that the properties with a column set to is_encrypted are encrypted
5. Update the configuration for an encrypted field in the OAEV interface
6. Check that it's still encrypted and different from the previous value
7. Start an executor
8. Check that the encrypted value in database is decrypted in the config in memory

### Related issues

* Closes #4605 
*

### Checklist

### Further comments

In your application.properties (or application-dev.properties), fill the openaev.admin.encryption_key and openaev.admin.encryption_salt properties. The documentation for the encryption library provided by spring is available here : https://docs.spring.io/spring-security/reference/features/integrations/cryptography.html#spring-security-crypto-encryption-text

### DIFF
--- a/openaev-front/src/admin/components/agents/InstructionSelector.tsx
+++ b/openaev-front/src/admin/components/agents/InstructionSelector.tsx
@@ -32,7 +32,7 @@ const InstructionSelector: React.FC<InstructionSelectorProps> = ({ userToken, pl
   const [selectedOption, setSelectedOption] = useState(USER);
   const [agentFolder] = useState<null | string>(null);
   const [arch, setArch] = useState<string>(x86_64);
-  const [calderaSettings, setCalderaSettings] = useState<CalderaSettings[]>(null);
+  const [calderaSettings, setCalderaSettings] = useState<null | CalderaSettings[]>(null);
 
   // Fetching data
   useEffect(() => {


openaev-api/src/main/resources/application.properties
```diff
@@ -21,6 +21,8 @@ openaev.extra-trusted-certs-dir=
 #openaev.admin.email=admin@openaev.io # mandatory
 #openaev.admin.password=ChangeMe # mandatory
 #openaev.admin.token=ChangeMe # mandatory
+#openaev.admin.encryption_key=ChangeMe #mandatory
+#openaev.admin.encryption_salt=ChangeMe #mandatory
```
RomuDeuxfois: question: This is a breaking changes for version upgrade right ?

openaev-api/src/main/java/io/openaev/integration/configuration/BaseIntegrationConfiguration.java
```diff
@@ -20,13 +20,17 @@
...
+@Slf4j
 public class BaseIntegrationConfiguration {
   private final ObjectMapper mapper = new ObjectMapper();
   @Getter @Setter private boolean enable = false;
 
   public static <T extends BaseIntegrationConfiguration> T fromConnectorInstanceConfigurationSet(
-      @NotNull Set<ConnectorInstanceConfiguration> configurations, Class<T> targetClass)
+      @NotNull ConnectorInstance instance,
+      Class<T> targetClass,
+      EncryptionService encryptionService)
```
RomuDeuxfois: It feels odd to pass a service as a method argument.
This breaks the dependency injection model.
We should move this logic into a dedicated service and embed the method there instead.

openaev-api/src/main/java/io/openaev/integration/configuration/BaseIntegrationConfiguration.java
```diff
@@ -20,13 +20,17 @@
...
 import lombok.Getter;
 import lombok.Setter;
+import lombok.extern.slf4j.Slf4j;
 
+@Slf4j
 public class BaseIntegrationConfiguration {
   private final ObjectMapper mapper = new ObjectMapper();
   @Getter @Setter private boolean enable = false;
 
   public static <T extends BaseIntegrationConfiguration> T fromConnectorInstanceConfigurationSet(
```
RomuDeuxfois: This method feels overly complex to me. Could we simplify it by passing the config and only decrypting the relevant keys?

openaev-api/src/main/resources/application.properties
Dimfacion: Yes, I have planned to add this to the doc and this will need to be correctly announced in the release note

openaev-api/src/main/java/io/openaev/integration/configuration/BaseIntegrationConfiguration.java
Dimfacion: Yeah, I do agree that was my initial intention but I had cyclic dependencies. I did not think of moving part of the code to a dedicated utils, I'll try to change that :)

openaev-api/src/main/java/io/openaev/integration/configuration/BaseIntegrationConfiguration.java
Dimfacion: This method do not only do decrypting. It's used to create a Configuration java model from the database model. I guess I can simplify it a bit with another private method but it can't be used only for the relevant keys.

openaev-api/src/main/java/io/openaev/integration/configuration/BaseIntegrationConfiguration.java
Dimfacion: Given how BaseIntegrationConfiguration is used, it seems like that's not doable. Having @Autowired components called from a static method does not work and I don't think we can get rid of the use of static either.

openaev-api/src/main/java/io/openaev/integration/configuration/BaseIntegrationConfiguration.java
Dimfacion: Should be slightly simpler now

openaev-api/src/main/resources/application.properties
MarineLeM: Question : If the encryption_salt changes after restarting the OpenAEV platform, will the values already stored in the database become impossible to decrypt?

Dimfacion: Yes. Same as if you changed the encryption_key. Basically, it's the couple key/salt that is used to generate the 256bit encryption key so if you change one of them, the key used changes too.

openaev-api/src/main/java/io/openaev/integration/configuration/BaseIntegrationConfiguration.java
Dimfacion: I managed to find a way :)

openaev-api/src/main/java/io/openaev/integration/configuration/BaseIntegrationConfiguration.java
```diff
@@ -3,68 +3,104 @@
...
               .filter(c -> c.getKey().equals(field.getAnnotation(IntegrationConfigKey.class).key()))
               .findFirst();
+      Object value = null;
       if (config.isPresent()) {
-        FieldUtils.setField(
-            newObj, field, JsonUtils.fromJsonNode(config.get().getValue(), field.getType()));
-      } else {
-        FieldUtils.setField(newObj, field, null);
+        // If the field is encrypted and can be decrypted
+        if (config.get().isEncrypted() && encryptionService != null) {
```
MarineLeM: Nested if statement make code harder to read and maintain

openaev-api/src/main/java/io/openaev/integration/configuration/BaseIntegrationConfiguration.java
```diff
@@ -3,68 +3,104 @@
...
+
+  public <T extends BaseIntegrationConfiguration> void fromConnectorInstanceConfigurationSet(
+      @NotNull ConnectorInstance instance, Class<T> targetClass) throws JsonProcessingException {
+    EncryptionService encryptionService = null;
+    if (instance instanceof ConnectorInstancePersisted) {
+      encryptionService =
+          encryptionFactory.getEncryptionService(
+              ((ConnectorInstancePersisted) instance).getCatalogConnector());
+    } else {
+      log.warn("The encryption service cannot be instanced. You might want to look into that.");
```
MarineLeM: Suggest: "Cannot instantiate encryption service: connectorInstance is not persisted"

However i think the real issue is : in this part of code, getEncryptionService() should always return NativeEncryptionService, so maybe argument inside the getEncryptionService function should be a boolean isManagerSupported ?

Dimfacion: We would still need to check if the instance is a ConnectorInstancePersisted since it's the only subclass of ConnectorInstance that implements isManagerSupported, no ?

openaev-api/src/main/java/io/openaev/integration/Integration.java
```diff
@@ -17,6 +18,7 @@ public abstract class Integration {
   private final ComponentRequestEngine componentRequestEngine;
   @Getter private ConnectorInstance connectorInstance;
   private final ConnectorInstanceService connectorInstanceService;
+  protected EncryptionService encryptionService;
```
MarineLeM: Why is this service set in this class? It doesn't seem to be used.

openaev-api/src/main/java/io/openaev/integration/IntegrationFactory.java
```diff
@@ -12,8 +13,9 @@
 
 @RequiredArgsConstructor
 public abstract class IntegrationFactory {
-  private final ConnectorInstanceService connectorInstanceService;
-  private final CatalogConnectorService catalogConnectorService;
+  protected final ConnectorInstanceService connectorInstanceService;
+  protected final CatalogConnectorService catalogConnectorService;
+  protected final HttpClientFactory httpClientFactory;
```
MarineLeM, line 18: Same question, Why is this service set in this class? It doesn't seem to be used.

openaev-api/src/main/java/io/openaev/integration/Integration.java
Dimfacion: Good call, it's not used anymore since we get it from the autowired EncryptionFactory. I'm removing it

openaev-api/src/main/java/io/openaev/integration/IntegrationFactory.java
Dimfacion: It's used by the various executor client (CrowdStrikeExecutorClient, CalderaExecutorClient, SentinelOneExecutorClient, ...).

openaev-api/src/main/java/io/openaev/service/CalderaSettingsService.java
```diff
@@ -0,0 +1,70 @@
...
+  @Resource protected ObjectMapper mapper;
+
+  /**
+   * Get Caldera settings
+   *
+   * @return caldera settings
+   */
+  public List<CalderaSettings> getCalderaSettings() {
+    return StreamSupport.stream(executorService.executors().spliterator(), false)
+        .filter(executor -> "openaev_caldera".equals(executor.getType()))
```
MarineLeM:
```suggestion
        .filter(executor -> CalderaContract.TYPE.equals(executor.getType()))
```

openaev-api/src/main/java/io/openaev/service/CalderaSettingsService.java
```diff
@@ -0,0 +1,70 @@
...
+              ConnectorInstance connectorInstance =
+                  connectorInstanceService.connectorInstanceById(
+                      connectorIds.getConnectorInstanceId());
+
+              // Getting the url in the configurations of the instance
+              String calderaPublicUrl =
+                  connectorInstance.getConfigurations().stream()
+                      .filter(
+                          configuration ->
+                              "EXECUTOR_CALDERA_PUBLIC_URL".equals(configuration.getKey()))
```
MarineLeM: can we avoid hardcoding "EXECUTOR_CALDERA_PUBLIC_URL" ?

openaev-front/src/admin/components/agents/InstructionSelector.tsx
```diff
@@ -31,6 +36,15 @@ const InstructionSelector: React.FC<InstructionSelectorProps> = ({ userToken, pl
   const [selectedOption, setSelectedOption] = useState(USER);
   const [agentFolder] = useState<null | string>(null);
   const [arch, setArch] = useState<string>(x86_64);
+  const dispatch = useAppDispatch();
+
+  // Fetching data
+  const { calderaSettings } = useHelper((helper: LoggedHelper) => ({ calderaSettings: helper.getCalderaSettings() }));
```
MarineLeM: I don't see why we need to add the Caldera settings to the React store ?

Issue #4680, opened 2026-01-06
[bug] Built-in integrations manager first initialisation may be too late


antoinemzs:
## Description

The built-in integrations manager is prodded via a scheduled job every few seconds, to start or stop integrations as needed.

When the plaform first starts, there might be a case where the first run of this scheduled job occurs after the inject execution job, and when this happens, injects may be wrongly put in error because no fulfilment injector is found.

PR #4696, opened 2026-01-08
[backend] fix(manager-factory): full initialization at construction and disallow concurrent calls (#4680)

RomuDeuxfois:
### Proposed changes

*
*

### Testing Instructions

1. Step-by-step how to test

### Related issues

* Closes https://github.com/OpenAEV-Platform/openaev/issues/4680
*

### Checklist

### Further comments

### DIFF
--- a/openaev-api/src/main/java/io/openaev/integration/ManagerFactory.java
+++ b/openaev-api/src/main/java/io/openaev/integration/ManagerFactory.java
@@ -12,22 +12,15 @@ import org.springframework.transaction.annotation.Transactional;
 @RequiredArgsConstructor
 public class ManagerFactory {
   private final List<IntegrationFactory> factories;
-  private Manager manager = null;
-
-  public Manager getManager() {
-    Manager local = manager;
-    if (local == null) {
-      throw new IllegalStateException("Manager not initialized yet");
-    }
-    return local;
-  }
+  private volatile Manager manager = null;
 
   @Transactional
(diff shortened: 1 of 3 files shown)


openaev-api/src/main/java/io/openaev/integration/ManagerFactory.java
```diff
@@ -15,6 +15,7 @@ public class ManagerFactory {
   public Manager getManager() throws Exception {
     if (this.manager == null) {
       this.manager = new Manager(factories);
+      this.manager.monitorIntegrations();
     }
     return this.manager;
```
antoinemzs, line 24: Possible race condition on `getManager()`, so there should be a lock here too like we discussed.

antoinemzs (approved):
Approving to expedite, but solving the race condition on `getManager()` is a must do. Thanks!

openaev-api/src/main/java/io/openaev/integration/ManagerFactory.java
RomuDeuxfois: Yes, after thinking it through and looking at how the Manager is accessed, it felt odd to duplicate logic instead of clearly separating responsibilities.

I’ve updated the design and introduced a clearer approach, let me know what you think

The manager field is initialized only via initializeAndSync(), which is protected by a distributed lock.
The getManager() method is read-only and never performs initialization.

monitorIntegrations() is intended to be called exclusively by ManagerFactory.

Issue #4635, opened 2025-12-22
Inject failed if the name of the expectation is too long


damgouj:
## Description

Inject failed if we create one with a expectation's name longer than 255 characters, we need to block this.

## Environment

OpenAEV version: 2.0.X

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Create a simulation
2. Create a mail inject in it with a manual expectation with a name longer than 255 characters
3. Launch the simulation, the inject is executed but with an error and with the PARTIAL status

## Expected Output

The inject is executed with no error.

## Actual Output

The inject is executed with a SQL error.

## Additional information

The inject expectation is not inserted with this bug so the simulation never ends and the user can't validate his expectation.

PR #4723, opened 2026-01-12
[frontend/backend] feat(simulation): fix inject name longer (#4635)

GaetanSantucci:
Inject failed if we create one with a expectation's name longer than 255 characters, we need to block this.

### Proposed changes

Implementation of double front-end and back-end validation to prevent entries exceeding 255 characters in the input name field

### Testing Instructions

Create a simulation
Create a mail inject in it with a manual expectation with a name longer than 255 characters
Launch the simulation, the inject is executed but with an error and with the PARTIAL status

### Related issues

* #4635

### Further comments

The remaining challenge will be to better manage the backend return error for a display that's better than a 500 error, but that will require a comprehensive platform-wide effort.
We initially decided to prevent the submission of strings longer than 255 characters to resolve the issue we were experiencing. However, after investigation and further consideration, a more permanent solution is needed. Most columns in PostgreSQL are varchar(255), but nothing on the front end indicates this limit, and error handling is not yet correct in all forms. Significant work is required to ensure consistent input and error handling that aligns with our database.

### DIFF
--- a/openaev-front/src/admin/components/simulations/simulation/ExerciseForm.tsx
+++ b/openaev-front/src/admin/components/simulations/simulation/ExerciseForm.tsx
@@ -56,9 +56,8 @@ const ExerciseForm: FunctionComponent<Props> = ({
     mode: 'onTouched',
     resolver: zodResolver(
       zodImplement<CreateExerciseInput>().with({
-        exercise_name: z.string().min(EXERCISE_NAME_MIN_LENGTH, { message: t('Should not be empty') }).max(EXERCISE_NAME_MAX_LENGTH, { message:t('Should not exceed {max_length} characters', {
-						max_length: EXERCISE_NAME_MAX_LENGTH.toString(),
-					}) }),
+        exercise_name: z.string().min(EXERCISE_NAME_MIN_LENGTH, { message: t('Should not be empty') })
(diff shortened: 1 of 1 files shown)


openaev-front/src/admin/components/simulations/simulation/ExerciseForm.tsx
```diff
@@ -55,7 +55,7 @@ const ExerciseForm: FunctionComponent<Props> = ({
     mode: 'onTouched',
     resolver: zodResolver(
       zodImplement<CreateExerciseInput>().with({
-        exercise_name: z.string().min(1, { message: t('Should not be empty') }),
+        exercise_name: z.string().min(1, { message: t('Should not be empty') }).max(255, { message: t('Should not exceed 255 characters') }),
```
antoinemzs: As with all magic numbers, wrapping it in a named constant and use it everywhere is a good practice.

antoinemzs: Could you interpolate the value 255 in the error message, as well as using it in the value field ?

openaev-api/src/main/java/io/openaev/rest/exercise/form/ExerciseInput.java
```diff
@@ -19,6 +19,7 @@
 public class ExerciseInput {
 
   @NotBlank(message = MANDATORY_MESSAGE)
+  @Size(max = 255, message = MAX_255_MESSAGE)
```
antoinemzs: Make constant and consider interpolation

PR #4729, opened 2026-01-13
[backend/frontend] feat(inject-expectations): add Injects Expectations from agents into endpoint tabs - chunk 1 (#4084)

### LINKED ISSUE #3593: Problem when update inject expectation
## Description
Updating an inject expectation on an endpoint causes it to duplicate on the agent side.

## Environment
1. OpenBAS version: 1.17.3

## Reproducible Steps

For sample go on this inject overview : 
1. Create a new prevention result on an endpoint target.
2. Update this result.

Image

3. Go to the agent target.
4. Check the result: The updated result is duplicated on the agent target

Image

## Expected Output

The result should not be duplicated when updated on the endpoint.

### LINKED ISSUE #4084: Add Injects Expectations from agents into endpoint tabs
Today when we run injects through asset like an endpoint with one ore more agents we can't see neither the number of alerts nor the links of alerts as we can't click on it.

We would like to avoid each time to switch betwenn tabs endpoints/agents

For now:

Image

What we would like:

Image

And if we click on it:

Image

damgouj:
### Proposed changes

* Tab endpoints: Rework the injects expectations for endpoints to put the agents injects expectations aggregated + remove the "validation rule" text
* Tab asset groups: remove the injects expectations list
* Tab agents: remove the shield "add a security platform"

### Testing Instructions

1. Create human and technical injects on atomic testings/exercises/scenarios
2. Check the injects expectations and add/update/delete them

### Related issues

* Closes #3593 
* Closes #4084 

### Checklist

### DIFF
--- a/openaev-front/src/admin/components/atomic_testings/atomic_testing/AtomicTesting.tsx
+++ b/openaev-front/src/admin/components/atomic_testings/atomic_testing/AtomicTesting.tsx
@@ -9,10 +9,10 @@ import Empty from '../../../../components/Empty';
 import { useFormatter } from '../../../../components/i18n';
 import Loader from '../../../../components/Loader';
 import { type InjectTarget, type SearchPaginationInput } from '../../../../utils/api-types';
+import { isAgentless } from '../../../../utils/target/TargetUtils';
 import { InjectResultOverviewOutputContext, type InjectResultOverviewOutputContextType } from '../InjectResultOverviewOutputContext';
 import PaginatedTargetTab from './PaginatedTargetTab';
 import TargetResultsDetail from './target_result/TargetResultsDetail';
-import {isAgentless} from "../../../../utils/target/TargetUtils";
 
 const useStyles = makeStyles()({
   chip: {


openaev-front/src/admin/components/atomic_testings/atomic_testing/AtomicTesting.tsx
```diff
@@ -282,7 +282,7 @@ const AtomicTesting = () => {
           {t('Results by target')}
         </Typography>
         {selectedTarget && !!injectResultOverviewOutput.inject_type && (
-          <TargetResultsDetail inject={injectResultOverviewOutput} target={selectedTarget} />
+          <TargetResultsDetail inject={injectResultOverviewOutput} target={selectedTarget} isAgentless={!hasAgents && !hasTeams} />
```
RomuDeuxfois: **nitpick**: This logic should be moved to a utils file since it’s internal logic and could be reused elsewhere.
!hasAgents && !hasTeams

openaev-front/src/actions/atomic_testings/atomic-testing-actions.ts
```diff
@@ -54,6 +54,11 @@ export const fetchTargetResult = (injectId: string, targetId: string, targetType
   return simpleCall(uri);
 };
 
+export const fetchTargetResultAssetWithAgents = (injectId: string, targetId: string, expectationType: string) => (dispatch: Dispatch) => {
+  const uri = `${ATOMIC_TESTING_URI}/${injectId}/target_results/${targetId}/asset_with_agents?expectationType=${expectationType}`;
+  return getReferential(schema.arrayOfInjectexpectations, uri)(dispatch);
```
RomuDeuxfois, line 59: **question**: Since you’re not returning the actual InjectExpectation object but a new DTO, I’m not sure it’s a good idea to use the store key that references the full object here.

The DTO is very close to the original object, but it introduces an additional property. If, in another file, the store tries to retrieve the InjectExpectationId while already holding this DTO, some properties may be null, which could lead to a NullPointerException.

I don’t remember if we’ve already addressed this kind of situation, but I think it’s something we should discuss with our Staff.

openaev-front/src/admin/components/atomic_testings/atomic_testing/context/InjectExpectationContext.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { type InjectExpectationResult } from '../../../../../utils/api-types';
+import type { InjectExpectationsStore } from '../../../common/injects/expectations/Expectation';
+
+type InjectExpectationContextType = {
+  onOpenDeleteInjectExpectationResult: (result: InjectExpectationResult | null, injectExpectationStore: InjectExpectationsStore | null) => void;
+  onOpenEditInjectExpectationResultResult: (result: InjectExpectationResult | null, injectExpectationStore: InjectExpectationsStore | null) => void;
+  onOpenSecurityPlatform: (result: InjectExpectationResult | null) => void;
+};
+
+const InjectExpectationContext = createContext<InjectExpectationContextType>({
```
RomuDeuxfois, line 12: **Question**: Just to be sure, are you creating this context to avoid drilling down through methods, or to avoid duplicating logic?
If it’s mainly to handle duplicated logic, I think it might be better to introduce a dedicated component instead (ButtonPopoverForThisCase).

openaev-front/src/actions/atomic_testings/atomic-testing-actions.ts
guillaumejparis: I spoke with Damien about it, and yes, it is a current problem with our DTO API and our store system. Later, we will move to a different system—API side, front side, or both—and at that moment, this kind of thing will be fixed. As this is a very low risk, I think we can continue to do it like that for now.

openaev-front/src/admin/components/atomic_testings/atomic_testing/context/InjectExpectationContext.tsx
damgouj: Yes it was to avoid the props drilling (down and up) :)

PR #4732, opened 2026-01-14
[frontend] feat(scv): improve to classify payload and contract management (#4119)

GaetanSantucci:
Quick update to add method in payloads update to check and remove To classify domains if another domains is linked to a payload

### Testing Instructions

1. Update a payload who has a "To classify" domains
2. Set another domain, this should remove the chips to classify

### Related issues

* #4119

### DIFF
--- a/openaev-front/src/components/DomainsAutocompleteField.tsx
+++ b/openaev-front/src/components/DomainsAutocompleteField.tsx
@@ -6,8 +6,8 @@ import AutocompleteField from './fields/AutocompleteField';
 
 type DomainValue = Domain | string;
 interface Props {
-    input: FieldRenderProps<DomainValue[]>['input'];
-    meta: FieldRenderProps<DomainValue[]>['meta'];
+  input: FieldRenderProps<DomainValue[]>['input'];
+  meta: FieldRenderProps<DomainValue[]>['meta'];
   domainOptions: Domain[];
   label: string;
   disabled?: boolean;


openaev-front/src/admin/components/integrations/injectors/injector_contracts/InjectorContractCustomForm.js
```diff
@@ -126,44 +125,61 @@ const InjectorContractForm = (props) => {
...
+
+                  const optionsToDisplay = hasToClassifySelected && toClassifyDomain
+                    ? [...filteredDomains, toClassifyDomain]
+                    : filteredDomains;
+
+                  const mappedOptions = optionsToDisplay.map(d => ({
+                    id: d.domain_id,
+                    label: d.domain_name,
+                  }));
+
```
gabriel-peze: is it possible to remove this js code in a method and call for this method, instead of doing it like that ?

here change it to `const { mappedOptions , currentIds } = method(...);`

```
const method = (...) => {
...
return { mappedOptions, currentIds };
```

openaev-front/src/admin/components/integrations/injectors/injector_contracts/InjectorContractForm.js
```diff
@@ -52,44 +52,56 @@ const InjectorContractForm = (props) => {
...
+                  if (d.domain_name === 'To classify') {
+                    return hasToClassifySelected;
+                  }
+                  return true;
+                });
+
+                const mappedOptions = optionsToDisplay.map(d => ({
                   id: d.domain_id,
                   label: d.domain_name,
                 }));
```
gabriel-peze: same here

openaev-front/src/admin/components/payloads/PayloadPopover.js
```diff
@@ -22,43 +21,50 @@ const PayloadPopover = ({ payload, onUpdate, onDelete, onDuplicate, disableUpdat
...
-        author_rule: value[1].author_rule,
-      }))),
-    )(data);
+    };
+
+    const inputValues = {
+      ...data,
+      payload_platforms: data.payload_platforms,
+      payload_tags: data.payload_tags,
+      payload_attack_patterns: data.payload_attack_patterns,
```
gabriel-peze: to test, but i don't think those lines are usefull here with the `...data` before.

...data already set all the attribute to the inputValues object, and after we overwrite the value payload_platforms of inputValues, who came from data, with the same value from data.

I think we can delete here those lines :
```
payload_platforms: data.payload_platforms,
payload_tags: data.payload_tags,
payload_attack_patterns: data.payload_attack_patterns,
executable_file: data.executable_file,
```

openaev-front/src/components/fields/DomainFieldController.tsx
```diff
@@ -27,41 +29,65 @@ const DomainFieldController = ({
...
+
+        const optionsToDisplay = [...filteredDomains];
+        if (hasToClassifySelected && toClassifyDomain) {
+          optionsToDisplay.push(toClassifyDomain);
+        }
+
+        const mappedOptions = optionsToDisplay.map(d => ({
+          id: d.domain_id,
+          label: d.domain_name,
+        }));
```
gabriel-peze: same here, maybe check if the methods can be mutualize

gabriel-peze (approved):

PR #4740, opened 2026-01-15
[frontend/backend] fix(filters): fix filter operators (#4283)

johanah29:
Try apply filters on domains in scenario/injects, simulation/injects, atomic testing and payloads pages.

### DIFF
--- a/openaev-api/src/main/java/io/openaev/rest/atomic_testing/form/InjectResultOutput.java
+++ b/openaev-api/src/main/java/io/openaev/rest/atomic_testing/form/InjectResultOutput.java
@@ -62,8 +62,6 @@ public class InjectResultOutput {
   @JsonProperty("inject_contract_domains")
   @Schema(description = "Domain of the inject")
   public String[] getDomains() {
-    return injectorContract != null && injectorContract.getPayload() != null
-        ? injectorContract.getPayload().getDomains()
-        : injectorContract.getDomains();
+    return injectorContract.getDomains();
   }
 }
--- a/openaev-api/src/main/java/io/openaev/service/InjectSearchService.java
+++ b/openaev-api/src/main/java/io/openaev/service/InjectSearchService.java
@@ -214,15 +214,6 @@ public class InjectSearchService {
         .toList();
   }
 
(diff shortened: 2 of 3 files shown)


openaev-api/src/main/java/io/openaev/rest/atomic_testing/form/InjectResultOutput.java
```diff
@@ -58,4 +58,12 @@ public class InjectResultOutput {
 
   @JsonProperty("inject_targets")
   private List<TargetSimple> targets = new ArrayList<>();
+
+  @JsonProperty("inject_contract_domains")
+  @Schema(description = "Domain of the inject")
+  public String[] getDomains() {
```
RomuDeuxfois, line 64: **Nitpick:** As is already done in `InjectorContract.java`, this logic should live in `InjectorContractSimple`. At this level, you should only need to call `injectContract.getDomains()`.

In `InjectorContractSimple`, you could implement something like:
```java
public String[] getDomains() {
  return injectorContract != null && injectorContract.getPayload() != null
      ? injectorContract.getPayload().getDomains()
      : injectorContract.getDomains();
}
```

openaev-api/src/main/java/io/openaev/service/InjectSearchService.java
```diff
@@ -204,6 +214,15 @@ private List<InjectOutput> execInject(TypedQuery<Tuple> query) {
         .toList();
   }
 
+  private List<String> resolveEffectiveDomains(String[] injectorDomains, String[] payloadDomains) {
```
RomuDeuxfois: Seems to be not used.

openaev-front/src/admin/components/atomic_testings/InjectResultList.tsx
```diff
@@ -99,12 +102,24 @@ const InjectResultList: FunctionComponent<Props> = ({
     },
     {
       field: 'inject_title',
-      label: 'Title',
+      label: 'Name',
```
RomuDeuxfois, line 105: **question**: Why changing this ?

johanah29: It was a request from product team

PR #4742, opened 2026-01-16
[backend] fix(domain): upserts

RomuDeuxfois:


openaev-api/src/main/java/io/openaev/rest/domain/DomainService.java
```diff
@@ -77,14 +76,26 @@ public Domain upsert(final Domain domainToUpsert) {
   }
 
   @Transactional
-  public Set<Domain> upserts(final Set<Domain> domains) {
-    if (domains == null) {
-      return new HashSet<>();
+  public Set<Domain> upserts(Set<Domain> domains) {
+    if (domains == null || domains.isEmpty()) {
+      return Set.of();
```
antoinemzs: Please keep the collection mutable: hibernate below 7.0 does not support immutables while persisting

openaev-api/src/main/java/io/openaev/rest/domain/DomainService.java
```diff
@@ -77,14 +76,26 @@ public Domain upsert(final Domain domainToUpsert) {
...
     return domains.stream()
-        .map(domain -> this.upsert(domain.getName(), domain.getColor()))
-        .collect(Collectors.toSet());
+        .map(
+            d ->
+                existing.computeIfAbsent(
+                    d.getName(),
+                    name ->
+                        domainRepository.save(
+                            new Domain(null, name, d.getColor(), Instant.now(), null))))
```
antoinemzs: the singular `upsert(String, String)` null-checks the `color` argument (3rd), it would be prudent to do the same in the plural upserts(Set<Domain>) version.

antoinemzs: Unrelated, but specifying values for the `createDate` and `updateDate` arguments is not useful due to the annotations on the class attributes which will superesede the provided values anyway.

nitpick: there can be an explicit ctor if it's useful, but not with providing unusable arguments which will create confusion. We can see that the `id`, `createDate`, `updateDate` are never set, or set with a default value, in the ctor, for good reason.

No need to fix this nitpick now IMO.

Image

antoinemzs (approved):
Approving to expedite, two small fixes before merge please. Thanks!

openaev-api/src/main/java/io/openaev/rest/domain/DomainService.java
RomuDeuxfois: Well seen.

openaev-api/src/main/java/io/openaev/rest/domain/DomainService.java
RomuDeuxfois: Well seen.

RomuDeuxfois: I saw that, but don't want to adress it now.

antoinemzs (commented):
lgtm thanks

PR #4755, opened 2026-01-16
[backend/frontend] chore(merge): merge back 20260116 (#1)

antoinemzs:
### Proposed changes

*
*

### Testing Instructions

1. Step-by-step how to test

### Related issues

* Closes #ISSUE-NUMBER
*

### Checklist

### Further comments

### DIFF
--- a/openaev-api/src/main/java/io/openaev/migration/V4_56__Implement_Domains_notion.java
+++ b/openaev-api/src/main/java/io/openaev/migration/V4_56__Implement_Domains_notion.java
@@ -117,7 +117,7 @@ public class V4_56__Implement_Domains_notion extends BaseJavaMigration {
               + "') ON CONFLICT (domain_name) DO NOTHING;");
 
       stmt.execute(
-            """
+          """
             WITH unknown_payloads AS (
               SELECT p.payload_id
               FROM payloads p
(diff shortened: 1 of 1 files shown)


openaev-model/src/main/java/io/openaev/healthcheck/enums/ExternalServiceDependency.java
```diff
@@ -33,7 +33,15 @@ public enum ExternalServiceDependency {
 
   /** Nmap network scanner service. */
   @JsonProperty("NMAP")
-  NMAP("openaev_nmap");
+  NMAP("openaev_nmap"),
+
+  /** OpenAEV Implant service. */
+  @JsonProperty("OpenAEV Email")
+  OPENAEV_EMAIL("openaev_email"),
```
MarineLeM, line 40: Currently in the database, the external dependencies for openaev_email are SMTP and IMAP. I'm not sure if the external dependencies field is actually used, but should we keep the same values as before?

openaev-model/src/main/java/io/openaev/healthcheck/enums/ExternalServiceDependency.java
```diff
@@ -33,7 +33,15 @@ public enum ExternalServiceDependency {
...
+
+  /** OpenAEV Implant service. */
+  @JsonProperty("OpenAEV Implant")
+  OPENAEV_IMPLANT("openaev_implant");
```
MarineLeM, line 44: same question that above, currently there is no external dependencies for openaev_implant

openaev-model/src/main/java/io/openaev/healthcheck/enums/ExternalServiceDependency.java
antoinemzs: Yes you are right, however fixing this requires a broader code change, and may also involve changing the contents of the starter pack zip files, which I don't want to be in the scope of this merge.

Adding these two values allows for a successful import.

MarineLeM: i don't get it why you said need to change the contents of starter pack, you need to let injector_type equals openaev_email but the externalServiceDependency must return SMTP and IMAP

PR #4786, opened 2026-01-23
[backend/frontend] fix(catalog): fix openaev implant and remove expose property (#4680)

RomuDeuxfois:


openaev-api/src/main/java/io/openaev/integration/impl/injectors/openaev/OpenaevInjectorIntegration.java
```diff
@@ -75,6 +91,9 @@ protected void innerStart() throws Exception {
         executorClearCommands,
         true,
         List.of());
+    this.openAEVImplantExecutor =
```
antoinemzs, line 94: @Dimfacion to answer your question: this was never assigned, so it wasn't possible to request a functioning component from the manager

PR #4795, opened 2026-01-26
[backend] fix(catalog): remove injector getBean (#4606)

MarineLeM:
### Proposed changes

* remove get bean

### DIFF
--- a/openaev-api/src/main/java/io/openaev/utils/InjectUtils.java
+++ b/openaev-api/src/main/java/io/openaev/utils/InjectUtils.java
@@ -73,12 +73,12 @@ public class InjectUtils {
       return inject.getStatus().get().getPayloadOutput();
     }
 
-    if (inject.getInjectorContract().isEmpty()
-        || inject.getInjectorContract().get().getPayload() == null) {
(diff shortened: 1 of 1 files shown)


openaev-api/src/main/java/io/openaev/utils/InjectUtils.java
```diff
@@ -75,139 +71,127 @@ public StatusPayload getStatusPayloadFromInject(final Inject inject) {
...
-            List.of(payloadCommandBlock),
-            payloadCommand.getCleanupExecutor());
+    }
+
+    if (inject.getInjectorContract().isEmpty()
+        || inject.getInjectorContract().get().getPayload() == null) {
+      return null;
+    }
+
+    InjectorContract injectorContract = inject.getInjectorContract().get();
```
Dimfacion: **question :** Can't getInjectorContract() return an empty optional ? Shouldn't we return a null ?

Dimfacion: Sorry, I'm a dum dum :p

PR #4813, opened 2026-01-28
[backend] fix(catalog): do not break init when single integration throws and exception (#1)

antoinemzs:
### Proposed changes

* Do not rethrow caught exception during batched init of integration in case there is one that throws an exception

### Testing Instructions

1. Force an exception during init of single integration
2. All other integrations can be interacted with

### DIFF
--- a/openaev-api/src/main/java/io/openaev/integration/impl/executors/caldera/CalderaExecutorIntegration.java
+++ b/openaev-api/src/main/java/io/openaev/integration/impl/executors/caldera/CalderaExecutorIntegration.java
@@ -144,5 +144,6 @@ public class CalderaExecutorIntegration extends Integration {
   protected void innerStop() {
     this.platformSettingsService.cleanMessage(BannerMessage.BANNER_KEYS.CALDERA_UNAVAILABLE);
     timers.forEach(timer -> timer.cancel(true));
+    timers.clear();
   }
 }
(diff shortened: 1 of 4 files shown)


openaev-api/src/main/java/io/openaev/integration/IntegrationFactory.java
```diff
@@ -33,15 +35,22 @@ public void initialise() throws Exception {
     runMigrations();
   }
 
-  @Transactional(rollbackFor = Exception.class)
   public List<Integration> sync(List<ConnectorInstance> instances) throws Exception {
```
MarineLeM: Since we do not rethrow any exception in the function, we do not need throws Exception here.

openaev-api/src/main/java/io/openaev/integration/Manager.java
```diff
@@ -79,40 +78,34 @@ public io.openaev.executors.Injector requestInjectorExecutorByType(String inject
...
-                            .toList();
-                    return factory.sync(newInstances).stream();
-                  } catch (Exception e) {
-                    log.error("There was a problem syncing integration factories.", e);
-                    throw new RuntimeException(e);
-                  }
-                })
-            .map(integration -> Map.entry(integration.getConnectorInstance(), integration))
-            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+  public void monitorIntegrations() throws Exception {
```
MarineLeM: same here

PR #4816, opened 2026-01-28
[backend] fix(executors): change slugs for openaev_crowdstrike executor and openaev_sentinelone executor (#1)

antoinemzs:
### Proposed changes

* Change "executor types" for crowdstrike, sentinel one


camrrx:
Test : NOK with a db which already has a crowdstrike executor

PR #4837, opened 2026-01-30
[backend] fix(catalog): don't collate log lines from xtm composer (#1)

antoinemzs:
### Proposed changes

* Do not collate log lines from xtm composer

Potential known issue: logs appear potentially out of order

### Testing Instructions

1. Step-by-step how to test

### Related issues

* Closes #ISSUE-NUMBER
*

### Checklist

### Further comments

### DIFF
--- a/openaev-api/src/main/java/io/openaev/service/connector_instances/ConnectorInstanceLogService.java
+++ b/openaev-api/src/main/java/io/openaev/service/connector_instances/ConnectorInstanceLogService.java
@@ -36,37 +36,36 @@ public class ConnectorInstanceLogService {
    * @param rawLogLines the set of raw log lines to transform
    * @return the formatted log string with lines separated by newlines
    */
-  public String transformRawLogsLineToLog(Set<String> rawLogLines) {
+  public Set<String> transformRawLogsLineToLog(Set<String> rawLogLines) {
     return rawLogLines.stream()
         .map(line -> line.replaceAll("^,", ""))
         .map(String::trim)
         .filter(line -> !line.isEmpty())
-        .collect(Collectors.joining("\n"));
+        .collect(Collectors.toSet());
   }
 
   /**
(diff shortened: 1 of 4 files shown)


openaev-api/src/main/java/io/openaev/service/connector_instances/ConnectorInstanceLogService.java
```diff
@@ -36,37 +36,36 @@ private void cleanupExcessLogs(String connectorInstanceId) {
...
+      ConnectorInstancePersisted connectorInstance, Set<String> rawLogs)
+      throws IllegalArgumentException {
+    if (rawLogs.isEmpty()) {
+      return;
+    }
+    for (String log : rawLogs) {
+      ConnectorInstanceLog logEntry = new ConnectorInstanceLog();
+      logEntry.setConnectorInstance(connectorInstance);
+      logEntry.setLog(log);
+      connectorInstanceLogRepository.save(logEntry);
```
RomuDeuxfois, line 65: **nitpick**: can we use saveAll here ?

antoinemzs: Given we are already in a transaction (`@Transactional` annotation on the method), I don't think there is a benefit for using `saveAll()`: it would make us loop through the list twice instead of once.

Here's the saveAll() source code:
```java
@Override
@Transactional
public <S extends T> List<S> saveAll(Iterable<S> entities) {

	Assert.notNull(entities, ENTITIES_MUST_NOT_BE_NULL);

	List<S> result = new ArrayList<>();

	for (S entity : entities) {
		result.add(save(entity));
	}

	return result;
}
```

RomuDeuxfois: That's true, ok for me then

PR #4841, opened 2026-02-02
[backend] feat(wf-engine): add datamodel and queue definitions (#4824)

impolitepanda:
### Proposed changes

* Add model for Workflow engine

### Testing Instructions

1. Check creation queues is correct
2. Check model is coherent (index, fk, pf..)

### Related issues

* Closes #3316 
* Closes #4824 

### Checklist

- [x] I consider the submitted work as finished
- [x] I tested the code for its functionality
- [x] I wrote test cases for the relevant uses case
- [ ] I added/update the relevant documentation (either on github or on notion)
- [x] Where necessary I refactored code to improve the overall quality
- [ ] For bug fix -> I implemented a test that covers the bug

### Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

Issue #4824, opened 2026-01-30
[Chaining] Basic Chaining Engine  


mergenhan:
Teams can run realistic simulations quickly and often without the overhead of a full pentest/red teaming.

Dod:

- OpenAev backend has a workflow engine that is able to support inject chaining 
- All new feature are feature flagged

Issue #3316, opened 2025-06-09
[EE] Include the notion of multiple executions in openbas code base + data model


savacano28:
## Context

We were able to generate structured output through output parsers or natively through some injectors like nmap. 
Now we want for our user to use these structured output as argument of my next inject. 

## Use case

We need to be able to trace multiples executions (n executions depending on n input (output from previous inject)

## Current Workaround

Generate n injects

## Proposed Solution

Evoluate data model : Inject 1 ->1 Injectstatus => inject 1 -> n Injectstatus (executions)

## If the feature request is approved, would you be willing to submit a PR?

Yes

savacano28:
@heditar @mergenhan I think this issue is a duplicated of basic chanining https://github.com/OpenAEV-Platform/openaev/issues/4824 or deprecated.

PR #4842, opened 2026-02-02
[backend] feat(wf-engine): init split service layer (#4824)

impolitepanda:
### Proposed changes

* Adds the service layer of the POC on the new engine
* Adds support for the new feature flag
* Adds unit test covering 90% of new devs and some on the existing batching queue services

### Testing Instructions

Not testable

### Related issues

* Relates to #3316 
* Relates to #4824 
* Continues work from PR #4841 

### Checklist

### DIFF
--- a/openaev-api/src/main/java/io/openaev/rest/exercise/service/ExerciseService.java
+++ b/openaev-api/src/main/java/io/openaev/rest/exercise/service/ExerciseService.java
@@ -34,6 +34,7 @@ import io.openaev.rest.inject.form.InjectExpectationResultsByAttackPattern;
 import io.openaev.rest.inject.service.InjectDuplicateService;
 import io.openaev.rest.inject.service.InjectService;
 import io.openaev.rest.scenario.service.ScenarioStatisticService;
+import io.openaev.rest.settings.PreviewFeature;
 import io.openaev.rest.team.output.TeamOutput;
 import io.openaev.service.*;
 import io.openaev.service.chaining.WorkflowService;
@@ -111,6 +112,7 @@ public class ExerciseService {
   private final ScenarioRecurrenceService scenarioRecurrenceService;
 
   private final WorkflowService workflowService;
+  private final PreviewFeatureService previewFeatureService;
 
   // region properties
   @Value("${openaev.mail.imap.enabled}")
@@ -152,7 +154,7 @@ public class ExerciseService {
   @Transactional(rollbackFor = Exception.class)
   public Exercise createSimulation(@NotNull final Exercise simulation, boolean isChaining) {
     Exercise savedSimulation = createExercise(simulation);
-    if (isChaining) {
+    if (isChaining && previewFeatureService.isFeatureEnabled(PreviewFeature.INJECT_CHAINING)) {
       workflowService.creationWorkflow(savedSimulation);
     }
     return savedSimulation;
--- a/openaev-api/src/test/java/io/openaev/rest/exercise/service/ExerciseServiceUnitTest.java
+++ b/openaev-api/src/test/java/io/openaev/rest/exercise/service/ExerciseServiceUnitTest.java
@@ -15,6 +15,7 @@ import io.openaev.rest.exception.ElementNotFoundException;
 import io.openaev.rest.exercise.form.ExercisesGlobalScoresInput;
 import io.openaev.rest.inject.service.InjectDuplicateService;
 import io.openaev.rest.inject.service.InjectService;
+import io.openaev.rest.settings.PreviewFeature;
 import io.openaev.service.*;
 import io.openaev.service.chaining.WorkflowService;
 import io.openaev.service.period.CronService;
@@ -75,6 +76,7 @@ class ExerciseServiceUnitTest {
 
   @Mock private WorkflowService workflowService;
   @Mock private LessonsService lessonsService;
+  @Mock private PreviewFeatureService previewFeatureService;
 
   @Spy @InjectMocks private ExerciseService mockedExerciseService;
 
@@ -333,7 +335,8 @@ class ExerciseServiceUnitTest {
 
       Exercise saved = mock(Exercise.class);
       doReturn(saved).when(mockedExerciseService).createExercise(exercise);
-
+      if (chaining)
+        doReturn(true).when(previewFeatureService).isFeatureEnabled(PreviewFeature.INJECT_CHAINING);
       Exercise result = mockedExerciseService.createSimulation(exercise, chaining);
 
       assertEquals(saved, result);
--- a/openaev-api/src/test/java/io/openaev/service/ExerciseServiceIntegrationTest.java
+++ b/openaev-api/src/test/java/io/openaev/service/ExerciseServiceIntegrationTest.java
@@ -48,7 +48,7 @@ class ExerciseServiceIntegrationTest extends IntegrationTest {
   @Mock EnterpriseEditionService enterpriseEditionService;
   @Mock InjectDuplicateService injectDuplicateService;
   @Mock VariableService variableService;
-
+  @Mock private PreviewFeatureService previewFeatureService;
   @Autowired private TeamService teamService;
   @Autowired private TagRuleService tagRuleService;
   @Autowired private DocumentService documentService;
@@ -114,7 +114,8 @@ class ExerciseServiceIntegrationTest extends IntegrationTest {
             lessonsService,
             injectExpectationMapper,
             scenarioRecurrenceService,
-            workflowService);
+            workflowService,
+            previewFeatureService);
   }
 
   @AfterAll


openaev-api/src/main/java/io/openaev/service/PauseSimulationService.java
heditar, line 1: Nit: Should it go to the service directory ?

openaev-api/src/main/java/io/openaev/api/chaining/ActionStep.java
heditar, line 1: Nit: These Step class are more related to our core logic than the api, should they be moved to the new chaining directory under service ?

openaev-api/src/main/java/io/openaev/aop/WorkflowUpdateEventAspect.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    if (hasInjectId) {
+      this.handleInjectIdParam(context, injectIdSPEL);
+    } else {
+      this.handleExpectationTracesParam(context, expectationIdsSPEL);
+    }
+
+    // Retry events that are stored in the unsent cache due to previous errors
+    sendEvents(unsentEventsCache);
```
heditar, line 100: What that is does ? Retry again all the failed events ?

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
```diff
@@ -0,0 +1,486 @@
...
+        executor.directExecute(executableInject);
+        String data = setInjectId(inject.getId(), readyStep.getData());
+        readyStep.setData(data);
+        return readyStep;
+      } catch (Exception e) {
+        log.warn(e.getMessage(), e);
+        injectStatusService.failInjectStatus(inject.getId(), e.getMessage());
+      }
+    } catch (IllegalArgumentException | JsonProcessingException e) {
+      log.warn(e.getMessage(), e);
```
heditar: Why are we swallowing the exception? Why we don't mark the inject as failed in that case ?

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
```diff
@@ -0,0 +1,486 @@
...
+
+      // TODO Check add documents? Executable Payloads
+      // executableInject.addDirectAttachment(inject.getDocuments());
+
+      try {
+        executor.directExecute(executableInject);
+        String data = setInjectId(inject.getId(), readyStep.getData());
+        readyStep.setData(data);
+        return readyStep;
+      } catch (Exception e) {
```
heditar: Nit: Maybe we should do some refacto to not have to catch Exception

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
```diff
@@ -0,0 +1,486 @@
...
+        log.warn(e.getMessage(), e);
+        injectStatusService.failInjectStatus(inject.getId(), e.getMessage());
```
heditar: How do we know which inject is going to be marked as failed ? the one in the step or the one newly created ?

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
```diff
@@ -0,0 +1,486 @@
...
+   * Runs a READY step by executing the corresponding to inject.
+   *
+   * <p>Handles deserialization of step data, creation of the inject, execution via {@link
+   * Executor}, and updates the step data with inject ID.
+   *
+   * @param readyStep the step currently in READY status
+   * @return the updated step with execution info, or null if execution fails
+   */
+  @Override
+  public Step run(Step readyStep) {
```
heditar: Nit: Can we extract some of the logic from this method ? The first half part is just to get the inject

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
```diff
@@ -0,0 +1,486 @@
...
+            .disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
+
+    try {
+      // GET INJECT FROM JSON
+      Inject inject = om.readValue(readyStep.getData(), Inject.class);
+
+      ObjectMapper mapper = new ObjectMapper();
+      JsonNode root = mapper.readTree(readyStep.getData());
+
+      Optional<InjectorContract> injectorContractOpt = inject.getInjectorContract();
```
heditar: Nit: should we just extract the value here instead of using an optional in the whole method?

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
```diff
@@ -0,0 +1,486 @@
...
+          root.path("inject_injector_contract").path("injector_contract_injector");
+
+      if (injectorNode.isMissingNode() && injectorNode.isEmpty()) {
+        log.info(
+            "Injector not found for injectorContractId {} & step (READY) id {}",
+            injectorContractOpt.get().getId(),
+            readyStep.getId());
+        return null;
+      }
+      if (injectorContractOpt.get().getInjector() == null) {
```
heditar: In what case we don't have injector?

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
heditar: No @Transactionnel needed ? If the inject is created but there is a failure before execution

openaev-api/src/main/java/io/openaev/api/chaining/dto/StepsCreateInput.java
```diff
@@ -0,0 +1,60 @@
...
+@Builder
+@AllArgsConstructor
+@NoArgsConstructor
+public class StepsCreateInput {
+  /** The Steps. */
+  @JsonProperty("steps")
+  public List<StepCreateInput> steps;
+
+  /** Workflow template ID */
+  @JsonProperty("workflow_id")
```
camrrx, line 21: Need the @Id ? To get it in the stream

openaev-api/src/main/java/io/openaev/rest/exercise/service/ExerciseService.java
```diff
@@ -141,6 +141,22 @@ public Exercise createExercise(@NotNull final Exercise exercise) {
     return exerciseRepository.save(exercise);
   }
 
+  /**
+   * Create a simulation with bthe chaining enabled OR a normal one
```
EvaE-Filigran: “bthe” → “the”

openaev-api/src/main/java/io/openaev/rest/exercise/service/ExerciseService.java
```diff
@@ -141,6 +141,22 @@ public Exercise createExercise(@NotNull final Exercise exercise) {
...
+   *
+   * @param simulation the simulation to create
+   * @param isChaining uses the chaining engine or not
+   * @return the created simulation
+   */
+  @Transactional(rollbackFor = Exception.class)
+  public Exercise createSimulation(@NotNull final Exercise simulation, boolean isChaining) {
```
EvaE-Filigran, line 155: Shouldn’t we use feature flag chaining?

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
EvaE-Filigran: The exception is intentionally not propagated because the step lifecycle is managed by the caller. Returning `null` indicates that the step execution is complete, and the inject itself is explicitly marked as failed.  

In our design, a failing step should simply stop; the workflow as a whole should not be treated as erroneous. An inject failure is non‑blocking at the workflow level. The step acts as an orchestrator, not as a guarantor of business success.    
However, this behavior could be made clearer. We can introduce a dedicated exception (e.g. `StepExecutionException`) and a `FAILED` step status to make the failure semantics explicit while preserving the current workflow behavior.

Proposed implementation:

``` java
catch (Exception e) {
    log.warn(e.getMessage(), e);
    injectStatusService.failInjectStatus(inject.getId(), e.getMessage());
    throw new StepExecutionException("Inject execution failed", e);
}
```
And in the caller:

```java
try {
    Step stepRun = actionStep.run(stepReady);
} catch (StepExecutionException e) {
    stepReady.setStatus(StepStatus.FAILED);
    saveStep(stepReady);
    return;
}
```

This keeps the workflow non‑blocking while making the failure state explicit and easier to reason about.

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
EvaE-Filigran: At that point in the code, there is only one inject: the one that has just been created.

The step template does not reference a persisted inject in the database — it only contains the template definition of an inject.

The flow is:

1. We create the inject from the template.
2. We persist this newly created inject.
3. The executed step then stores the JSON of the executed inject.

The exception is propagated during the direct execution of this inject. Therefore, when we mark the inject as failed, it can only be the newly created and persisted inject instance associated with that specific step execution.

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
EvaE-Filigran: done

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
EvaE-Filigran: The Exception come from the directExecution method use in Executor class

openaev-api/src/main/java/io/openaev/rest/exercise/service/ExerciseService.java
EvaE-Filigran: done

EvaE-Filigran: done

openaev-api/src/main/java/io/openaev/api/chaining/dto/StepsCreateInput.java
camrrx: Never mind, it’s useless in this file since it’s the input used to create the form

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
EvaE-Filigran: The deserialization fail for Injector.

openaev-api/src/main/java/io/openaev/service/PauseSimulationService.java
EvaE-Filigran: done

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
EvaE-Filigran: done

openaev-api/src/main/java/io/openaev/service/chaining/WorkflowService.java
```diff
@@ -0,0 +1,134 @@
...
+   * workflow run is created as a copy of the template with RUN status.
+   *
+   * @param simulationId the ID of the simulation to launch the workflow for
+   * @return the created workflow run, or null if no template exists
+   */
+  public Workflow launchWorkflow(String simulationId) {
+    // 1 WORKFLOW TEMPLATE / SIMULATION
+    Workflow workflowTemplate =
+        workflowRepository.findBySimulation_IdAndStatus(simulationId, WorkflowStatus.TEMPLATE);
+    if (workflowTemplate == null) return null;
```
EvaE-Filigran: We should throw an exception if the workflow template is not found and handle it in the calling method.

openaev-api/src/main/java/io/openaev/service/chaining/StepService.java
```diff
@@ -0,0 +1,805 @@
...
+    }
+  }
+
+  /**
+   * Start workflow for given simulation
+   *
+   * @param simulationId id of the simulation to start
+   */
+  public void startWorkflow(String simulationId) {
+    Workflow workflowTemplate = workflowService.findWorkflowTemplateBySimulationId(simulationId);
```
EvaE-Filigran: We should throw an exception if the workflow template is not found and handle it in the calling method.

openaev-api/src/main/java/io/openaev/service/chaining/StepService.java
```diff
@@ -0,0 +1,805 @@
...
+
+      Step finalstepReady = stepReady;
+
+      // For each step template, IF condition is valid, create condition execution
+      conditionExecution.forEach(condition -> condition.setStep(finalstepReady));
+      conditionService.saveAllConditions(conditionExecution);
+      try {
+        queueChainingService.readyStep(finalstepReady, workflowRun);
+      } catch (IOException e) {
+        // TODO exception management
```
EvaE-Filigran: Todo exception management

openaev-api/src/main/java/io/openaev/service/chaining/ConditionService.java
```diff
@@ -0,0 +1,251 @@
...
+      Instant goal = start.plus(value, ChronoUnit.MILLIS);
+
+      Condition timeConditionValid = isTimeConditionValid(condition, workflowRun, now, goal);
+      if (timeConditionValid == null) {
+        long delay = ChronoUnit.MILLIS.between(now, goal);
+        try {
+          queueChainingService.delayStep(nextStepTemplateToExecute, workflowRun, delay);
+        } catch (IOException e) {
+          // TODO: better exception management
+          throw new RuntimeException(e);
```
EvaE-Filigran: exception management

openaev-api/src/main/java/io/openaev/service/chaining/ConditionService.java
```diff
@@ -0,0 +1,251 @@
...
+   * Retrieves all conditions associated with a step.
+   *
+   * @param stepId step identifier
+   * @return list of conditions linked to the step
+   */
+  public List<Condition> findAllByStepId(String stepId) {
+    return conditionRepository.findAllByStep_Id(stepId);
+  }
+
+  public List<Condition> checkCondition(
```
EvaE-Filigran, line 137: Doc java missing

openaev-api/src/main/java/io/openaev/service/chaining/StepService.java
```diff
@@ -0,0 +1,804 @@
...
+                (a, b) -> {
+                  throw new IllegalArgumentException("Only 1 condition can be first parent");
+                })
+            .orElseThrow(
+                () -> new IllegalArgumentException("Only 1 condition can be first parent"));
+
+    Step stepFrom =
+        firstCondition.getStepFrom() != null
+            ? stepRepository.findById(firstCondition.getStepFrom()).orElse(null)
+            : null;
```
EvaE-Filigran: if stepFrom not null and not find throw an Exception

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
EvaE-Filigran: done

openaev-api/src/main/java/io/openaev/service/chaining/ConditionService.java
EvaE-Filigran: done

openaev-api/src/main/java/io/openaev/service/chaining/StepService.java
EvaE-Filigran: done

openaev-api/src/main/java/io/openaev/service/chaining/ConditionService.java
EvaE-Filigran: done

openaev-api/src/main/java/io/openaev/service/chaining/StepService.java
EvaE-Filigran: done

openaev-api/src/main/java/io/openaev/service/chaining/StepService.java
EvaE-Filigran: done

openaev-api/src/main/java/io/openaev/service/chaining/WorkflowService.java
EvaE-Filigran: done

openaev-api/src/main/java/io/openaev/rest/exercise/form/ExerciseInput.java
```diff
@@ -59,4 +59,8 @@ public class ExerciseInput {
 
   @JsonProperty("exercise_message_footer")
   private String footer;
+
+  /** Indicate if this simulation will use the chaining engine or the legacy one */
+  @JsonProperty("exercise_is_chaining")
+  private Boolean isChaining = false;
```
savacano28, line 65: If we don’t need null values, I'd prefer use boolean

openaev-api/src/main/java/io/openaev/rest/exercise/service/ExerciseService.java
```diff
@@ -439,6 +456,46 @@ public Page<ExerciseSimple> exercisesWithEmptyGlobalScore(
...
 
+  /**
+   * Find a simulation by it's ID
+   *
+   * @param simulationId ID of the simulation to fetch
+   * @return the simulation found
+   * @throws ElementNotFoundException if no simulation matches the given ID
+   */
+  public Exercise findById(String simulationId) {
+    return exerciseRepository.findById(simulationId).orElseThrow(ElementNotFoundException::new);
```
savacano28: Maybe we should try pass message: "Exercise not found with id: " + simulationId

EvaE-Filigran: Done : ()->new ElementNotFoundException("Simulation not found with ID: " + simulationId));

openaev-api/src/main/java/io/openaev/rest/exercise/service/ExerciseService.java
impolitepanda: So, I hadn’t added it initially because this function was 100% new and therefore only called by us. But you’re absolutely right that it should be feature-flagged, just in case someone starts using it once we’ve merged into release/current.

openaev-api/src/main/java/io/openaev/aop/WorkflowUpdateEventAspect.java
impolitepanda: Exactly, yes. It's not supposed to happen at all, only if, for some reason, rabbit become unreachable or unable to accept new event. It behaves like a stack that stores unsent event and rensend them when a new event happen. It act as a buffer and should last enough time for the infra team to fix the rabbitmq problem.

openaev-api/src/main/java/io/openaev/rest/exercise/service/ExerciseService.java
EvaE-Filigran: done

openaev-api/src/main/java/io/openaev/aop/WorkflowUpdateEventAspect.java
heditar: If for some reason a specific event fails to be processed, will it be retried indefinitely, or is there a maximum number of retries?

Issue #4834, opened 2026-01-30
OAEV Trials - Create 2 metrics for the Hubspot nurturing flow


EllynBsc:
## Context

In the context of OAEV trials, we need to create 2 metrics on OAEV, that will be used to lead the nurturing hubspot flow:

-number_of_users_trials, on Hubspot the corresponding property is: number_of_users_saas_trial
-avg_logins_per_day_trials, on Hubspot the corresponding property : average_daily_logins_on_saas_trial

PR #4860, opened 2026-02-03
[backend] feat(user-event): add user event logic (#4834)

RomuDeuxfois:
### Proposed changes

* Introduce user event tracking (starting with LOGIN)
* Add one new metric: count_logins_per_day

### Testing Instructions

1. Log in to OpenAEV with an user
2. Verify that a LOGIN event is created on DB
3. Check telemetry export to confirm the new metric is emitted

### Related issues

* Closes https://github.com/OpenAEV-Platform/openaev/issues/4834

### Checklist


openaev-api/src/main/java/io/openaev/migration/V4_68__Add_user_event.java
```diff
@@ -0,0 +1,37 @@
...
+public class V4_68__Add_user_event extends BaseJavaMigration {
+  @Override
+  public void migrate(Context context) throws Exception {
+    try (Statement stmt = context.getConnection().createStatement()) {
+      // Table creation
+      stmt.execute(
+          """
+        CREATE TABLE user_events (
+          user_event_id varchar(255) PRIMARY KEY,
+          user_id varchar(255) NOT NULL constraint user_id_fk REFERENCES users(user_id) ON DELETE CASCADE,
```
Dimfacion: **Suggestion :** Do you think we should prepare the future and make it nullable in case an action has been made by a job of openaev for instance ?

openaev-api/src/main/java/io/openaev/rest/user/UserApi.java
```diff
@@ -88,6 +90,7 @@ public User login(@Valid @RequestBody LoginUserInput input) {
       User user = optionalUser.get();
       if (userService.isUserPasswordValid(user, input.getPassword())) {
         userService.createUserSession(user);
+        userEventService.createLoginEvent(user);
```
Dimfacion: **Suggestion :** Since we're doing modification in this method, how about logging a different event for login successful and login failed ?

openaev-api/src/main/java/io/openaev/migration/V4_68__Add_user_event.java
Dimfacion: **Thoughts :** I think the ON DELETE CASCADE should be removed as we probably want to keep the history of what he did. 
For instance, if a malicious actor creates an account, he could do bad things and then deletes his account and there would be no traces.

openaev-api/src/main/java/io/openaev/service/settings/SettingService.java
```diff
@@ -0,0 +1,59 @@
...
+import org.springframework.stereotype.Service;
+
+@RequiredArgsConstructor
+@Service
+@Slf4j
+public class SettingService {
+
+  private final SettingRepository settingRepository;
+
+  public boolean getBoolean(SettingKeys key) {
```
Dimfacion, line 19: **Nitpick :** We should as much as possible add javadoc on public methods of services (I won't add a comment on all the methods that need it, I'll let you check). It's not much but I feel it helps a bit with readibility for newcomers.

openaev-api/src/main/java/io/openaev/service/user_events/UserEventRetentionService.java
```diff
@@ -0,0 +1,45 @@
...
+@Service
+@RequiredArgsConstructor
+@Slf4j
+public class UserEventRetentionService {
+
+  private final UserEventRepository userEventRepository;
+  private final UserEventRetentionSettingsService settingsService;
+
+  @Transactional
+  public void applyRetention() {
```
Dimfacion: **Very Nitpicky :** Maybe deleteOldEvents would be a bit more explicit than applyRetention as method name ?

openaev-api/src/main/java/io/openaev/service/user_events/UserEventRetentionSettingsService.java
```diff
@@ -0,0 +1,38 @@
...
+import io.openaev.service.settings.SettingService;
+import java.util.Objects;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Service;
+
+@Service
+@RequiredArgsConstructor
+@Slf4j
+public class UserEventRetentionSettingsService {
```
Dimfacion: **Nitpick :** Maybe we should name it  UserEventRetentionConfig as it's focused on getting the config for UserEventRetention ? I think it clarifies it's purpose

openaev-api/src/main/java/io/openaev/service/user_events/UserEventService.java
```diff
@@ -0,0 +1,65 @@
...
+
+  private final UserEventRepository userEventRepository;
+
+  // -- CRUD --
+
+  public void createLoginEvent(User user) {
+    this.createEvent(user, LOGIN);
+  }
+
+  private void createEvent(User user, UserEventType type) {
```
Dimfacion: **Question :** Do you think we could maybe make this async ? I don't think it would make a big difference but we might as well prevent adding workload onto the database during a potentially intensive request being handled

openaev-api/src/main/java/io/openaev/telemetry/metric_collectors/UserMetricCollector.java
```diff
@@ -0,0 +1,22 @@
...
+public class UserMetricCollector {
+
+  private final MetricRegistry metricRegistry;
+  private final UserEventService userEventService;
+
+  @PostConstruct
+  public void init() {
+    metricRegistry.registerGauge(
+        "avg_logins_per_day",
+        "Average daily logins for OAEV users",
```
Dimfacion: **Very Nitpicky :** That's not an average, it's a count

openaev-api/src/main/resources/application.properties
```diff
@@ -352,7 +352,7 @@ lade.password=<password>
 
 ### Expectation
 
-openaev.expectation.technical.expiration-time=21600
+openaev.expectation.technical.expiration-time=10
```
Dimfacion: **Question :** Is this supposed to be part of the PR ?

openaev-api/src/main/java/io/openaev/migration/V4_68__Add_user_event.java
RomuDeuxfois: Good point !
I’ll make both changes, thanks for this heads-up :)

openaev-api/src/main/java/io/openaev/rest/user/UserApi.java
RomuDeuxfois: Based on this idea, I’ll improve our login handling and add more granular retention rules.
I’ll inform you once it’s done.

openaev-api/src/main/java/io/openaev/service/user_events/UserEventService.java
RomuDeuxfois: Good point !
I’ll take a look at what can be done.

openaev-api/src/main/resources/application.properties
RomuDeuxfois: Thanks for pointing this out !

Dimfacion:
It seems to be working well, just two remarks :
image
If I connect with the SSO, the user_id is not visible.
If I try to login but fail, the user id is also not visible (we might want to have it in the database to know if someone is trying to bruteforce his way with a user email ?)

PR #4887, opened 2026-02-05
[frontend] fix(catalog): fix stale injector content flash when navigating between connectors (#4835)

### LINKED ISSUE #4835: Detail catalog connector page briefly shows previous content when navigating between items
## Description

When navigating from one detail page to another, the previous page’s content is briefly displayed before the new content loads.

## Reproducible Steps

1. Open a detail page of one connector in the catalog (connector A)
2. Go back on the catalog page
3. Navigate to another detail page (connector B)
4. Observe a brief flash of detail A before detail B is displayed.

## Expected Output

The detail page should immediately show the new content without displaying the previous page’s data.

## Actual Output

The previous detail page content is briefly visible before the new data is rendered.

## Screenshots (optional)
https://github.com/user-attachments/assets/82f2f1c9-9740-4ad9-b981-a05a44706363

camrrx:
### Proposed changes

* Clear previous values before fetching new ones.

### Testing Instructions

1. Open a detail page of one connector in the catalog (connector A)
2. Go back on the catalog page
3. Navigate to another detail page (connector B)
4. Verify there is no brief flash of Item A content while Item B loads

### Related issues

* Closes #4835

### DIFF
--- a/openaev-front/src/admin/components/integrations/common/ConnectorLayout.tsx
+++ b/openaev-front/src/admin/components/integrations/common/ConnectorLayout.tsx
@@ -62,15 +62,14 @@ const ConnectorLayout = () => {
   useEffect(() => {
     isXtmComposerIsReachable().then(({ data }) => setIsXtmComposerUp(data));
 
+    setIsRelatedIdsLoaded(false);
     setLoading(true);
     setRelatedIds(undefined);
-    setIsRelatedIdsLoaded(false);
 
     if (!connectorId) {
       setLoading(false);
       return;
     }
-    setIsRelatedIdsLoaded(false);
     apiRequest.getRelatedIds(connectorId).then(({ data }: { data: ConnectorIds }) => {
       setRelatedIds(data);
       setIsRelatedIdsLoaded(true);


openaev-front/src/admin/components/integrations/common/ConnectorLayout.tsx
```diff
@@ -61,6 +61,11 @@ const ConnectorLayout = () => {
 
   useEffect(() => {
     isXtmComposerIsReachable().then(({ data }) => setIsXtmComposerUp(data));
+
+    setLoading(true);
+    setRelatedIds(undefined);
+    setIsRelatedIdsLoaded(false);
```
guillaumejparis: Already in line 73

openaev-front/src/admin/components/integrations/common/ConnectorLayout.tsx
```diff
@@ -61,6 +61,11 @@ const ConnectorLayout = () => {
 
   useEffect(() => {
     isXtmComposerIsReachable().then(({ data }) => setIsXtmComposerUp(data));
+
+    setLoading(true);
```
guillaumejparis, line 66: Question: if setRelatedIds(data); with data undefined you will never go to the useDataLoader and never put the setLoading to false, are you  sure about that ?

camrrx: You’re right, if getRelatedIds returns undefined or fails, we never set isRelatedIdsLoaded to true, then useDataLoader won’t run and loading could stay true.
To make it safe, we should add a catch/finally on getRelatedIds to always flip isRelatedIdsLoaded and loading ! What do you think?

Issue #3505, opened 2025-07-02
OAEV Multi tenancy [EE] 


Seb-MIGUEL:
## Context

For our MSSP customers we need multi tenancy management in OAEV

With a platform in EE:
- I have CRUD capa to manage multiple tenants
- I manage my platform settings
- I manage my SSO configurations
- As a user invited in multiple tenants I can switch from one tenant to the other

PR #4894, opened 2026-02-06
[backend/frontend] feature(multi-tenancy): Introduce feature flag (#3505)

RomuDeuxfois:
### Proposed changes

* Introduce feature flag for multi-tenancy

### DIFF
--- a/openaev-api/src/test/java/io/openaev/service/PreviewFeatureServiceTest.java
+++ b/openaev-api/src/test/java/io/openaev/service/PreviewFeatureServiceTest.java
@@ -1,7 +1,6 @@
 package io.openaev.service;
 
-import static io.openaev.rest.settings.PreviewFeature.FEATURE_FLAG_ALL;
-import static io.openaev.rest.settings.PreviewFeature.MULTI_TENANCY;
+import static io.openaev.rest.settings.PreviewFeature.*;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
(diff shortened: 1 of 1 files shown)


openaev-api/src/main/resources/application.properties
```diff
@@ -411,4 +411,9 @@ pyroscope.profile.export.timeout=300s
...
 #management.endpoints.web.exposure.include=metrics,prometheus
 management.endpoint.health.show-details=always
-management.server.port=8080
\ No newline at end of file
+management.server.port=8080
+
+#############################
+# FEATURE UNDER DEVELOPMENT #
+#############################
+openaev.enabled-dev-features
```
antoinemzs, line 419: does it need an equal sign `=` at the end ?

openaev-api/src/main/java/io/openaev/rest/settings/PreviewFeature.java
```diff
@@ -9,15 +11,24 @@
...
  * and their corresponding flag will be removed from this enum.
  */
+@Getter
 public enum PreviewFeature {
   // Reserved for internal use.
-  _RESERVED,
-  STIX_SECURITY_COVERAGE_FOR_VULNERABILITIES,
-  LEGACY_INGESTION_EXECUTION_TRACE;
+  _RESERVED("_RESERVED"),
+  FEATURE_FLAG_ALL("*"),
```
impolitepanda, line 18: question: shouldn't this be tested in PreviewFeatureService isFeatureEnabled as well?

RomuDeuxfois:
@antoinemzs & @impolitepanda I fixed the missing usage on PreviewFeatureService

openaev-model/src/main/java/io/openaev/database/model/Tenant.java
```diff
@@ -0,0 +1,28 @@
...
+import lombok.Data;
+import lombok.Getter;
+import lombok.Setter;
+import org.hibernate.annotations.UuidGenerator;
+
+@Data
+@Entity
+@Table(name = "tenants")
+@EntityListeners(ModelBaseListener.class)
+public class Tenant  implements Base {
```
antoinemzs: Accidental, to remove

openaev-api/src/test/java/io/openaev/service/PreviewFeatureServiceTest.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+  }
+
+  @Test
+  void should_enable_feature_when_feature_flag_all_is_enabled() {
+    // ARRANGE
+    openAEVConfig.setEnabledDevFeatures(FEATURE_FLAG_ALL.getValue());
+
+    // ACT & ASSERT
+    assertTrue(previewFeatureService.isFeatureEnabled(MULTI_TENANCY));
```
antoinemzs: Just a heads up, the policy re: feature flags is to remove them from the code once we consider a feature fully done; hence these tests risk breaking when we retire the MULTI_TENANCY flag.

Not a huge deal, but there is the _RESERVED flag available, which will remain always there (one of its purpose is to populate the enum when there are no flags available, but it's a good use case to rely on it for tests).

RomuDeuxfois: Good call, I made the changes.

PR #4905, opened 2026-02-09
[backend] fix(obfuscator): obfuscator options on cmd and psh (#3795)

### LINKED ISSUE #3795: In inject view, it is indicated CMD does not support base64 obfuscation
## Description

In inject view, it is indicated CMD does not support base64 obfuscation:

Image

This should not be displayed (and it is irrelevant here as the payload is using powershell).

We should be able to not propose base64 obfuscation when the payload is using cmd, and propose it when it is compatible.

gabriel-peze:
### Proposed changes

* Remove unecessarry warning message on psh base64 obfuscator option, and remove base64 option on obfuscator when cmd

### Testing Instructions

1. Before running the platform with these changes, start the platform and create a few injects with Command payloads with cmd and psh, warning messages and base64 option should be present.
2. Restart the platform with these changes, new injects shouldn't have the base64 option for cmd, and warning message shouldn't be here for psh.
3. Previously created injects should be fixed too.

### Related issues

* Closes #3795

### DIFF
--- a/openaev-api/src/main/java/io/openaev/injectors/openaev/util/OpenAEVObfuscationMap.java
+++ b/openaev-api/src/main/java/io/openaev/injectors/openaev/util/OpenAEVObfuscationMap.java
@@ -48,7 +48,8 @@ public class OpenAEVObfuscationMap {
   public Map<String, String> getAllObfuscationInfo() {
     Map<String, String> keyInfoMap = new HashMap<>();
     for (Map.Entry<String, OpenAEVObfuscation> entry : obfuscationMap.entrySet()) {
-      keyInfoMap.put(entry.getKey(), "");
+      // Key is used for both label and value (common use case where they're identical)
+      keyInfoMap.put(entry.getKey(), entry.getKey());
     }
     return keyInfoMap;
   }
--- a/openaev-framework/src/main/java/io/openaev/injector_contract/fields/ContractChoiceInformation.java
+++ b/openaev-framework/src/main/java/io/openaev/injector_contract/fields/ContractChoiceInformation.java
@@ -33,19 +33,16 @@ public class ContractChoiceInformation extends ContractCardinalityElement {
   public static class ChoiceItem {
     private final String label;
     private final String value;
-    private final String information;
 
     /**
      * Creates a new choice item.
      *
      * @param label the display label for the choice
      * @param value the actual value stored when this choice is selected
-     * @param information additional context or help text for this choice
      */
-    public ChoiceItem(String label, String value, String information) {
+    public ChoiceItem(String label, String value) {
       this.label = label;
       this.value = value;
-      this.information = information;
     }
   }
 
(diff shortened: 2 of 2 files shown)


openaev-api/src/main/java/io/openaev/injectors/openaev/util/OpenAEVObfuscationMap.java
```diff
@@ -21,11 +21,12 @@ public OpenAEVObfuscation(String information, BiFunction<String, String, String>
     }
   }
 
-  public OpenAEVObfuscationMap() {
+  public OpenAEVObfuscationMap(String executor) {
     this.obfuscationMap = new HashMap<>();
     this.registerObfuscation("plain-text", "", this::obfuscatePlainText);
-    this.registerObfuscation(
-        "base64", "CMD does not support base64 obfuscation", this::obfuscateBase64);
+    if (!"cmd".equals(executor)) {
```
MarineLeM: "cmd" should probably be a constant ?

openaev-api/src/main/java/io/openaev/injectors/openaev/util/OpenAEVObfuscationMap.java
```diff
@@ -21,11 +24,12 @@ public OpenAEVObfuscation(String information, BiFunction<String, String, String>
     }
```
MarineLeM, line 22: it seems that "information" is now useless, isn't it ? may be we should remove it ? what do you think?

MarineLeM (changes requested):
Execution no longer works for the existing inject with CMD payload and Base64 obfuscation.

openaev-api/src/main/java/io/openaev/injectors/openaev/util/OpenAEVObfuscationMap.java
```diff
@@ -47,7 +48,7 @@ public String executeObfuscation(String key, String command, String executor) {
   public Map<String, String> getAllObfuscationInfo() {
     Map<String, String> keyInfoMap = new HashMap<>();
     for (Map.Entry<String, OpenAEVObfuscation> entry : obfuscationMap.entrySet()) {
-      keyInfoMap.put(entry.getKey(), entry.getValue().getInformation());
+      keyInfoMap.put(entry.getKey(), "");
```
MarineLeM: since we remove information on CMD obfuscation, this function getAllObfuscationInfo become completely obsolete

openaev-api/src/main/java/io/openaev/rest/payload/service/PayloadService.java
```diff
@@ -126,8 +126,8 @@ private void updateInjectorContract(Injector injector, Payload payload) {
     injectorContractRepository.save(injectorContractToUpdate);
   }
 
-  private ContractChoiceInformation obfuscatorField() {
-    OpenAEVObfuscationMap obfuscationMap = new OpenAEVObfuscationMap();
+  private ContractChoiceInformation obfuscatorField(String executor) {
+    OpenAEVObfuscationMap obfuscationMap = new OpenAEVObfuscationMap(executor);
     Map<String, String> obfuscationInfo = obfuscationMap.getAllObfuscationInfo();
     return ContractChoiceInformation.choiceInformationField(
         "obfuscator", "Obfuscators", obfuscationInfo, obfuscationMap.getDefaultObfuscator());
```
MarineLeM, line 133: With no additional details on CMD obfuscation, using ContractChoiceInformation for the obfuscator field is no longer relevant. We should switch it to ContractSelect.

If that’s out of scope for this PR, we can log it as a pain point to address later.

MarineLeM (approved):

openaev-api/src/main/java/io/openaev/rest/payload/service/PayloadService.java
gabriel-peze: Tech Pain have been created :)

PR #4912, opened 2026-02-09
[backend] chore(ee-annotation): add ee flag on RBAC annotation and rename files

RomuDeuxfois:
### Proposed changes

* Rename RBAC to AccessControl
* Rename EE to EnterpriseEditionService
* Add ee flag to AccessControl annotation

### DIFF
--- a/openaev-api/src/main/java/io/openaev/service/connectors/ConnectorOrchestrationService.java
+++ b/openaev-api/src/main/java/io/openaev/service/connectors/ConnectorOrchestrationService.java
@@ -33,7 +33,7 @@ import org.springframework.transaction.annotation.Transactional;
 public class ConnectorOrchestrationService {
   private final ConnectorInstanceService connectorInstanceService;
   private final XtmComposerService xtmComposerService;
-  private final EnterpriseEditionService enterpriseEditionServiceService;
+  private final EnterpriseEditionService enterpriseEditionService;
   private final CatalogConnectorService catalogConnectorService;
 
   private final CollectorService collectorService;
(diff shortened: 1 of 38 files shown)


openaev-api/src/test/java/io/openaev/service/ScenarioServiceTest.java
```diff
@@ -53,7 +53,7 @@ class ScenarioServiceTest extends IntegrationTest {
   @Autowired private LessonsCategoryRepository lessonsCategoryRepository;
   @Autowired private HealthCheckUtils healthCheckUtils;
 
-  @Mock Ee eeService;
+  @Mock EnterpriseEditionService enterpriseEditionServiceService;
```
impolitepanda: I think the var name is a bit long. Maybe just one "service" would be enough ? :D (it appears in other files as well)

openaev-api/src/main/java/io/openaev/service/connectors/ConnectorOrchestrationService.java
```diff
@@ -80,7 +80,8 @@ public XtmComposerInstanceOutput updateConnectorInstanceStatus(
   }
 
   private void throwIfEnterpriseLicenseNotActive() throws LicenseRestrictionException {
-    if (!eeService.isLicenseActive(licenseCacheManager.getEnterpriseEditionInfo())) {
+    if (!enterpriseEditionServiceService.isLicenseActive(
```
impolitepanda: Technically not linked to your work, but worth a look at: is there really a need to pass the licenseCacheManager.getEnterpriseEditionInfo() as param of enterpriseEditionServiceService.isLicenseActive ?
It could be better if this call was done directly inside isLicenseActive, maybe ?

What do you think? (almost all calls of this are done like that)

openaev-api/src/test/java/io/openaev/service/ScenarioServiceTest.java
RomuDeuxfois: Oh.. good shot !

openaev-api/src/main/java/io/openaev/service/connectors/ConnectorOrchestrationService.java
RomuDeuxfois: I saw it, but it introduces circular dependencies. I’ll take a closer look during the multi-tenancy work.

PR #4932, opened 2026-02-11
[frontend] fix: create list widget

MarineLeM:
### Proposed changes

* Fix : be able to create list widget

fix a regression where list widgets could no longer be created on prerelease. The issue was caused by a Renovate PR that upgraded the Zod dependency.
In the Renovcate PR the field "serie" has been added in the zod schema but this field does not exist on list configuration

### DIFF
--- a/openaev-api/src/main/java/io/openaev/migration/V4_70__Clean_list_configuration_data.java
+++ b/openaev-api/src/main/java/io/openaev/migration/V4_70__Clean_list_configuration_data.java
@@ -1,0 +1,23 @@
+package io.openaev.migration;
+
+import java.sql.Statement;
+import org.flywaydb.core.api.migration.BaseJavaMigration;
+import org.flywaydb.core.api.migration.Context;
+import org.springframework.stereotype.Component;
+
+@Component
+public class V4_70__Clean_list_configuration_data extends BaseJavaMigration {
+
+  @Override
+  public void migrate(Context context) throws Exception {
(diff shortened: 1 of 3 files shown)


openaev-front/src/utils/api-types-custom.d.ts
```diff
@@ -123,7 +123,7 @@ export type WidgetInput = Omit<ApiTypes.WidgetInput, 'widget_config'> & {
       widget_configuration_type: 'temporal-histogram';
     }
     | ApiTypes.FlatConfiguration & { widget_configuration_type: 'flat' }
-    | ApiTypes.ListConfiguration & { widget_configuration_type: 'list' }
+    | Omit<ApiTypes.ListConfiguration, 'series'> & { widget_configuration_type: 'list' }
```
guillaumejparis: Question : The problem is not in the frontend definition but on the backend no?

camrrx: If ListConfiguration does not need a series field, it probably shouldn’t be defined in WidgetConfiguration but in FlatConfiguration for example? But it will probably make regression?

MarineLeM: @camrrx i updated by PR, and created a WidgetConfihurationsWithSerie, all kind of config extend from it except the ListConfiguration

camrrx: Nice, I’ve re-tested it and everything is working !!

PR #4950, opened 2026-02-13
[backend] feat(stix): artifact management (#3511)

### LINKED ISSUE #3511: Import OpenCTI artifacts into OpenBAS scenarios
When Threat actor is referenced with artifacts like executable or others we then will be able to import that file as a file drop payload into OpenBAS scenario. Be sure to cleanup that file.

gabriel-peze:
### Proposed changes

* Add artifacts management on stix bundle

### Testing Instructions

1. On OCTI, linked to OAEV, create a security coverage from a report with artifacts
2. On OAEV, you should see a scenario created from the report information
3. Play with the report artifacts, changes on report on OCTI should be handle on OAEV

### Related issues

* Closes #3511 

### Checklist

### DIFF
--- a/openaev-front/src/admin/components/common/injects/Injects.tsx
+++ b/openaev-front/src/admin/components/common/injects/Injects.tsx
@@ -15,6 +15,7 @@ import PaginationComponentV2 from '../../../../components/common/queryable/pagin
 import { buildSearchPagination } from '../../../../components/common/queryable/QueryableUtils';
 import SortHeadersComponentV2 from '../../../../components/common/queryable/sort/SortHeadersComponentV2';
 import { useQueryableWithLocalStorage } from '../../../../components/common/queryable/useQueryableWithLocalStorage';
+import CustomTooltip from '../../../../components/CustomTooltip';
 import { useFormatter } from '../../../../components/i18n';
 import ItemBoolean from '../../../../components/ItemBoolean';
 import ItemDomains from '../../../../components/ItemDomains';
@@ -46,7 +47,6 @@ import InjectorContract from './InjectorContract';
 import InjectPopover from './InjectPopover';
 import InjectsListButtons from './InjectsListButtons';
 import UpdateInject from './UpdateInject';
-import CustomTooltip from "../../../../components/CustomTooltip";
 
 const useStyles = makeStyles()(() => ({
   disabled: {


openaev-api/src/main/java/io/openaev/service/stix/SecurityCoverageInjectService.java
```diff
@@ -102,6 +111,155 @@ private void cleanInjectPlaceholders(String scenarioId) {
...
+        scenario.getInjects().stream()
+            .filter(inject -> !deletedInjects.contains(inject.getId()))
+            .toList();
+    List<String> documentIdsInScenario =
+        currentInjects.stream()
+            .filter(
+                inject ->
+                    inject.getPayload().isPresent()
+                        && inject.getPayload().get() instanceof FileDrop)
+            .map(fileDrop -> ((FileDrop) fileDrop.getPayload().get()).getFileDropFile().getId())
```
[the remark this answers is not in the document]
gabriel-peze: Not applicable anymore

openaev-api/src/main/java/io/openaev/rest/document/DocumentApi.java
```diff
@@ -131,84 +131,12 @@ public Document upsertDocument(
...
-        if (!input.getScenarioIds().isEmpty()) {
-          document.setScenarios(
-              iterableToSet(scenarioRepository.findAllById(input.getScenarioIds())));
-        }
-        document.setTags(iterableToSet(tagRepository.findAllById(input.getTagIds())));
-        document.setType(file.getContentType());
-        return documentRepository.save(document);
-      }
-    }
+    return documentService.upsert(
```
savacano28, line 134: I think you could just pass file, input

openaev-api/src/main/java/io/openaev/rest/document/DocumentService.java
```diff
@@ -133,4 +242,12 @@ public List<RawDocument> documentsForChallenge(@NotBlank String challengeId) {
   public List<RawDocument> documentsForPayload(@NotBlank String payloadId) {
     return this.documentRepository.rawAllDocumentsByPayloadId(payloadId);
   }
+
+  public List<Document> findAllDistinctOnInjectsByScenarioId(@NotBlank String scenarioId) {
+    return this.documentRepository.findAllDistinctOnInjectsByScenarioId(scenarioId);
+  }
+
+  public Document save(Document document) {
```
savacano28, line 254: Maybe we could reuse this save() method in the other places too? could replace them? ty,

openaev-api/src/main/java/io/openaev/opencti/service/OpenCTIService.java
```diff
@@ -244,4 +255,55 @@ public void createReport(
...
+      try {
+        return documentService.upsert(
+            name, octiFile.getInputStream(), octiFile.getSize(), mimeType, documentCreateInput);
+      } catch (Exception e) {
+        String errorMessage =
+            String.format(
+                "Error while upserting document from OpenCTI file (uri=%s, name=%s, mimeType=%s)",
+                uri, name, mimeType);
+        log.error(errorMessage, e);
+        throw new RuntimeException(errorMessage, e);
```
savacano28: Question: if one document fails the whole batch will be rolled back. Is that the intended behavior?

openaev-api/src/main/java/io/openaev/opencti/client/response/File.java
```diff
@@ -0,0 +1,14 @@
+package io.openaev.opencti.client.response;
+
+import java.io.InputStream;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@NoArgsConstructor
+public class File {
```
savacano28: Maybe you could called FileResponse (to be different from io.File)

openaev-api/src/main/java/io/openaev/opencti/client/OpenCTIClient.java
```diff
@@ -49,6 +54,30 @@ public Response execute(String url, String authToken, String mutationBody, JsonN
...
+      try (CloseableHttpResponse res = client.execute(req)) {
+        int statusCode = res.getCode();
+        if (statusCode != 200) {
+          log.error(
+              String.format("Error downloading file from %s with status code %s", url, statusCode));
+          return null;
+        }
+
+        HttpEntity entity = res.getEntity();
+        byte[] content = entity.getContent().readAllBytes();
```
savacano28, line 71: Question: Do you know the typical size of the files being downloaded here?

openaev-api/src/main/java/io/openaev/opencti/client/OpenCTIClient.java
```diff
@@ -49,6 +54,30 @@ public Response execute(String url, String authToken, String mutationBody, JsonN
...
+
+        File file = new File();
+        file.setInputStream(new ByteArrayInputStream(content));
+        file.setSize(entity.getContentLength());
```
savacano28: Here you can use directly: content.length

openaev-api/src/main/java/io/openaev/opencti/client/OpenCTIClient.java
```diff
@@ -49,6 +54,30 @@ public Response execute(String url, String authToken, String mutationBody, JsonN
...
+    try (CloseableHttpClient client = httpClientFactory.httpClientCustom()) {
+      HttpGet req = new HttpGet(url);
+      req.addHeader(HttpHeaders.AUTHORIZATION, "Bearer %s".formatted(authToken));
+
+      try (CloseableHttpResponse res = client.execute(req)) {
+        int statusCode = res.getCode();
+        if (statusCode != 200) {
+          log.error(
+              String.format("Error downloading file from %s with status code %s", url, statusCode));
+          return null;
```
savacano28, line 67: Maybe here we should throw an exception instead of returning null? Returning null on failure forces every caller to handle it explicitly, and it's easy to forget. An exception would make the error impossible to silently ignore.
What should be the behavoir here?

openaev-api/src/main/java/io/openaev/opencti/client/OpenCTIClient.java
gabriel-peze: I don't really know, it could vary i think.
But there won't be a lot of artifacts.

openaev-api/src/main/java/io/openaev/opencti/service/OpenCTIService.java
gabriel-peze: After a check with product, the behavior is to create a placeholder if the download failed :)
image

openaev-api/src/main/java/io/openaev/rest/document/DocumentApi.java
gabriel-peze: i will have to destruct it just after in all cases, so i think it's cleaner like that, we can discuss about it if you want to :)

Issue #4302, opened 2025-10-23
Shodan Injector add asset-creation option


Seb-MIGUEL:
Into Shodan Injector we want to have the ability to select if the contract will create asset (as the logical we have for AWS injector) but within the UI instead of only the docker side

PR #4960, opened 2026-02-16
[backend] refactor(contract-types): extract ContractOutputType business logic into handler classes (#4302)

savacano28:
### Proposed changes

Chunk 1: Part 1 

* Cleaned up the ContractOutputType enum by moving all business logic into separate handler classes. 
* Created handler classes for each output type (Text, Number, Port, PortScan, IPv4, IPv6, Credentials, CVE, Asset)
* Services now fetch handlers from a factory instead of calling enum functions

### Testing Instructions
[Test refact_2026-02-18T08_40_12.089231500Z_(with_teams & with_players & with_variable_values).zip](https://github.com/user-attachments/files/25387446/Test.refact_2026-02-18T08_40_12.089231500Z_.with_teams.with_players.with_variable_values.zip)

1. Atomic Testing with Payload and Output Parsers
You should be able to execute an atomic test using a payload with output parsers configured.
The test should work as expected, with normal execution behavior.

2. Inject Execution with Nuclei
You should be able to execute an inject operation Nuclei.
The execution should complete successfully and produce:
* Findings (including CVEs, if applicable)
* Execution traces
* No unexpected errors

3. Atomic Testing with Vulnerability Expectation (CVE Validation)
You should be able to execute an atomic test with a defined vulnerability expectation.
If the payload uses an output parser of type cve, the system should correctly validate the vulnerability expectation against the detected CVE.

### Related issues

* Related #4302 

### Checklist

### Further comments

### DIFF
--- a/openaev-api/src/main/java/io/openaev/output_processor/AbstractOutputProcessor.java
+++ b/openaev-api/src/main/java/io/openaev/output_processor/AbstractOutputProcessor.java
@@ -1,0 +1,122 @@
+package io.openaev.output_processor;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import io.openaev.database.model.ContractOutputField;
+import io.openaev.database.model.ContractOutputTechnicalType;
+import io.openaev.database.model.ContractOutputType;
+import jakarta.validation.constraints.NotBlank;
+import jakarta.validation.constraints.NotNull;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import lombok.extern.slf4j.Slf4j;
+
+/** Abstract base class providing common functionality for structured output processor handlers. */
+@Slf4j
+public abstract class AbstractOutputProcessor implements OutputProcessor {
+
+  protected final ContractOutputType type;
+  protected final ContractOutputTechnicalType technicalType;
+  protected final List<ContractOutputField> fields;
+  protected final boolean isFindingCompatible;
+
+  protected AbstractOutputProcessor(
+      ContractOutputType type,
+      ContractOutputTechnicalType technicalType,
+      List<ContractOutputField> fields,
+      boolean isFindingCompatible) {
+    this.type = type;
+    this.technicalType = technicalType;
+    this.fields = fields;
+    this.isFindingCompatible = isFindingCompatible;
+  }
+
+  @Override
+  public ContractOutputType getType() {
+    return type;
+  }
+
+  @Override
+  public ContractOutputTechnicalType getTechnicalType() {
+    return technicalType;
+  }
+
+  @Override
+  public List<ContractOutputField> getFields() {
+    return fields;
+  }
+
+  @Override
+  public boolean isFindingCompatible() {
+    return isFindingCompatible;
+  }
+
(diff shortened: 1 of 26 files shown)


openaev-model/src/main/java/io/openaev/output_processor/OutputProcessorHandlerRegistry.java
```diff
@@ -0,0 +1,26 @@
...
+public class OutputProcessorHandlerRegistry {
+
+  private final Map<ContractOutputType, OutputProcessorHandler> cache;
+
+  public OutputProcessorHandlerRegistry(ApplicationContext applicationContext) {
+    this.cache =
+        Arrays.stream(ContractOutputType.values())
+            .collect(
+                Collectors.toUnmodifiableMap(
+                    type -> type, type -> applicationContext.getBean(type.getHandlerClass())));
```
heditar: I think we should avoid this kind of method (getBean), can we maybe use an attribute in the handler  and put it as the key in the map ?

openaev-model/src/main/java/io/openaev/database/model/ContractOutputType.java
```diff
@@ -1,221 +1,51 @@
...
-      "number",
-      ContractOutputTechnicalType.Number,
-      null,
-      true,
-      Objects::nonNull,
-      ContractOutputType::buildString,
-      null,
-      null,
-      null),
+  Number("number", NumberOutputProcessorHandler.class),
```
heditar: Not a fan of this kind of design (with .class), do you think there is another way to do this mapping ?

openaev-api/src/main/java/io/openaev/output_processor/AbstractOutputProcessor.java
heditar, line 1: I think most. of the files here should not be in model as they hold parsing logic that should be in the api module.

openaev-api/src/main/java/io/openaev/output_processor/AssetOutputProcessorHandler.java
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  @Override
+  public boolean validate(JsonNode jsonNode) {
+    return jsonNode != null;
+  }
+
+  @Override
+  public Asset toAsset(JsonNode jsonNode) {
+    return new Asset();
```
heditar: why is not implemnted ?

openaev-api/src/main/java/io/openaev/rest/finding/FindingService.java
```diff
@@ -147,27 +152,36 @@ List<Finding> getFindingsFromInjectorContract(
           if (!contractOutput.isFindingCompatible()) {
             return;
           }
+          OutputProcessorHandler handler =
+              outputProcessorHandlerRegistry.getHandler(contractOutput.getType());
+
+          // Check if handler supports FINDING context
+          if (!(handler instanceof FindingCapable findingHandler)) {
```
heditar: if we know we need a FindingCapable why not just get a FindingCapable ?

PR #4985, opened 2026-02-19
[backend] feat(sso): add group mapping (#4832)

### LINKED ISSUE #4832: OAEV Trials - ability to to do group mapping for the SSO between Hub & OAEV Trials
## Context
Ability to to do group mapping for the SSO between Hub & OAEV Trials

Dimfacion:
### Proposed changes

* Adding group mapping for SSO
*

### Testing Instructions

1. Configure your SAML2 or Auth0 SSO as well as your group mapping
2. Connect to OpenAEV using SSO
3. Check that the user is now in a group as configured

### Related issues

* Closes #4832 
*

### Checklist

### DIFF
--- a/openaev-api/src/main/java/io/openaev/rest/domain/DomainApi.java
+++ b/openaev-api/src/main/java/io/openaev/rest/domain/DomainApi.java
@@ -36,10 +36,7 @@ public class DomainApi extends RestBehavior {
 
   @Operation(summary = "Get a Domain by ID", description = "Fetches detailed Domain info by ID")
   @GetMapping(DOMAIN_URI + "/{domainId}")
-  @RBAC(
-      resourceId = "#domainId",
-      actionPerformed = Action.READ,
-      resourceType = ResourceType.DOMAIN)
+  @RBAC(resourceId = "#domainId", actionPerformed = Action.READ, resourceType = ResourceType.DOMAIN)
   public Domain getDomain(@PathVariable String domainId) {
     return domainService.findById(domainId);
   }


openaev-api/src/main/java/io/openaev/config/AppSecurityConfig.java
```diff
@@ -18,13 +15,12 @@
...
 import jakarta.servlet.http.HttpServletRequest;
-import java.io.IOException;
+import jakarta.validation.constraints.NotBlank;
 import java.util.ArrayList;
-import java.util.Base64;
 import java.util.List;
-import java.util.stream.Stream;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
+import org.jetbrains.annotations.NotNull;
```
damgouj: I don't remember what Yann said after Springboot upgrade but the NotNull and NotBlank don't come for the same package here, is it OK ?

openaev-api/src/main/java/io/openaev/config/AppSecurityConfig.java
```diff
@@ -201,32 +167,28 @@ public User userOauth2Management(
...
+  public OidcUser oidcUserManagement(ClientRegistration clientRegistration, OAuth2User user) {
+    User loginUser = userOauth2Management(clientRegistration, user);
     return new OpenAEVOidcUser(loginUser);
   }
 
-  public OAuth2User oAuth2UserManagement(
-      OAuth2AccessToken accessToken, ClientRegistration clientRegistration, OAuth2User user) {
-    User loginUser = userOauth2Management(accessToken, clientRegistration, user);
+  public OAuth2User oAuth2UserManagement(ClientRegistration clientRegistration, OAuth2User user) {
+    User loginUser = userOauth2Management(clientRegistration, user);
```
damgouj, line 170: Same method for OIDC and OAuth2, I think it's normal but just a message to check with you ?

openaev-api/src/main/java/io/openaev/sso/GroupMapping.java
```diff
@@ -0,0 +1,17 @@
+package io.openaev.sso;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.Data;
+
+@Data
+public class GroupMapping {
+
+  @JsonProperty("idpGroup")
+  private String idpGroup;
```
damgouj, line 10: What is "idp" for ?

openaev-api/src/main/java/io/openaev/service/UserMappingService.java
```diff
@@ -0,0 +1,71 @@
...
+              user.setGroups(userGroups);
+            }
+          } else {
+            if (autoCreate) {
+              Group newGroup = new Group();
+              newGroup.setName(userGroup);
+              groupRepository.save(newGroup);
+              List<Group> userGroups = user.getGroups();
+              userGroups.add(newGroup);
+              user.setGroups(userGroups);
```
damgouj, line 62: Maybe I miss something but when the user is saved in the group ?

openaev-api/src/main/java/io/openaev/sso/GroupMapping.java
Dimfacion: Identity Provider

openaev-api/src/main/java/io/openaev/config/AppSecurityConfig.java
Dimfacion: It doesn't look like it has changed from before this PR but you're right, this is weird. I'll try to check if it makes sense ...

openaev-api/src/main/java/io/openaev/service/UserMappingService.java
Dimfacion: Right after the call to mapCurrentUserWithGroup in SecurityService.userManagement, there is a call to update the user.
That being said, it is not called on newly created user so I'm fixing this.
Good call !

openaev-api/src/main/java/io/openaev/config/AppSecurityConfig.java
Dimfacion: Looks like we don't pass in this method when logging in so I'm unsure if it does anything. Maybe we could remove it later but I would rather not touch it given the urgency of this PR

PR #5012, opened 2026-02-25
[backend] feat(multi-tenancy): add auto default tenant filters and save in Hibernate requests (#3505)

damgouj:
### Proposed changes

* Add auto default tenant filters and save in Hibernate requests

### Testing Instructions

1. OpenAEV must work as before

### Related issues

* 3505

### Checklist

### DIFF
--- a/openaev-api/src/main/java/io/openaev/importer/V1_DataImporter.java
+++ b/openaev-api/src/main/java/io/openaev/importer/V1_DataImporter.java
@@ -1164,7 +1164,8 @@ public class V1_DataImporter implements Importer {
                 allTeams,
                 enabled,
                 dependsDuration,
-                content);
+                content,
+                TenantContext.getCurrentTenant());
           }
           baseIds.put(id, new BaseHolder(injectId));
           originalIds.add(id);
@@ -1275,15 +1276,15 @@ public class V1_DataImporter implements Importer {
   }
 
   /**
-   * Used to create a dummy injector to be able to import injector contract from the starterpack
-   * before the real contract is created by the real injector
+   * Used to create or get a dummy injector to be able to import injector contract from the
+   * starterpack before the real contract is created by the real injector
    *
    * @param importNode contract node
    * @return
    */
-  private Injector createDummyInjector(JsonNode importNode) {
+  private Injector createOrGetDummyInjector(JsonNode importNode) {
 
-    return injectorService.createDummyInjector(
+    return injectorService.createOrGetDummyInjector(
         importNode.get("injector_contract_injector_type").asText(),
         importNode.get("injector_contract_injector_type_name").asText());
   }
@@ -1300,7 +1301,7 @@ public class V1_DataImporter implements Importer {
     injectorContract.setId(importNode.get("injector_contract_id").textValue());
     injectorContract.setCustom(false);
     injectorContract.setContent(importNode.get("injector_contract_content").textValue());
-    injectorContract.setInjector(createDummyInjector(importNode));
+    injectorContract.setInjector(createOrGetDummyInjector(importNode));
     injectorContract.setConvertedContent((ObjectNode) importNode.get("convertedContent"));
     injectorContract.setExternalId(importNode.get("injector_contract_external_id").textValue());
     injectorContract.setLabels(
--- a/openaev-api/src/main/java/io/openaev/service/InjectorService.java
+++ b/openaev-api/src/main/java/io/openaev/service/InjectorService.java
@@ -117,21 +117,26 @@ public class InjectorService extends AbstractConnectorService<Injector, Injector
   }
 
   /**
-   * Create a dummmy injector, that is used when importing the starter pack before the real
+   * Create or get a dummmy injector, that is used when importing the starter pack before the real
    * injectors are registered
    *
    * @param injectorType
    * @param injectorName
    * @return
    */
-  public Injector createDummyInjector(
+  public Injector createOrGetDummyInjector(
       @NotBlank final String injectorType, @NotBlank final String injectorName) {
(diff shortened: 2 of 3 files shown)


openaev-model/src/main/java/io/openaev/database/audit/TenantBaseListener.java
```diff
@@ -0,0 +1,20 @@
...
+import jakarta.persistence.PreUpdate;
+import lombok.RequiredArgsConstructor;
+import org.springframework.stereotype.Component;
+
+@Component
+@RequiredArgsConstructor
+public class TenantBaseListener<T extends TenantBase> {
+
+  @PrePersist
+  @PreUpdate
```
RomuDeuxfois: **issue(security):** `@PreUpdate` should not overwrite the tenant

An entity should have immutability on his tenant. 
We can add an assertion on `@PreUpdate`

```
@PreUpdate
private void assertTenant(T entity) {
  Object current = TenantContext.getCurrentTenant();
  if (!current.equals(entity.getTenant())) {
    throw new IllegalStateException("Tenant is immutable);
  }
}
```

openaev-model/src/main/java/io/openaev/config/HibernateFilterTransactionAspect.java
```diff
@@ -0,0 +1,29 @@
+package io.openaev.config;
```
RomuDeuxfois: Question : not sure we want to add this aspect on a config package.

openaev-model/src/main/java/io/openaev/config/HibernateFilterTransactionAspect.java
```diff
@@ -0,0 +1,29 @@
...
+
+  @Before(
+      "@annotation(org.springframework.transaction.annotation.Transactional) || "
+          + "@annotation(jakarta.transaction.Transactional)")
+  public void enableFilters() {
+    Session session = entityManager.unwrap(Session.class);
+    if (session.getEnabledFilter("tenantFilter") == null) {
+      session
+          .enableFilter("tenantFilter")
+          .setParameter("tenantId", TenantContext.getCurrentTenant());
```
RomuDeuxfois: Todo: we should always set the tenant id with the current context to avoid side effect I think.

openaev-model/src/main/java/io/openaev/database/model/Agent.java
```diff
@@ -150,7 +152,7 @@ public boolean isActive() {
   @JoinColumn(name = "tenant_id")
   @JsonIgnore
   @NotNull
-  private Tenant tenant = new Tenant(DEFAULT_TENANT_UUID);
+  private Tenant tenant;
```
RomuDeuxfois, line 154: **issue:** `@NotNull` should be replaced with `@Column(nullable = false)`

Now that the tenant is set by `TenantBaseListener` in `@PrePersist`, the field is intentionally `null` between object creation and flush. Keeping `@NotNull` risks a premature `ConstraintViolationException` if Bean Validation triggers before the listener (e.g. @valid).

What do you think?

openaev-model/src/main/java/io/openaev/database/model/TenantBase.java
```diff
@@ -0,0 +1,10 @@
+package io.openaev.database.model;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
```
RomuDeuxfois: Suggestion : seems not usefull

openaev-model/src/main/java/io/openaev/database/model/TenantBase.java
```diff
@@ -0,0 +1,10 @@
...
+public interface TenantBase {
```
RomuDeuxfois: **suggestion:** `TenantBase` should extend `Base`

Every entity implementing `TenantBase` also implements `Base`.

openaev-model/src/main/java/io/openaev/database/model/Tag.java
```diff
@@ -105,7 +107,7 @@ public class Tag implements Base {
   @JoinColumn(name = "tenant_id")
   @JsonIgnore
   @NotNull
-  private Tenant tenant = new Tenant(DEFAULT_TENANT_UUID);
+  private Tenant tenant;
```
RomuDeuxfois, line 111: **suggestion:** use `@Getter`/`@Setter` on the `tenant` instead of manual override

openaev-api/src/main/java/io/openaev/rest/finding/FindingService.java
```diff
@@ -105,7 +105,8 @@ public void buildFinding(
         inject.getId(),
         contractOutputElement.getName(),
         asset.getId(),
-        tagIds);
+        tagIds,
+        inject.getTenant().getId());
```
RomuDeuxfois, line 109: **praise:** good choice to use `inject.getTenant().getId()` instead of `TenantContext.getCurrentTenant()`

A finding must belong to the same tenant as the inject that produced it. I am wondering if we need tenant id on finding class...

damgouj: Thanks ! Yes I put tenant_id on findings because we have a specific Menu "Findings" in our UI so It will be easier for the SQL requests (in my opinion). I added a comment to clarify what we have discussed about it.

openaev-model/src/main/java/io/openaev/database/model/Agent.java
damgouj: Good catch, thanks !

PR #5084, opened 2026-03-04
[frontend] fix(xtm-hub): hide register/unregister buttons in demo mode (#5083)

### LINKED ISSUE #5083: [xtmhub] Prevent user from seeing 'register'/'unregister' buttons in the demo instance
## Description

On the demo platform (`demo.openaev.io`), the "Register in XTM Hub" and "Unregister from XTM Hub" buttons are displayed and clickable. These actions should not be available in demo mode, as registration/unregistration should not be triggered from the demo environment.

## Environment

1. OS (where OpenAEV server runs): Any
2. OpenAEV version: Current
3. OpenAEV client: Frontend
4. Other environment details: Demo platform (`https://demo.openaev.io`)

## Reproducible Steps

1. Go to `https://demo.openaev.io`
2. Navigate to Settings > Experience > XTM Hub tab
3. Observe the "Register in XTM Hub" (or "Unregister from XTM Hub" if already registered) button is visible and clickable

## Expected Output

The Register/Unregister XTM Hub buttons should not be displayed when the platform is running in demo mode.

## Actual Output

The buttons are displayed and functional, allowing users to trigger registration or unregistration from the demo platform.

## Additional information

Related to https://github.com/FiligranHQ/xtm-hub/issues/1784

jbanety:
### Proposed changes

* Hide the "Register in XTM Hub" and "Unregister from XTM Hub" buttons when the platform is running in demo mode (`demo.openaev.io`)
* Add an early return in `XtmHubTab` component after all hooks when demo mode is detected

### Testing Instructions

1. Deploy or configure the platform with `platform_base_url` set to `https://demo.openaev.io`
2. Navigate to Settings > Experience > XTM Hub tab
3. Verify that neither the "Register in XTM Hub" nor the "Unregister from XTM Hub" button is displayed
4. Switch back to a non-demo `platform_base_url` and verify the buttons appear normally

### Related issues

* Closes #5083
* https://github.com/FiligranHQ/xtm-hub/issues/1784

### Checklist

### Further comments

Minimal change — 3 lines added. The early return is placed after all hooks (`useState`, `useCallback`, `useMemo`, `useExternalTab`) to respect React's rules of hooks. This follows the same pattern used elsewhere in the codebase for demo platform checks (e.g. `StartTrialBanner.tsx`).


openaev-front/src/admin/components/settings/experience/xtm_hub/XtmHubTab.tsx
```diff
@@ -32,6 +32,7 @@ const XtmHubTab: React.FC = () => {
   const [showConfirmation, setShowConfirmation] = useState(false);
   const { settings } = useAuth();
   const isEnterpriseEdition = settings.platform_license?.license_is_validated === true;
+  const isDemoMode = settings.platform_base_url === 'https://demo.openaev.io';
```
RomuDeuxfois: I'm not a big fan of hardcoding a URI directly in the front end. Do we have any way to detect that we're in demo mode, maybe through a license or some other flag?

openaev-front/src/public/components/trialbanners/StartTrialBanner.tsx
```diff
@@ -2,14 +2,15 @@ import { useTheme } from '@mui/material/styles';
...
+import { DEMO_PLATFORM_URL } from '../../../utils/Environment';
 import { isEmptyField } from '../../../utils/utils';
 import TopBanner from './TopBanner';
 
 const StartTrialBanner = (settings: { settings: PlatformSettings }) => {
   const { t } = useFormatter();
   const theme = useTheme();
 
-  if (!settings || isEmptyField(settings.settings?.xtm_hub_url) || settings.settings.platform_base_url !== 'https://demo.openaev.io') return <></>;
+  if (!settings || isEmptyField(settings.settings?.xtm_hub_url) || settings.settings.platform_base_url !== DEMO_PLATFORM_URL) return <></>;
```
antoinemzs: I think it's worse than the previous version, sorry! If you want the instance to self-identify as a demo instance, the stop-gap solution of using the URL should be confined behind an abstraction that implements an `isDemoInstance()` function.

Ok for having a hardcoded temporary constant, however you should encapsulate the implementation logic for defining a demo environment in a function away from the rest of the code, and call that abstraction in the various (3) places where you currently need it.

Cheers!

openaev-front/src/admin/components/settings/experience/xtm_hub/XtmHubTab.tsx
jbanety: Fixed.

openaev-front/src/public/components/trialbanners/StartTrialBanner.tsx
jbanety: Done! I've encapsulated the demo detection behind an `isDemoInstance()` function in `Environment.ts`:

```
const DEMO_PLATFORM_URL = 'https://demo.openaev.io';
export const isDemoInstance = (platformBaseUrl?: string) => platformBaseUrl === DEMO_PLATFORM_URL;
```
 
The constant is now private, and the 3 call sites (`XtmHubTab`, `StartTrialBanner`, `systembanners/utils`) all go through `isDemoInstance()`. When we switch to an env var, there's only one place to change.

antoinemzs: Hi,
> When we switch to an env var, there's only one place to change.

that's not correct, there will be 4 places to change: the call sites passing the url string, and the method itself.

```typescript
if (!settings || isEmptyField(settings.settings?.xtm_hub_url) || !isDemoInstance(settings.settings.platform_base_url)) return <></>;
```
```typescript
const DEMO_PLATFORM_URL = 'https://demo.openaev.io';
export const isDemoInstance = (platformBaseUrl?: string) => platformBaseUrl === DEMO_PLATFORM_URL;
```

I assume the demo mode setting will be backend-driven, so you will obtain it through the standard platform settings. You already rely on the useAuth hook to obtain a `PlatformSettings` instance, so IMO it makes sense that the `isDemoMode` method in `Environment.ts` takes the whole settings object and use the appropriate contents to return the boolean.

I've pushed a commit to show you how to put the URL concerns away :)

PR #5091, opened 2026-03-04
[backend] feat(multi-tenancy): manage tag_rules and tags tables (#3505)

damgouj:
### Proposed changes

* Add tenant_id on native requests that can't be automatically managed on tag_rules and tags tables

### Testing Instructions

1. OpenAEV must work as before

### Related issues

* 3505

### Checklist

### DIFF
--- a/openaev-model/src/main/java/io/openaev/database/repository/TagRepository.java
+++ b/openaev-model/src/main/java/io/openaev/database/repository/TagRepository.java
@@ -20,13 +20,9 @@ public interface TagRepository extends CrudRepository<Tag, String>, JpaSpecifica
   Optional<Tag> findById(@NotNull String id);
 
   @NotNull
-  @Query(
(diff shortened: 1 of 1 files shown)


openaev-model/src/main/java/io/openaev/database/repository/TagRepository.java
```diff
@@ -20,9 +20,13 @@ public interface TagRepository extends CrudRepository<Tag, String>, JpaSpecifica
   Optional<Tag> findById(@NotNull String id);
 
   @NotNull
+  @Query(
```
RomuDeuxfois: Why reworking this query ?

damgouj: Because it's a native query and tomorrow we are going to update the unique tag name in the database

RomuDeuxfois: At first, this is not a native query, and entity tag has the tenant filter

PR #5125, opened 2026-03-10
[backend] feat(injects): target a specific injector instance for execution (#4862)

### LINKED ISSUE #4862: Implement catalog multi-deployment [backend]
## Context

To implement multi tenancy we need as prerequisits to be able to deploy multiple times any intergration.

Dimfacion:
### Proposed changes

* Have the injector referenced in the injects and use it during execution
*

### Testing Instructions

1. Start injects
2. Check it's working as usual (no changes expected)

### Related issues

* Closes #4862 
*

### Checklist

### DIFF
--- a/openaev-api/src/main/java/io/openaev/migration/V4_77__Add_inject_injector_column.java
+++ b/openaev-api/src/main/java/io/openaev/migration/V4_77__Add_inject_injector_column.java
@@ -26,13 +26,7 @@ public class V4_77__Add_inject_injector_column extends BaseJavaMigration {
             AND i.inject_injector IS NULL;
           """);
 
-      // 3. For injects without a contract (manual injects), inject_injector stays NULL.
-      //    We cannot make the column NOT NULL if such rows exist.
-      //    So we only set NOT NULL if there are no NULLs remaining.
-      //    In practice, all injects with a contract now have an injector.
-      //    Injects without a contract are "orphan" channel/manual injects — leave them.
-
-      // 4. Add FK constraint (allows NULLs — injects without contract have no injector)
+      // 3. Add FK constraint (allows NULLs — injects without contract have no injector)
       stmt.execute(
           """
           ALTER TABLE injects


openaev-model/src/main/java/io/openaev/database/model/Inject.java
```diff
@@ -174,6 +174,15 @@ public void setDependsOn(List<InjectDependency> dependsOn) {
   @Queryable(filterable = true, dynamicValues = true, path = "injectorContract.injector.id")
   private InjectorContract injectorContract;
 
+  @Getter
+  @ManyToOne(fetch = FetchType.LAZY)
+  @JoinColumn(name = "inject_injector")
+  @JsonSerialize(using = MonoIdSerializer.class)
```
damgouj, line 180: question: do you really need the injector in the Front ?

openaev-api/src/main/java/io/openaev/migration/V4_77__Add_inject_injector_column.java
```diff
@@ -0,0 +1,44 @@
...
+          WHERE i.inject_injector_contract = ic.injector_contract_id
+            AND ic.injector_id IS NOT NULL
+            AND i.inject_injector IS NULL;
+          """);
+
+      // 3. For injects without a contract (manual injects), inject_injector stays NULL.
+      //    We cannot make the column NOT NULL if such rows exist.
+      //    So we only set NOT NULL if there are no NULLs remaining.
+      //    In practice, all injects with a contract now have an injector.
+      //    Injects without a contract are "orphan" channel/manual injects — leave them.
```
damgouj: question: just to be sure, you don't forget any request here ?

openaev-api/src/main/java/io/openaev/migration/V4_77__Add_inject_injector_column.java
```diff
@@ -0,0 +1,44 @@
...
+
+      // 4. Add FK constraint (allows NULLs — injects without contract have no injector)
+      stmt.execute(
+          """
+          ALTER TABLE injects
+          ADD CONSTRAINT fk_inject_injector
+          FOREIGN KEY (inject_injector) REFERENCES injectors(injector_id) ON DELETE CASCADE;
```
damgouj, line 34: question: the injector is deleted if the inject is deleted ?

openaev-model/src/main/java/io/openaev/database/model/Inject.java
Dimfacion: At some point, yes, we will need to select the right injector in the front but also to display which one is being used when we will have multi connector in the same tenant. I thought it would be future proofing to add this, especially since it's just an id that is being sent back. Do you think we should rather remove it for now ?

openaev-api/src/main/java/io/openaev/migration/V4_77__Add_inject_injector_column.java
Dimfacion: Sorry, I removed the query but not the comment :p I'm removing it now

openaev-api/src/main/java/io/openaev/migration/V4_77__Add_inject_injector_column.java
Dimfacion: This is the other way around unless I'm mistaken : if an injector is deleted, the inject linked is deleted. Do you think we should maybe keep it but with an injector set to null ?

openaev-model/src/main/java/io/openaev/database/model/Inject.java
damgouj: Discussed together, OK.

openaev-api/src/main/java/io/openaev/migration/V4_77__Add_inject_injector_column.java
damgouj: Discussed together, OK.

PR #5137, opened 2026-03-13
[backend/frontend] feat(catalog): injectors share injector contracts (#4862)

Dimfacion:
### Proposed changes

* Possibility to have multiple injectors for each injector contracts 
*

### Testing Instructions

1. Check injects are working as usual
2. There shouldn't be a visible change

### Related issues

* Issue #4862 
*

### Checklist

### DIFF
--- a/openaev-model/src/main/java/io/openaev/database/repository/InjectStatusRepository.java
+++ b/openaev-model/src/main/java/io/openaev/database/repository/InjectStatusRepository.java
@@ -2,11 +2,13 @@ package io.openaev.database.repository;
 
 import io.openaev.database.model.InjectStatus;
 import jakarta.validation.constraints.NotNull;
+import java.util.List;
 import java.util.Optional;
 import org.springframework.data.jpa.repository.JpaSpecificationExecutor;
 import org.springframework.data.jpa.repository.Modifying;
 import org.springframework.data.jpa.repository.Query;
 import org.springframework.data.repository.CrudRepository;
+import org.springframework.data.repository.query.Param;
 import org.springframework.stereotype.Repository;
 
 @Repository


openaev-model/src/main/java/io/openaev/database/specification/InjectSpecification.java
```diff
@@ -113,8 +113,12 @@ public static Specification<Inject> fromContract(@NotBlank final String contract
       new HashSet<>(Arrays.asList("openaev_email", "openaev_ovh_sms"));
 
   public static Specification<Inject> testable() {
-    return (root, query, cb) ->
-        root.get("injectorContract").get("injector").get("type").in(VALID_TESTABLE_TYPES);
+    return (root, query, cb) -> {
+      if (query != null) {
```
RomuDeuxfois, line 118: **question**: Why do we need this changes ?

openaev-model/src/main/java/io/openaev/database/repository/InjectStatusRepository.java
```diff
@@ -15,6 +17,11 @@ public interface InjectStatusRepository
   @NotNull
   Optional<InjectStatus> findById(@NotNull String id);
 
+  @Query(
+      value =
+          "select c from InjectStatus c where c.name = 'PENDING' and c.inject.injector.type = :injectType")
+  List<InjectStatus> pendingForInjectType(@Param("injectType") String injectType);
```
RomuDeuxfois: **todo**: Seems unused

openaev-model/src/main/java/io/openaev/database/repository/InjectorContractRepository.java
```diff
@@ -90,11 +90,15 @@ List<RawInjectorsContracts> getAllRawInjectorsContractsWithoutPayloadOrGranted(
   Optional<InjectorContract> findByIdOrExternalId(String id, String externalId);
 
   @NotNull
-  List<InjectorContract> findInjectorContractsByInjector(@NotNull Injector injector);
+  @Query("SELECT ic FROM InjectorContract ic JOIN ic.injectors i WHERE i = :injector")
```
RomuDeuxfois: **question**: I am curious, why did you write yourself the query ?

openaev-model/src/main/java/io/openaev/database/model/InjectorContract.java
```diff
@@ -223,14 +241,26 @@ public boolean getImportAvailableEffective() {
...
+
   @JsonProperty("injector_contract_injector_type")
   private String getInjectorType() {
-    return this.getInjector() != null ? this.getInjector().getType() : null;
+    Injector first = getFirstInjector();
+    return first != null ? first.getType() : null;
   }
 
   @JsonProperty("injector_contract_injector_type_name")
   private String getInjectorName() {
```
RomuDeuxfois: **question**: I saw that it's used on front-end. But, finally, we want one couple injector-contract / injector on front-end no ?
So, we need to changes the all mechanism

openaev-model/src/main/java/io/openaev/database/model/Inject.java
```diff
@@ -506,8 +506,11 @@ public List<AttackPattern> getAttackPatterns() {
   @JsonProperty("inject_type")
   @Queryable(filterable = true, path = "injectorContract.labels", clazz = Map.class)
   public String getType() {
+    if (this.injector != null) {
+      return this.injector.getType();
+    }
     return getInjectorContract()
```
RomuDeuxfois, line 520: **nitpick**: I really don't know why we have this method
getInjectorContract()
But it's really confusing.

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
```diff
@@ -448,30 +453,31 @@ private Inject getInjectFromDataStep(Step step) throws ChainingException {
         throw new ChainingException(
             "Injector contract not found for step (READY) ID: " + step.getId());
 
-      InjectorContract injectorContract = inject.getInjectorContract().get();
-      JsonNode injectorNode =
-          root.path("inject_injector_contract").path("injector_contract_injector");
+      JsonNode injectorNode = root.path("inject_injector");
```
RomuDeuxfois: **questions**: I suppose if your simulation is on the way, and the step is already compute, it willl break due to the changes on the data model ?

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
```diff
@@ -448,30 +453,31 @@ private Inject getInjectFromDataStep(Step step) throws ChainingException {
...
-
-        if (injector == null)
-          throw new ChainingException(
-              "Injector not found for injectorId "
-                  + injectorId
-                  + " and step (READY) ID "
-                  + step.getId());
-
-        injectorContract.setInjector(injector);
+        Injector injector =
```
RomuDeuxfois: **nitpick**: you can use InjectorService.injector with a try-catch

openaev-api/src/main/java/io/openaev/executors/caldera/service/CalderaExecutorContextService.java
```diff
@@ -72,7 +72,9 @@ public void launchExecutorSubprocess(
 
     inject
         .getInjectorContract()
```
RomuDeuxfois, line 74: **question**: Can you use inject.getInjector() direclty ? Same for the other context service.

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
```diff
@@ -260,8 +264,9 @@ private String stepData(StepsCreateInput.StepCreateInput step, Exercise simulati
 
     InjectorContract injectorContract =
         this.injectorContractService.injectorContract(data.getInjectorContract());
-    Inject inject = data.toInject(injectorContract);
-    inject.setInjector(injectorContract.getInjector());
+
+    Injector injector = injectUtils.resolveInjector(data.getInjectorId(), injectorContract);
```
RomuDeuxfois: **nitpick**: This kind of thing will kill me :D fetching a whole entity just to set the relationship..
If injectorReposiroty extends JPArepository instead of CrudRepository, you can use injectorRepository.getReferenceById().

openaev-api/src/main/java/io/openaev/executors/sentinelone/service/SentinelOneExecutorContextService.java
```diff
@@ -68,12 +68,16 @@ public List<Agent> launchBatchExecutorSubprocess(
     // Sometimes, assets from agents aren't fetched even with the EAGER property from Hibernate
     sentinelOneAgents.forEach(agent -> agent.setAsset((Asset) Hibernate.unproxy(agent.getAsset())));
 
-    Injector injector =
-        inject
-            .getInjectorContract()
-            .map(InjectorContract::getInjector)
-            .orElseThrow(
-                () -> new UnsupportedOperationException("Inject does not have a contract"));
+    Injector injector = inject.getInjector();
```
RomuDeuxfois: **question:** Should be more readable, what do you think ?
    Optional<InjectorContract> injectorContract = inject.getInjectorContract();
    Injector injector = inject.getInjector();
    if (injector == null && injectorContract.isEmpty()) {
      new UnsupportedOperationException("Inject does not have a contract");
    }

openaev-api/src/main/java/io/openaev/executors/Executor.java
```diff
@@ -93,24 +93,25 @@ public InjectStatus execute(ExecutableInject executableInject) throws Exception
...
-
     // Resolve the injector instance from the inject entity directly
     Injector injector = inject.getInjector();
     if (injector == null) {
       // Fallback for legacy injects that may not have the field populated
       injector =
           injectorRepository
               .findByTypeAndTenantId(
-                  injectorContract.getInjector().getType(), inject.getTenant().getId())
+                  injectorContract.getFirstInjector().getType(), inject.getTenant().getId())
```
RomuDeuxfois: **nitpick**: rely on existing method in the class, injectorContract.getInjectorType()

openaev-api/src/main/java/io/openaev/importer/V1_DataImporter.java
```diff
@@ -1318,7 +1318,9 @@ private InjectorContract importInjectorContractFromStarterPack(
         new ObjectMapper()
             .convertValue(importNode.get("injector_contract_labels"), new TypeReference<>() {}));
     injectorContract.setPayload(payload);
-    return injectorContractRepository.save(injectorContract);
+    InjectorContract saved = injectorContractRepository.save(injectorContract);
+    dummyInjector.getContracts().add(saved);
+    return saved;
```
RomuDeuxfois, line 1325: **Question**: Why we need to do that ? we already set this on line 1306

openaev-api/src/main/java/io/openaev/rest/injector_contract/InjectorContractService.java
```diff
@@ -264,20 +263,39 @@ public InjectorContract createNewInjectorContract(InjectorContractAddInput input
     setVulnerabilitiesFromExternalOrInternalIds(
         input.getVulnerabilityExternalIds(), input.getVulnerabilityIds(), injectorContract);
 
-    injectorContract.setInjector(
-        updateRelation(input.getInjectorId(), injectorContract.getInjector(), injectorRepository));
+    // Resolve the injector specified in the input
+    Injector injector =
+        injectorRepository
```
RomuDeuxfois: **nitpick**: you can use the InjectorService direclty

openaev-api/src/main/java/io/openaev/rest/payload/service/PayloadService.java
```diff
@@ -82,7 +82,32 @@ public class PayloadService {
...
+      contract = existingContracts.getFirst();
+    } else {
+      contract = new InjectorContract();
+      contract.setId(String.valueOf(UUID.randomUUID()));
+    }
+
+    // Use the first injector for building contract content (they all share the same type)
+    Injector referenceInjector = injectors.isEmpty() ? null : injectors.getFirst();
+    if (referenceInjector != null) {
+      setInjectorContractPropertyBasedOnPayload(contract, payload, referenceInjector);
```
RomuDeuxfois, line 101: **question**: We set the injector line 119 on PayloadService. We are ok with it ?

openaev-api/src/main/java/io/openaev/scheduler/jobs/ComchecksExecutionJob.java
```diff
@@ -56,7 +56,7 @@ private Inject buildComcheckEmail(Comcheck comCheck) {
     InjectorContract contract =
         injectorContractRepository.findById(EmailContract.EMAIL_DEFAULT).orElseThrow();
     emailInject.setInjectorContract(contract);
-    emailInject.setInjector(contract.getInjector());
+    emailInject.setInjector(contract.getFirstInjector());
```
RomuDeuxfois, line 59: **todo**: emailInject.getInjector()

openaev-api/src/test/java/io/openaev/injects/InjectCrudTest.java
```diff
@@ -46,9 +48,5 @@ void createInjectSuccess() {
     // -- EXECUTE --
     Inject injectCreated = this.injectRepository.save(inject);
     assertNotNull(injectCreated);
-
-    // -- CLEAN --
-    this.injectRepository.delete(injectCreated);
```
RomuDeuxfois, line 51: **praise**: good improvment

openaev-model/src/main/java/io/openaev/database/specification/InjectSpecification.java
Dimfacion: We move from a get to a join because we can have multiple injectors. We also add a distinct to avoid returning multiple times the same type

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
Dimfacion: Do you mean if we upgrade oaev between the moment a simulation is scheduled and it's execution ?

openaev-api/src/main/java/io/openaev/executors/caldera/service/CalderaExecutorContextService.java
Dimfacion: I had planned on doing it in the last PR (you can see it here : https://github.com/OpenAEV-Platform/openaev/pull/5158/changes#diff-d8500d479e2e42f03f5b8b092eb76725733bae1e30ff09d83b8af87686673918R72). Is that fine with you if we let it like that for this one ?

openaev-api/src/main/java/io/openaev/scheduler/jobs/ComchecksExecutionJob.java
Dimfacion: AFAIK, there can be only one injector per tenant for the email injector, no ?

openaev-model/src/main/java/io/openaev/database/model/Inject.java
Dimfacion: I guess it's a shortcut to Optional.ofNullable(this.injectorContract). Do you want to remove it ? Because there are 100+ usage of this overall

openaev-model/src/main/java/io/openaev/database/model/InjectorContract.java
Dimfacion: Well, while we want to have multiple injectors per injector contracts, they should all be of the same type and name, no ? Or do you think it would be possible to have injector of different types fulfill an injector contracts ?

openaev-model/src/main/java/io/openaev/database/repository/InjectorContractRepository.java
Dimfacion: Why not ? ;)
Okay, I'm switching to a more automatic query

openaev-api/src/main/java/io/openaev/rest/injector_contract/InjectorContractService.java
Dimfacion: Actually, this lead to cyclic dependency issues

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
RomuDeuxfois: Exaclty.

openaev-api/src/main/java/io/openaev/executors/caldera/service/CalderaExecutorContextService.java
RomuDeuxfois: Fine

openaev-api/src/main/java/io/openaev/executors/sentinelone/service/SentinelOneExecutorContextService.java
```diff
@@ -68,12 +69,12 @@ public List<Agent> launchBatchExecutorSubprocess(
     // Sometimes, assets from agents aren't fetched even with the EAGER property from Hibernate
     sentinelOneAgents.forEach(agent -> agent.setAsset((Asset) Hibernate.unproxy(agent.getAsset())));
 
+    Optional<InjectorContract> injectorContract = inject.getInjectorContract();
```
RomuDeuxfois, line 72: **TODO**: This should be handled the same way as in that PR.
https://github.com/OpenAEV-Platform/openaev/pull/5158/changes#diff-d8500d479e2e42f03f5b8b092eb76725733bae1e30ff09d83b8af87686673918R72
Maybe we should introduce a utility class to centralize the shared logic.

openaev-api/src/main/java/io/openaev/rest/inject/SimulationInjectApi.java
```diff
@@ -240,11 +242,12 @@ public InjectStatus executeInject(
       @PathVariable @NotBlank final String exerciseId,
       @Valid @RequestPart("input") DirectInjectInput input,
       @RequestPart("file") Optional<MultipartFile> file) {
-    Inject inject =
-        input.toInject(
-            this.injectorContractRepository
-                .findById(input.getInjectorContract())
-                .orElseThrow(() -> new ElementNotFoundException("Injector contract not found")));
+    InjectorContract injectorContract =
```
RomuDeuxfois, line 248: **todo** PLease use InjectorContractService.injectorContract

openaev-api/src/main/java/io/openaev/rest/injector_contract/output/InjectorContractFullOutput.java
```diff
@@ -81,16 +81,18 @@ public InjectorContractFullOutput(
...
+    Injector injector = sourceContract.getFirstInjector();
     return new InjectorContractFullOutput(
         sourceContract.getId(),
         sourceContract.getExternalId(),
         sourceContract.getLabels(),
         sourceContract.getContent(),
         sourceContract.getPlatforms(),
         sourceContract.getPayload() == null ? null : sourceContract.getPayload().getType(),
-        sourceContract.getInjector().getName(),
+        injector != null ? injector.getName() : null,
```
RomuDeuxfois: **Nitpick**: Using Optional feels cleaner and easier to read. What do you think?

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
Dimfacion: True, I'll put a fallback in place

openaev-api/src/main/java/io/openaev/rest/injector_contract/InjectorContractService.java
RomuDeuxfois: What about introducing an InjectorCRUDService to start moving toward something more suitable?

openaev-api/src/main/java/io/openaev/scheduler/jobs/ComchecksExecutionJob.java
RomuDeuxfois: That’s true for today, but it may not hold tomorrow.
You could deploy an email injector using a different SMTP configuration.

openaev-api/src/main/java/io/openaev/service/InjectImportService.java
```diff
@@ -740,7 +740,9 @@ private ImportRow importRow(
 
     // Once it's done, we set the injectorContract
     inject.setInjectorContract(injectorContract);
-    inject.setInjector(injectorContract.getInjector());
+    // Given that we import from XLS, we can't know the injector that will be wanted by the user so
+    // we get one that is available
+    inject.setInjector(injectorContract.getFirstInjector());
```
RomuDeuxfois, line 747: **question**: Could we leave it empty so the user can choose one when the simulation is launch ?

openaev-api/src/main/java/io/openaev/service/ScenarioToExerciseService.java
```diff
@@ -235,7 +235,8 @@ public Exercise toExercise(
               .getInjectorContract()
               .ifPresentOrElse(
                   injectorContract -> {
-                    if (ChannelContract.TYPE.equals(injectorContract.getInjector().getType())) {
+                    if (ChannelContract.TYPE.equals(
+                        injectorContract.getFirstInjector().getType())) {
```
RomuDeuxfois: **TODO:** Use `injectorContract.getInjectorType()` to hide the internal implementation details.

openaev-model/src/main/java/io/openaev/database/repository/InjectorContractRepository.java
```diff
@@ -90,11 +90,13 @@ List<RawInjectorsContracts> getAllRawInjectorsContractsWithoutPayloadOrGranted(
   Optional<InjectorContract> findByIdOrExternalId(String id, String externalId);
 
   @NotNull
-  List<InjectorContract> findInjectorContractsByInjector(@NotNull Injector injector);
+  List<InjectorContract> findByInjectorsContaining(@NotNull Injector injector);
 
   @NotNull
+  @Query(
```
RomuDeuxfois: **Question:** Why hardcode this? It gives the same result when SQL is generated.

openaev-api/src/test/java/io/openaev/utils/fixtures/composers/InjectorContractComposer.java
```diff
@@ -81,7 +81,7 @@ public Composer withChallenge(ChallengeComposer.Composer challengeComposer) {
       this.injectorContract.setId(challengeInjectorContract.getId());
       this.injectorContract.setContent(challengeInjectorContract.getContent());
       this.injectorContract.setConvertedContent(challengeInjectorContract.getConvertedContent());
-      this.injectorContract.setInjector(challengeInjectorContract.getInjector());
+      this.injectorContract.addInjector(challengeInjectorContract.getFirstInjector());
```
RomuDeuxfois: **nitpick**: why not using this.injectorContract.addInjectors(challengeInjectorContract.getInjectors());

openaev-api/src/test/java/io/openaev/utils/fixtures/composers/InjectorContractComposer.java
```diff
@@ -144,11 +145,19 @@ public Composer persist() {
       attackPatternComposer.forEach(AttackPatternComposer.Composer::persist);
       vulnerabilityComposer.forEach(VulnerabilityComposer.Composer::persist);
       if (!WELL_KNOWN_CONTRACT_IDS.contains(injectorContract.getId())) {
-        entityManager.persist(injectorContract.getInjector());
-        injectorRepository.save(injectorContract.getInjector());
+        Injector injector = injectorContract.getFirstInjector();
```
RomuDeuxfois: **nitpick**: Same here, can we have a loop ?

openaev-api/src/test/java/io/openaev/service/ScenarioServiceTest.java
```diff
@@ -462,14 +462,17 @@ public void testRunChecksForTeamsIssue() {
     Scenario scenario = new Scenario();
 
     Injector injector = new Injector();
+    injector.setId(java.util.UUID.randomUUID().toString());
```
RomuDeuxfois: **question**: Why do we need this ?

openaev-model/src/main/java/io/openaev/database/model/InjectorContract.java
```diff
@@ -129,15 +128,34 @@ public Payload.PAYLOAD_EXECUTION_ARCH getArch() {
...
+   * Hibernate auto-flush issues.
+   *
+   * <p>After the contract is persisted, update the owning side directly via {@code
+   * injector.getContracts().add(contract)} for join-table persistence.
+   */
+  public void addInjector(Injector injector) {
+    if (injector != null && !this.injectors.contains(injector)) {
+      this.injectors.add(injector);
+    }
+  }
```
RomuDeuxfois, line 191: **suggestion**: can you add a check here to validate every injector have the same type ?
public void addInjector(Injector injector) {
    if (injector != null && !this.injectors.contains(injector)) {
        if (!this.injectors.isEmpty() 
            && !this.injectors.getFirst().getType().equals(injector.getType())) {
            throw new IllegalArgumentException(
                "Cannot link injector of type " + injector.getType() 
                + " to contract already linked to type " + this.injectors.getFirst().getType());
        }
        this.injectors.add(injector);
    }
}

openaev-model/src/main/java/io/openaev/database/model/InjectorContract.java
```diff
@@ -129,15 +128,34 @@ public Payload.PAYLOAD_EXECUTION_ARCH getArch() {
...
+
+  /**
+   * Convenience method: returns the first linked injector, or null. All injectors sharing a
+   * contract have the same type, so this is safe for type/name lookups. TODO : remove this method
+   * when multi connector is ready
+   */
+  @JsonIgnore
+  @Deprecated
+  public Injector getFirstInjector() {
+    return (injectors != null && !injectors.isEmpty()) ? injectors.getFirst() : null;
```
RomuDeuxfois, line 168: **nitpick**: you can use optional here

openaev-model/src/main/java/io/openaev/database/model/InjectorContract.java
RomuDeuxfois: The type will be the same, but not the name right ?

openaev-model/src/main/java/io/openaev/database/model/Inject.java
RomuDeuxfois: I think you can remove this part since you made a full migration to link inject on injector.

return getInjectorContract() .map(InjectorContract::getFirstInjector)
        .map(Injector::getType)
        .orElse(null);

openaev-api/src/main/java/io/openaev/importer/V1_DataImporter.java
RomuDeuxfois: I think it's better to introduce internal class method to handle that

public void linkContract(InjectorContract contract) {
    this.contracts.add(contract);
    contract.getInjectors().add(this);
}

public void unlinkContract(InjectorContract contract) {
    this.contracts.remove(contract);
    contract.getInjectors().remove(this);
}

https://vladmihalcea.com/the-best-way-to-use-the-manytomany-annotation-with-jpa-and-hibernate/

openaev-api/src/main/java/io/openaev/rest/inject/SimulationInjectApi.java
RomuDeuxfois: If you’re dealing with a circular dependency, I’d suggest creating a dedicated CRUD service. I’d rather start with a solid pattern and accept a bit of code duplication than introduce a poor design choice.

openaev-api/src/main/java/io/openaev/rest/injector_contract/output/InjectorContractFullOutput.java
RomuDeuxfois: It’s better to keep the mechanism internal and call sourceContract.getInjectorName() directly.

openaev-model/src/main/java/io/openaev/database/model/Inject.java
Dimfacion: Unfortunately, exported injects be it in scenario, simulation, ... won't have their injector set so letting this allow for retro compatibility

openaev-api/src/main/java/io/openaev/service/InjectImportService.java
Dimfacion: We could but that would mean that for XLS import, you would have to add it on all injects. Or maybe you mean to let it empty if there are more than one ?

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
Dimfacion: I didn't knew about getReferenceById, I'll make sure to use it when possible now

openaev-api/src/main/java/io/openaev/executors/sentinelone/service/SentinelOneExecutorContextService.java
Dimfacion: Is that fine if we do it in the other PR, since it will be merged after this one ?

openaev-api/src/main/java/io/openaev/rest/payload/service/PayloadService.java
Dimfacion: Sorry, I'm not sure to understand what you mean. Can you explain a bit more your reasoning ?

openaev-api/src/test/java/io/openaev/service/ScenarioServiceTest.java
Dimfacion: Not needed, must be something I forgot to clean

openaev-model/src/main/java/io/openaev/database/model/InjectorContract.java
Dimfacion: Do you mean as a return type of getFirstInjector ?

openaev-front/src/admin/components/common/injects/CreateInject.tsx
```diff
@@ -220,7 +220,7 @@ const CreateInject: FunctionComponent<Props> = ({
 
   const availableFilterNames = [
     'injector_contract_attack_patterns',
-    'injector_contract_injector',
+    'injector_contract_injectors',
```
RomuDeuxfois, line 223: **todo**: missing i18n

RomuDeuxfois, line 223: **todo**: filter is not working anymore

Image

openaev-front/src/admin/components/integrations/injectors/InjectorContracts.jsx
```diff
@@ -134,8 +134,8 @@ const InjectorContracts = () => {
       mode: 'and',
       filters: [
         {
-          key: 'injector_contract_injector',
-          operator: 'eq',
+          key: 'injector_contract_injectors',
```
RomuDeuxfois, line 137: **todo**: logo is not working

Image

openaev-model/src/main/java/io/openaev/database/model/InjectorContract.java
Dimfacion: This is the name of the type so if we only have one type, we should have only one type name, don't you think ? 
Tbh, the more I think about it, the more I wonder if we shouldn't do the same as for collectors and have a table of injector types. It's not needed but it would be cleaner. What do you think about it ?

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
Dimfacion: We should be good now

openaev-api/src/main/java/io/openaev/rest/injector_contract/InjectorContractService.java
Dimfacion: I'm honestly not a huge fan of splitting InjectorService, especially since we don't really have a clear CRUD. I'm not a fan either but I'd rather have a @Lazy on the constructor. Would you be ok with that ?

openaev-model/src/main/java/io/openaev/database/model/InjectorContract.java
RomuDeuxfois: Discussed today: remove on chunk 5

PR #5157, opened 2026-03-17
[backend] feat(catalog): renaming rabbitmq queue (#4862)

Dimfacion:
### Proposed changes

* Renaming rabbitmq queues from injector type to injector id
*

### Testing Instructions

1. Check injectors still work as intended
2. 

### Related issues

* Issue #4862
*

### Checklist


Dimfacion:
Hi @corinnekrych 
This seems to be a behavior that is already existing. I think the injector is not created per se until the container is started. Could you try starting it and check if that solves the issue ?

Dimfacion:
Thanks for the approval ! I'll merge it right after the chunk 3 has been merged if that's fine with you

PR #5158, opened 2026-03-17
[backend] feat(catalog): multi-tenant executor routing (#4862)

Dimfacion:
### Proposed changes

* Final PR of the multi connector
*

### Testing Instructions

1. Activate the feature flag to have the UI available
2. Have multiple connectors (injectors, collectors, executors) of the same type
3. Check they work properly
4. Delete them and make sure they are properly deleted

### Related issues

* Issue #4862 
*

### Checklist

### DIFF
--- a/openaev-api/src/main/java/io/openaev/migration/V4_91__Migrate_collector_type_fk_to_uuid.java
+++ b/openaev-api/src/main/java/io/openaev/migration/V4_91__Migrate_collector_type_fk_to_uuid.java
@@ -19,7 +19,7 @@ import org.springframework.stereotype.Component;
  * </ul>
  */
 @Component
-public class V4_90__Migrate_collector_type_fk_to_uuid extends BaseJavaMigration {
+public class V4_91__Migrate_collector_type_fk_to_uuid extends BaseJavaMigration {
 
   @Override
   public void migrate(Context context) throws Exception {
--- a/openaev-api/src/main/java/io/openaev/executors/openaev/service/OpenAEVExecutorContextService.java
+++ b/openaev-api/src/main/java/io/openaev/executors/openaev/service/OpenAEVExecutorContextService.java
@@ -37,17 +37,6 @@ public class OpenAEVExecutorContextService extends ExecutorContextService {
                   () -> new UnsupportedOperationException("Inject does not have a contract"));
     }
 
-    Injector injector = inject.getInjector();
-    if (injector == null) {
-      // Fallback for legacy injects without inject_injector populated
-      injector =
-          inject
-              .getInjectorContract()
-              .map(InjectorContract::getFirstInjector)
-              .orElseThrow(
-                  () -> new UnsupportedOperationException("Inject does not have a contract"));
-    }
-
     return switch (platform) {
       case Windows, Linux, MacOS -> {
         String executorCommandKey = platform.name() + "." + arch.name();


openaev-api/src/main/java/io/openaev/service/connectors/ConnectorOrchestrationService.java
```diff
@@ -114,38 +114,27 @@ public ConnectorInstancePersisted updateRequestedStatus(
...
+  /**
+   * Validates that no connector instance already exists for this catalog connector.
+   *
+   * <p>This check is intentionally disabled: multi-connector support allows deploying multiple
+   * instances of the same catalog connector. Kept as a no-op for documentation purposes.
+   */
+  @SuppressWarnings("unused")
+  private void throwIfConnectorInstanceAlreadyExist(String catalogId) {
+    // Multi-connector is now always allowed — no duplicate check needed.
   }
```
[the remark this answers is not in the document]
Dimfacion: The point of the feature flag is to show or hide the frontend so that's irrelevant

openaev-api/src/main/java/io/openaev/executors/caldera/service/CalderaExecutorContextService.java
```diff
@@ -69,26 +69,27 @@ public void launchExecutorSubprocess(
...
-                calderaExecutorClient.exploit(
-                    "base64",
-                    agent.getExternalReference(),
-                    this.injectorExecutorAbilities.get(injector.getId()).getAbility_id(),
-                    additionalFields);
-              }
-            });
+    Injector injector = inject.getInjector();
+    if (injector == null) {
+      // Fallback for legacy injects without inject_injector populated
```
RomuDeuxfois, line 74: **Question**: Should this fallback be handled through a migration, or should we update the injector so it does not end up here next time?

openaev-model/src/main/java/io/openaev/database/repository/ExecutorRepository.java
```diff
@@ -12,7 +12,6 @@ public interface ExecutorRepository extends CrudRepository<Executor, String> {
   @NotNull
   Optional<Executor> findById(@NotNull String id);
 
-  // TODO multi-tenancy: Multi executors dev
   @NotNull
   Optional<Executor> findByTypeAndTenantId(@NotNull String type, @NotNull String tenantId);
```
RomuDeuxfois, line 16: **question**: Executor add @Filter(name = "tenantFilter", condition = "tenant_id = :tenantId")
So, @NotNull String tenantId seems not usefull

openaev-api/src/main/java/io/openaev/rest/settings/PreviewFeature.java
```diff
@@ -22,7 +22,8 @@ public enum PreviewFeature {
   SENTINEL_ONE_EXECUTOR("SENTINEL_ONE_EXECUTOR"),
   PALO_ALTO_CORTEX_EXECUTOR("PALO_ALTO_CORTEX_EXECUTOR"),
   OPENAEV_TRIALS_XTMHUB("OPENAEV_TRIALS_XTMHUB"),
-  INJECT_CHAINING("INJECT_CHAINING");
+  INJECT_CHAINING("INJECT_CHAINING"),
+  MULTI_CONNECTOR("MULTI_CONNECTOR");
```
RomuDeuxfois: **nitpick**: Why is this in FF? Does this block wrap up your multi-exec work?

openaev-api/src/main/java/io/openaev/execution/ExecutionExecutorService.java
```diff
@@ -47,73 +41,81 @@ public void launchExecutorContext(Inject inject) {
...
-          } catch (AgentException e) {
-            log.error("launchExecutorContextForAgent error: {}", e.getMessage());
-            saveAgentErrorTrace(e, injectStatus);
-          }
-        });
+
+    // Group remaining agents by their executor entity for per-instance routing.
+    // Each executor entity maps to exactly one ConnectorInstance (and therefore one Integration
+    // with its own API client/config)
+    Map<io.openaev.database.model.Executor, Set<Agent>> agentsByExecutor =
```
RomuDeuxfois, line 59: **nitpick**: Good move

openaev-api/src/main/java/io/openaev/execution/ExecutionExecutorService.java
```diff
@@ -47,73 +41,81 @@ public void launchExecutorContext(Inject inject) {
...
+        Manager manager = managerFactory.getManager();
+        ExecutorContextService executorContextService;
+        try {
+          // Resolve the ConnectorInstance that owns this executor
+          ConnectorInstancePersisted instance =
+              connectorInstanceService.findByExecutorId(executor.getId());
+          executorContextService =
+              manager.requestForInstance(instance, ExecutorContextService.class);
+        } catch (NoSuchElementException instanceNotFound) {
+          if (instanceNotFound.getMessage().startsWith("No component found")) {
```
RomuDeuxfois: **todo**: This is unstable because it relies on a string defined in another class, which may change over time.

openaev-api/src/main/java/io/openaev/executors/caldera/service/CalderaExecutorContextService.java
```diff
@@ -69,26 +69,27 @@ public void launchExecutorSubprocess(
...
+      injector =
+          inject
```
RomuDeuxfois, line 76: **question**: Why not using inject.getInjector direclty ? Same for the other context service

openaev-api/src/main/java/io/openaev/executors/ExecutorService.java
```diff
@@ -164,17 +164,6 @@ public Executor register(
 
     Executor executor = executorRepository.findById(id).orElse(null);
     if (executor == null) {
-      Executor executorChecking =
```
RomuDeuxfois, line 167: **question:** Why removing this ?

openaev-api/src/main/java/io/openaev/importer/V1_DataImporter.java
```diff
@@ -1285,9 +1285,11 @@ private void importInjects(
...
    */
   private Injector createOrGetDummyInjector(JsonNode importNode) {
-    return injectorService.createOrGetDummyInjector(
-        importNode.get("injector_contract_injector_type").asText(),
-        importNode.get("injector_contract_injector_type_name").asText());
+    String injectorType = importNode.get("injector_contract_injector_type").asText();
+    JsonNode nameNode = importNode.get("injector_contract_injector_type_name");
+    String injectorName =
+        (nameNode != null && !nameNode.isNull()) ? nameNode.asText() : injectorType;
+    return injectorService.createOrGetDummyInjector(injectorType, injectorName);
```
RomuDeuxfois: **question**: Why does this change? We control the data we send, so is there an issue with the dataset we provide?

openaev-api/src/main/java/io/openaev/integration/impl/executors/caldera/CalderaExecutorIntegration.java
```diff
@@ -91,13 +89,16 @@ public CalderaExecutorIntegration(
...
+    String executorName =
+        connectorInstanceService.getConnectorInstanceConfigurationsByIdAndKey(
+            getConnectorInstance().getId(), "EXECUTOR_NAME");
 
     Executor executor =
         executorService.register(
             executorId,
             CALDERA_EXECUTOR_TYPE,
-            CALDERA_EXECUTOR_NAME,
+            executorName != null ? executorName : CALDERA_EXECUTOR_NAME,
```
RomuDeuxfois: **question**: If executorName is null, doesn’t that point to a bigger issue? Shouldn’t we throw an exception, or simply let Hibernate fail, since executor_name should never be blank?

openaev-api/src/main/java/io/openaev/execution/ExecutionExecutorService.java
```diff
@@ -47,73 +41,81 @@ public void launchExecutorContext(Inject inject) {
...
-        ExecutorContextService executorContextService =
-            managerFactory
-                .getManager()
-                .request(new ComponentRequest(executorName), ExecutorContextService.class);
-        executorContextService.launchBatchExecutorSubprocess(inject, agents, injectStatus);
+        Manager manager = managerFactory.getManager();
+        ExecutorContextService executorContextService;
+        try {
+          // Resolve the ConnectorInstance that owns this executor
+          ConnectorInstancePersisted instance =
```
RomuDeuxfois, line 87: **question**: Why not using 
                  manager.request(
                          new ComponentRequest(executor.getName()), ExecutorContextService.class);

openaev-api/src/main/java/io/openaev/migration/V4_91__Migrate_collector_type_fk_to_uuid.java
```diff
@@ -0,0 +1,186 @@
...
+@Component
+public class V4_90__Migrate_collector_type_fk_to_uuid extends BaseJavaMigration {
+
+  @Override
+  public void migrate(Context context) throws Exception {
+    try (Statement stmt = context.getConnection().createStatement()) {
+
+      // ---- detection_remediations: name → UUID ----
+
+      // 1a. Drop old FK
```
RomuDeuxfois, line 30: **question**: I don’t understand why we have a dedicated CollectorType class, repository, and migration instead of using Collector directly. Since CollectorType is only a subset of Collector, why not use the main class directly?

openaev-api/src/main/java/io/openaev/executors/caldera/service/CalderaExecutorContextService.java
Dimfacion: Thing is, if we don't let fallbacks, previously exported injects won't work anymore :/

openaev-model/src/main/java/io/openaev/database/repository/ExecutorRepository.java
Dimfacion: Maybe I misunderstand this filter but afaik, it will filter on the current tenant but this call is used by the cron jobs of openaev, not during a call to the API so you won't have a current tenant

openaev-api/src/main/java/io/openaev/rest/settings/PreviewFeature.java
Dimfacion: Only the UI part as this was not previously planned

openaev-api/src/main/java/io/openaev/executors/ExecutorService.java
Dimfacion: This is a check to make sure we can't have multiple connectors. If I don't remove this, then you can't have the same type multiple times in the same tenant which defeat the purpose of the multi connector

openaev-api/src/main/java/io/openaev/execution/ExecutionExecutorService.java
Dimfacion: Because we want specifically the integration linked to the id of the executor, not his name. Otherwise, if we have several executor that have the same name we would have a hard time figuring out which one to use

openaev-api/src/main/java/io/openaev/executors/caldera/service/CalderaExecutorContextService.java
Dimfacion: We're using inject.getInjector() (line 72). This is just so that if we import an inject that was exported before the multi connector and it does not have an injector selected, we fallback to this method

openaev-api/src/main/java/io/openaev/migration/V4_91__Migrate_collector_type_fk_to_uuid.java
Dimfacion: Because payloads and detection_remediations are linked to a type and not a collector itself. With your solution, if you have two collectors of the same type and a payload that is pointing to one of them just to reference the collector type it is linked to, then what happens when that collectors is deleted ? You have no choice but to change the FK to another row with the same collector_type

In the end, I think it would work but it would not be very clean and any collector deletion would lead to some shenanigans for every table that reference a collector just to reference it's type

That being said, I'm open to suggestions if you feel there is a better way. Please feel free to tell me if you want to discuss it further

openaev-api/src/main/java/io/openaev/rest/settings/PreviewFeature.java
RomuDeuxfois: Can you schedule a demo with Seb ? If he is ok, we can remove the feature flag and release this part e2e.

openaev-api/src/main/java/io/openaev/executors/caldera/service/CalderaExecutorContextService.java
RomuDeuxfois: What do you think to handle it direclty in the getInjector method to remove complexity here and centralize this behavior accross ContextService.

openaev-front/src/admin/components/common/injects/form/InjectForm.tsx
```diff
@@ -482,6 +512,15 @@ const InjectForm = ({
         <TextFieldController name="inject_description" label={t('Description')} multiline rows={2} disabled={isSubmitting || disabled || permissions.readOnly} />
         <TagFieldController name="inject_tags" label={t('Tags')} disabled={isSubmitting || disabled || permissions.readOnly} />
 
+        {showInjectorSelector && (
```
RomuDeuxfois: **todo**: Should be required*

Image

openaev-front/src/admin/components/common/injects/form/InjectForm.tsx
```diff
@@ -101,11 +105,26 @@ const InjectForm = ({
...
+  const injectorIds = Object.keys(injectorNames);
+  const showInjectorSelector = multiConnectorEnabled && injectorIds.length > 1;
+  const injectorItems = useMemo(() => {
+    if (!showInjectorSelector) return [];
+    return injectorIds
+      .map(id => ({
+        value: id,
+        label: injectorNames[id] ?? id,
+      }));
+  }, [showInjectorSelector, injectorIds, injectorNames]);
```
RomuDeuxfois: **nitpick**: Could be a dedicated component to isolate the logic

openaev-front/src/admin/components/integrations/common/ConnectorPopover.tsx
```diff
@@ -45,13 +47,14 @@ const ConnectorPopover = ({ connectorInstanceId, connectorName, disabled = false
...
-      //   userRight: ability.can(ACTIONS.MANAGE, SUBJECTS.PLATFORM_SETTINGS),
-      // }, {
       label: 'Update',
       action: () => onOpenUpdateConnectorInstanceDrawer(),
       userRight: ability.can(ACTIONS.MANAGE, SUBJECTS.PLATFORM_SETTINGS),
+    },
+    {
+      label: 'Delete',
+      action: handleDelete,
+      userRight: ability.can(ACTIONS.MANAGE, SUBJECTS.PLATFORM_SETTINGS),
```
RomuDeuxfois: **todo**: Should be ACTIONS.DELETE

openaev-front/src/admin/components/integrations/common/DeployButton.tsx
```diff
@@ -27,39 +31,47 @@ const DeployButton = ({ onDeployBtnClick, style = {}, deploymentCount }: Props)
...
+        endIcon={isEnterpriseEdition ? null : <span><EEChip /></span>}
+      >
+        {t('Deploy')}
+      </Button>
+      <Badge
+        badgeContent={deploymentCount}
+        color="warning"
+        sx={{
+          position: 'absolute',
+          top: '0px',
```
RomuDeuxfois: **question**: If it's 0 it can be removed ?

RomuDeuxfois: Hum, I have a strange behavior here

Image

openaev-api/src/main/java/io/openaev/execution/ExecutionExecutorService.java
Dimfacion: Yes, this is ugly, sorry for letting this one slide

Dimfacion:
@RomuDeuxfois Ready for round 2 ?

RomuDeuxfois (approved):

Issue #4868, opened 2026-02-03
Platform RBAC: Implement platform user and capabilities management 


jborozco:
## Context

At platform level I can manage my the user able to interact with the platform settings and give them the following capabilities:
- Bypass
- Access platform settings (mandatory)
    - Can manage parameters (licence key, EE, logo)
    - Create tenant
    - Delete tenant
    - Access all tenants
    - Edit tenants
    - Can Access User & capabilities
        - Manage platform user capabilities
        - Create user groups
        - Delete user groups
    - Can Access SSO configuration
        - Can create SSO configuration
        - Can delete SSO configuration
        - Can Manage SSO configuration

PR #5162, opened 2026-03-18
[backend/frontend] feat(rbac-platform): rework capabilities & add new capas (#4868)

RomuDeuxfois:
### Proposed changes

* Migrate capabilities definition from front-end to API
Replace the static JSON file ([capabilities.json](https://github.com/OpenAEV-Platform/openaev/blob/3ab89fdce7fb8b6d6e8f7c298345424be75c06ca/openaev-front/src/admin/components/settings/roles/capabilities.json)) previously used to define capabilities in the front-end. Capabilities are now served dynamically through a dedicated API endpoint.

* Introduce platform-level capabilities with scoped model
Add a scope property to the capability object (e.g. PLATFORM, TENANT) and define the following platform-level capability tree:
- [ ]  Bypass (user has all rights)
    
    Platform settings
    
    - [ ]  Access platform settings
        - [ ]  Manage platform settings

Tenants

- [ ]  Access tenants
    - [ ]  Manage tenants
        - [ ]  Delete tenants

Platform groups & roles

- [ ]  Access platform groups and roles
    - [ ]  Manage platform groups and roles
        - [ ]  Delete platform groups and roles

### Testing Instructions

1. Step-by-step how to test

### Related issues

* Related to https://github.com/OpenAEV-Platform/openaev/issues/4868

### Checklist

### Further comments


openaev-model/src/main/java/io/openaev/database/model/Capability.java
```diff
@@ -226,7 +277,48 @@ MANAGE_SECURITY_PLATFORMS, pair(ResourceType.SECURITY_PLATFORM, Action.DELETE)),
...
+      ACCESS_TENANTS,
+      pair(ResourceType.TENANT, Action.WRITE),
+      pair(ResourceType.TENANT, Action.CREATE)),
+  DELETE_TENANTS(MANAGE_TENANTS, pair(ResourceType.TENANT, Action.DELETE)),
+
+  // Platform Groups & Roles
+  ACCESS_PLATFORM_GROUPS_AND_ROLES(
+      null,
+      CapabilityGroup.PLATFORM_GROUPS_AND_ROLES,
+      EnumSet.of(CapabilityScope.PLATFORM),
```
camrrx, line 298: Just to confirm: is it expected that `TENANTS` and `PLATFORM_GROUPS_AND_ROLES` do not appear in the frontend tree ? 
With the current logic, that seems normal since both roots are `PLATFORM` scoped and we are filtering by `TENANT` scope? But to be sure, they are not supposed to appear in the same capabilities list?

RomuDeuxfois: Yes, you’re right! We’ll have another capabilities list for the platform only.

camrrx: Ok so on this PR it's expected that we do not have the platform capabilities list on the ui?

RomuDeuxfois: Not at this time; it will be included in a second PR once group and role are available at the platform level. Capabilities will appear on the role during creation.

.github/instructions/backend.instructions.md
```diff
@@ -0,0 +1,92 @@
...
+import static io.openaev.api.platform.PlatformRoleMapper.toOutput;
+public PlatformRoleOutput findById(...) { return toOutput(service.findById(id)); }
+public Page<PlatformRoleOutput> search(...) { return service.search(input).map(Mapper::toOutput); }
+```
+
+## Entities
+
+- Tenant-scoped: `TenantBase` + `@Filter("tenantFilter")` + `TenantBaseListener`
+- Platform-level: `Base` only + `ModelBaseListener`
+- Audit timestamps: implement `Auditable` + add `AuditableListener` (do **not** use Hibernate `@CreationTimestamp`/`@UpdateTimestamp`)
```
corinnekrych, line 60: I could not find those `AuditableListener` in the code base. Is it already available?

.github/instructions/backend.instructions.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Hibernate
+
+- Collections must be mutable — never `List.of()` directly on entity fields
+- Prefer unidirectional relationships
+- `deleteById()` does a SELECT first — use native `@Query @Modifying` for perf-critical deletes
+
+## Services
+
+- `@Service @RequiredArgsConstructor @Transactional(rollbackFor = Exception.class)`
```
corinnekrych:
```suggestion
- Every new service class should have these annotations:
@Service — marks the class as a Spring-managed service bean
@RequiredArgsConstructor — Lombok generates a constructor for all private final fields (replacing @Autowired)
- Methods on Service class should uses
@Transactional(rollbackFor = Exception.class) — wraps every public method in a transaction that rolls back on any exception (not just unchecked ones, which is the Spring default)
```

openaev-front/src/admin/components/settings/roles/RoleForm.tsx
```diff
@@ -112,9 +113,11 @@ const RoleForm: FC<RoleFormProps> = ({
 
         {currentTab === 'Capabilities' && (
           <>
-            {capabilities.map(cap => (
-              <CapabilitiesTab capability={cap} key={cap.name} capabilities={capabilities} />
-            ))}
+            {loading
+              ? <span>{t('Loading...')}</span>
```
corinnekrych: suggestion: could we reuse Loading component?

.github/instructions/backend.instructions.md
RomuDeuxfois: I didn’t include it in this PR since it wasn’t being used yet. However, it will be included in the next one, where it will be used by the new platform group and platform role entities.
Would you prefer that I add it now?

corinnekrych: got it. thanks

PR #5166, opened 2026-03-18
[backend] feat(chaining): replace RabbitMQ delay queue with DB

EvaE-Filigran:
## Summary

This PR replaces the RabbitMQ-based delay queue (`workflows-delay`) with a DB-backed polling scheduler for handling time-conditioned workflow steps.

---

## Context

Previously, when a step had an `AFTER` time condition that was not yet valid, it was pushed into a RabbitMQ delay queue (`workflows-delay`) and consumed by `handleDelayStepEvent`. This approach had several drawbacks:
- Required RabbitMQ configuration for a feature that is purely time-based
- No persistence — delayed steps were lost on restart
- No visibility into what was delayed and when

---

### New: DB-backed delay queue (`steps_delay_queue`)

A new table `steps_delay_queue` stores delayed step executions with the following fields:
- `steps_delay_queue_input` — input payload for the step
- `steps_delay_queue_now` — timestamp when the step was enqueued
- `steps_delay_queue_goal` — target execution timestamp
- `steps_delay_queue_delay` — delay in milliseconds
- `steps_delay_queue_condition_id` — FK to the `AFTER` condition (`ON DELETE SET NULL`)
- `steps_delay_queue_step_template_id` — FK to the step template (`ON DELETE CASCADE`)
- `steps_delay_queue_workflow_run_id` — FK to the workflow run (`ON DELETE CASCADE`)

Indexes are created on `goal`, `step_template_id`, and `workflow_run_id`.

### New: `StepDelayQueueService`

Handles all interactions with the `steps_delay_queue` table:
- `pushStepTemplateIntoStepDelayQueue(...)` — creates and persists a new queue entry
- `deleteStepsDelayQueue(...)` — removes a processed entry
- Add deleteStepsDelayQueue with atomic popNextPerWorkflowRun, using DELETE ... RETURNING to atomically retrieve and remove eligible entries, returning one entry per workflow run ordered by oldest goal first

### New: `StepsDelayQueueService`
- Rename StepsDelayQueue to StepDelayQueue for naming consistency
- Remove delayCondition field from StepDelayQueue — condition is no longer stored in the queue, decoupling condition management from the queue model

### New: QueueChainingJob (Quartz) 
- controlled by @DisallowConcurrentExecution(configurable via 
`openaev.cron.config.steps.delay.queue.polling.interval`, default 10s) that:
1. Polls  and delete`steps_delay_queue` by workflow run id for the oldest entry whose `goal <= now()`
2. Calls `stepService.ready(...)` to attempt step execution
- Add InjectChainingCondition to gate queueChainingTrigger registration on openaev.enabled-dev-features=* or INJECT_CHAINING
- Add queueChainingJobDetail and queueChainingTrigger to platform scheduler
- 
### Modified: `ConditionService`

- Replaced `QueueChainingService` dependency with `StepDelayQueueService`
- `checkCondition()` now calls `pushStepTemplateIntoStepDelayQueue()` directly instead of publishing to RabbitMQ

### Modified: `StepService`

- Removed `handleDelayEvent()` and `handleDelayStepEvent()`
- Added `delayRateTimeCondition()` as a delegate to `ConditionService` to avoid direct field access from `QueueChainingScheduler`

### Modified: `QueueChainingService`

- Removed `delayQueueService` (`BatchQueueService<StepEvent>`)
- Removed `delayStep()` method
- Removed `setCallbackForDelayQueue()` method
- Removed `workflows-delay` queue configuration check on startup
- 
### Modified: 
- Removed `StepEventHandler.handleDelayStepEvent()` method
- Removed `QueueChainingServiceCallbackRegistrar.setCallbackForDelayQueue()` registration

### Modified: `Workflow` and `Step` models

- Added `@OnDelete` on `Workflow.workflowTemplate` and `Step.stepTemplate` to comply with Hibernate 6.6 FK constraint enforcement

---

## Error handling

- `ChainingException` from `ready()` is caught and logged — the entry is deleted anyway to prevent infinite reprocessing loops
- `IllegalArgumentException` from `delayRateTimeCondition()` is caught and logged — the entry is deleted anyway for the same reason
- Both cases follow the same invariant: **always delete after processing**

---

## Tests

| File | Type | Coverage |
|---|---|---|
| `StepDelayQueueServiceTest` | Unit | `pushStepTemplateIntoStepDelayQueue` save assertion |
| `StepDelayQueueIntegrationTest` | Integration | `findFirstByGoalLessThanEqualOrderByGoalAsc` — 4 scenarios |
| `StepServiceTest` — `ProcessDelayStep` | Unit | 3 parametrized scenarios: success, `ChainingException`, queue empty |
| `StepServiceTest` — `DelayRateTimeCondition` | Unit | delegation + exception propagation |
| `ConditionServiceTest` | Unit | `pushStepTemplateIntoStepDelayQueue` called with correct args |
| `QueueChainingServiceTest` | Unit | Removed delay queue tests |
| `QueueChainingServiceCallbackRegistrarTest` | Unit | Removed delay callback tests |

---

## Known limitations / follow-up

- The scheduler processes one entry per tick — high volumes of delayed steps may require batching in a follow-up
- `TODO`: add check whether the workflow contains a delay rate limit before calling `delayRateTimeCondition` (tracked separately)

### Related issues

* Closes #(https://github.com/OpenAEV-Platform/filigran-private/issues/21)

### Checklist

- [X] I added/update the relevant documentation (either on github or on notion)

### DIFF
--- a/openaev-api/src/main/java/io/openaev/migration/V4_82__Add_steps_delay_queue.java
+++ b/openaev-api/src/main/java/io/openaev/migration/V4_82__Add_steps_delay_queue.java
@@ -6,7 +6,7 @@ import org.flywaydb.core.api.migration.Context;
 import org.springframework.stereotype.Component;
 
 @Component
-public class V4_78__Add_steps_delay_queue extends BaseJavaMigration {
+public class V4_82__Add_steps_delay_queue extends BaseJavaMigration {
 
   @Override
   public void migrate(Context context) throws Exception {


openaev-api/src/main/java/io/openaev/service/chaining/QueueChainingScheduler.java
```diff
@@ -0,0 +1,84 @@
...
+   *
+   * <ul>
+   *   <li>Check if the step is ready via {@link StepService#ready(Step,
+   *       io.openaev.database.model.Workflow, String)}.
+   *   <li>If ready, update the delay rate condition using {@link
+   *       io.openaev.service.chaining.ConditionService#delayRateTimeCondition(Condition, Instant,
+   *       Long)}.
+   *   <li>Delete the step from the queue after processing. If the step is not ready, it may be
+   *       re-queued inside {@code ready()}.
+   * </ul>
```
antoinemzs: That's literally the code in the method; are these comments really useful?

openaev-api/src/main/java/io/openaev/service/chaining/StepService.java
```diff
@@ -951,6 +913,22 @@ public void handleExternalUpdateEvent(ExternalUpdateEvent stepEvent) {
     }
   }
 
+  /**
+   * Delegates the delay rate time condition update to {@link ConditionService}.
+   *
+   * <p>This method exists to avoid direct access to {@link ConditionService} from {@link
```
antoinemzs: Encapsulation is when you don't have direct access attributes of objects and instead pass messages via a public interface. Here we are breaking it by handling StepDelayQueue.delayCondition directly. The claims look a bit exaggerated.

In our product we generally consider direct access to class members ok from within the associated service which we subsume as part of the object's public interface, but any direct access (e.g. via getters) elsewhere is technically breaking encapsulation.

openaev-api/src/main/java/io/openaev/service/chaining/QueueChainingScheduler.java
```diff
@@ -0,0 +1,84 @@
...
+   *
+   * Exceptions during step readiness evaluation are logged but do not stop the scheduler.
+   */
+  @Scheduled(fixedDelayString = "${openaev.cron.config.steps.delay.queue.polling.interval:10000}")
+  public void processDelayStep() {
```
antoinemzs: This isn't thread-safe so a useful comment would be to not try to paralllelise this method without reworking it to make it thread-safe

openaev-api/src/test/java/io/openaev/service/chaining/StepDelayQueueIntegrationTest.java
```diff
@@ -0,0 +1,100 @@
...
+import java.util.List;
+import java.util.Optional;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+
+@SpringBootTest
+public class StepDelayQueueIntegrationTest {
```
antoinemzs, line 24: Integration test harnesses should be transactional as per standard

openaev-model/src/main/java/io/openaev/database/model/Step.java
```diff
@@ -107,6 +107,7 @@ public class Step implements Base {
   @JoinColumn(name = "step_template_id")
   @ManyToOne(fetch = FetchType.LAZY)
   @JsonIgnore
+  @OnDelete(action = OnDeleteAction.SET_NULL)
```
antoinemzs: Interested to see if this really works. #5141 was designed because this specific behaviour was not firing. However, we may be avoiding the error case because the fetch mode is lazy (as it should). What do you think? Is there a test that proves it is working?

openaev-api/src/main/java/io/openaev/service/chaining/QueueChainingScheduler.java
antoinemzs: This is catching only two exception types, so we may miss the deletion (consume) if an unexpected exception type arises.

openaev-api/src/main/java/io/openaev/service/chaining/QueueChainingScheduler.java
```diff
@@ -0,0 +1,84 @@
...
+                  "Invalid condition type for delay rate, entry will be discarded. "
+                      + "Condition ID: {}, Type: {}",
+                  stepDelay.getDelayCondition().getId(),
+                  stepDelay.getDelayCondition().getType(),
+                  e);
+            }
+          }
+          // Always deleted after treatment, if needed the method ready() will delay the step again
+          stepDelayQueueService.deleteStepsDelayQueue(stepDelay);
+        });
```
antoinemzs: This claim looks like a good case for a `finally` block.

openaev-api/src/main/java/io/openaev/service/chaining/StepDelayQueueService.java
```diff
@@ -0,0 +1,91 @@
...
+  public void pushStepTemplateIntoStepDelayQueue(
+      Step stepTemplate,
+      Instant now,
+      String input,
+      long delay,
+      Condition delayCondition,
+      Workflow workflowRun,
+      Instant goal) {
+    log.info(
+        "DELAY STEP TEMPLATE : {} CONDITION TIME AFTER: {} + {} milliseconds => Goal: {}",
```
antoinemzs: Consider lower case 🙏 also consider whether it isn't rather debug level also

openaev-api/src/main/java/io/openaev/service/chaining/QueueChainingScheduler.java
```diff
@@ -0,0 +1,84 @@
...
+   *   <li>If ready, update the delay rate condition using {@link
+   *       io.openaev.service.chaining.ConditionService#delayRateTimeCondition(Condition, Instant,
+   *       Long)}.
+   *   <li>Delete the step from the queue after processing. If the step is not ready, it may be
+   *       re-queued inside {@code ready()}.
+   * </ul>
+   *
+   * Exceptions during step readiness evaluation are logged but do not stop the scheduler.
+   */
+  @Scheduled(fixedDelayString = "${openaev.cron.config.steps.delay.queue.polling.interval:10000}")
```
antoinemzs: I wonder if we wouldn't rather want to have a quartz job scheduler so that we don't multiply the scheduling frameworks?
My thinking is that we might not, at first glance, understand from the codebase all the background work being done globally on the platform. Why choosing `@Scheduled` rather than a quartz job?

openaev-api/src/main/java/io/openaev/service/chaining/QueueChainingScheduler.java
```diff
@@ -0,0 +1,84 @@
...
+  public void processDelayStep() {
+
+    // Retrieve the next delayed step whose goal time is reached
+    Optional<StepsDelayQueue> stepDelayOptional = stepDelayQueueService.findNextToProcess();
+    stepDelayOptional.ifPresent(
```
antoinemzs: Probably clearer to return early rather than `ifPresent()`, also potential traces would be cleaner

openaev-api/src/main/java/io/openaev/service/chaining/StepDelayQueueService.java
```diff
@@ -0,0 +1,91 @@
...
+   *
+   * <p>Retrieves the first {@link StepsDelayQueue} entry whose goal time is less than or equal to
+   * the current timestamp, ordered by ascending goal time.
+   *
+   * @return an {@link Optional} containing the next ready {@link StepsDelayQueue}, or empty if none
+   *     are ready
+   */
+  public Optional<StepsDelayQueue> findNextToProcess() {
+    return stepDelayQueueRepository.findFirstByGoalLessThanEqualOrderByGoalAsc(Instant.now());
+  }
```
antoinemzs: _not_ thread safe, please comment

openaev-api/src/main/java/io/openaev/service/chaining/QueueChainingScheduler.java
```diff
@@ -0,0 +1,84 @@
...
+              .isPresent()) { // todo add check if workflow contain delay rate limit (means its a
+            // chaining event base)
+
+            // Update the delay condition for all steps in the workflow run.
+            // This ensures the rate limit is respected for subsequent step executions.
+            try {
+              stepService.delayRateTimeCondition(
+                  stepDelay.getDelayCondition(), Instant.now(), stepDelay.getDelay());
+            } catch (IllegalArgumentException e) {
+              // Programming error — condition is not of type AFTER
```
antoinemzs: Implementation detail

openaev-api/src/main/java/io/openaev/service/chaining/ConditionService.java
```diff
@@ -263,4 +261,14 @@ public Condition isDependOn(String idStepFromTemplate) {
         .value(idStepFromTemplate)
         .build();
   }
+
+  public void delayRateTimeCondition(Condition condition, Instant lastExecution, Long timeRate) {
+    if (!condition.getType().equals(ConditionType.AFTER))
```
antoinemzs: always better to introduce the constant first, ie 
```suggestion
    if (!ConditionType.AFTER.equals(condition.getType()))
```
avoids a lot of NPEs

openaev-api/src/main/java/io/openaev/service/chaining/QueueChainingScheduler.java
EvaE-Filigran: For me, it was the method isn’t simple, which is why I added many comments.

openaev-api/src/main/java/io/openaev/service/chaining/QueueChainingScheduler.java
EvaE-Filigran: I will implement quartz job for this scheduler

openaev-api/src/main/java/io/openaev/service/chaining/StepDelayQueueService.java
EvaE-Filigran: Done

openaev-api/src/main/java/io/openaev/service/chaining/QueueChainingScheduler.java
EvaE-Filigran: Update Java doc to describe behavior rather than implementation details.

openaev-api/src/main/java/io/openaev/service/chaining/StepService.java
EvaE-Filigran: delayCondition removed

openaev-api/src/main/java/io/openaev/service/chaining/QueueChainingScheduler.java
EvaE-Filigran: Implementation of Quartz Scheduler using @DisallowConcurrentExecution and synchronized method when needed (StepDelayQueueService.popNextToProcess())

openaev-model/src/main/java/io/openaev/database/model/Step.java
EvaE-Filigran: Change @OnDelete (hibernate )by @OnDelete (oaev)

openaev-api/src/main/java/io/openaev/service/chaining/QueueChainingScheduler.java
EvaE-Filigran: fix

openaev-api/src/main/java/io/openaev/service/chaining/StepDelayQueueService.java
EvaE-Filigran: resolved

openaev-model/src/main/java/io/openaev/database/repository/StepDelayQueueRepository.java
```diff
@@ -0,0 +1,28 @@
...
+    WITH next_per_run AS (
+        SELECT DISTINCT ON (steps_delay_queue_workflow_run_id) steps_delay_queue_id
+        FROM steps_delay_queue
+        WHERE steps_delay_queue_goal <= now()
+        ORDER BY steps_delay_queue_workflow_run_id, steps_delay_queue_goal
+    )
+    DELETE FROM steps_delay_queue
+    WHERE steps_delay_queue_id IN (SELECT steps_delay_queue_id FROM next_per_run)
+    RETURNING *
+    """,
```
antoinemzs, line 25: nice

Issue #5109, opened 2026-03-06
Security platform score in stix bundle send to OCTI is incorrect


MarineLeM:
## Description
The STIX bundle sent to OpenCTI does not properly set the score for each security platform on the relationship with the security coverage.

Additionally, the bundle should not include any security platform that has no link to the simulation

Impact : Creates an incoherent state => in OpenCTI the security coverage shows "detected" or "prevented" results, but none of the security platforms have a score.

<img width="861" height="826" alt="Image" src="https://github.com/user-attachments/assets/4fb8a0b7-c225-436d-9dc0-4884a8bde1a1" />

## Environment

1. OpenAEV version: 2.2.1
2. OpenCTI version : 7.0

## Reproducible Steps

Create a security coverage,  OpenAEV create and run a simulation, "prevented" and "detected" expectation results are collected from Microsoft Defender for sample.
Once all expectations are completed, the simulation finishes and OpenAEV computes a summary as a STIX bundle and sends it to OpenCTI.
Observe the STIX bundle sent to OpenCTI.

<img width="637" height="380" alt="Image" src="https://github.com/user-attachments/assets/a58e7a29-0a70-4bed-ad2e-a3d6f7b99917" />

## Expected Output

The STIX bundle includes a score on the relationship between each security platform and the security coverage.
Only security platforms that have a link to the simulation are included in the bundle.

## Actual Output

No score is set.
Security platforms with no link to the simulation are still included in the bundle.

PR #5173, opened 2026-03-19
[backend] feat (interco octi) : fix security platform send in stix bundle (#5109)

GaetanSantucci:
The STIX bundle sent to OpenCTI does not properly set the score for each security platform on the relationship with the security coverage.

Additionally, the bundle should not include any security platform that has no link to the simulation

Impact : Creates an incoherent state => in OpenCTI the security coverage shows "detected" or "prevented" results, but none of the security platforms have a score.

### Proposed changes

* Verify and send only the security platforms related to the simulation

### Testing Instructions

You need OpenCTI and OpenAEV to test
1. In OpenCTI, create a report and add a Security Coverage, that will create a scenario/simulation in OpenAEV
2.  When simulation is finished, verified on Security coverage overview (OpenCTI) you only need to find the Security Platform related to the inject in the simulation on OpenAEV side.

OpenAEV side 
image

OpenCTI side
image

### Related issues

* #5109 

### Checklist

### DIFF
--- a/openaev-api/src/main/java/io/openaev/service/AssetService.java
+++ b/openaev-api/src/main/java/io/openaev/service/AssetService.java
@@ -13,7 +13,6 @@ import jakarta.validation.constraints.NotBlank;
 import jakarta.validation.constraints.NotNull;
 import java.util.List;
 import java.util.Set;
-
 import lombok.RequiredArgsConstructor;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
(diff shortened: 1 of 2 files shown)


openaev-api/src/main/java/io/openaev/service/stix/SecurityCoverageService.java
```diff
@@ -565,6 +567,16 @@ private void processCoverageRefs(
...
   }
 
+  private List<String> getPlatformNames(List<Inject> injects) {
+    return injects.stream()
+        .flatMap(inject -> inject.getExpectations().stream())
+        .flatMap(expectation -> expectation.getResults().stream())
+        .map(InjectExpectationResult::getSourceName)
+        .filter(Objects::nonNull)
+        .distinct()
+        .toList();
```
gabriel-peze: You can remove the .distinct().toList() and use .toSet() directly and use Set on the process

PR #5189, opened 2026-03-20
[backend] feat(multi-tenancy): add tenant context to elasticsearch and opensearch queries (#3505)

johanah29:
### Proposed changes

Add tenant context to OpenAEV queries, so that Elasticsearch and Opensearch separate the data. 

### Testing Instructions

1. Set the config file to consider elasticsearch as engine
2. Create a dashboard: the result should be the same than on the main brain
3. Do the same with opensearch as the engine

### Testing Instructions : endpoints case
1. Create a number widget with the endpoints entity
2. Create a new tenant and get the id from your database
3. In your database create a new asset with type **Endpoint** and add the previous id as its tenant_id
4. The number of endpoints in your number widget should be the same.


openaev-model/src/main/java/io/openaev/engine/model/vulnerableendpoint/EsVulnerableEndpoint.java
```diff
@@ -87,4 +87,8 @@ public class EsVulnerableEndpoint extends EsBase {
   @Queryable(label = "tags", filterable = true, dynamicValues = true)
   @EsQueryable(keyword = true)
   private Set<String> base_tags_side; // Must finish by _side
+
+  @Queryable(label = "tenant", filterable = true, dynamicValues = true)
```
RomuDeuxfois: **todo**: Can you create a EsTenantBase like we have for PostgreSQL entity ? (TenantBase)

Something like that
@Getter
@Setter
public class EsTenantBase extends EsBase {

  @Queryable(label = "tenant", filterable = true, dynamicValues = true)
  @EsQueryable(keyword = true)
  private String base_tenant_side; // Must finish by _side
}

openaev-model/src/main/java/io/openaev/database/raw/impl/SimpleRawInjectExpectation.java
```diff
@@ -56,4 +56,5 @@ public class SimpleRawInjectExpectation implements RawInjectExpectation {
   public Set<String> security_platform_ids;
   private String inject_title;
   private Instant tracking_sent_date;
+  private String tenant_id;
```
RomuDeuxfois: **question**: Why adding this here ? this bean seems not used.

openaev-model/src/main/java/io/openaev/database/raw/RawAsset.java
```diff
@@ -84,4 +84,11 @@ public interface RawAsset {
    * @return set of associated tag IDs
    */
   Set<String> getAsset_tags();
+
```
RomuDeuxfois: **nitpick**: Can you align bean name with others ? RawEndpointIndexing
Same with others, if you have time. It's really a nitpick :)

openaev-model/src/main/java/io/openaev/database/raw/RawAsset.java
```diff
@@ -84,4 +84,11 @@ public interface RawAsset {
...
+  /**
+   * Returns the tenant ID to which this asset belongs.
+   *
+   * @return the tenant ID
+   */
+  String getTenant_id();
```
RomuDeuxfois: **todo**: Can we have a RawTenantIndexing to implement ?
public interface RawTenantIndexing {

  String getTenant_id();
}

openaev-model/src/main/java/io/openaev/database/raw/impl/SimpleRawInjectExpectation.java
johanah29: This bean implements RawInjectExpectation interface with getTenant_id methode. If it is not used, i migth as well delete it

RomuDeuxfois: I confirm, you can deleted it

openaev-model/src/main/java/io/openaev/database/raw/RawAssetGroupIndexing.java
RomuDeuxfois: **todo**: Should extends RawTenant no ?

openaev-model/src/main/java/io/openaev/database/raw/RawDomain.java
RomuDeuxfois: **todo**: Should extends RawTenant no ?

openaev-model/src/main/java/io/openaev/engine/model/endpoint/EsEndpoint.java
```diff
@@ -5,15 +5,15 @@
...
+import io.openaev.engine.model.tenant.EsTenantBase;
 import java.util.Set;
 import lombok.Getter;
 import lombok.Setter;
 
 @Getter
 @Setter
 @Indexable(index = "endpoint", label = "Endpoint")
-public class EsEndpoint extends EsBase {
+public class EsEndpoint extends EsTenantBase {
```
RomuDeuxfois, line 16: **question**: Why we don't set tenant id on EndpointHandler handler ?

openaev-model/src/main/java/io/openaev/engine/model/securityplatform/SecurityPlatformHandler.java
```diff
@@ -19,7 +19,7 @@ public class SecurityPlatformHandler implements Handler<EsSecurityPlatform> {
   @Override
   public List<EsSecurityPlatform> fetch(Instant from) {
     Instant queryFrom = from != null ? from : Instant.ofEpochMilli(0);
-    List<RawAsset> forIndexing = securityPlatformRepository.findForIndexing(queryFrom);
+    List<RawAssetIndexing> forIndexing = securityPlatformRepository.findForIndexing(queryFrom);
```
RomuDeuxfois, line 22: **question**: Why we don't set tenant id here ?

openaev-model/src/main/java/io/openaev/database/raw/RawTenant.java
```diff
@@ -0,0 +1,20 @@
+package io.openaev.database.raw;
+
+/**
+ * Spring Data projection interface for tenant data.
+ *
+ * <p>This interface defines a projection for retrieving tenant information including metadata and
+ * timestamps. It serves as a base projection for tenant-related queries and can be extended with
+ * additional fields as needed.
```
RomuDeuxfois: **todo**: comments are not aligned whith current code. We are not retrieving metadata and timestamps but only tenant_id property

openaev-model/src/main/java/io/openaev/engine/model/endpoint/EsEndpoint.java
johanah29: Line 51

PR #5207, opened 2026-03-23
[backend] fix(auth): improve code quality and test coverage (#5206)

### LINKED ISSUE #5206: JWT auth is too verbose, code is redundant, missing test coverage
## Description
A dozen warning-level log lines per auth in the new platform JWT auth, plus the code is duplicating functionality and does not integrate well with the auth middleware.

Missing tests.

Trusted issuer is marked as "filigran-copilot"

# Code quality improvements
* Remove the ad hoc crypto handling and use JJWT (our JWT handling library)
* Move to a new class implementing ExtractorBase
* Integrate within main auth token handling process `private String parseAuthorization(String value)` as one of the fallback attempts
* Consider adding config value for trusted issuer; need confirmation

Note that there isn't currently a known agreement on the format of the secret configuration key that verifies the JWT signature. We need to clarify that first.

# Excessive verbosity
* Downgrade the logs to debug
* Use token interpolation rather than concatenation

# Test coverage
* Implement tests

antoinemzs:
### Proposed changes

* Fix asymmetrical JWT signature verification
* Reconcile auth methods and eliminate duplicated code
* Add tests

### Testing Instructions

1. Usual regression testing: cookie auth, plain API token auth (integrations), OpenAEV Coverage connector auth with OpenCTI
2. Configure XTM One integration
3. Generate a key pair, configure the encoded `Base64(DER(public key))` as the "token"
4. Generate and sign (with the private key) a JWT with email claim ("email") for existing OpenAEV user
5. Authenticate against the API using header `Authorization: Bearer <jwt>`

### Related issues

* Closes #5206

### Checklist

### Further comments


openaev-api/src/main/java/io/openaev/security/error/AuthenticationError.java
```diff
@@ -0,0 +1,7 @@
+package io.openaev.security.error;
+
+public class AuthenticationError extends Exception {
```
RomuDeuxfois, line 3: **Nitpick**: Spring already provides AuthenticationException; would that be sufficient?

openaev-api/src/main/java/io/openaev/security/token/PlatformJwtExtractor.java
```diff
@@ -0,0 +1,68 @@
...
+
+    Jws<Claims> jws = Jwts.parser().verifyWith(pubkey).build().parseSignedClaims(value);
+
+    Claims claims = jws.getPayload();
+
+    if (!TRUSTED_ISSUERS.contains(claims.getIssuer())) {
+      throw new JwtException("Issuer not trusted.");
+    }
+
+    String emailClaim = claims.get("email", String.class);
```
RomuDeuxfois, line 50: **nitpick**: We lost some specific errors from the old code.

openaev-api/src/main/java/io/openaev/security/token/PlatformJwtExtractor.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /*
+     JWT is assumed signed with the EdDSA algorithm as per Filigran standard.
+     The configured key material retrieved with `XtmOneConfig.getToken()` must
+     be an ASN.1/DER encoded public key, itself Base64-encoded.
+    */
+    X509EncodedKeySpec keySpec =
+        new X509EncodedKeySpec(Base64.getDecoder().decode(xtmOneConfig.getToken()));
+    KeyFactory keyFactory = KeyFactory.getInstance("EdDSA");
```
RomuDeuxfois: **todo**: Can you us Jwts.SIG.EdDSA const ?

antoinemzs: better yet I think we should read the `alg` header from the JWT itself (possibly vaildate against a known list of algorithms)

antoinemzs: would use constants too, sure, thanks for this :+1:

openaev-api/src/main/java/io/openaev/security/token/PlatformJwtExtractor.java
antoinemzs: Verifying the JWT syntax and signature is handled by the parser invocation with specific error messages (we definitely shouldn't be reimplementing this sort of thing ourselves), so no loss in this regard.

I will put a special handling with retrieving the email claim though, good call.

All other possibly unexpected exceptions are logged by the caller with cause.

openaev-api/src/main/java/io/openaev/security/error/AuthenticationError.java
antoinemzs: I contemplated this, but made the call; they are very tied to occurrences in Spring Security and I don't really have the correct context to fill them out, e.g. the `Authentication` object: 
```
public abstract class AuthenticationException extends RuntimeException {

	@Serial
	private static final long serialVersionUID = 2018827803361503060L;

	private Authentication authenticationRequest;

	/**
	 * Constructs an {@code AuthenticationException} with the specified message and root
	 * cause.
	 * @param msg the detail message
	 * @param cause the root cause
	 */
	public AuthenticationException(String msg, Throwable cause) {
		super(msg, cause);
	}

	/**
	 * Constructs an {@code AuthenticationException} with the specified message and no
	 * root cause.
	 * @param msg the detail message
	 */
	public AuthenticationException(String msg) {
		super(msg);
	}

	/**
	 * Get the {@link Authentication} object representing the failed authentication
	 * attempt.
	 * <p>
	 * This field captures the authentication request that was attempted but ultimately
	 * failed, providing critical information for diagnosing the failure and facilitating
	 * debugging
	 * @since 6.5
	 */
	public Authentication getAuthenticationRequest() {
		return this.authenticationRequest;
	}
```
Perhaps my own error class is not best named as such? Or should we use a java generic exception type?

openaev-api/src/main/java/io/openaev/security/token/PlatformJwtExtractor.java
antoinemzs: Done in 55c358201

openaev-api/src/main/java/io/openaev/security/token/PlatformJwtExtractor.java
antoinemzs: Added exception in 55c358201

openaev-api/src/main/java/io/openaev/security/error/AuthenticationError.java
antoinemzs: I've kept the in-house exception unless there is a strong objection? Not too keen on using RuntimeException or the Spring Security collection of exceptions

RomuDeuxfois: I didn’t investigate the AuthenticationException in depth, so I’m fine with your implementation.

RomuDeuxfois: Thanks for the clarification.

PR #5211, opened 2026-03-23
[backend] fix(security-coverage): return current simulation security platforms (#5109)

### LINKED ISSUE #5109: Security platform score in stix bundle send to OCTI is incorrect

gabriel-peze:
### Proposed changes

* Fix Security Coverage bundle generation to return only used security platforms into the bundle, instead of all.

### Testing Instructions

1. Create an automatic Scenario from an OCTI instance
2. Run the Scenario
3. Only the security platform used in the simulation should be returned to OCTI

### Related issues

* Closes #5109

### DIFF
--- a/openaev-api/src/main/java/io/openaev/service/stix/SecurityCoverageService.java
+++ b/openaev-api/src/main/java/io/openaev/service/stix/SecurityCoverageService.java
@@ -14,7 +14,6 @@ import io.openaev.aop.lock.Lock;
 import io.openaev.aop.lock.LockResourceType;
 import io.openaev.config.OpenAEVConfig;
 import io.openaev.database.model.*;
-import io.openaev.database.repository.PayloadRepository;
 import io.openaev.database.repository.ScenarioRepository;
 import io.openaev.database.repository.SecurityCoverageRepository;
 import io.openaev.opencti.connectors.impl.SecurityCoverageConnector;
@@ -26,7 +25,6 @@ import io.openaev.rest.inject.service.InjectService;
 import io.openaev.rest.settings.PreviewFeature;
 import io.openaev.rest.tag.TagService;
 import io.openaev.rest.vulnerability.service.VulnerabilityService;
-import io.openaev.service.AssetService;
 import io.openaev.service.PreviewFeatureService;
 import io.openaev.service.scenario.ScenarioService;
 import io.openaev.service.stix.error.BundleValidationError;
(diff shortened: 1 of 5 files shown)


openaev-api/src/main/java/io/openaev/utils/inject_expectation_result/ExpectationResultBuilder.java
```diff
@@ -185,12 +191,19 @@ private static InjectExpectationResult buildForMediaPressure(
...
         .result(result)
         .date(Instant.now().toString())
         .score(score)
         .build();
   }
 
+  private static String getSourceAssetId(Collector collector) {
+    return collector.isExternal() && collector.getSecurityPlatform() != null
+        ? collector.getSecurityPlatform().getId()
+        : collector.getId();
```
antoinemzs: Let's not put the collector ID here if it's not related to an actual asset (i.e. security platform)

Leave null if there isn't an asset related to the collector

openaev-api/src/main/java/io/openaev/service/stix/SecurityCoverageService.java
```diff
@@ -565,6 +567,15 @@ private void processCoverageRefs(
...
   }
 
+  private Set<String> getSecurityPlatformAssetIds(List<Inject> injects) {
+    return injects.stream()
+        .flatMap(inject -> inject.getExpectations().stream())
+        .flatMap(expectation -> expectation.getResults().stream())
+        .map(InjectExpectationResult::getSourceAssetId)
+        .filter(Objects::nonNull)
+        .collect(Collectors.toSet());
+  }
```
antoinemzs: nitpick: favour encapsulation by not drilling down the attributes but have service methods return the correct information, and each service calls the lower level service if needed

Issue #4864, opened 2026-02-03
Implement core multi tenancy


jborozco:
## Context

With a platform in EE:
- I have CRUD capa to manage multiple tenants
- I manage my platform settings
- I manage my SSO configurations
- As a user invited in multiple tenants I can switch from one tenant to the other

RomuDeuxfois:
Some bugs raised and we are working on it:
- injector contract exception error on grafana main-ff
- xtm composer is not working propery on multi-tenancy
- expiration manager is not multi-tenant
- ...

PR #5214, opened 2026-03-23
[backend/frontend] feat(tenant-switcher): add tenant switcher API (#4864)

RomuDeuxfois:
### Proposed changes

* Wire tenant switcher to real backend API — replaces mock data
* Auto-link creator to tenant on creation via `DependenciesManager` pattern
* Tenant switcher reloads on create/delete, handles empty state gracefully

### Testing Instructions

1. Create a tenant → appears immediately in switcher with check mark
2. Switch tenant → check mark moves, persists after reload
3. Delete selected tenant → switcher auto-selects another or shows placeholder
4. Delete last tenant → switcher shows "No tenant available" (disabled)
5. Create first tenant from empty state → auto-selected immediately

### Checklist

### DIFF
--- a/openaev-api/src/main/java/io/openaev/service/tenants/UserTenantService.java
+++ b/openaev-api/src/main/java/io/openaev/service/tenants/UserTenantService.java
@@ -12,7 +12,7 @@ import org.springframework.stereotype.Service;
 @Slf4j
 @Service
 @RequiredArgsConstructor
-public class UserTenantDependencyManager implements DependenciesManager {
+public class UserTenantService implements DependenciesManager {
 
   private final TenantRepository tenantRepository;
 
--- a/openaev-front/src/utils/hooks/useTenant.ts
+++ b/openaev-front/src/utils/hooks/useTenant.ts
@@ -45,6 +45,8 @@ const useTenant = (me: User | undefined, logged: unknown) => {
     }
   }, [me, logged, loadUserTenants]);
 
+  // TODO multi-tenancy: Multi executors dev
+  // When switching tenants we need to navigate to the new tenant URL prefix and reload tenant-scoped data
   const switchUserTenant = useCallback(async (tenantId: string) => {
     if (tenantId === currentUserTenant?.tenant_id) {
       return;
--- a/openaev-front/src/admin/components/platform/tenants/tenant/TenantUpdate.tsx
+++ b/openaev-front/src/admin/components/platform/tenants/tenant/TenantUpdate.tsx
@@ -5,6 +5,7 @@ import Drawer from '../../../../../components/common/Drawer';
 import { useFormatter } from '../../../../../components/i18n';
 import { type TenantInput, type TenantOutput } from '../../../../../utils/api-types';
 import { useAppDispatch } from '../../../../../utils/hooks';
+import useAuth from '../../../../../utils/hooks/useAuth';
 import TenantForm from './TenantForm';
 
 interface Props {
@@ -23,6 +24,7 @@ const TenantUpdate: FunctionComponent<Props> = ({
   // Standard hooks
   const { t } = useFormatter();
   const dispatch = useAppDispatch();
+  const { reloadUserTenants } = useAuth();
 
   // Form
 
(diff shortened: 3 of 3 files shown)


openaev-api/src/main/java/io/openaev/service/tenants/UserTenantDependencyManager.java
```diff
@@ -0,0 +1,28 @@
...
+import io.openaev.database.repository.TenantRepository;
+import io.openaev.multitenancy.DependenciesManager;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Service;
+
+@Slf4j
+@Service
+@RequiredArgsConstructor
+public class UserTenantDependencyManager implements DependenciesManager {
```
corinnekrych: Suggestion: I find the name confusing I'd not used `Dependency` int he context of dealing with user-tenant relationship. suggested names: `UserTenantService `, `TenantMembershipService`

openaev-api/src/main/java/io/openaev/service/tenants/UserTenantService.java
```diff
@@ -0,0 +1,28 @@
...
+  private final TenantRepository tenantRepository;
+
+  @Override
+  public void createDependencyForTenant(Tenant tenant) {
+    tenantRepository.addUserToTenant(currentUser().getId(), tenant.getId());
+  }
+
+  @Override
+  public void deleteDependencyForTenant(String tenantId) {
+    // users_tenants rows are cascade-deleted via FK on tenants table
```
corinnekrych, line 26: I think @damgouj suggested a pattern for to do like: `// TODO multi-tenancy:`

openaev-front/src/admin/components/platform/tenants/tenant/TenantCreate.tsx
```diff
@@ -34,6 +36,7 @@ const TenantCreate: FunctionComponent<Props> = ({ onCreate }) => {
 
       const createdTenant = result.entities.tenants[result.result];
       onCreate(createdTenant);
+      await reloadUserTenants();
```
corinnekrych, line 39: Edge case:
1/ i create a tenat `titi`
2/ the teant user can see `titi` in the list of available tenant for a giver user in top bar switcher
3/ update titi to `titi2` => name is not update

we should reload in TeantUpdate

openaev-front/src/utils/hooks/useTenant.ts
```diff
@@ -51,20 +50,18 @@ const useTenant = (me: User | undefined, logged: unknown) => {
...
-      }
-      // TODO multi-tenancy: tenant routing
-      // window.location.replace(window.location.href);
-    }, 0);
-  }, [currentUserTenant, userTenants]);
+    const current = userTenants.find(t => (t.tenant_id === tenantId));
+    if (current) {
+      setCurrentUserTenant(current);
+      setCurrentTenantStorage(current);
+    }
```
corinnekrych, line 59: I think there is a need for a TODO here when we do the routing we will need to relaod the page, correct?

openaev-api/src/main/java/io/openaev/service/tenants/UserTenantDependencyManager.java
RomuDeuxfois: I agree, I will made the changes on the next commit.

openaev-api/src/main/java/io/openaev/service/tenants/UserTenantService.java
RomuDeuxfois: This will not be handled later, since it is already part of the original behavior. Therefore, there is no need to implement this part.
Perhaps the comment is not clear enough?

openaev-front/src/utils/hooks/useTenant.ts
RomuDeuxfois: That’s a good catch. I’ll add it.

openaev-front/src/admin/components/platform/tenants/tenant/TenantCreate.tsx
RomuDeuxfois: That’s a good catch. I’ll add it.

openaev-api/src/main/java/io/openaev/service/tenants/UserTenantService.java
corinnekrych: ah ok. I was wondering what is the need to have deleteDependencyForTenant in the interface DependenciesManager if no implementation is required?

RomuDeuxfois: It’s supported by MinIO, but you’re right that we don’t always implement it.
To me, this should be treated as a contract for developers: creation and deletion should be implemented, but if they aren’t, we simply skip them.

PR #5221, opened 2026-03-24
feat(audit-log): PoC to validate log event format and Search engine log (#5241)

corinnekrych:
### Proposed changes

This is a PoC for Audit log event format, the goal is: 
* capture user login, logout (administration for event_access)
* capture mutation event on entity (extended for event_access)
* to match OCTI format as much as possible,
* capture input/old_value in the log in `context_data` to capture preview/new value

### Testing Instructions

Pre-requisite:
payload created BOO
payload created ECHO Corinne

Use case: create start a scenario and launch it
1/ user login
2/ user creates a scenario “tuesday24-5”
3/ user adds an inject “BOO” to the scenario
4/ user adds an inject “ECHO Corinne” to the scenario
5/ user deletes ECHO Corinne
6/ user start scenario from launch button
7/ user logout

See logs file in log/audit.log. Here is an attached version
[audit.log](https://github.com/user-attachments/files/26208565/audit.log)

This PoC also include Elastic Search PoC
http://localhost:9200/openaev_audit-log-000001/_search?pretty  => to watch log event

### Related issues

* Closes #ISSUE-NUMBER
*

### Checklist

### Further comments


a19836:
Maybe I misunderstood this method, or perhaps I haven’t fully understood all the logic behind it yet. I’m also looking at this only from a very high-level perspective and without running any tests so far, but at first glance it seems that `AuditLogAspect.detectChildResource` may not be fully reliable because it extracts the entity ID and then tries to infer the entity type based solely on that value.

Since these IDs are UUIDs generated automatically, all entities follow the same ID generation process. This means that, although unlikely, the same UUID could theoretically appear in different tables. If that happens, `detectChildResource` may return the wrong entity type, and all the subsequent logic inside `AuditLogAspect` could fail because it depends on the returned `ChildResourceInfo`.

A more technical explanation:

If `paramValue` is treated as a generic identifier that could potentially exist across multiple entity types, then this approach becomes fragile because we may resolve the wrong entity.

A more robust solution would probably be to infer the entity type from the URL pattern, controller mapping, method signature, method return type, or even the `AccessControl` annotation, instead of relying only on the ID value. That said, this would require more complex logic and additional assumptions about the API structure.

For now, the implementation seems to rely on the assumption that path variable IDs are globally unique across entity types, which is usually true with UUIDs — but not something we can guarantee forever. If this ever becomes a real issue, we may need to revisit this logicå and introduce additional heuristics or metadata to disambiguate entity types.

A tiny edge case today… a future debugging adventure tomorrow 😄

If my understanding is correct, this method is probably fine for now given that this is still a POC. However, during the US1 implementation, it may be worth considering improving it to make the resolution logic a bit more robust and future-proof.

a19836:
For the US1 implementation, in the `AuditLogService.populateUserMetadata` method, we should probably also consider the following improvements:
- Encrypt the email according to the GDPR requirements defined in the NAG guidelines - unless **those rules apply only to emails stored inside entity objects?**  
- Improve client IP detection so we don’t rely only on `X-Forwarded-For`, `X-Real-IP`, and `REMOTE_ADDR`, but also consider additional headers such as:  
  `Forwarded`, `CF-Connecting-IP`, `True-Client-IP`, `Fastly-Client-IP`, `X-Azure-ClientIP`, `Client-IP`, `Proxy-Client-IP`, `WL-Proxy-Client-IP`, `HTTP_CLIENT_IP`, and `HTTP_X_FORWARDED_FOR`.

Just a couple of small hardening ideas for future-us to thank present-us later 😄

a19836:
In `AuditLogAspect.auditAround`, around line 216, the condition:

```java
if ("update".equals(eventScope))
```

made me wonder if it should actually be:

```java
if ("create".equals(eventScope))
```

If I understood the flow correctly, this section seems to handle bulk child creation, where we need to process each individual entity separately. If that’s the case, then the `eventScope` would logically be `create` rather than `update`, correct?

Maybe I’m missing part of the intended behavior, but at first glance it feels slightly suspicious... :)

Issue #4891, opened 2026-02-06
XLS import creates empty expectations on injects


Lhorus6:
## Description

When using the XLS import feature to create injects, it is possible to end up with injects that contain an expectation with no values (no name, no description, no score). This should not be possible through the UI or normal workflows.

In addition to being invalid from a data perspective, this behavior introduces an issue: the simulation never ends and remains stuck in an "ongoing" state.

## Environment

OAEV 2.0.14

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Create a mapper with a unique inject: Send Individual Mail.
2. Map only the mandatory fields for the inject (do not map any expectation-related fields).
3. Create a new simulation.
4. Go to the Injects tab.
5. Import an XLS file (created beforehand to match your mapper) that creates a mail inject.

- The XLS file itself does not define any expectations.
- The mapper does not map any expectations either.
-> But after the import, inspect the created mail inject. The inject contains an expectation object, but it is empty: No name, no description, no success score

<img width="3293" height="760" alt="Image" src="https://github.com/user-attachments/assets/3f8e86c4-781d-4343-b4c9-49c33ce94bf2" />

## Expected Output

- If no expectation is defined in the mapper or the XLS file, no expectation should be created on the inject.

## Actual Output

- An empty expectation is automatically created on the inject.

## Additional information

In addition to creating an empty expectation, there is an inconsistency in how the expiration time is handled. When inspecting the expectation before running the simulation, the expiration time is 0 days, 0 hours, 0 minutes.

<img width="1301" height="157" alt="Image" src="https://github.com/user-attachments/assets/a91c37d5-1e77-408d-808a-5ee28029ba97" />

Which suggest that the expiration time is configured to expire directly after the execution of the inject.

However, after the simulation is executed, if we inspect the inject logs, the expectation of the executed inject appears to expire 24 hours later, as if a value of 1 day had been applied.

<img width="1625" height="312" alt="Image" src="https://github.com/user-attachments/assets/f317d28d-d7f8-4568-98f6-e2525a89c9db" />

This behavior introduces an additional issue:
- The simulation remains in an "ongoing" state until all expectations are expired. As a result, even though no expectation was intended, the simulation stays open for 24 hours after the last inject execution. This is unexpected, as the simulation should have completed immediately after the last mail was sent.

PR #5236, opened 2026-03-24
[backend] fix(inject): xls import creates empty expectations causing 24h stuck simulations (#5229)

RomuDeuxfois:
### Proposed changes

* Filter on not valid expectation

### Testing Instructions

Follow instructions on github issue.

### Related issues

* Closes https://github.com/OpenAEV-Platform/openaev/issues/4891

### Checklist

### Further comments

### DIFF
--- a/openaev-front/yarn.lock
+++ b/openaev-front/yarn.lock
@@ -7046,9 +7046,9 @@ __metadata:
   linkType: hard
 
 "flatted@npm:^3.2.9":
-  version: 3.4.1
-  resolution: "flatted@npm:3.4.1"
-  checksum: 10c0/3987a7f1e39bc7215cece001354313b462cdb4fb2dde0df4f7acd9e5016fbae56ee6fb3f0870b2150145033be8bda4f01af6f87a00946049651131bbfca7dfa6
+  version: 3.4.2
+  resolution: "flatted@npm:3.4.2"
+  checksum: 10c0/a65b67aae7172d6cdf63691be7de6c5cd5adbdfdfe2e9da1a09b617c9512ed794037741ee53d93114276bff3f93cd3b0d97d54f9b316e1e4885dde6e9ffdf7ed
   languageName: node
   linkType: hard
 


openaev-api/src/main/java/io/openaev/utils/InjectImportUtils.java
```diff
@@ -20,8 +21,39 @@ public final class InjectImportUtils {
...
+   */
+  public static boolean isExpectationValid(InjectExpectation expectation) {
+    if (expectation == null) {
+      return false;
+    }
+    boolean hasName = expectation.getName() != null && !expectation.getName().isBlank();
+    boolean hasDescription =
+        expectation.getDescription() != null && !expectation.getDescription().isBlank();
+    boolean hasScore = expectation.getExpectedScore() != null && expectation.getExpectedScore() > 0;
+    return hasName || hasDescription || hasScore;
```
antoinemzs: No need for assignment
```suggestion
return !StringUtils.isBlank(expectation.getName())
  || !StringUtils.isBlank(expectation.getDescription())
  || (expectation.getExpectedScore() != null && expectation.getExpectedScore() > 0);
```

openaev-api/src/main/java/io/openaev/service/InjectImportService.java
```diff
@@ -765,7 +765,7 @@ private ImportRow importRow(
     // No dependencies
     inject.setDependsOn(null);
 
-    if (expectation.get() != null) {
+    if (expectation.get() != null && InjectImportUtils.isExpectationValid(expectation.get())) {
```
antoinemzs: nitpick, to use the `Optional<T>` interface
```suggestion
    if (expectation.isPresent() && InjectImportUtils.isExpectationValid(expectation.get())) {
```

openaev-api/src/main/java/io/openaev/utils/InjectImportUtils.java
```diff
@@ -20,8 +21,39 @@ public final class InjectImportUtils {
...
+  // -- VALIDATION: RULE ATTRIBUTES --
+
   /**
-   * Returns the date as string from an excel cell
+   * Checks whether a rule attribute has usable data, i.e. mapped columns or a non-blank default
+   * value.
+   */
+  public static boolean hasColumnsOrDefaultValue(RuleAttribute ruleAttribute) {
+    return ruleAttribute.getColumns() != null
+        || (ruleAttribute.getDefaultValue() != null && !ruleAttribute.getDefaultValue().isBlank());
```
antoinemzs: Should `getColumns()` also not be empty to satisfy "has columns"?
```
import io.openaev.utils.StringUtils;
```
```suggestion
    return (ruleAttribute.getColumns() != null && !ruleAttribute.getColumns().isEmpty())
        || !StringUtils.isBlank(ruleAttribute.getDefaultValue());
```

openaev-api/src/main/java/io/openaev/utils/InjectImportUtils.java
```diff
@@ -20,8 +21,39 @@ public final class InjectImportUtils {
...
+  }
+
+  // -- VALIDATION: EXPECTATIONS --
+
+  /**
+   * Validates that an expectation has meaningful data. An expectation without at least a name,
+   * description, or positive score is considered empty/invalid and should not be persisted — see
+   * OpenAEV-Platform/openaev#4891.
+   */
+  public static boolean isExpectationValid(InjectExpectation expectation) {
```
antoinemzs: This looks like it could be part of the `InjectExpectation` class, or at least the `InjectExpectationService` class to include the null check (but not mandatory since you are already nullchecking on an `Optional<InjectExpectation>` at the call site for example), to preserve encapsulation

openaev-api/src/main/java/io/openaev/utils/InjectImportUtils.java
```diff
@@ -20,8 +21,39 @@ public final class InjectImportUtils {
...
   private InjectImportUtils() {}
 
+  // -- VALIDATION: RULE ATTRIBUTES --
+
   /**
-   * Returns the date as string from an excel cell
+   * Checks whether a rule attribute has usable data, i.e. mapped columns or a non-blank default
+   * value.
+   */
+  public static boolean hasColumnsOrDefaultValue(RuleAttribute ruleAttribute) {
```
antoinemzs: Could this be part of the `RuleAttribute` class?

openaev-api/src/main/java/io/openaev/service/InjectImportService.java
```diff
@@ -931,83 +931,85 @@ private List<ImportMessage> addFields(
...
-                      new ImportMessage(
-                          ImportMessage.MessageLevel.WARN,
-                          ImportMessage.ErrorCode.EXPECTATION_SCORE_UNDEFINED,
-                          Map.of(
-                              "column_type_num",
-                              String.join(", ", columns),
-                              "row_num",
-                              String.valueOf(row.getRowNum()))));
-                  return importMessages;
+        if (InjectImportUtils.hasColumnsOrDefaultValue(ruleAttribute)) {
```
antoinemzs: You should break early to avoid wrapping the long method in the `if` statement; also helps with readability, including the diff 🙏

openaev-api/src/test/java/io/openaev/service/InjectImportServiceTest.java
```diff
@@ -300,4 +310,45 @@ void testGetInjectTimeUndetectedWithTimePattern() throws Exception {
...
+        InjectImportService.class.getDeclaredMethod(
+            "addFields",
+            Inject.class,
+            RuleAttribute.class,
+            Row.class,
+            Map.class,
+            AtomicReference.class,
+            ImportMapper.class,
+            Map.class);
+    addFieldsMethod.setAccessible(true);
```
antoinemzs: We shouldn't let the bot output tests that do this sort of thing, i.e. breaking visibility with reflection. Tests should use the public interface unless there is a very good, documented reason why we don't, and document the private method in production code that there are fragile tests that will break if we change it.

openaev-api/src/test/java/io/openaev/utils/mockMapper/MockMapperUtils.java
```diff
@@ -46,4 +52,30 @@ public static RuleAttribute createRuleAttribute() {
...
+    injectorContract.setContent("{\"fields\":[]}");
+    ObjectNode convertedContent = MAPPER.createObjectNode();
+    convertedContent.set("fields", MAPPER.createArrayNode());
+    injectorContract.setConvertedContent(convertedContent);
+
+    Inject inject = new Inject();
+    inject.setContent(MAPPER.createObjectNode());
+    inject.setInjectorContract(injectorContract);
+    return inject;
+  }
```
antoinemzs: nitpick: I would prefer these methods to be in a `*Fixture.java` file so that others find them easily; and there might be a satisfying fixture existing already?

openaev-api/src/main/java/io/openaev/service/InjectImportService.java
RomuDeuxfois: This is not an optional but a AtomicReference<InjectExpectation>

PR #5242, opened 2026-03-25
[backend/frontend] feat(multi-tenancy): e2e for tag entity (#4864)

RomuDeuxfois:
### Proposed changes

* Turn tag API to use multi-tenancy mechanism

### Testing Instructions

1. Connect to the application.
2. Create a tenant.
3. Create a tag and verify that it appears.
4. Create another tenant.
5. Verify that the previously created tag does not appear, and that tags can only be created for the currently selected tenant.

### Related issues

* Related to https://github.com/OpenAEV-Platform/openaev/issues/4864

### DIFF
--- a/openaev-api/src/main/java/io/openaev/api/tenants/TenantApi.java
+++ b/openaev-api/src/main/java/io/openaev/api/tenants/TenantApi.java
@@ -84,17 +84,36 @@ public class TenantApi {
     return toOutput(tenantService.update(tenantId, TenantMapper.fromInput(tenantId, input)));
   }
 
+  @Operation(
+      summary = "Reactivate a soft-deleted tenant",
+      description =
+          "Reactivates a previously soft-deleted tenant within the 30-day grace period."
+              + " Fails if the grace period has expired.")
+  @AccessControl(
+      resourceId = "#tenantId",
+      actionPerformed = Action.WRITE,
+      resourceType = ResourceType.TENANT,
+      isEnterpriseEdition = true)
+  @PostMapping("/{tenantId}/reactivate")
+  public TenantOutput reactivate(@PathVariable String tenantId) {
+    return toOutput(tenantService.reactivate(tenantId));
+  }
+
   // -- DELETE --
 
-  @Operation(summary = "Delete a tenant", description = "Deletes a tenant by its ID")
+  @Operation(
+      summary = "Soft-delete a tenant",
+      description =
+          "Marks a tenant as deleted. Data is preserved for 30 days."
+              + " An admin can reactivate the tenant within this grace period."
+              + " After 30 days, the tenant and all associated data are permanently removed.")
   @AccessControl(
       resourceId = "#tenantId",
       actionPerformed = Action.DELETE,
       resourceType = ResourceType.TENANT,
       isEnterpriseEdition = true)
   @DeleteMapping("/{tenantId}")
-  @ResponseStatus(HttpStatus.NO_CONTENT)
-  public void delete(@PathVariable String tenantId) throws DependenciesManagerException {
-    tenantService.delete(tenantId);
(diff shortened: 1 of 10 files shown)


openaev-front/scripts/auto-translate.js
```diff
@@ -8,7 +8,7 @@ import { DEFAULT_LANG } from './constants/Lang.js';
 const execAsync = promisify(exec);
 
 // deepl-free API key
-const subscriptionKey = process.env.SUBSCRIPTION_KEY;
+const subscriptionKey = '78423651-dfe0-0c59-6b16-92ccfb754c2e:fx';
```
corinnekrych: why do we want to remove env variable process.env.SUBSCRIPTION_KEY ?

line 17 on this file, we still have `  throw new Error('SUBSCRIPTION_KEY environment variable is not set. Aborting.');`

RomuDeuxfois: Sorry, this file should not be committed.

openaev-front/src/utils/tenant-uri.ts
```diff
@@ -0,0 +1,34 @@
...
+  } catch {
+    // malformed JSON — fall back
+  }
+  return DEFAULT_TENANT_UUID;
+};
+
+/**
+ * Builds a tenant-scoped API URI using the active tenant from local storage.
+ */
+export const tenantUri = (path: string, tenantId?: string): string =>
```
corinnekrych: From usage, i got the feeling you never call `tenantUri(PATH, tenantId)`
Suggestion: remove unused tanenatId param, rename currentTenantUri

corinnekrych:
@RomuDeuxfois why openaev-api/src/main/resources/implants/caldera/macos/arm64/oaev-implant-caldera-macos was removed? is it anything to do with this PR's scope?

openaev-front/src/utils/tenant-uri.ts
RomuDeuxfois: You’re right, and we should never call it with a dedicated tenantId. Everything should be handled through the useTenant hook. Thanks for pointing that out.

openaev-front/src/reducers/Referential.ts
```diff
@@ -114,6 +114,11 @@ const referential = (state: any = Map({}), action: any = {}) => {
       return state;
     }
 
+    case Constants.TENANT_SWITCH_SUCCESS: {
```
corinnekrych, line 132: We can group case as we're doing the same reset for both
```suggestion
    case Constants.TENANT_SWITCH_SUCCESS: 
    case Constants.IDENTITY_LOGOUT_SUCCESS: {
```

RomuDeuxfois:
@corinnekrych PR ready for a second round.

openaev-front/src/utils/hooks/useTenant.ts
```diff
@@ -54,10 +93,10 @@ const useTenant = (me: User | undefined, logged: unknown) => {
 
     const current = userTenants.find(t => (t.tenant_id === tenantId));
     if (current) {
-      setCurrentUserTenant(current);
+      setTenant(current);
       setCurrentTenantStorage(current);
```
corinnekrych, line 97: l98 let's add: ` window.location.reload();` so that we can test e2e:
1/ create tenantA: see we land on tenant A
2/ create a taxicomany "aaa"
3/ create tenantb: see we land on tenant B
4/ create a taxicomany "bbb"
5/ switch to tenant A, see page is reloaded and list of label display "aaa" (without this change the reload wont happen and we're stuck to old teant data).

corinnekrych, line 97:
```suggestion
      setCurrentTenantStorage(current);
      window.location.reload();
```

PR #5252, opened 2026-03-26
[backend] feat(multi-tenancy): manage unique constraints and datapack when tenant is created (#4864)

damgouj:
### Proposed changes

* Manage unique constraints to add the tenant_id
* Process datapack when tenant is created

### Testing Instructions

1. Can't be tested yet, waiting for another PR

### Related issues

* #4864 

### Checklist

### DIFF
--- a/openaev-api/src/main/java/io/openaev/migration/V4_80__Update_datapack_key.java
+++ b/openaev-api/src/main/java/io/openaev/migration/V4_80__Update_datapack_key.java
@@ -6,7 +6,7 @@ import org.flywaydb.core.api.migration.Context;
 import org.springframework.stereotype.Component;
 
 @Component
-public class V4_79__Update_datapack_key extends BaseJavaMigration {
+public class V4_80__Update_datapack_key extends BaseJavaMigration {
 
   @Override
   public void migrate(Context context) throws Exception {
--- a/openaev-api/src/main/java/io/openaev/migration/V4_81__Update_unique_constraints_for_tenants.java
+++ b/openaev-api/src/main/java/io/openaev/migration/V4_81__Update_unique_constraints_for_tenants.java
@@ -6,7 +6,7 @@ import org.flywaydb.core.api.migration.Context;
 import org.springframework.stereotype.Component;
 
 @Component
-public class V4_80__Update_unique_constraints_for_tenants extends BaseJavaMigration {
+public class V4_81__Update_unique_constraints_for_tenants extends BaseJavaMigration {
 
   @Override
   public void migrate(Context context) throws Exception {


openaev-model/src/main/java/io/openaev/database/model/DatapackTenantId.java
```diff
@@ -0,0 +1,38 @@
...
+
+  @Column(name = "datapack_id", updatable = false, nullable = false)
+  @JsonProperty("datapack_id")
+  @NotBlank
+  private String id;
+
+  @ManyToOne
+  @JoinColumn(name = "tenant_id", updatable = false, nullable = false)
+  @JsonIgnore
+  private Tenant tenant;
```
RomuDeuxfois, line 28: **question**: Why not using only tenantId here ?

openaev-model/src/main/java/io/openaev/database/repository/DataPackRepository.java
```diff
@@ -12,5 +13,5 @@ public interface DataPackRepository
     extends CrudRepository<DataPack, String>, JpaSpecificationExecutor<DataPack> {
 
   @NotNull
-  Optional<DataPack> findById(@NotNull String id);
+  Optional<DataPack> findByCompositeId(DatapackTenantId datapackTenantId);
```
RomuDeuxfois: **nitpick**: @NotNull can stay

openaev-model/src/main/java/io/openaev/context/TenantContext.java
```diff
@@ -15,6 +15,13 @@ public static String getCurrentTenant() {
     return tenant != null ? tenant : Tenant.DEFAULT_TENANT_UUID;
   }
 
+  /**
+   * DO NOT USE except to set the tenant id from the URL (TenantInterceptor) AND in very specific
```
RomuDeuxfois, line 19: **praise**: Thanks for this clarification

openaev-api/src/test/java/io/openaev/service/tenants/TenantServiceTest.java
```diff
@@ -83,6 +89,13 @@ void should_create_and_find_tenant() throws Exception {
     Session session = entityManager.unwrap(Session.class);
     session.enableFilter("tenantFilter").setParameter("tenantId", created.getId());
     assertThat(domainRepository.findAll()).hasSize(9);
+    // Verify datapack added, can't because tests are disabled
+    /*assertThat(customDashboardRepository.findAll()).hasSize(6);
+    assertThat(scenarioRepository.findAll()).hasSize(3);
+    assertThat(tagRepository.findAll()).hasSize(6);
+    assertThat(tagRuleRepository.findAll()).hasSize(4);
+    assertThat(endpointRepository.findAll()).hasSize(1);
+    assertThat(assetGroupRepository.findAll()).hasSize(7);*/
```
RomuDeuxfois: **question**: Not sure to understand why you comment this line

openaev-api/src/main/java/io/openaev/datapack/packs/V20260101_Starter_pack.java
```diff
@@ -248,15 +237,4 @@ private void setDefaultDashboard(String filename, String dashboardId) {
       settingRepository.save(defaultDashboardSetting);
     }
   }
-
-  private void createSetting() {
```
RomuDeuxfois, line 252: **question** Why removing all this logic ?

openaev-api/src/main/java/io/openaev/datapack/packs/V20260107_Tags_and_tagrules_and_assetgroups.java
```diff
@@ -90,5 +90,6 @@ public void doProcess() {
                 });
       }
     }
+    return true;
```
RomuDeuxfois: **question**: In which case are we returning false ? If we always return the same value, why returning something ? Try catch concept is maybe more accurate in this case.

openaev-api/src/main/java/io/openaev/datapack/packs/V20260101_Starter_pack.java
```diff
@@ -86,20 +85,12 @@ private static final class AllEndpointsAssetGroup {
...
     // early break for when the starter pack was already run
     if (!isStarterPackEnabled) {
       log.info("Starter pack is disabled by configuration");
-      return;
-    }
-
-    if (this.settingRepository.findByKey(Config.STARTER_PACK_KEY).isPresent()) {
-      log.info("Starter pack already initialized");
-      return;
+      return false;
```
antoinemzs, line 92: This is important to support installations that had run the Starter Pack initally before the data packs were a thing, so that there is no replay, I suppose it's better if we keep this around.

openaev-api/src/main/java/io/openaev/datapack/packs/V20260101_Starter_pack.java
antoinemzs: This might not be needed anymore indeed. This was preventing the starter pack replay in the inital design, but now the replays are prevented by the generic data pack logic.

openaev-model/src/main/java/io/openaev/database/model/DatapackTenantId.java
damgouj: I do the same for each database java object and it's the Hibernate recommendation. When it's a foreign key, the java object want the object and not a String.

openaev-api/src/test/java/io/openaev/service/tenants/TenantServiceTest.java
damgouj: I would like to test the Datapack when we create a tenant but I can't. There is the annotation @ Profile("!test") on the DatapackProcessor so the dependency is not taken into account during the tests. It's just to think about it and to remember it is initialized but I can remove or clarify the comment if you want ?

openaev-api/src/main/java/io/openaev/datapack/packs/V20260101_Starter_pack.java
damgouj: Because it's not useful anymore. We already set on the DataPack table when a DataPack migration is done, the settings to say "Datapack is done" are deprecated.

openaev-api/src/main/java/io/openaev/datapack/packs/V20260101_Starter_pack.java
damgouj: Good call, thanks !

openaev-api/src/test/java/io/openaev/service/tenants/TenantServiceTest.java
RomuDeuxfois: In my opinion, it's better to remove this kind of code to avoid confusion.

openaev-model/src/main/java/io/openaev/database/model/DataPack.java
```diff
@@ -16,14 +14,28 @@
...
-  @JsonProperty("datapack_id")
-  @NotBlank
-  private String id;
 
-  @ManyToOne
-  @JoinColumn(name = "tenant_id", updatable = false, nullable = false)
-  @JsonIgnore
-  private Tenant tenant;
+  @EmbeddedId
+  @JsonProperty("datapackTenantId_relationship")
```
RomuDeuxfois: **todo**: should be standardized datapack_tenant_id_relationship

openaev-api/src/main/java/io/openaev/datapack/packs/V20260101_Starter_pack.java
```diff
@@ -131,11 +127,11 @@ protected void doProcess() {
 
       this.importScenariosFromResources(honeyScanMeEndpoint, allEndpointAssetGroup);
       this.importDashboardsFromResources();
+      return true;
     } catch (Exception e) {
-      recordError("Unexpected error during StarterPack initialization.", e);
+      recordError("Unexpected error during DataPack 20260101 initialization.", e);
```
RomuDeuxfois: **todo**: Remove recordError and just log it ?

openaev-api/src/main/java/io/openaev/migration/V4_80__Update_datapack_key.java
```diff
@@ -0,0 +1,52 @@
...
+                        ALTER TABLE datapacks
+                        ADD CONSTRAINT datapacks_pkey UNIQUE (datapack_id, tenant_id);
+                        """);
+      statement.execute(
+          """
+                        DELETE FROM parameters WHERE parameter_key = 'starterpack';
+                        """);
+      statement.execute(
+          """
+              ALTER TABLE scenarios_teams
```
RomuDeuxfois, line 30: **question**: We miss this case at first right ? What about user -> team releation ? everything is already ON CASCADE ?

damgouj: Yes you are right, I don't know how to proceed ? Maybe we need to check all the many<-->many relations to check that all is cascading ? In a new PR ?

RomuDeuxfois: Let's have a look in a new PR yes.

openaev-api/src/main/java/io/openaev/datapack/packs/V20260107_Tags_and_tagrules_and_assetgroups.java
antoinemzs: @damgouj I'm interested to know the rationale behind this change, as it remained with no explanation :) The design goal is that if there wasn't an exception, the datapack is reputed passed.

openaev-api/src/test/java/io/openaev/service/tenants/TenantServiceTest.java
antoinemzs: Yes the `DatapackProcessor` bean is not loaded at test time to avoid importing the datapacks in the test database.
It's always possible to intantiate it explicitly and call the `process()` method though, so you can test whether all fixtured tenants are being served the correct datapacks.

antoinemzs: Oh I see the new system (sorry ignore the previous comment). You would indeed test it in several locations.
1. Testing the `TenantService` iterating over `List<DependenciesManager>`, with a specific mock `DependenciesManager` bean that you can verify was called, on your required event (tenant creation)
2. Asserting that `DatapackProcessor` implements `DependenciesManager`
3. Testing that DatapackProcessor indeed does iterate over all datapacks and runs them (already implemented?)
4. Testing that individual datapacks create the required dataset (already implemented?)

Alternatively, you may run a spring boot test under another profile name (i.e. not "test"), so that all the beans load.

damgouj:
Manage unique constraints -> before, unique on the name. After, unique on the name and tenant id to manage for example one Admin role in each tenant (2 lines in the database).
Datapack -> execute datapack when a tenant is created to migrate the first data.

PR #5301, opened 2026-03-31
[backend] feat(multi-tenancy): add missing migration when a tenant is created (#4864)

damgouj:
### Proposed changes

* Add missing datapack migration to add roles/groups/user admin/vulnerabilities/cwes to the tenant created
* Add migration to update the unique external id for vulnerabilities and cwes

### Testing Instructions

1. Create a tenant
2. Check if the tables are filled with the data
3. Nothing else must change

### Related issues

* #4864

### Checklist

### DIFF
--- a/openaev-api/src/main/java/io/openaev/datapack/PresetTenantData.java
+++ b/openaev-api/src/main/java/io/openaev/datapack/PresetTenantData.java
@@ -28,6 +28,7 @@ public class PresetTenantData {
                   .cisaActionDue(Instant.parse("2021-03-02T00:00:00Z"))
                   .description("Microsoft Exchange Server ProxyLogon SSRF leading to RCE.")
                   .cvssV31(new BigDecimal("9.8"))
+                  .vulnStatus(Vulnerability.VulnerabilityStatus.ANALYZED)
                   .cisaRequiredAction("Apply updates per vendor instructions")
                   .cisaVulnerabilityName(
                       "Microsoft Exchange Server Remote Code Execution Vulnerability")
@@ -48,6 +49,7 @@ public class PresetTenantData {
                   .description(
                       "Cisco IOS XE Web UI remote code execution via unauthenticated command injection.")
                   .cvssV31(new BigDecimal("10.0"))
+                  .vulnStatus(Vulnerability.VulnerabilityStatus.ANALYZED)
                   .cisaRequiredAction("Verify compliance with BOD 23-02 and apply mitigations.")
                   .cisaVulnerabilityName("Cisco IOS XE Web UI Privilege Escalation Vulnerability")
                   .referenceUrls(
@@ -66,6 +68,7 @@ public class PresetTenantData {
                   .cisaActionDue(Instant.parse("2024-01-12T00:00:00Z"))
                   .description("Ivanti Connect Secure/Policy Secure gateway authentication bypass.")
                   .cvssV31(new BigDecimal("8.2"))
+                  .vulnStatus(Vulnerability.VulnerabilityStatus.ANALYZED)
                   .cisaRequiredAction(
                       "Apply mitigations per vendor instructions or discontinue product if unavailable.")
                   .cisaVulnerabilityName(
@@ -86,6 +89,7 @@ public class PresetTenantData {
                   .description(
                       "Improper neutralization of SQL elements in Fortinet FortiClientEMS allows RCE/commands.")
                   .cvssV31(new BigDecimal("9.8"))
+                  .vulnStatus(Vulnerability.VulnerabilityStatus.ANALYZED)
                   .cisaRequiredAction(
                       "Apply mitigations per vendor instructions or discontinue product if unavailable.")
                   .cisaVulnerabilityName("Fortinet FortiClient EMS SQL Injection Vulnerability")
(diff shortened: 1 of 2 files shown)


openaev-api/src/test/java/io/openaev/service/tenants/TenantServiceTest.java
```diff
@@ -57,6 +69,17 @@ void should_create_and_find_tenant() throws Exception {
 
     // -- ACT --
     Tenant created = tenantService.create(tenant);
+    TenantContext.setCurrentTenant(tenant.getId());
```
RomuDeuxfois, line 70: question: Should already be handled by DefaultTenantExtension no ?

openaev-api/src/main/java/io/openaev/datapack/PresetTenantData.java
```diff
@@ -0,0 +1,155 @@
+package io.openaev.datapack;
+
+import io.openaev.database.model.Capability;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public class PresetTenantData {
+
+  public record DefaultVulnerability(
```
RomuDeuxfois: **TODO:** This should be linked to the existing `Vulnerabilities` class in some way. Otherwise, if that class changes, the insert may fail.

openaev-api/src/main/java/io/openaev/datapack/PresetTenantData.java
```diff
@@ -0,0 +1,155 @@
...
+      String published,
+      String description,
+      String cvss,
+      String requiredAction,
+      String vulnerabilityName,
+      String cweExternalId,
+      String cweSource,
+      List<String> referenceUrls) {}
+
+  public static final List<DefaultVulnerability> DEFAULT_VULNERABILITIES =
```
RomuDeuxfois: **TODO:** This data is the same as in the existing migration. We should have a single source of truth, and the migration should use that code directly. The same applies to roles.

openaev-api/src/main/java/io/openaev/datapack/packs/V20260330_Default_tenant_data.java
```diff
@@ -0,0 +1,97 @@
...
+        // tenant created
+        PresetTenantData.DEFAULT_ROLES.forEach(
+            (roleName, capabilities) -> {
+              Role role = roleService.createRole(roleName, roleName, capabilities);
+              Group group = new Group();
+              group.setName(roleName);
+              group.setDescription(roleName);
+              group.setDefaultUserAssignation(false);
+              group.setRoles(List.of(role));
+              if ("Admin".equals(roleName)) {
```
RomuDeuxfois: **todo**: you should use a constant for "Admin"

openaev-api/src/test/java/io/openaev/service/tenants/TenantServiceTest.java
```diff
@@ -57,6 +69,17 @@ void should_create_and_find_tenant() throws Exception {
...
+    // Simulate for tenant creation because Dataprocessor has @Profile("!test")
+    V20260330_Default_tenant_data datapack =
```
RomuDeuxfois: **nitpick**: Can you use @Autowired instead of constructor ?

openaev-api/src/test/java/io/openaev/service/tenants/TenantServiceTest.java
damgouj: No because we don't use the DatapackProcessor class in tests so the tenant context is not set

openaev-api/src/test/java/io/openaev/service/tenants/TenantServiceTest.java
damgouj: Good call, thanks !

openaev-api/src/main/java/io/openaev/datapack/PresetTenantData.java
damgouj: Builder or constructor Vulnerability/Cwe/...

corinnekrych: 👍 in the existing folder `openaev-api/resources/starterpack` we could add a json or yaml file and then as you suggested use a Jackson-friendly DTO Class to build a java object

openaev-api/src/main/java/io/openaev/datapack/PresetTenantData.java
```diff
@@ -0,0 +1,155 @@
...
+              "Cisco IOS XE Web UI insufficient input validation leading to root command injection.",
+              "7.2",
+              "Verify compliance with BOD 23-02 and apply mitigations.",
+              "Cisco IOS XE Web UI Command Injection Vulnerability",
+              "CWE-78",
+              "Nist",
+              List.of(
+                  "https://sec.cloudapps.cisco.com/security/center/content/CiscoSecurityAdvisory/cisco-sa-iosxe-webui-privesc-j22SaA4z")));
+
+  public static final Map<String, Set<Capability>> DEFAULT_ROLES =
```
corinnekrych, line 154: Same idea of builder/loader here, in the end this class could just be:
```
public class PresetTenantData {

  /** Default vulnerabilities loaded from {@code datapack/default-vulnerabilities.json}. */
  public static final List<DefaultVulnerability> DEFAULT_VULNERABILITIES =
      DefaultVulnerabilityLoader.load();

  /** Default roles loaded from {@code datapack/default-roles.json}. */
  public static final Map<String, Set<Capability>> DEFAULT_ROLES = DefaultRoleLoader.load();
}
```
with associated json/yamls files

damgouj: Oups thanks Corinne for the idea but I didn't see your comment and updated my Java code like PresetDomain if you want to check. If it's not OK for you, I can update my code to do your idea ?

PR #5308, opened 2026-03-31
[frontend] feat(multi-tenancy): embed tenantId in API - initial setup for multiple PRs (#4864)

corinnekrych:
### Proposed changes

This pull request introduces:

1/  'isFeatureFlagEnabled` It allows FE to get features flags injected by Spring configuration instead of accessing them thro /api/setting - as for mutitenancy we need to access the flag "MULTI_TENANT" to configure the API path before doing an API call (chicken-egg issue)

2/ add a white list TODO of API path that will need migration to support tenant scoping

  - The `buildUri` function in `Action.ts` now conditionally rewrites API paths using `buildTenantApiPath` if the `MULTI_TENANCY` feature flag is enabled and if the API is not in the TODO list
  - All tag-related API actions are updated to use the new `/api/tags` path, removing dependency on the old tenant URI utility.

We want to iteratively move each REST API (excepted some platform specific API s(see Note 1) to a tenant scope API URL. We an have a whitelist of TENANT_MIGRATION_TODO in the FE, As each BE controller is migrated to accept TENANT_PREFIX, we remove the matching entries from TENANT_MIGRATION_TODO — the FE automatically starts sending tenant-prefixed URLs.

| PR | Group | Prefixes to remove from TODO |
|---|---|---|
| ~~PR0~~ | ~~Tags~~ | ~~`/api/tags`~~ ✅ |
| PR1 | initial setup| Define TENANT_MIGRATION_TODO  👀 |
| PR2 | Scenarios & Exercises | `/api/scenarios`, `/api/exercises`, `/api/simulations` |
| PR3 | Injects & lifecycle | `/api/injects`, `/api/injector_contracts`, `/api/atomic-testings`, `/api/inject-expectations-traces` |
| PR4 | Teams & Players | `/api/teams`, `/api/players`, `/api/organizations` |
| PR5 | Assets | `/api/endpoints`, `/api/asset_groups`, `/api/security_platforms` |
| PR6 | Components | `/api/channels`, `/api/challenges`, `/api/payloads`, `/api/documents` |
| PR7 | Findings & Lessons | `/api/findings`, `/api/detection-remediations`, `/api/notification-rules`, `/api/vulnerabilities` |
| PR8 | Integrations | `/api/injectors`, `/api/collectors`, `/api/executors`, `/api/connector-instances`, `/api/catalog-connector` |
| PR9 | Reference data & Misc | `/api/attack_patterns`, `/api/kill_chain_phases`, `/api/domains`, `/api/mappers`, `/api/tag-rules`, `/api/dashboards`, `/api/fulltextsearch`, `/api/roles`, `/api/groups`, `/api/users`, `/api/capabilities`, `/api/xtmhub`, `/api/variables`, `/api/reports` |

### Testing Instructions

1. create one tenant
2. add one tag in taxomonie menu => check API call with tenant_id
3. add one payload (or any other entities) => check API call does not have tenant_id

### Checklist

### Further comments


openaev-front/vite.config.ts
```diff
@@ -52,7 +52,8 @@ export default ({ mode }: { mode: string }) => {
             .replace(/%APP_TITLE%/g, 'OpenAEV Dev')
             .replace(/%APP_DESCRIPTION%/g, 'OpenAEV Development platform')
             .replace(/%APP_FAVICON%/g, `${basePath}/static/favicon.png`)
-            .replace(/%APP_MANIFEST%/g, `${basePath}/static/manifest.json`);
+            .replace(/%APP_MANIFEST%/g, `${basePath}/static/manifest.json`)
```
RomuDeuxfois: **Question**: Would changing this mechanism prevent us from creating feature flags dynamically? Now, we are using environment variable instead of calling API. What do you think ?

openaev-front/src/utils/tenant-url-helper.ts
```diff
@@ -0,0 +1,78 @@
...
+  '/api/platform-roles',
+];
+
+/**
+ * Rewrites an API path to include the tenant prefix.
+ *
+ * This is the FE equivalent of the BE's TenantInterceptor:
+ * one place that applies the tenant prefix to all API calls.
+ */
+export const buildTenantApiPath = (uri: string): string => {
```
RomuDeuxfois, line 64: **Question**: I’m not sure I fully understand the goal here. Doing this seems less disruptive than changing all the URIs on the front end, but is the goal to update the URIs in the -action files, or to let this hook handle it automatically?

RomuDeuxfois: Because when I read tag_uri now, it feels a bit magical to me.

corinnekrych: It's similar to an AOP approach. All URL get rewritten from `/api/tags` to `/api/tenants/${tenantId}/tags`
We could have prefix all URLs occurences. There is pro and con on both approaches.

* This AOP style feels more magical but is more condense and less error prone (not forgetting tenant_id) as default is tenant_id based.
* The other static prefix approach is a massive file changes. It could be error prone if you forget to add tenant_id

openaev-front/vite.config.ts
corinnekrych: Yep this one was tricky.
I thought we need to be behind FF for this rewrite of the API URL but it's a chicken-egg issue: the useFeatureFlag need to access /api/settings 
Here it's a specific case where i need the flag early on in the initialisation of the app.

I dont see it as a replacement for accessing FF dynamically more a edge case for this flag. 

We could decide not to feature flag this change of API routing since our delivery of multi-tenancy is key to the milestone and remove this code on accessing Flag statically

RomuDeuxfois: Since we already have default routing on the back end, maybe the feature flag is not actually necessary for this.

openaev-front/src/utils/tenant-url-helper.ts
RomuDeuxfois: You convinced me that the tenant ID is too error-prone. It’s better to be strict about this and follow your approach.

openaev-front/.env
```diff
@@ -1 +1,2 @@
-HTTPS=false
\ No newline at end of file
+HTTPS=false
+VITE_ENABLED_DEV_FEATURES=MULTI_TENANCY
```
damgouj: question: is it normal you commit that ?

damgouj (commented):
I am a bit late but tests OK and just 1 comment more ;)

PR #5323, opened 2026-04-01
[backend] feat(chaining): add flag local/global in condition mappings (#4824)

savacano28:
### Proposed changes

* Add condition mapping type (default, global, local)
* Set default value when a condition type mapper is defined but not the mapping type

### Testing Instructions

1. Create a condition with this new attribute
2. Update
3. Fetch

### Related issues

* Related to #4824 

### Checklist

### Further comments


openaev-model/src/main/java/io/openaev/database/model/MappingType.java
```diff
@@ -0,0 +1,7 @@
+package io.openaev.database.model;
+
+public enum MappingType {
+  DEFAULT,
```
camrrx, line 4: I’m not sure the default value is useful. Is the default mapping local?

savacano28: The default value is used when we don't have an available mapping (events), then we get the default value from the payload and also allow the user to provide it (ui).

Issue #5303, opened 2026-03-31
Support UBI 9 as base image for platform


xfournet:
## Use case

OpenAEV platform Docker image should support UBI 9 (Universal Base Image 9) by Red Hat as an additional base image option, alongside the current base image.

PR #5324, opened 2026-04-01
[docker] chore: create UBI9 Docker image (#5303)

xfournet:
### Proposed changes

* Add a new Dockerfile to build UBI9 based docker image

### Related issues

* #5303 

### Checklist

### DIFF
--- a/Dockerfile_ubi9
+++ b/Dockerfile_ubi9
@@ -33,7 +33,8 @@ FROM registry.access.redhat.com/ubi9/openjdk-21-runtime AS app
 USER root
 
 RUN set -eux; \
-    curl -L -o /usr/bin/tini https://github.com/krallin/tini/releases/latest/download/tini; \
+    curl -L -o /usr/bin/tini https://github.com/krallin/tini/releases/download/v0.19.0/tini; \
+    echo "93dcc18adc78c65a028a84799ecf8ad40c936fdfc5f2a57b1acda5a8117fa82c /usr/bin/tini" | sha256sum -c -; \
     chmod +x /usr/bin/tini;
 
 WORKDIR /opt/openaev


Dockerfile_ubi9
```diff
@@ -0,0 +1,43 @@
...
+COPY --from=front-builder /opt/openaev-build/openaev-front/builder/prod/build ./openaev-front/builder/prod/build
+RUN mvn install -DskipTests -Pdev
+
+
+FROM registry.access.redhat.com/ubi9/openjdk-21-runtime AS app
+
+USER root
+
+RUN set -eux; \
+    curl -L -o /usr/bin/tini https://github.com/krallin/tini/releases/latest/download/tini; \
```
antoinemzs: I'm concerned with just downloading a random binary from some github repo. This theoretically does only improve signal handling for the java process (we don't handle OS signals anyway in openaev), thus mostly making the container restart and stop times faster by avoiding the container runtime reaping the process, but at the same time this is a non-trivial attack surface (compromised init process).

If we really need to retain an init process that isn't just `bash` (default for ubi9 images), please consider https://catalog.redhat.com/en/software/containers/ubi9/ubi-init/615bdc22075b022acc111bf6

xfournet:
> Is the intended scope just adding the dockerfile?
> 
> To be able to say that we support UBI9, we should at least:
> 
> * build all the images in UBI9 (not just the core server, but also injectors, collectors)

i'm waiting for the complete list of OAEV related images to create with UBI9

> * integrate with a CI pipeline (build)
> * deploy on a stack

yes it will be done in a 2nd step, first is just about producing a first batch of UBI9  images

 
> Importantly, this has an impact with the Catalog feature, the manifest suggests an image tag and version, but we need to also support the "base image" sort of switch, which isn't currently handled by the manifest format.

for the moment the composer is not in the scope. It will be upgraded later to be able to select the image flavor.

Dockerfile_ubi9
xfournet: ubi-init is much overkill for what we need.
the question is do we need tini ? i just keep it like in the other image.

regarding the download link (and if we still require tini) i will check how we can secure that

xfournet: i secured the download by pinning the version instead of using latest and enforcing the sha256sum (hardcoded in the Dockerfile)

PR #5326, opened 2026-04-01
[backend] feat(telemetry): add a new metric: number of active instance tenants (#4868)

RomuDeuxfois:
### Proposed changes

* add a new metric: number of active instance tenants


openaev-api/src/test/java/io/openaev/telemetry/metric_collectors/GlobalMetricCollectorTest.java
```diff
@@ -0,0 +1,85 @@
+package io.openaev.telemetry.metric_collectors;
```
MarineLeM, line 1: 👍

PR #5332, opened 2026-04-01
[backend] feat: process outputs as inputs (#4824)

savacano28:
### Proposed changes

* Add workflowState entity (to save global and local inputs)
* Save scope data in workflowState as global when workflow is started
* Trigger events or actions which could use scope data directly
* Save outputs from executions as global data
* Chaining of creation of local states from steps which their events can use the previous outputs
* Generate n executions for n inputs (input from combinations of global, local and default values)

image
image
image
image
image
image
image

### What is not covered in this pr

* Scope Execution : Setting targets, apply blacklist
* Validation of execution of workflow (step configuration is completed, EE for assets, Agroups..)
* Chaining for Correlated values; Evaluation of conditions type filter or even the tracking of these correlated values.
* Improving Condition Evaluation: IsConditionfilterValid

### Testing Instructions

1. Create Scenario/Simulation
2. Create workflow
3. Add scope
4. Create some steps with payload with arguments type : text, IPv4, ex, and mapping conditions for these steps and events related to these steps.
5. Launch Scenario/simulation
6. Check Creation Scope in workflowState
7. Check if any local data was saved in workflowState related to your steps
8. Check Steps (created)
9. Check Conditions (created)

### Related issues

* Related to #4824 

### Checklist

### Further comments


openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
```diff
@@ -426,7 +457,7 @@ public static StepInput toStepInput(InjectInput input) {
    *
    * @param step the {@link Step} containing the JSON data for the inject
    * @return the deserialized {@link Inject} object with its injector set if found; {@code null} if
-   *     the injector or contract is missing or if an exception occurs during deserialization
+   *     the injector or r or contract is missing or if an exception occurs during deserialization
```
camrrx, line 460: oups

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
```diff
@@ -442,40 +473,104 @@ private Inject getInjectFromDataStep(Step step) throws ChainingException {
       ObjectMapper mapper = new ObjectMapper();
       JsonNode root = mapper.readTree(step.getData());
 
-      if (inject.getInjectorContract().isEmpty())
```
EvaE-Filigran, line 445: Every change into getInjectFromDataStep() will make conflict with workflow crud branch

openaev-api/src/main/java/io/openaev/service/chaining/StepService.java
```diff
@@ -790,7 +805,7 @@ private static JsonPrimitive toJsonPrimitive(Object primitiveObject) {
    *
    * @param stepFromId the identifier of the step referenced by a condition; may be {@code null}
    * @return the persisted {@link Step} corresponding to the given identifier, or {@code null} if
-   *     the identifier is {@code null} {@link Step} is found
+   *     the identifier is r is {@code null} {@link Step} is found
```
EvaE-Filigran, line 808: oups 😅

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
```diff
@@ -451,7 +492,7 @@ public static StepsCreateInput.StepInput getInjectAsStepsCreateInput(InjectInput
           ConditionCreateInput.builder()
               .temporaryId("0")
               .type(ConditionType.AFTER)
-              .key(null)
+              .keyType(null)
```
camrrx: duplicate

openaev-api/src/main/java/io/openaev/service/chaining/WorkflowStateService.java
```diff
@@ -0,0 +1,487 @@
...
+    WorkflowState globalState = initializeGlobalState(workflowRun, scopeEntries);
+
+    save(globalState);
+    log.debug("Initialized Global WorkflowState for workflow execution {}", workflowRun.getId());
+
+    return globalState;
+  }
+
+  public void save(WorkflowState state) {
+    workflowStateRepository.save(state);
```
heditar: Return the worflowstate saved ?

openaev-api/src/main/java/io/openaev/service/chaining/ConditionService.java
```diff
@@ -406,6 +411,26 @@ public boolean isTimeCondition(Condition condition) {
...
+   *
+   * <p>TODO: this is for legacy behavior only (compare from start of workflow instead of previous
+   * step.
+   *
+   * @param conditionTemplate the condition to evaluate
+   * @param now current instant
+   * @param goal target instant
+   * @return {@code true} if the condition is valid
+   */
+  public Boolean isTimeConditionValid(Condition conditionTemplate, Instant now, Instant goal) {
```
heditar: Maybe "now" should be "executionTime" or something

openaev-api/src/main/java/io/openaev/service/chaining/StepService.java
```diff
@@ -142,6 +148,9 @@ private void startWorkflow(Workflow workflowRun, Workflow workflowTemplate)
       return;
     }
 
+    // Save data from Scope in WorkflowState
```
heditar: Should we start the workflow in Workflow service ?

openaev-api/src/main/java/io/openaev/service/chaining/WorkflowStateService.java
```diff
@@ -0,0 +1,487 @@
...
+
+  /**
+   * Syncs a completed inject's structured output traces into the workflow's global state entries.
+   * Exits early if the inject has no status, no traces, or no matching global state.
+   *
+   * @param inject the completed inject whose output traces are synchronized
+   * @param stepRun the RUN step used to locate the workflow's global state
+   */
+  @Transactional
+  public JsonObject syncInjectOutputToGlobalState(Inject inject, Step stepRun) {
```
heditar: Do we need both inject and steprun ?

openaev-api/src/main/java/io/openaev/service/chaining/WorkflowStateService.java
```diff
@@ -0,0 +1,487 @@
...
+    Optional<InjectStatus> injectStatus = inject.getStatus();
+    JsonObject allStepParsed = new JsonObject();
+
+    if (injectStatus.isPresent() && injectStatus.get().getTraces() != null) {
+      Map<String, ContractOutputType> fieldTypeMap = new HashMap<>();
+
+      inject
+          .getPayload()
+          .ifPresentOrElse(
```
heditar: Very hard to read, thhe code is too complex to be in lambdas

openaev-api/src/main/java/io/openaev/service/chaining/StepService.java
```diff
@@ -175,20 +184,17 @@ public Optional<Step> ready(Step nextStepTemplateToExecute, Workflow workflowRun
     Step nextStepTemplateToExecutePersisted =
         findByIdAndStatus(nextStepTemplateToExecute.getId(), StepStatus.TEMPLATE);
 
-    // CHECK CONDITIONS
-    List<Condition> conditionExecution =
-        conditionService.checkCondition(
```
heditar, line 180: That should be added back as it prevents all the steps to be executed at the beginning

openaev-model/src/main/java/io/openaev/database/model/WorkflowState.java
```diff
@@ -0,0 +1,55 @@
...
+  @UpdateTimestamp
+  private Instant updatedAt;
+
+  @ManyToOne(fetch = FetchType.LAZY)
+  @JoinColumn(name = "workflow_execution_id", nullable = false)
+  private Workflow workflowExecution;
+
+  @ManyToOne(fetch = FetchType.LAZY)
+  @JoinColumn(name = "workflow_step_template_id") // Nullable for Global
+  private Step stepTemplate;
```
heditar, line 54: Should we just have the step template ? Why do we also need workflow

openaev-api/src/main/java/io/openaev/service/chaining/WorkflowStateService.java
```diff
@@ -0,0 +1,487 @@
...
+                } catch (Exception e) {
+                  log.error("Failed to sync trace {} to global state", trace.getId(), e);
+                }
+              });
+
+      // Save JSON back to DB
+      globalState.setEntries(gson.toJson(entries));
+      save(globalState);
+    }
+    return allStepParsed;
```
heditar: if injectStatus is null we return allStepParsed?

openaev-api/src/main/java/io/openaev/service/chaining/WorkflowStateService.java
savacano28: Yes, because we save traces from this executed inject into stepRun

openaev-model/src/main/java/io/openaev/database/model/WorkflowState.java
savacano28: Because we need a workflow state for each workflow execution / RUN (for snapshots).

openaev-api/src/main/java/io/openaev/service/chaining/WorkflowStateService.java
savacano28: Done

openaev-api/src/main/java/io/openaev/service/chaining/StepService.java
savacano28: Thank you, done!

openaev-api/src/main/java/io/openaev/service/chaining/WorkflowStateService.java
savacano28: Done

openaev-model/src/main/java/io/openaev/database/model/WorkflowState.java
heditar: I mean a step is always link to 1 workflow

openaev-api/src/main/java/io/openaev/service/chaining/WorkflowOrquestor.java
heditar, line 1: Typo in filename -> not matching class name

openaev-api/src/main/java/io/openaev/service/chaining/WorkflowService.java
```diff
@@ -512,4 +519,79 @@ public void isPreviewFeatureChainingEnable() throws ChainingException {
...
+  @Transactional(rollbackFor = Exception.class)
+  public void startWorkflowBySimulationId(String simulationId) throws ChainingException {
+    Workflow workflowTemplate =
+        findWorkflowTemplateBySimulationId(simulationId)
+            .orElseThrow(
+                () ->
+                    new ElementNotFoundException(
+                        "Workflow (TEMPLATE) not found. Simulation ID: " + simulationId));
+    Workflow workflowRun = launchWorkflowSimulation(workflowTemplate);
+    initializeStateFromScopeDefinition(workflowRun);
```
heditar: Can't find this method

openaev-api/src/main/java/io/openaev/service/chaining/StepService.java
```diff
@@ -26,97 +27,59 @@
...
 @Service
-public class StepService implements StepEventHandler, ExternalUpdateEventHandler {
-  private final InjectExecutionStep injectExecutionStep;
+public class StepService {
 
-  private final WorkflowService workflowService;
-  private final StepRepository stepRepository;
+  private final InjectExecutionStep injectExecutionStep;
 
+  public final InjectService injectService;
```
impolitepanda: Should be set to private, not public. Same for line 35

openaev-api/src/main/java/io/openaev/service/chaining/StepEventService.java
```diff
@@ -0,0 +1,162 @@
...
+    return events;
+  }
+
+  /**
+   * Handle ready event and run the corresponding step.
+   *
+   * @param stepEvent event to handle
+   */
+  @Override
+  public void handleReadyStepEvent(StepEvent stepEvent) {
```
impolitepanda, line 49: This is becoming a top level method, called by the queue consumer. I believe it should be set as Transactional, then.

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
```diff
@@ -246,14 +262,67 @@ public Optional<Step> update(Step stepRun) throws ChainingException {
...
+              JsonArray valuesArray = children.getAsJsonObject(key).getAsJsonArray("_children");
+              for (JsonElement item : valuesArray) {
+                String val = item.getAsJsonObject().get("_value").getAsString();
+                // SyncState keys are usually Uppercase (e.g., "IP")
+                result.computeIfAbsent(key, k -> new ArrayList<>()).add(val);
+              }
+            }
+          }
+        } catch (Exception e) {
+          log.error("Failed to parse structured output for syncState", e);
```
impolitepanda: This makes the interation fail and return an empty result for this iteration, which mean the caller functions will never know if the input they received in response is complete or not. Is that the behavior we expect?

openaev-api/src/main/java/io/openaev/service/chaining/ConditionService.java
```diff
@@ -718,4 +661,196 @@ public void unlinkFromStep(Condition condition, String stepId) {
...
+   * partition (GLOBAL or LOCAL), computes the Cartesian product of all dynamic values, merges
+   * DEFAULT mapper values, and keeps only combinations that satisfy required execution keys. Unique
+   * combinations are tracked via hash to avoid duplicate executions, persisted into LOCAL state,
+   * and returned as ready-to-run input batches with resolved mapper conditions.
+   *
+   * @param stepTemplate step template for which input combinations are generated
+   * @param workflowRun active workflow run used to resolve global/local workflow states
+   * @return list of execution batches; empty when no mapper-driven execution is currently possible
+   */
+  public List<ConditionService.ExecutionBatch> extractInputsForStepExecution(
```
impolitepanda: This function does many things and doesn't match the extractXXX name. I believe it would be better to rename it to something like prepareInputsForStepExecution. Or, even better, we should probably split it into two methods : one to prepare/plan the execution batches and a second to actually save them.
Something like:
-  public BatchPlan planExecutionBatches(Step stepTemplate, Workflow workflowRun, List<Condition> mappers) {...}
- public void saveExecutedBatches(BatchPlan plan)

(name are only suggestion so change them if you xant, same for the new BatchPlan object, i just wrote that to save space)

openaev-api/src/main/java/io/openaev/service/chaining/WorkflowStateService.java
```diff
@@ -0,0 +1,280 @@
...
+    return workflowStateRepository.save(state);
+  }
+
+  /**
+   * Returns or creates the global state for a workflow execution.
+   *
+   * @param workflow the running workflow
+   * @return existing global state, or a new (unsaved) one with empty entries
+   */
+  private WorkflowState getOrCreateGlobalState(Workflow workflow) {
```
impolitepanda: I have an issue with the getOrCreate name. Usually, it's a database operation. Here, it's just an object instantiation. Something like loadOrBuildGlobalState would probably be better ?

openaev-api/src/main/java/io/openaev/service/chaining/WorkflowStateService.java
```diff
@@ -0,0 +1,280 @@
...
+   * Syncs structured output data into the global workflow state entries.
+   *
+   * @param dataToSync JSON element containing output data to merge
+   * @param fieldTypeMap mapping from field name to its contract output type
+   * @param workflowRun the running workflow whose global state is updated
+   */
+  public void syncState(
+      JsonElement dataToSync, Map<String, ContractOutputType> fieldTypeMap, Workflow workflowRun) {
+    WorkflowState globalState = getOrCreateGlobalState(workflowRun);
+    if (globalState == null) {
```
impolitepanda: I don't think this check is necessary. It doesn't seem like getOrCreateGlobalState can retun null

openaev-api/src/main/java/io/openaev/service/chaining/WorkflowStateService.java
impolitepanda: Or maybe we should actually persist the scope here? I don't know which is better, to be honest

openaev-api/src/main/java/io/openaev/service/chaining/WorkflowStateService.java
```diff
@@ -0,0 +1,280 @@
...
+  }
+
+  /**
+   * Returns the local {@link WorkflowState} for a step template within a workflow execution.
+   *
+   * @param targetTemplate the step template ID
+   * @param workflowExecution the workflow execution ID
+   * @return the local state, or {@code null} if not yet initialized
+   */
+  public WorkflowState getLocalStateByWorkflowAndStep(
```
impolitepanda: I feel like this function behaves in a similar way to getOrCreateGlobalState. They should probably be named similarly as well and behave the same. So if my renaming suggestion is taken into account, this should probably be renamed loadOrBuildLocalState, for instance. I dropped the ByWorkflowAndStep from the name as I don't think a local state can belong to anything else, but maybe I'm wrong.

Also, if getOrCreateGlobalState is modified to persist the store, this one should do the same as well.

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
```diff
@@ -59,7 +67,7 @@
 @Component
 @Slf4j
 public class InjectExecutionStep implements ActionStep {
```
impolitepanda, line 69: 17 deps. This class definitely need splitting ^^

openaev-api/src/main/java/io/openaev/service/chaining/WorkflowStateService.java
savacano28: I save the state in the end of this method.

openaev-api/src/main/java/io/openaev/service/chaining/ConditionService.java
savacano28: Next Pr

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
savacano28: Next Pr

impolitepanda (approved):

Issue #45, opened 2019-07-01
Login Credentials for Demo not working


sikula:
## Description
When on the demo site, the login credentials that appear at the bottom of the login form do not work.

## Environment

1. OS (where OpenEx server runs): Ubuntu 16.4
2. OpenEx version: { e.g. OpenEx 1.0.2 }
3. OpenEx component: frontend
4. Other environment details:

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Go to the demo app
2. Use the credentials
3. "Invalid Credentials" message

## Expected Output

Able to signin.

## Actual Output

Invalid Credentials

PR #5351, opened 2026-04-02
[backend] feat(chaining): adding RBAC resource types for chaining (#4824)

heditar:
### Proposed changes

* Adding the new resource types for chaining and linked them to the capa, as it is this change as no impact but it will allow the resource type ti be used in the chaining apis coming in the next PRs
* This PR doesn't include Grant or Filtering which are also necessary to have the whole RBAC functionalities to be applied to chaining, these will be done in separate PRs

### Testing Instructions

1. Can't be tested as it is, it will require to change an existing api with the new resource type and test

### Related issues

* #4824
*  [#45](https://github.com/OpenAEV-Platform/filigran-private/issues/45)

### Checklist


openaev-model/src/main/java/io/openaev/database/model/Capability.java
```diff
@@ -41,12 +45,20 @@ public enum Capability {
...
+      pair(ResourceType.WORKFLOW, Action.DUPLICATE),
+      pair(ResourceType.WORKFLOW, Action.CREATE)),
   DELETE_ASSESSMENT(
       MANAGE_ASSESSMENT,
       pair(ResourceType.SCENARIO, Action.DELETE),
       pair(ResourceType.SIMULATION, Action.DELETE),
-      pair(ResourceType.ATOMIC_TESTING, Action.DELETE)),
+      pair(ResourceType.ATOMIC_TESTING, Action.DELETE),
+      pair(ResourceType.STEP, Action.DELETE),
+      pair(ResourceType.WORKFLOW, Action.DELETE)),
```
savacano28: We have also condition

heditar: added

PR #5361, opened 2026-04-03
[backend/frontend] feat(settings): introduce tenant-scoped settings (#4864)

RomuDeuxfois:
### Proposed changes

* Move global settings under platform settings
* Let dashboard settings onto dedicated tenant settings

### DIFF
--- a/openaev-api/src/main/java/io/openaev/rest/attack_pattern/AttackPatternApi.java
+++ b/openaev-api/src/main/java/io/openaev/rest/attack_pattern/AttackPatternApi.java
@@ -1,5 +1,6 @@
 package io.openaev.rest.attack_pattern;
 
+import static io.openaev.config.TenantUriUtils.TENANT_PREFIX;
 import static io.openaev.database.specification.AttackPatternSpecification.byName;
 import static io.openaev.helper.DatabaseHelper.updateRelation;
 import static io.openaev.helper.StreamHelper.fromIterable;
@@ -39,6 +40,7 @@ import org.springframework.web.multipart.MultipartFile;
 
 @RestController
 @RequiredArgsConstructor
+@RequestMapping({AttackPatternApi.ATTACK_PATTERN_URI, TENANT_PREFIX + "/attack_patterns"})
 public class AttackPatternApi extends RestBehavior {
 
   public static final String ATTACK_PATTERN_URI = "/api/attack_patterns";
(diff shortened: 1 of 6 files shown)


corinnekrych:
@RomuDeuxfois I'm trying to test your PR but i might be missing some context/understanding.
1/ I log in as platofm admin
2/ create a tenant A
3/ go to Platform | Parameters see:
Screenshot 2026-04-09 at 14 25 25
4/ go to Settings | Parameters
Screenshot 2026-04-09 at 14 26 19

Small bug: <Typography variant="h4">{t('OpenAEV platform')}</Typography> is missing in 3/

Question:
1/ how do i get dashboard setting displayed?
2/ is the left menu
* Settings menu => for tenant user: we could have a read only view of platform settings
* Platform menu => platform user

RomuDeuxfois:
@corinnekrych PR ready for a secound round

openaev-front/src/actions/settings/tenant-settings-action.ts
```diff
@@ -0,0 +1,18 @@
+import { type Dispatch } from 'redux';
+
+import { getReferential, simplePutCall } from '../../utils/Action';
+import { getCurrentTenantId } from '../../utils/tenant-url-helper';
+import * as schema from '../Schema';
+import {TenantSettingsUpdateInput} from "../../utils/api-types";
+
+const tenantSettingsUri = () => `/api/tenants/${getCurrentTenantId()}/settings`;
```
corinnekrych: So here we have a special case where /api/settings could be used for either tenant-scope or platform-scope therefore https://github.com/OpenAEV-Platform/openaev/blob/418fa1eb393c13e709ced6010917e8cd4f206e8d/openaev-front/src/utils/tenant-url-helper.ts#L124-L134 won't work...
Suggestion: I wonder if a separate name `api/tenant-settings` would be cleaner, since we're also adding anew spring REST endpoint

corinnekrych:
@RomuDeuxfois here is the test i did:
1/ create tenant A
2/ create a new dashboard "TOTO"
3/ go to Settings | Parameter, set home dashboard to TOTO
4/ go to Home => blank screen with "select dashboard"
Looking at network tab api/settings/home-dashboard does not return any response

is it expected?

RomuDeuxfois:
@corinnekrych ready for the second round

PR #5396, opened 2026-04-08
feat(tanium): add TTR detach script for Linux and macOS (#4710)

### LINKED ISSUE #4710: Add Tanium scripts for Linux & MacOS
## Context

As we did for Windows, today CrowdStrike executor don't generate logs inside Tanium as it' excluded itself, we need to add a new scripts as we did on Windows to create a cron tab & execute it to spawn a new process which will not be tanium and then be recorded.

Seb-MIGUEL:
Closes #4710

### DIFF
--- a/.github/instructions/backend.instructions.md
+++ b/.github/instructions/backend.instructions.md
@@ -116,24 +116,3 @@ public Page<{Entity}Output> search(...) { return service.search(input).map({Enti
   created with `new Connector()`). Comments should explain *why*, not *what*.
 - **Don't duplicate computed values**: compute once (e.g. `String logoFilename = getLogoFilename()`)
   and pass as an argument — never repeat the same `.formatted(...)` expression in multiple places.
-## Tanium Executor Scripts
-
-When adding or modifying Tanium package scripts (`.sh`, `.ps1`) and their JSON package definition:
-
-### Fail-fast on missing argument
-Both `.sh` and `.ps1` scripts must **fail fast** with `exit 1` + explicit error message when the payload argument is missing.
-- **✅ Correct (`.sh`)**: `if [ -z "${1:-}" ]; then echo "[ERROR] no payload argument" >&2; exit 1; fi`
-- **✅ Correct (`.ps1`)**: `if(-not $arg) { Write-Host "[ERROR] No payload argument supplied"; exit 1 }`
-- **❌ Wrong**: silent NOOP substitution (`$arg = "NOOP"`) that masks the error and reports success
-
-### Post-launch verification
-After submitting the payload, confirm it actually started (not just "accepted"):
-- **nohup/setsid**: capture launcher PID, `sleep 1`, then verify with `kill -0 $PID` or `wait`
-- **systemd-run**: use `--wait` flag so the call blocks until the transient unit exits (real exit code)
-- **`at` / fire-and-forget**: acceptable when the scheduler provides no status API
-
-### Package JSON integrity
-Every file entry in the Tanium package JSON must include `hash` (SHA-256 hex) and `size` (bytes) fields, matching the actual script content. **Recompute hash and size after any script change.**
-```json
-{ "hash": "<sha256>", "name": "openaev-ttr.sh", "source": "", "size": 4648, ... }
-```
--- a/.github/instructions/backend.instructions.md
+++ b/.github/instructions/backend.instructions.md
@@ -116,3 +116,24 @@ public Page<{Entity}Output> search(...) { return service.search(input).map({Enti
   created with `new Connector()`). Comments should explain *why*, not *what*.
 - **Don't duplicate computed values**: compute once (e.g. `String logoFilename = getLogoFilename()`)
   and pass as an argument — never repeat the same `.formatted(...)` expression in multiple places.
+## Tanium Executor Scripts
+
+When adding or modifying Tanium package scripts (`.sh`, `.ps1`) and their JSON package definition:
+
+### Fail-fast on missing argument
+Both `.sh` and `.ps1` scripts must **fail fast** with `exit 1` + explicit error message when the payload argument is missing.
+- **✅ Correct (`.sh`)**: `if [ -z "${1:-}" ]; then echo "[ERROR] no payload argument" >&2; exit 1; fi`
+- **✅ Correct (`.ps1`)**: `if(-not $arg) { Write-Host "[ERROR] No payload argument supplied"; exit 1 }`
+- **❌ Wrong**: silent NOOP substitution (`$arg = "NOOP"`) that masks the error and reports success
+
+### Post-launch verification
+After submitting the payload, confirm it actually started (not just "accepted"):
+- **nohup/setsid**: capture launcher PID, `sleep 1`, then verify with `kill -0 $PID` or `wait`
+- **systemd-run**: use `--wait` flag so the call blocks until the transient unit exits (real exit code)
+- **`at` / fire-and-forget**: acceptable when the scheduler provides no status API
+
+### Package JSON integrity
+Every file entry in the Tanium package JSON must include `hash` (SHA-256 hex) and `size` (bytes) fields, matching the actual script content. **Recompute hash and size after any script change.**
+```json
+{ "hash": "<sha256>", "name": "openaev-ttr.sh", "source": "", "size": 4648, ... }
+```
--- a/openaev-api/src/main/java/io/openaev/executors/tanium/openaev-tanium-packages-TTR.json
+++ b/openaev-api/src/main/java/io/openaev/executors/tanium/openaev-tanium-packages-TTR.json
@@ -1,5 +1,5 @@
 {
-  "comment": "Exported from Tanium Server 7.8.1.3109 at 2025-08-26T11:42:48",
+  "comment": "OpenAEV TTR package definition — manually maintained, not a Tanium Server export",
   "version": 2,
   "object_list": {
     "package_specs": [
@@ -11,10 +11,10 @@
         "display_name": "OpenAEV Subprocessor (Unix) TTR",
         "files": [
           {
-            "hash": "41e2e27dfb399a4431caa5efe12d6b5749b9f23800185fdf7a5c39786026b033",
+            "hash": "da49982e61cb09b562d95d9b53d65e3caf2568ac4a53108499ff3fa9d3cd6617",
             "name": "openaev-ttr.sh",
             "source": "",
-            "size": 4648,
+            "size": 4667,
             "download_seconds": 0,
             "direct_download_flag": false
           }
--- a/openaev-api/src/main/java/io/openaev/executors/tanium/openaev-tanium-packages-TTR.json
+++ b/openaev-api/src/main/java/io/openaev/executors/tanium/openaev-tanium-packages-TTR.json
@@ -9,10 +9,17 @@
         },
         "name": "OpenAEV Subprocessor (Unix) TTR",
         "display_name": "OpenAEV Subprocessor (Unix) TTR",
-        "files": [],
+        "files": [
+          {
+            "name": "openaev-ttr.sh",
+            "source": "",
+            "download_seconds": 0,
+            "direct_download_flag": false
+          }
+        ],
         "file_templates": [],
-        "command": "/bin/sh -c \"/bin/echo $1 | sed 's#%2b#+#gi' | sed 's#%3d#=#gi' | sed 's#%2f#/#gi' | base64 -d | sh & exit 0\"",
-        "command_timeout": 60,
+        "command": "/bin/sh openaev-ttr.sh $1",
+        "command_timeout": 120,
         "expire_seconds": 660,
         "hidden_flag": 0,
         "process_group_flag": 1,
--- a/openaev-api/src/main/java/io/openaev/executors/tanium/openaev-tanium-packages-TTR.json
+++ b/openaev-api/src/main/java/io/openaev/executors/tanium/openaev-tanium-packages-TTR.json
@@ -11,8 +11,10 @@
         "display_name": "OpenAEV Subprocessor (Unix) TTR",
         "files": [
           {
+            "hash": "41e2e27dfb399a4431caa5efe12d6b5749b9f23800185fdf7a5c39786026b033",
             "name": "openaev-ttr.sh",
             "source": "",
+            "size": 4648,
             "download_seconds": 0,
             "direct_download_flag": false
           }
--- a/openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
+++ b/openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
@@ -26,10 +26,12 @@ if [ -z "${1:-}" ]; then
 fi
 
 # ── 1. URL-decode ─────────────────────────────────────────────────────────────
-# %20 is stripped (not replaced): space is not valid base64.
+# Tanium URL-encodes the base64 payload; restore the 3 non-alphanumeric
+# characters of standard base64: + = /
+# Note: [xX] bracket notation is used instead of sed's 'i' flag because
+# the 'i' flag is a GNU extension not available on macOS/BSD sed.
 DECODED=$(echo -n "$1" \
-  | sed -e 's#%2[bB]#+#g' -e 's#%3[dD]#=#g' -e 's#%2[fF]#/#g' \
-        -e 's#%0[aA]##g' -e 's#%0[dD]##g' -e 's#%20##g')
+  | sed -e 's#%2[bB]#+#g' -e 's#%3[dD]#=#g' -e 's#%2[fF]#/#g')
 
 # ── 2. Base64-decode ──────────────────────────────────────────────────────────
 PAYLOAD=$(echo -n "$DECODED" | base64 -d 2>/dev/null) \
@@ -46,7 +48,7 @@ trap 'rm -f "$PAYLOAD_SCRIPT"' EXIT INT TERM
 
 {
   echo '#!/bin/sh'
-  printf '%s\n' "$PAYLOAD"
+  echo "$PAYLOAD"
 } > "$PAYLOAD_SCRIPT"
 
 chmod 700 "$PAYLOAD_SCRIPT"
@@ -66,6 +68,7 @@ detach_with_nohup() {
 }
 
 detach_with_systemd() {
+  # Requires elevated privileges (Tanium agent runs as root)
   command -v systemd-run >/dev/null 2>&1 || return 1
   systemd-run --quiet /bin/sh -c "$WRAPPED_CMD" 2>/dev/null
 }
--- a/openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
+++ b/openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
@@ -28,8 +28,8 @@ fi
 # ── 1. URL-decode ─────────────────────────────────────────────────────────────
 # Tanium URL-encodes the base64 payload; restore the 3 non-alphanumeric
 # characters of standard base64: + = /
-# Note: [xX] bracket notation is used instead of sed's 'i' flag because
-# the 'i' flag is a GNU extension not available on macOS/BSD sed.
+# Note: [bB]/[dD]/[fF] bracket notation is used instead of sed's 'i' flag
+# because the 'i' flag is a GNU extension not available on macOS/BSD sed.
 DECODED=$(echo -n "$1" \
   | sed -e 's#%2[bB]#+#g' -e 's#%3[dD]#=#g' -e 's#%2[fF]#/#g')
 
@@ -48,7 +48,7 @@ trap 'rm -f "$PAYLOAD_SCRIPT"' EXIT INT TERM
 
 {
   echo '#!/bin/sh'
-  echo "$PAYLOAD"
+  printf '%s\n' "$PAYLOAD"
 } > "$PAYLOAD_SCRIPT"
 
 chmod 700 "$PAYLOAD_SCRIPT"
--- a/openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
+++ b/openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
@@ -1,0 +1,94 @@
+#!/bin/sh
+# =============================================================================
+# openaev-ttr.sh — OpenAEV Tanium TTR Payload Launcher (Linux / macOS)
+# =============================================================================
+# Purpose:
+#   Execute an OpenAEV inject payload OUTSIDE the Tanium process tree so that
+#   Tanium Threat Response (TTR) can detect it.  Mirrors the logic of the
+#   Windows counterpart openaev-ttr.ps1 which uses a Scheduled Task.
+#
+# Usage (called by the Tanium package):
+#   /bin/sh openaev-ttr.sh <url-encoded-base64-payload>
+#
+# Detach strategy (first available wins):
+#   1. nohup/setsid  – double-fork, reparented to PID 1 (most portable)
+#   2. systemd-run   – transient service under PID 1 (Linux with systemd)
+#   3. at now        – spawned by atd (if installed)
+#   4. (fallback)    – direct execution, not detached
+# =============================================================================
+
+set -eu
+
+# ── 0. Validate input ────────────────────────────────────────────────────────
+if [ -z "${1:-}" ]; then
+  echo "[openaev-ttr] ERROR: no payload argument supplied" >&2
+  exit 1
+fi
+
+# ── 1. URL-decode ─────────────────────────────────────────────────────────────
+# %20 is stripped (not replaced): space is not valid base64.
+DECODED=$(echo -n "$1" \
+  | sed -e 's#%2[bB]#+#g' -e 's#%3[dD]#=#g' -e 's#%2[fF]#/#g' \
+        -e 's#%0[aA]##g' -e 's#%0[dD]##g' -e 's#%20##g')
+
+# ── 2. Base64-decode ──────────────────────────────────────────────────────────
+PAYLOAD=$(echo -n "$DECODED" | base64 -d 2>/dev/null) \
+  || { echo "[openaev-ttr] ERROR: base64 decode failed" >&2; exit 1; }
+
+# ── 3. Write payload to a temp script ─────────────────────────────────────────
+umask 077
+PAYLOAD_SCRIPT=$(mktemp /tmp/openaev-payload-XXXXXXXXXX.sh) \
+  || { echo "[openaev-ttr] ERROR: mktemp failed" >&2; exit 1; }
+TASK_ID=$(basename "$PAYLOAD_SCRIPT" .sh)
+
+# Ensure cleanup if the script is interrupted before detach
+trap 'rm -f "$PAYLOAD_SCRIPT"' EXIT INT TERM
+
+{
+  echo '#!/bin/sh'
+  printf '%s\n' "$PAYLOAD"
+} > "$PAYLOAD_SCRIPT"
+
+chmod 700 "$PAYLOAD_SCRIPT"
+
+# ── 4. Detach from Tanium process tree ────────────────────────────────────────
+# Cleanup via trap so it runs even if the payload calls exit or exec.
+WRAPPED_CMD="trap 'rm -f \"${PAYLOAD_SCRIPT}\"' EXIT INT TERM; /bin/sh '${PAYLOAD_SCRIPT}'"
+
+detach_with_nohup() {
+  command -v nohup >/dev/null 2>&1 || return 1
+  if command -v setsid >/dev/null 2>&1; then
+    nohup setsid /bin/sh -c "$WRAPPED_CMD" >/dev/null 2>&1 &
+  else
+    # macOS: no setsid, double-fork via subshell
+    ( nohup /bin/sh -c "$WRAPPED_CMD" >/dev/null 2>&1 & ) &
+  fi
+}
+
+detach_with_systemd() {
+  command -v systemd-run >/dev/null 2>&1 || return 1
+  systemd-run --quiet /bin/sh -c "$WRAPPED_CMD" 2>/dev/null
+}
+
+detach_with_at() {
+  command -v at >/dev/null 2>&1 || return 1
+  echo "$WRAPPED_CMD" | at now 2>/dev/null
+}
+
+# Try methods from most common to least common.
+# nohup is POSIX-mandated and expected to always be available.
+if detach_with_nohup; then
+  echo "[openaev-ttr] Payload launched via nohup (task: $TASK_ID)"
+elif detach_with_systemd; then
+  echo "[openaev-ttr] Payload launched via systemd-run (task: $TASK_ID)"
+elif detach_with_at; then
+  echo "[openaev-ttr] Payload scheduled via at (task: $TASK_ID)"
+else
+  # Fallback: direct execution (not detached from Tanium process tree)
+  echo "[openaev-ttr] WARN: all detach methods failed, executing directly (task: $TASK_ID)" >&2
+  /bin/sh -c "$WRAPPED_CMD"
+fi
+
+# Disarm the main-script trap — the detached process handles its own cleanup
+trap - EXIT INT TERM
+exit 0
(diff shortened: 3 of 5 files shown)


openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
```diff
@@ -0,0 +1,107 @@
...
+# ── 3. Write payload to a temp script ─────────────────────────────────────────
+umask 077
+
+if command -v uuidgen >/dev/null 2>&1; then
+  TASK_ID=$(uuidgen)
+elif [ -f /proc/sys/kernel/random/uuid ]; then
+  TASK_ID=$(cat /proc/sys/kernel/random/uuid)
+else
+  TASK_ID="openaev-$(date +%s)-$$"
+fi
```
antoinemzs: looks unreasonably fiddly. How about we assume none of the above dependencies exist and just fall back to the date based stamp?

Alternatively a coreutils-based option with` $(dd if=/dev/urandom bs=1 count=128 2>/dev/null | sha256sum | head -c 40)`, with /dev/urandom a CSPRNG so virtually no collision.
(coreutils available on most linux, macos, some BSD)

Choose one single method for all

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
```diff
@@ -0,0 +1,107 @@
...
+  exit 1
+fi
+
+# ── 1. URL-decode ─────────────────────────────────────────────────────────────
+ENCODED="$1"
+DECODED=$(printf '%s' "$ENCODED" \
+  | sed \
+    -e 's/%2[bB]/+/g'  \
+    -e 's/%3[dD]/=/g'   \
+    -e 's/%2[fF]/\//g'  \
```
antoinemzs: Note you can use arbitrary characters as separators in sed commands to avoid mangling or escaping the contents;
from the original code: `sed 's#%2f#/#gi'`

also the `i` (case-insensitive) switch avoids having to do regex

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
```diff
@@ -0,0 +1,107 @@
...
+
+# ── 0. Validate input ────────────────────────────────────────────────────────
+if [ -z "$1" ]; then
+  echo "[openaev-ttr] ERROR: no payload argument supplied" >&2
+  exit 1
+fi
+
+# ── 1. URL-decode ─────────────────────────────────────────────────────────────
+ENCODED="$1"
+DECODED=$(printf '%s' "$ENCODED" \
```
antoinemzs: looks like a verbose way to avoid doing `echo -n "$ENCODED"`

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
```diff
@@ -0,0 +1,107 @@
...
+  else
+    # macOS: no setsid, double-fork via subshell
+    ( nohup /bin/sh "$PAYLOAD_SCRIPT" >/dev/null 2>&1 & ) &
+  fi
+}
+
+detach_with_cron() {
+  CRON_TAG="# openaev-${TASK_ID}"
+  CRON_LINE="* * * * * /bin/sh $PAYLOAD_SCRIPT ; (crontab -l 2>/dev/null | grep -v '$TASK_ID' | crontab -) ${CRON_TAG}"
+  ( crontab -l 2>/dev/null; printf '%s\n' "$CRON_LINE" ) | crontab - 2>/dev/null
```
antoinemzs: extremely fragile and prone to breaking in situations of race conditions, could reschedule past tasks

I doubt there is a safe way to do this. I wouldn't recommend doing this.

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
```diff
@@ -0,0 +1,107 @@
...
+rm -f "\$0"
+PAYLOAD_EOF
+
+chmod 700 "$PAYLOAD_SCRIPT"
+
+# ── 4. Detach from Tanium process tree ────────────────────────────────────────
+
+detach_with_at() {
+  echo "/bin/sh $PAYLOAD_SCRIPT" | at now 2>/dev/null
+}
```
antoinemzs: `at` is not a default utility on a lot of platforms

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
```diff
@@ -0,0 +1,107 @@
...
+  ( crontab -l 2>/dev/null; printf '%s\n' "$CRON_LINE" ) | crontab - 2>/dev/null
+}
+
+# Try methods in priority order
+if detach_with_at; then
+  echo "[openaev-ttr] Payload scheduled via 'at now' (task: $TASK_ID)"
+elif detach_with_systemd; then
+  echo "[openaev-ttr] Payload launched via 'systemd-run' (task: $TASK_ID)"
+elif detach_with_nohup; then
+  echo "[openaev-ttr] Payload launched via nohup/double-fork (task: $TASK_ID)"
```
antoinemzs: questioning the order:

* `nohup` is prevalent on most *nix systems including macos;
* systemd is mostly just linux, and not all of them at that;
* `at` is not distributed on all systems; installed on macos but optional on most linux distros

is there a reason to start with the less common to the more common commands?

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
```diff
@@ -0,0 +1,107 @@
...
+elif detach_with_cron; then
+  echo "[openaev-ttr] Payload scheduled via cron one-shot (task: $TASK_ID) – runs within 1 min"
+else
+  echo "[openaev-ttr] ERROR: all detach methods failed" >&2
+  rm -f "$PAYLOAD_SCRIPT"
+  exit 1
+fi
```
antoinemzs, line 100: Wouldn't we attempt an undetached execution if all fails?

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
```diff
@@ -0,0 +1,107 @@
...
+  TASK_ID="openaev-$(date +%s)-$$"
+fi
+
+PAYLOAD_SCRIPT="/tmp/openaev-payload-${TASK_ID}.sh"
+
+cat > "$PAYLOAD_SCRIPT" <<PAYLOAD_EOF
+#!/bin/sh
+# --- OpenAEV detached payload – auto-cleanup ---
+$PAYLOAD
+rm -f "\$0"
```
antoinemzs: if you escape `\$0` you will try to remove the file named `$0`, not the current script

antoinemzs: if the payload contents contain an explicit exit you will fail to cleanup the script. This would better be included in the scehduled command, not the script itself.

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
```diff
@@ -0,0 +1,107 @@
...
+    -e 's/%3[dD]/=/g'   \
+    -e 's/%2[fF]/\//g'  \
+    -e 's/%0[aA]//g'    \
+    -e 's/%0[dD]//g'    \
+    -e 's/%20/ /g')
+
+# ── 2. Base64-decode ──────────────────────────────────────────────────────────
+# macOS base64 uses -D, GNU uses -d.
+PAYLOAD=$(printf '%s' "$DECODED" | base64 -d 2>/dev/null) \
+  || PAYLOAD=$(printf '%s' "$DECODED" | base64 -D 2>/dev/null) \
```
antoinemzs: in which case do you need the `-D` (capital D) switch that the `-d` cannot do?
macOS `base64` impelmentation takes both capital and minuscule form example https://ss64.com/mac/base64.html

also recommending using `echo -n` instead of `printf '%s'`

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
```diff
@@ -0,0 +1,107 @@
...
+# ── 1. URL-decode ─────────────────────────────────────────────────────────────
+ENCODED="$1"
+DECODED=$(printf '%s' "$ENCODED" \
+  | sed \
+    -e 's/%2[bB]/+/g'  \
+    -e 's/%3[dD]/=/g'   \
+    -e 's/%2[fF]/\//g'  \
+    -e 's/%0[aA]//g'    \
+    -e 's/%0[dD]//g'    \
+    -e 's/%20/ /g')
```
antoinemzs: Are we certain we will be POSTing base64 content with these encoded entities? This would mean the base64-encoded scripts string (valid base64) would have been broken up in lines before being escaped via the API POST

antoinemzs: If we do find a `%20` (space char) in the escaped string it should be suppressed, not added as a space character which is not valid base64.

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
```diff
@@ -0,0 +1,107 @@
...
+PAYLOAD_SCRIPT="/tmp/openaev-payload-${TASK_ID}.sh"
+
+cat > "$PAYLOAD_SCRIPT" <<PAYLOAD_EOF
+#!/bin/sh
+# --- OpenAEV detached payload – auto-cleanup ---
+$PAYLOAD
+rm -f "\$0"
+PAYLOAD_EOF
+
+chmod 700 "$PAYLOAD_SCRIPT"
```
antoinemzs, line 54: we already have `umask 077` but why not

antoinemzs (dismissed):

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
Seb-MIGUEL: Hello based on Windows script yes (.Replace('%3d','='))

antoinemzs: I was referring specifically to the CR, LF and space characters :)
It's a weird behaviour of the Tanium black box that the base64 string would be URL-encoded somehow in the process, so we had provisioned for converting back the three non alphanumeric characters in the base64 encoding standard. I wasn't expecting that we'd find CR, LF or space in the URL-encoded base64 string though.

antoinemzs:
How would you test this? I suspect this has to be manually installed on the Tanium console?

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
```diff
@@ -0,0 +1,94 @@
...
+if [ -z "${1:-}" ]; then
+  echo "[openaev-ttr] ERROR: no payload argument supplied" >&2
+  exit 1
+fi
+
+# ── 1. URL-decode ─────────────────────────────────────────────────────────────
+# %20 is stripped (not replaced): space is not valid base64.
+DECODED=$(echo -n "$1" \
+  | sed -e 's#%2[bB]#+#g' -e 's#%3[dD]#=#g' -e 's#%2[fF]#/#g' \
+        -e 's#%0[aA]##g' -e 's#%0[dD]##g' -e 's#%20##g')
```
antoinemzs: Still questioning whether the last three substitutions are necessary. Why are they included?

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
```diff
@@ -0,0 +1,94 @@
...
+PAYLOAD_SCRIPT=$(mktemp /tmp/openaev-payload-XXXXXXXXXX.sh) \
+  || { echo "[openaev-ttr] ERROR: mktemp failed" >&2; exit 1; }
+TASK_ID=$(basename "$PAYLOAD_SCRIPT" .sh)
+
+# Ensure cleanup if the script is interrupted before detach
+trap 'rm -f "$PAYLOAD_SCRIPT"' EXIT INT TERM
+
+{
+  echo '#!/bin/sh'
+  printf '%s\n' "$PAYLOAD"
```
antoinemzs, line 51: can still use `echo -n`. Avoid mixing many ways to do the same thing

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
antoinemzs: Also why use regex rather than case-insensitive matching?

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
```diff
@@ -0,0 +1,94 @@
...
+  else
+    # macOS: no setsid, double-fork via subshell
+    ( nohup /bin/sh -c "$WRAPPED_CMD" >/dev/null 2>&1 & ) &
+  fi
+}
+
+detach_with_systemd() {
+  command -v systemd-run >/dev/null 2>&1 || return 1
+  systemd-run --quiet /bin/sh -c "$WRAPPED_CMD" 2>/dev/null
+}
```
antoinemzs, line 81: Note this only works with elevated privileges, so I assume Tanium always runs under a privileged user.

antoinemzs (commented):
Tested on linux via nohup, systemd, atd and fallback ok; not tested on macos (no access to a resource).

Note I was unable to test in integration via the Tanium executor so I executed the proposed package command manually in a shell and verified the process trees.

**nohup double fork**
```shell
obas@vm20:~$ sh -c "./openaev-ttr.sh $(echo 'pstree -ps $$ > .secret' | base64)"; cat .secret
[openaev-ttr] Payload launched via nohup (task: openaev-payload-8lDnjGpzUX)
systemd(1)---sh(606452)---sh(606454)---pstree(606455)
```

**systemd**
```shell
obas@vm20:~$ sh -c "./openaev-ttr.sh $(echo 'pstree -ps $$ > .secret' | base64)"; cat .secret
[openaev-ttr] Payload launched via nohup (task: openaev-payload-BOeU7iyktF)
systemd(1)---sh(606790)---sh(606792)---pstree(606793)
```

**atd**
```shell
obas@vm20:~$ sh -c "./openaev-ttr.sh $(echo 'pstree -ps $$ > .secret' | base64)"; cat .secret
[openaev-ttr] Payload scheduled via at (task: openaev-payload-GEQkpiEx0a)
systemd(1)---atd(615763)---atd(616058)---sh(616060)---sh(616061)---pstree(616062)
```

**direct (fallback)**
```shell
obas@vm20:~$ sh -c "./openaev-ttr.sh $(echo 'pstree -ps $$ > .secret' | base64)"; cat .secret
[openaev-ttr] WARN: all detach methods failed, executing directly (task: openaev-payload-jtchkVZBNL)
systemd(1)---sshd(11950)---sshd(538437)---sshd(538457)---bash(538458)---sh(618459)---openaev-ttr.sh(618460)---sh(618470)---sh(61847+
```

Please address the remaining comments before we can approve the PR.

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
Seb-MIGUEL: Removed in 9fbe8cd — you had confirmed during testing that Tanium only URL-encodes the 3 base64 special chars (+, =, /). The CR/LF/space substitutions were defensive but never actually observed, so they were just noise.

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
Seb-MIGUEL: Fixed in 9fbe8cd — now using \cho\ consistently throughout (echo already appends the newline we need for the script content).

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
Seb-MIGUEL: Correct — Tanium agent always runs as root/SYSTEM. Added a comment in the \detach_with_systemd()\ function to make this assumption explicit.

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
Seb-MIGUEL: The sed \i\ flag for case-insensitive matching is a GNU extension — it's not available on macOS/BSD sed. The \[xX]\ bracket notation is POSIX-portable and works on all target platforms. Added a comment in the script explaining this.

guillaumejparis:
⚠️ You're still targeting master, which is a now-decommissioned branch — please target main and rebase your work.

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-tanium-packages-TTR.json
```diff
@@ -9,10 +9,17 @@
         },
         "name": "OpenAEV Subprocessor (Unix) TTR",
         "display_name": "OpenAEV Subprocessor (Unix) TTR",
-        "files": [],
+        "files": [
```
RomuDeuxfois, line 12: **nitpick**: Missing hash/size on openaev-ttr.sh in the Unix package JSON (unlike openaev-ttr.ps1 on Windows). Worth confirming Tanium doesn't require them for integrity verification before deploying to the endpoint. It seems to be a security practice to have it.

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
```diff
@@ -0,0 +1,97 @@
+#!/bin/sh
```
RomuDeuxfois, line 1: **suggestion**: No post-launch verification here, unlike openaev-ttr.ps1 which polls Get-ScheduledTaskInfo for up to 180s to confirm the task actually ran. Right now this script reports success as soon as the submission command is accepted (e.g. at now), even if it never actually executes.

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
```diff
@@ -0,0 +1,97 @@
...
+#   1. nohup/setsid  – double-fork, reparented to PID 1 (most portable)
+#   2. systemd-run   – transient service under PID 1 (Linux with systemd)
+#   3. at now        – spawned by atd (if installed)
+#   4. (fallback)    – direct execution, not detached
+# =============================================================================
+
+set -eu
+
+# ── 0. Validate input ────────────────────────────────────────────────────────
+if [ -z "${1:-}" ]; then
```
RomuDeuxfois, line 23: **suggestion**: openaev-ttr.ps1 silently substitutes "NOOP" when no argument is supplied and still reports [OK] Task has executed (the Get-ScheduledTaskInfo poll only checks LastRunTime, not the actual exit code of the decoded payload), a false success on what is effectively a missing-argument error.
openaev-ttr.sh handles this correctly (explicit exit 1 with a clear stderr message on missing argument).
Suggest aligning .ps1 to fail fast the same way, rather than loosening the Unix script's stricter (and better) behavior.

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-tanium-packages-TTR.json
Seb-MIGUEL: Fixed: added hash (SHA-256) and size fields to the openaev-ttr.sh entry in the JSON package, matching the Windows entry format. Hash and size are now up-to-date with the current script content (commit ed9c29813).

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
Seb-MIGUEL: Fixed: added post-launch verification to each detach method. For nohup/setsid: capture launcher PID, sleep 1s, then verify it exited cleanly (non-zero exit means spawn failed). For systemd-run: added --wait so the call blocks until the transient unit exits and returns the real exit code. The at method remains fire-and-forget as atd does not expose job status via shell primitives.

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
Seb-MIGUEL: Fixed: aligned .ps1 with the stricter .sh behavior. Replaced the silent NOOP substitution with an explicit error and exit 1 when no argument is supplied. Agreed the .sh behavior is correct and the .ps1 should match it.

.github/instructions/backend.instructions.md
```diff
@@ -101,6 +95,24 @@ public Page<{Entity}Output> search(...) { return service.search(input).map({Enti
...
+After submitting the payload, confirm it actually started (not just "accepted"):
+- **nohup/setsid**: capture launcher PID, `sleep 1`, then verify with `kill -0 $PID` or `wait`
+- **systemd-run**: use `--wait` flag so the call blocks until the transient unit exits (real exit code)
+- **`at` / fire-and-forget**: acceptable when the scheduler provides no status API
+
+### Package JSON integrity
+Every file entry in the Tanium package JSON must include `hash` (SHA-256 hex) and `size` (bytes) fields, matching the actual script content. **Recompute hash and size after any script change.**
+```json
+{ "hash": "<sha256>", "name": "openaev-ttr.sh", "source": "", "size": 4648, ... }
+```
```
RomuDeuxfois: **todo**: not needed

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-tanium-packages-TTR.json
Seb-MIGUEL: Fixed in ed9c29813 — added both \hash\ (SHA-256) and \size\ fields to the Unix entry in the JSON. Hashes have also been recomputed in the latest commit after further script changes.

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
Seb-MIGUEL: Added in ed9c29813 — nohup captures the launcher PID, sleeps 1s then verifies with \kill -0\ (nohup path), and \systemd-run\ uses \--wait\ so it blocks until the unit actually exits. The \t\ path is fire-and-forget by nature since atd has no status API, which is acceptable as last resort.

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
Seb-MIGUEL: Fixed in ed9c29813 — removed the silent \\ = \\NOOP\\\ substitution from \openaev-ttr.ps1\ and replaced it with \xit 1\ + an explicit error message, matching the stricter behavior of the Unix script.

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
```diff
@@ -0,0 +1,97 @@
...
+# Tanium URL-encodes the base64 payload; restore the 3 non-alphanumeric
+# characters of standard base64: + = /
+# Note: [xX] bracket notation is used instead of sed's 'i' flag because
+# the 'i' flag is a GNU extension not available on macOS/BSD sed.
+DECODED=$(echo -n "$1" \
+  | sed -e 's#%2[bB]#+#g' -e 's#%3[dD]#=#g' -e 's#%2[fF]#/#g')
+
+# ── 2. Base64-decode ──────────────────────────────────────────────────────────
+PAYLOAD=$(echo -n "$DECODED" | base64 -d 2>/dev/null) \
+  || { echo "[openaev-ttr] ERROR: base64 decode failed" >&2; exit 1; }
```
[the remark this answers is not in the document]
Seb-MIGUEL: Actually, \ase64 -d\ works on both GNU and macOS. The macOS \ase64\ man page explicitly lists both \-d\ and \-D\ as synonyms for decode (https://ss64.com/mac/base64.html), and this was confirmed by @antoinemzs during manual testing on Linux. No change needed.

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
```diff
@@ -0,0 +1,97 @@
...
+PAYLOAD_SCRIPT=$(mktemp /tmp/openaev-payload-XXXXXXXXXX.sh) \
+  || { echo "[openaev-ttr] ERROR: mktemp failed" >&2; exit 1; }
+TASK_ID=$(basename "$PAYLOAD_SCRIPT" .sh)
+
+# Ensure cleanup if the script is interrupted before detach
+trap 'rm -f "$PAYLOAD_SCRIPT"' EXIT INT TERM
+
+{
+  echo '#!/bin/sh'
+  echo "$PAYLOAD"
```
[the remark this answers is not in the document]
Seb-MIGUEL: Fixed in the latest commit — changed to \printf '%s\n' \\\\\\ which is POSIX-portable and handles payloads containing backslashes or leading hyphens correctly.

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
```diff
@@ -0,0 +1,97 @@
...
+if [ -z "${1:-}" ]; then
+  echo "[openaev-ttr] ERROR: no payload argument supplied" >&2
+  exit 1
+fi
+
+# ── 1. URL-decode ─────────────────────────────────────────────────────────────
+# Tanium URL-encodes the base64 payload; restore the 3 non-alphanumeric
+# characters of standard base64: + = /
+# Note: [xX] bracket notation is used instead of sed's 'i' flag because
+# the 'i' flag is a GNU extension not available on macOS/BSD sed.
```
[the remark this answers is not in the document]
Seb-MIGUEL: Fixed in the latest commit — updated the comment to say \[bB]/[dD]/[fF] bracket notation\ to match the actual patterns in the sed expressions.

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-tanium-packages-TTR.json
```diff
@@ -9,10 +9,17 @@
...
+            "source": "",
+            "download_seconds": 0,
+            "direct_download_flag": false
+          }
+        ],
         "file_templates": [],
-        "command": "/bin/sh -c \"/bin/echo $1 | sed 's#%2b#+#gi' | sed 's#%3d#=#gi' | sed 's#%2f#/#gi' | base64 -d | sh & exit 0\"",
-        "command_timeout": 60,
+        "command": "/bin/sh openaev-ttr.sh $1",
+        "command_timeout": 120,
```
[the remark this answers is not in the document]
Seb-MIGUEL: Fixed in the latest commit — updated the \comment\ field to \OpenAEV TTR package definition — manually maintained, not a Tanium Server export\.

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
Seb-MIGUEL: Simplified in 9fbe8cd - task ID is now derived directly from mktemp's XXXXXXXXXX suffix (TASK_ID derived from basename of PAYLOAD_SCRIPT). The OS-provided entropy from mktemp is sufficient and avoids any external dependency on urandom/sha256sum/date.

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
Seb-MIGUEL: Fixed in 9fbe8cd - the sed commands now use '#' as separator (e.g. 's#%2[bB]#+#g') as you suggested. The [bB]/[dD]/[fF] bracket notation is intentional (POSIX-portable) since sed's 'i' flag is GNU-only and doesn't work on macOS/BSD sed.

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
Seb-MIGUEL: Fixed in 9fbe8cd - now using echo -n to pipe the URL-encoded input into sed, as you suggested.

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
Seb-MIGUEL: Agreed - "at" is kept only as 3rd-tier fallback (after nohup and systemd-run), with nohup first since it is POSIX-mandated. The race condition risk is acknowledged; it is acceptable as last-resort because the alternative (the final fallback) executes directly in the Tanium process tree which is worse for TTR detection. If at is not available, the script falls through to direct execution.

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
Seb-MIGUEL: Agreed - "at" is 3rd-tier fallback only, guarded by "command -v at" availability check. On systems where it is not installed the script moves on to direct execution.

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
Seb-MIGUEL: Fixed in 9fbe8cd - order is now: 1) nohup (POSIX-mandated, most portable), 2) systemd-run (Linux with systemd), 3) at (if installed), 4) direct fallback. The comment in the script says "Try methods from most common to least common. nohup is POSIX-mandated and expected to always be available."

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
Seb-MIGUEL: Yes - line 96-99 shows the final else branch: it logs a WARN to stderr and executes the payload directly (in the Tanium process tree). This is intentional as a last resort - the payload still runs, just without detach.

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
Seb-MIGUEL: Fixed - we no longer use $0 for cleanup. The cleanup trap uses the absolute path stored in the PAYLOAD_SCRIPT variable. The same path is embedded in WRAPPED_CMD so the detached process also cleans up via its own trap, even if the payload calls exit.

Seb-MIGUEL: Addressed - cleanup is inside WRAPPED_CMD (the command passed to the detached process): "trap ... EXIT INT TERM; /bin/sh PAYLOAD_SCRIPT". This means the trap fires even if the payload calls exit directly. The main script trap on line 47 is only a safety net for failures before detach.

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
Seb-MIGUEL: Using -d (lowercase) as you suggested - macOS base64 accepts both -d and -D as confirmed in the ss64.com reference and by @antoinemzs testing. Also switched from printf to echo -n for the pipe into base64 as recommended.

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
Seb-MIGUEL: Removed in 9fbe8cd - the %20/%0a/%0d substitutions were defensive but after confirmation that Tanium only URL-encodes the 3 standard base64 special characters (+, =, /), they were removed as noise. A space/CR/LF in a base64 string would indicate data corruption upstream, not something we should silently suppress.

openaev-api/src/main/java/io/openaev/executors/tanium/openaev-ttr.sh
Seb-MIGUEL: Defence in depth - umask 077 ensures the file is created with restrictive permissions, and chmod 700 makes the execute bit explicit. The redundancy is intentional: some implementations of mktemp may not respect umask consistently.

Seb-MIGUEL:
@guillaumejparis Done - rebased on main and changed the PR base to main.

.github/instructions/backend.instructions.md
Seb-MIGUEL: Done in 60901269f — removed the ## Tanium Executor Scripts section from ackend.instructions.md. The file is now identical to main.

RomuDeuxfois (approved):

PR #5448, opened 2026-04-13
[backend] feat(auth): implement JWKS endpoint and JWT validation for cross-platform authentication (#5331)

### LINKED ISSUE #5331: [Auth] Cross-platform JWT trust : expose JWKS endpoint and enforce incoming JWT signature verification
## Context
As part of the Filigran cross-platform authentication initiative, each product must be able to emit signed JWTs on behalf of a logged-in user, and expose its public keys so that other products can validate those JWTs.

## Goals
* Expose a JWKS endpoint at `/xtm/auth/jwks` so that other Filigran products (XTM One, OpenCTI, OpenGRC, ...) can retrieve the * public key(s) used to verify JWTs emitted by OpenAEV.
* When doing cross-product communication, generate a JWT signed with this key
* Accept JWT from other platform based on an issuer trust list

### Key Algorithm
- **Algorithm**: Ed25519 (EdDSA)
- **JWKS key type**: `OKP` with `crv: Ed25519`
- **JWT signing algorithm**: `EdDSA`

### JWKS Endpoint
- **Path:** `GET /xtm/auth/jwks`
- **Auth:** none (public endpoint)
- **Response:** standard JWKS format

```json
{
  "keys": [
    {
      "kty": "OKP",
      "crv": "Ed25519",
      "use": "sig",
      "kid": "<key-id>",
      "alg": "EdDSA",
      "x": "<base64url-encoded-public-key>"
    }
  ]
}
```

### JWT Emission
When OpenAEV needs to call another Filigran product on behalf of a logged-in user, it must emit a signed JWT with the following claims:
- `iss`: `https://<openaev-base-url>` (e.g. `https://openaev.filigran.io`)
- `sub`: OpenAEV id of the logged-in user (can be used to identify user of own JWT)
- `email`: email of the logged-in user (can be used to identify user for 3rd XTM platforms)
- `aud`: base URL of the target product
- `iat` / `exp`: short TTL (max 10 minutes)
- `jti`: unique token ID (for future replay protection)
- `kid` in JWT header: must match a key in the JWKS
- **Header `alg`**: `EdDSA`

### Trusted Platforms (consumer side)
OpenAEV must also be able to **validate** incoming JWTs from other trusted Filigran products:
- Configuration: a list of trusted issuer base URLs (e.g. `https://xtmone.filigran.io`)
- On receiving a JWT: extract `iss`, check it's in the trusted list, fetch `{iss}/xtm/auth/jwks`, cache the JWKS (TTL ~1h), validate the JWT signature using Ed25519
- Force-refresh the JWKS cache if an unknown `kid` is received

## Acceptance Criteria
- [ ] `GET /xtm/auth/jwks` returns a valid JWKS payload with `OKP`/`Ed25519` keys
- [ ] Each key in the JWKS has a `kid`
- [ ] OpenAEV can emit a signed JWT using `EdDSA` (Ed25519) with the correct claims
- [ ] OpenAEV can validate an incoming JWT from a trusted Filigran issuer using Ed25519
- [ ] The list of trusted issuer base URLs is configurable via environment variables
- [ ] A JWT from an issuer not in the trusted list is rejected
- [ ] JWKS is cached on the consumer side with force-refresh on unknown `kid`

## Notes
- **Key choice**: Ed25519 (EdDSA) — compact, fast, and modern. Preferred over RSA/RS256.
- This is a transitional design. A centralized auth service will eventually replace per-product key management. The JWKS endpoint path `/xtm/auth/jwks` is the agreed Filigran internal convention.
- `sub` uses email for now — may migrate to a stable UUID in a future iteration.

xfournet:
This pull request introduces a new cross-platform JWT authentication mechanism using JWKS discovery, exposes the platform's public JWKS for verification by other services, and refactors the authentication flow to support these changes. It removes the old static key-based JWT validation and replaces it with dynamic JWKS-based validation, ensuring improved security and interoperability. Additionally, it introduces deterministic Ed25519 keypair derivation for consistent JWT signing and public key exposure.

Fix #5331 

**Cross-platform JWT authentication and JWKS support:**

* Added the `/xtm/auth/jwks` endpoint via the new `XtmAuthApi` controller to expose the platform's public JWKS for JWT verification by other Filigran products. [[1]](diffhunk://#diff-c990efb0cef7bbf5b162ef492a0bd8e081d10d2822ff806cf20374f177265003R1-R54) [[2]](diffhunk://#diff-c5e7ec24e4ae276754259782dfef8285dca127363e345bc52a31df54a15cb277R100-R101)
* Implemented `XtmJwksExtractor`, a new JWT extractor that validates incoming JWTs using JWKS discovery, issuer validation, and key caching, replacing the previous static key-based validation.

**Key management and JWT issuance:**

* Added `XtmAuthKeyService` to deterministically derive an Ed25519 keypair from the platform's encryption key using HKDF-SHA256, ensuring HA-safe and consistent JWT signing.
* Updated `XtmOneClient` to sign authentication JWTs with the derived Ed25519 key, set the correct issuer/audience, and include a unique `kid` in the JWT header. [[1]](diffhunk://#diff-b9c16fb8f05fbd2d3bf4857f01aaef194e47c9d86e9b9e6dafc9401646e0e401R5-R18) [[2]](diffhunk://#diff-b9c16fb8f05fbd2d3bf4857f01aaef194e47c9d86e9b9e6dafc9401646e0e401R30-R54)

**Authentication flow refactoring and cleanup:**

* Refactored `TokenAuthenticationFilter` to replace the old `PlatformJwtExtractor` with the new `XtmJwksExtractor` throughout the codebase. [[1]](diffhunk://#diff-7c4e92a14d2291ee5f931ec14c96f76ac9f6ae0727f1199e3caba5026453ec5fL10-R10) [[2]](diffhunk://#diff-7c4e92a14d2291ee5f931ec14c96f76ac9f6ae0727f1199e3caba5026453ec5fL35-R35) [[3]](diffhunk://#diff-7c4e92a14d2291ee5f931ec14c96f76ac9f6ae0727f1199e3caba5026453ec5fL48-R49) [[4]](diffhunk://#diff-7c4e92a14d2291ee5f931ec14c96f76ac9f6ae0727f1199e3caba5026453ec5fL59-R59)
* Removed legacy classes `PlatformJwtExtractor` and `KeyFromAlgHeader` that handled static key-based JWT validation. [[1]](diffhunk://#diff-dba08002471102a567f622e45422cf562c30dae641b4893f2b1b783979703c35L1-L60) [[2]](diffhunk://#diff-0eb3d61d3a43c4faa68f27e473e5bb3f4fc25f04b485836f40e43f9418a4415eL1-L59)


laugiov:
Thanks for the walkthrough on this PR, nice entry point into the XTM auth flow.
One technical question and a broader one:
- On audience validation: I can see `XtmOneClient` sets an `aud` when issuing, but I didn't spot the check on the `XtmJwksExtractor` side. Is that intentional, or would it be worth adding a `.requireAudience(...)`? Asking because as the XTM suite grows, a JWT issued for another service being accepted here would become an interesting vector (token replay).
- On the broader angle: this HKDF+Ed25519 pattern is borrowed from OpenCTI if I understand correctly. Does Filigran already have a shared spec for XTM inter-service auth, or is it still handled case by case between products? I'm asking because that's typically the kind of topic I can be useful on from a cross-cutting angle, but I'd like to understand where things stand before proposing anything.

laugiov:
Tested the full JWKS auth flow locally with OpenAEV (this branch) + filigran-copilot on the right branch.

Setup: OpenAEV on :8080, Copilot on :8100, registered OpenAEV as a platform via /api/v1/platform/register.

Direction 1 : OpenAEV → Copilot: Confirmed. Every chatbot request sends an EdDSA JWT with the correct kid, iss (OpenAEV base URL), aud (Copilot URL), email claim, 10-min TTL, and unique jti. Copilot accepts and processes the requests without auth errors.

Direction 2 : Copilot → OpenAEV: Confirmed. Asked "list my scenarios" in the chatbot, Copilot called search_openaev_scenarios against the OpenAEV API and returned real data (3 scenarios found). The tool call completed successfully .

Also verified:

JWKS endpoint (/xtm/auth/jwks) returns valid OKP/Ed25519 key on both sides
Unit tests pass (9/9 on TokenAuthenticationFilterTest)
Key derivation is deterministic from encryption_key as expected
LGTM on the auth flow.

openaev-api/src/main/java/io/openaev/rest/xtm_auth/XtmAuthApi.java
```diff
@@ -0,0 +1,49 @@
...
+
+  @GetMapping(value = XTM_AUTH_JWKS_URI, produces = MediaType.APPLICATION_JSON_VALUE)
+  @Operation(
+      summary = "Get JWKS",
+      description =
+          "Returns the public key(s) used to verify JWTs emitted by this OpenAEV instance. "
+              + "Standard JWKS format with OKP/Ed25519 keys.")
+  @ApiResponse(responseCode = "200", description = "JWKS payload")
+  @LogExecutionTime
+  public Map<String, Object> jwks() {
```
antoinemzs, line 38: TODO: in general, prefer a strongly typed DTO as the output contract; part of the OpenAEV coding standard.

You may define the DTO in the same package as the endpoint itself; use `jackson` for declarative serialisation

openaev-api/src/main/java/io/openaev/security/token/XtmJwksExtractor.java
```diff
@@ -0,0 +1,172 @@
...
+      throw new AuthenticationError("Failed to extract issuer from JWT: " + e.getMessage());
+    }
+  }
+
+  private static String normalizeUrl(String url) {
+    if (url == null) {
+      return "";
+    }
+    return url.endsWith("/") ? url.substring(0, url.length() - 1) : url;
+  }
```
antoinemzs, line 171: This should be implemented in `XtmOneConfig` and used as a filter for retrieving the URL, so that it is always normalised throughout

openaev-api/src/main/java/io/openaev/security/token/XtmJwksExtractor.java
```diff
@@ -0,0 +1,172 @@
...
+        .filter(k -> kid.equals(k.getId()))
+        .findFirst()
+        .map(jwk -> (Key) jwk.toKey())
+        .orElse(null);
+  }
+
+  private void refreshJwks(String issuer) {
+    try {
+      String jwksUrl = issuer + "/xtm/auth/jwks";
+      String jwksJson = restTemplate.getForObject(jwksUrl, String.class);
```
antoinemzs, line 135: We should always use this for initiating outbound http requests:
https://github.com/OpenAEV-Platform/openaev/blob/e8056fb155b7f89aa1e4b1ba9aed447a10608fb0/openaev-api/src/main/java/io/openaev/authorisation/HttpClientFactory.java#L16-L24

and instantiate it as a try-with-resource assignment (https://www.baeldung.com/java-try-with-resources) (this is the "closeable" part of `CloseableHttpClient`), for example:
https://github.com/OpenAEV-Platform/openaev/blob/b4e323f96f96ebdaaa3d69f812ef58f9f54d0e5e/openaev-api/src/main/java/io/openaev/executors/crowdstrike/client/CrowdStrikeExecutorClient.java#L232

The reason is that the platform supports adding discrete TLS certificates for e.g. custom CAs, and this requires using the custom `SslContext` wrapped in this client. Obtain a client with the bean `HttpClientFactory`.

openaev-api/src/main/java/io/openaev/security/token/XtmJwksExtractor.java
```diff
@@ -0,0 +1,172 @@
...
+    if (!issuer.equals(trustedIssuer)) {
+      throw new AuthenticationError("Untrusted JWKS issuer: " + issuer);
+    }
+
+    Claims claims =
+        Jwts.parser()
+            .keyLocator(header -> resolveKey(issuer, (String) header.get("kid")))
+            .build()
+            .parseSignedClaims(value)
+            .getPayload();
```
antoinemzs, line 77: It isn't useful to explicitly parse the `iss` claim if we have a single authorised issuer (currently identified as `XtmOneConfig.getUrl()`).
```suggestion
    String trustedIssuer = normalizeUrl(xtmOneConfig.getUrl());

    Claims claims =
        Jwts.parser()
            .requireIssuer(trustedIssuer)
            .keyLocator(header -> resolveKey(trustedIssuer, (String) header.get("kid")))
            .build()
            .parseSignedClaims(value)
            .getPayload();
```

openaev-api/src/main/java/io/openaev/security/token/XtmJwksExtractor.java
```diff
@@ -0,0 +1,172 @@
...
+    if (cached.fetchedAt().plus(JWKS_CACHE_TTL).isBefore(Instant.now())) {
+      refreshJwks(issuer);
+      cached = jwksCache.get(issuer);
+    }
+    if (cached == null) {
+      return null;
+    }
+
+    return Jwks.setParser().build().parse(cached.jwksJson()).getKeys().stream()
+        .filter(k -> kid.equals(k.getId()))
```
antoinemzs, line 126: NPE if `kid` is not part of the header, which is not controlled by us because part of the submitted JWT.

Prefer the opposite check. It's more likely that `k.getId()` would not be null (because we specifically get it from a JWKS endpoint).

Suggestion for illustration, do feel free to improve the null checking here as well.
```suggestion
        .filter(k -> k.getId().equals(kid))
```
or better
```suggestion
        .filter(k -> !StringUtils.isBlank(kid) && kid.equals(k.getId()))
```

openaev-api/src/main/java/io/openaev/service/xtm_auth/XtmAuthKeyService.java
```diff
@@ -0,0 +1,110 @@
...
+  /**
+   * HKDF-SHA256 Extract-and-Expand (RFC 5869) to derive a 32-byte Ed25519 seed from the platform
+   * encryption key.
+   */
+  private static byte[] deriveEd25519Seed(String encryptionKey) {
+    byte[] ikm = encryptionKey.getBytes(java.nio.charset.StandardCharsets.UTF_8);
+    byte[] seed = new byte[ED25519_SEED_LENGTH];
+
+    HKDFBytesGenerator hkdf = new HKDFBytesGenerator(new SHA256Digest());
+    hkdf.init(new HKDFParameters(ikm, /* salt: empty */ new byte[0], HKDF_INFO));
```
antoinemzs, line 71: question: if we are going to repurpose the platform-specific encryption key, why not repurpose the platform-specific salt `adminConfig.getEncryptionSalt()`?

Note that if there is a risk of accidentally duplicating both components (e.g. platform cloning) rather than only one, so there's a caveat.

openaev-api/src/main/java/io/openaev/service/xtm_auth/XtmAuthKeyService.java
```diff
@@ -0,0 +1,110 @@
...
+    hkdf.generateBytes(seed, 0, seed.length);
+    return seed;
+  }
+
+  /**
+   * Builds a JCA {@link KeyPair} from a raw Ed25519 seed using BouncyCastle's lightweight API, then
+   * converts to standard JCA types via PKCS8/X509 encoding for jjwt compatibility.
+   */
+  private static KeyPair buildJcaKeyPair(byte[] seed) throws Exception {
+    Ed25519PrivateKeyParameters bcPriv = new Ed25519PrivateKeyParameters(seed, 0);
```
antoinemzs, line 81: preferable not to use the overloaded ctor if not using a meaningful value for `offset`.

openaev-api/src/main/java/io/openaev/xtmone/XtmOneClient.java
```diff
@@ -26,37 +27,31 @@ public class XtmOneClient {
...
+        .header()
+        .keyId(keyService.getKid())
+        .and()
+        .issuer(openAEVConfig.getBaseUrl())
         .subject(userId)
         .claim("name", userName)
         .claim("email", userEmail)
-        .issuer("openaev")
+        .audience()
+        .add(config.getUrl())
```
antoinemzs, line 49: <strike>Since we're creating a (transitional) many emitters, many consumers scheme, it seems odd to hardcode the audience to `XtmOneCOnfig.getUrl()` (also note it isn't normalised here).

It should perhaps be at least based on the configured trusted issuer (even if it's currently the same value, but different interface), and potentially soon many trusted issuers, as part of the same "federation"?</strike>

openaev-api/src/main/java/io/openaev/xtmone/XtmOneClient.java
```diff
@@ -26,37 +27,31 @@ public class XtmOneClient {
 
   private final XtmOneConfig config;
   private final ObjectMapper objectMapper;
+  private final XtmAuthKeyService keyService;
+  private final OpenAEVConfig openAEVConfig;
   private final HttpClient httpClient =
       HttpClient.newBuilder()
           .version(HttpClient.Version.HTTP_1_1)
           .connectTimeout(Duration.ofSeconds(15))
           .build();
```
antoinemzs, line 36: Incidentally I notice the same remark applies to this http client, we should always use the one from the `HttpClientFactory` bean.

antoinemzs (changes requested):
On design: an open question on whether it's intended to have a single "trusted issuer" hardcoded to `XtmOneConfig.getUrl()` and not an array, as I seemed to understand we wanted to enable this scheme across many compatible products. This choice makes it moot to dig up the `iss` claim for getting the keys because the url is directly accessible from the config object.

On the rest: some housekeeping, using the correct libraries and standards.

openaev-api/src/main/java/io/openaev/xtmone/XtmOneClient.java
antoinemzs: Ok, so it's in the XtmOneClient so it makes sense that we'd only set the configured XTM One audience.

nitpick: JWT crafting should be in its own service

openaev-api/src/main/java/io/openaev/rest/xtm_auth/XtmAuthApi.java
xfournet: DTO added

openaev-api/src/main/java/io/openaev/service/xtm_auth/XtmAuthKeyService.java
xfournet: Fixed

openaev-api/src/main/java/io/openaev/service/xtm_auth/XtmAuthKeyService.java
xfournet: salt doesn't bring security improvement and complexify the backup (need to save both key and salt). I don't known what was the intention to have both key & salt, so for the moment i aligned on OpenCTI key derivation approach. Let me know if we need to discuss of that

openaev-api/src/main/java/io/openaev/security/token/XtmJwksExtractor.java
xfournet: fixed

openaev-api/src/main/java/io/openaev/security/token/XtmJwksExtractor.java
xfournet: i rather aligned on the way we have done on OpenCTI/XTMOne using a list (even if more the moment we only have one element in it)

openaev-api/src/main/java/io/openaev/security/token/XtmJwksExtractor.java
xfournet: fixed according your suggestion

openaev-api/src/main/java/io/openaev/xtmone/XtmOneClient.java
xfournet: Fixed

openaev-api/src/main/java/io/openaev/xtmone/XtmOneClient.java
xfournet: it was a question for me to create a 'universal' JWT that could be forwarded through many services, or just re-emit new token based on the same user.
Since XTM One was not ready for the first option, i do the 2nd one for the moment. This may change later.

openaev-api/src/main/java/io/openaev/security/token/XtmJwksExtractor.java
xfournet: moved to config

openaev-api/src/main/java/io/openaev/service/xtm_auth/XtmAuthKeyService.java
xfournet: to confirm, encryptionKey is supposed to be randomly generated ?

antoinemzs: yes, the instruction is to have a random string generated using an underlying CSPRNG (though currently outside of the platform's control), so I agree with your statement about security if we accept this as fact. Though we shouldn't reuse salts, so there's also that.

We have both encryptionKey and salt defined because they were the two components of the encryption lib we used for the catalog secrets. We opted for the users to provide both values, which has its drawbacks.

Yoiur approach is good. Note we already have the de facto requirement to back up both config values (key and salt).

xfournet: yes my concern was more a OCTI/OAEV alignment concern

openaev-api/src/main/java/io/openaev/security/token/XtmJwksExtractor.java
```diff
@@ -0,0 +1,196 @@
...
+                  return null;
+                }
+                return EntityUtils.toString(response.getEntity(), StandardCharsets.UTF_8);
+              });
+      if (jwksJson != null) {
+        jwksCache.put(issuer, new CachedJwks(Instant.now(), jwksJson));
+        log.debug("Refreshed JWKS cache for issuer {}", issuer);
+      }
+    } catch (Exception e) {
+      log.warn("Failed to fetch JWKS from {}: {}", issuer, e.getMessage());
```
antoinemzs: Use the signature that takes the throwable so that we get the full picture in case of breakage. (sorry I missed it the first time)
```suggestion
      log.warn("Failed to fetch JWKS from {}.", issuer, e);
```

openaev-api/src/main/java/io/openaev/security/token/XtmJwksExtractor.java
```diff
@@ -0,0 +1,196 @@
...
+      JsonNode payload = objectMapper.readTree(payloadJson);
+      JsonNode issNode = payload.get("iss");
+      if (issNode == null || issNode.isNull()) {
+        throw new AuthenticationError("JWT has no 'iss' claim");
+      }
+      return issNode.asText();
+    } catch (AuthenticationError e) {
+      throw e;
+    } catch (Exception e) {
+      throw new AuthenticationError("Failed to extract issuer from JWT: " + e.getMessage());
```
antoinemzs: nitpick: we should expose the overloaded ctor in `AuthenticationError` to embed `e` in the wrapping `AuthenticationError` exception object.

e.g.
```java
public AuthenticationError(String message, Throwable cause) {
  super(message, cause);
}
```

antoinemzs (approved):
couple nitpicks which would be nice addressing; otherwise good to go I think, thanks for the efforts 🙏

openaev-api/src/main/java/io/openaev/security/token/XtmJwksExtractor.java
xfournet: Fixed

openaev-api/src/main/java/io/openaev/security/token/XtmJwksExtractor.java
xfournet: Fixed

PR #5451, opened 2026-04-14
[frontend/backend] feat(threat-arsenal): add import and export process (#4458)

### LINKED ISSUE #4458: [SCV]  Consolidate Payloads & Contracts into a Unified Page
The system currently exposes **Payloads,  Contract** separately

The user should instead interact with **one entity: an Action**

We need to simplify the UI by merging the *Payloads list+* the Injectors *Contracts in one page*

The system must continue to manage payloads and contracts internally, but the user should only interact with a unified and simplified representation 1 **Actions**

Each Action must expose the metadata required for simulations and we will implement the information that will miss (e.g Nuclei that not have Source, Status etc)

GaetanSantucci:
#### Goal
Preserve all existing import/export capabilities currently available on the Payload page and 
the Injector page, now consolidated under the Threat Arsenal page.

---

#### CSV Export
Export all actions globally as a CSV file with the following columns:

`type` · `name` · `domains` · `platforms` · `status` · `tags` · `updated_at` · `description` · `source` · `created_at` · `attack_pattern` · `origin`

---

#### JSON Export
- Export actions as JSON — **payload-based actions only**
- Actions originating from an injector **cannot** be exported in this format

---

#### Import
Support two JSON import scenarios:

| Scenario | Description |
|----------|-------------|
| **Threat Arsenal action** | Export a JSON action from Threat Arsenal, then re-import it |
| **Legacy Payload** | Export a JSON payload from a previous OpenAEV version, then import it into the Threat Arsenal page of the new version |

### Related issues

* Closes #4458


antoinemzs:
@GaetanSantucci please can you add the test cases to cover in the OP thanks!

antoinemzs:
Please also increase the test coverage, with help from codecov for seeing the areas to improve

openaev-api/src/main/java/io/openaev/utils/TargetType.java
```diff
@@ -30,5 +30,8 @@ public enum TargetType {
   TEAMS,
 
   /** Endpoint devices in the infrastructure. */
-  ENDPOINTS
+  ENDPOINTS,
+
+  /** Injector contracts that define the rules and parameters for inject execution. */
+  INJECTOR_CONTRACTS
```
RomuDeuxfois: **question**: I am not sure to understant why Injector Contract can be a target type. Can you explain it ?

openaev-api/src/main/java/io/openaev/api/payload/PayloadImportService.java
```diff
@@ -0,0 +1,170 @@
...
+   * @return the import result containing the persisted payload and the synchronised injector
+   *     contract
+   */
+  public PayloadImportResult importPayload(MultipartFile file) throws Exception {
+    ZipJsonService.ImportOutput<Payload> response = zipJsonApi.handleImport(file, "payload_name");
+
+    List<AttackPattern> attackPatterns =
+        extractRelationshipObjects(
+            "attack_patterns", this::handleAttackPatternImport, response.sourceDocument());
+    List<Domain> domains =
```
RomuDeuxfois, line 71: I don't understand why we handle it here.
Payload doesn"t have domain properties for example, but we have it on the UI ?How it's possible ?

openaev-api/src/main/java/io/openaev/api/threat_arsenal/ThreatArsenalApiExporter.java
```diff
@@ -0,0 +1,51 @@
...
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+@RequestMapping({ThreatArsenalApi.THREAT_ARSENAL_URL, ThreatArsenalApi.TENANT_THREAT_ARSENAL_URL})
+@RequiredArgsConstructor
+public class ThreatArsenalApiExporter {
+
+  private final ZipJsonApi<InjectorContract> zipJsonApi;
+  private final InjectorContractRepository injectorContractRepository;
```
RomuDeuxfois: **nitpick**: please use services instead of repositories here

openaev-api/src/main/java/io/openaev/api/threat_arsenal/ThreatArsenalApiExporter.java
```diff
@@ -0,0 +1,51 @@
...
+  public ResponseEntity<byte[]> export(@PathVariable @NotBlank final String actionId)
+      throws IOException {
+    Map<String, Boolean> opts = new HashMap<>();
+    opts.put("exclude from action export", false);
+    IncludeOptions includeOptions = IncludeOptions.of(opts);
+    InjectorContract injectorContract =
+        injectorContractRepository.findById(actionId).orElseThrow(ElementNotFoundException::new);
+    if (injectorContract.getPayload() == null) {
+      throw new ElementNotFoundException(
+          "Only injector contract based on payload can be exported ");
```
RomuDeuxfois, line 46: **question**: Why we can't export email or SMS for instance ? Maybe a product question

openaev-api/src/main/java/io/openaev/rest/attack_pattern/service/AttackPatternService.java
```diff
@@ -255,4 +257,29 @@ public Map<String, AttackPattern> fetchInternalAttackPatternIds(
...
     return attackPatternRepository.findAllByIdIn(idsAttackPattern);
   }
+
+  /**
+   * Upsert an attack pattern
+   *
+   * @param input to handle upsert values
+   * @return found or created attack pattern
+   */
+  public AttackPattern upsert(AttackPatternCreateInput input) {
```
RomuDeuxfois: **question**: Seems to be used only on test, but it can be used on AttackPatternApi

openaev-model/src/main/java/io/openaev/database/repository/AttackPatternRepository.java
```diff
@@ -27,6 +27,8 @@ public interface AttackPatternRepository
   List<AttackPattern> findAllByExternalIdInIgnoreCaseAndTenantId(
       List<String> externalIds, String tenantId);
 
+  Optional<AttackPattern> findByExternalIdIgnoreCaseAndTenantId(String externalId, String tenantId);
```
RomuDeuxfois: **nitpick**: Why not using findAllByExternalIdInIgnoreCaseAndTenantId with List.of(ID) to not duplicate method ?

openaev-api/src/main/java/io/openaev/api/payload/PayloadImportService.java
```diff
@@ -0,0 +1,170 @@
...
+   * <p>Legacy payload exports may contain {@code payload_attack_patterns}, {@code payload_domains},
+   * and {@code payload_tags} as relationships. Since these fields now live on {@code
+   * InjectorContract}, they are extracted from the source document and passed to the
+   * synchronisation method.
+   *
+   * @param file the ZIP file containing the JSON:API payload document
+   * @return the import result containing the persisted payload and the synchronised injector
+   *     contract
+   */
+  public PayloadImportResult importPayload(MultipartFile file) throws Exception {
```
RomuDeuxfois, line 64: Ok, I think we need to change something in the code to improve the logic here.

If a payload can have domains and TTPs in the UI, then Payload should directly own these relationships in the Java class. That would simplify the code: we could export the payload with its domains and TTPs, and handle it normally.

Synchronization would not be necessary, since InjectorContract could expose methods like:

getDomains() {
  return payload != null ? payload.getDomains() : this.getDomains();
}

The UI should reflect the same logic as well.

antoinemzs: Actually we have made the schema leaner in the Threat Arsenal feature; it was always an anomaly that the `Payload` class would bear some `InjectorContract` concerns directly such as the `Domain` and `AttackPattern` relationships.

The Threat Arsenal is now subsuming all `InjectorContract` objects including those linked to `Payload` as `Action` objects.

This is an API level contract; the underlying data model is still split between `InjectorContract` and an optional `Payload`.

>Synchronization would not be necessary, since InjectorContract could expose methods like:

We preferred opting for a fully normalised db schema to ease on other use cases such as querying the SQL directly (e.g. elastic sync) and not create surprises (100% consistent) and not require going through the java layer to have consistent data.

openaev-api/src/main/java/io/openaev/api/payload/PayloadImportService.java
antoinemzs: It's the `Action` API contract that does this. The scheme is similar to how we treat the `InjectTarget` objects, which are an API layer contract and map to various database entities (Asset Group, Endpoint, Agent, Team, Player).

antoinemzs: Note this addition is a backwards compatibility measure. The export/import function is tied to the database schema (annotations on the JPA entities), so when modifying the db schema, we broke the contract for already exported `Payload` objects. The older schema specified some attributes on `Payload` that were moved to `InjectorContract`, so this is bridging the gap.

Newer exports won't need this.

openaev-api/src/main/java/io/openaev/api/payload/PayloadImportService.java
RomuDeuxfois: All right, if InjectorContract is responsible for handling Domain and AttackPattern, for example, then these fields should not be present in payload creation, import, or export, in order to remain consistent.

openaev-api/src/main/java/io/openaev/api/payload/PayloadImportService.java
RomuDeuxfois: I don’t agree.
Payload should use Domain and TTPs as first-class fields, given what we want to do with the payload repository (atomic red team like)

MarineLeM: Previously, we maintained relationships between **Payload ↔ Domains** and **Payload ↔ AttackPatterns**, but these were duplicates of the relationships already defined in **InjectorContract ↔ Domains** and **InjectorContract ↔ AttackPatterns**, and were kept in sync with them.

Having duplicate relation tables in the database is unnecessary and error-prone.
Therefore, in this first chunk, we remove the redundant Payload-side relation tables and keep the single source of truth on the **InjectorContract** side.

openaev-api/src/main/java/io/openaev/api/threat_arsenal/ThreatArsenalApiExporter.java
MarineLeM: The issue was to preserve the existing behavior from master, and the JSON export from InjectorContract based on an injector is not enabled.

openaev-api/src/main/java/io/openaev/api/payload/PayloadImportService.java
MarineLeM: With the Threat Arsenal issue, we no longer call the PayloadApi directly from the FrontEnd — we keep it only for users who interact with the OpenAEV API directly.

To avoid breaking the existing API, the Payload creation DTO still exposes domains and attackPatterns fields, but those relations are now linked to the InjectorContract instead. In practice, the behavior has not changed: previously, creating a payload in the backend would also create an InjectorContract along with four relationships — InjectorContract ↔ Domains, InjectorContract ↔ AttackPatterns, Payload ↔ Domains, and Payload ↔ AttackPatterns. We simply removed the last two, as they were duplicates.

MarineLeM:
@RomuDeuxfois ready for a second round 🚀

RomuDeuxfois (approved):

RomuDeuxfois:
@MarineLeM I don't have the time to fully tests you works, but I don't want to be a blocker.
So, it's good to go on my side.

MarineLeM:
@RomuDeuxfois I will re-test before merging

PR #5516, opened 2026-04-17
[tool] chore: add ci for testing-xtm-one (#1)

raugier:
https://github.com/FiligranHQ/platform/issues/2435

### DIFF
--- a/.github/workflows/deploy-testing-xtm-one.yml
+++ b/.github/workflows/deploy-testing-xtm-one.yml
@@ -74,7 +74,7 @@ jobs:
 
   deploy:
       name: Restart testing-xtm-one instance
-      needs: wf-build-image
+      needs: build-image
       runs-on: ubuntu-latest
       steps:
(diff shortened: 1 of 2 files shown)


efaure (changes requested):
It's not a bad idea to reuse existing workflow, but I'm not sure that it's relevant to use the feature flag deployment for this testing-xtm-one env. Feature flag instance are specific (deleted after one week of unactivity, the host name won't be controllable). 
I think the ideal would be to 
 - extract from existing feautre branch workflow the "build" part (as it's done for opencti) 
 - use this build part both in your xtm one workflow and in feature branch
 - add to xtm one a step that will just trigger a restart of openaev instance to refresh the image

In order to reduce risk of breaking what is working, I suggest to start by creating a separate workflow for build, use it in xtm one, and only once it's clear that it work fine, use it in feature branch (no urgency on this last point)

efaure (approved):

Issue #5515, opened 2026-04-17
[Multi-tenancy] Email login check tenant membership


corinnekrych:
Tenant isolation issue:
I log in as admin user:
- create tenant A
- land in tenant A, go to Setting | security, create a user a@gmail.com associated with tenant A
- create tenant B
- land in tenant B, go to Setting | security, create a user b@gmail.com associated with tenant B
- logout => the url is http://localhost:3001/a716d13f-3e52-4dc7-b223-b9bc7a18fa28 with tenant A id.
EXPECTED: logout should forward to http://localhost:3001/, not tenant A url
- login to http://localhost:3001/a716d13f-3e52-4dc7-b223-b9bc7a18fa28 (tenant A URL) with b@gmail.com user
ACTUAL: login succeeds, and land in tenant A URL, all API call will use tenant A id, Tenant Switcher will display tenant B 
EXPECTED: login should fail, because b@gmail.com is associated with tenant B, not tenant A

PR #5517, opened 2026-04-17
[backend/frontend] fix(multi-tenant): enforce tenant isolation (#5515)

corinnekrych:
…gin and logout

- Backend: validate user belongs to requested tenant during local login, return 404 if not (authz should not give hint on grant so no 403 but 404 as implemented in code base)
- Backend: add tenantId field to LoginUserInput
- Backedn 401 when user A try to connect to tenant B (no membership for this tenant)
- Frontend: send tenant ID from URL during login
- Frontend: logout redirects to root URL instead of tenant-scoped URL

Screenshot 2026-04-20 at 12 33 32

### Proposed changes

*
*

### Testing Instructions

1. Step-by-step how to test

### Related issues

* Closes #ISSUE-NUMBER
*

### Checklist

### Further comments


openaev-api/src/main/java/io/openaev/config/TenantInterceptor.java
```diff
@@ -1,30 +1,54 @@
...
-      TenantContext.setCurrentTenant(pathVariables.get(TENANT_ID_PATH_VARIABLE));
+      String tenantId = pathVariables.get(TENANT_ID_PATH_VARIABLE);
+
+      // Validate the authenticated user belongs to this tenant
+      Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
+      if (authentication != null
+          && authentication.isAuthenticated()
+          && !ANONYMOUS_USER.equals(authentication.getPrincipal())) {
+        OpenAEVPrincipal principal = (OpenAEVPrincipal) authentication.getPrincipal();
+        if (!principal.isAdmin()
```
damgouj: question: admin user can access the tenant he wants ?

openaev-api/src/main/java/io/openaev/config/TenantInterceptor.java
```diff
@@ -1,30 +1,54 @@
...
+            && !tenantRepository.existsByUserIdAndTenantId(principal.getId(), tenantId)) {
```
damgouj: suggestion: maybe we could try to improve this request with cache or something else ? Each time we have a HTTP call, we check the tenant of the user, is it not too costly ?

openaev-api/src/main/java/io/openaev/rest/user/UserApi.java
```diff
@@ -77,6 +79,17 @@ public User login(@Valid @RequestBody LoginUserInput input) {
     if (optionalUser.isPresent()) {
       User user = optionalUser.get();
       if (userService.isUserPasswordValid(user, input.getPassword())) {
+        // Verify tenant membership if a tenant ID was provided
+        if (input.getTenantId() != null && !input.getTenantId().isBlank()) {
+          boolean belongsToTenant =
+              tenantService.findTenantsByUserId(user.getId()).stream()
```
damgouj: suggestion: maybe you can directly use existsByUserIdAndTenantId here ?

damgouj (changes requested):
Tests OK with local auth and SSO :)

openaev-api/src/main/java/io/openaev/config/TenantInterceptor.java
corinnekrych: Good point! 

isAdmin is set to true only for user creted at platform level with the checkbox "is admin" true. 
if we consider this as a super admin with all grants, we could say "yes, all admin have access to all tenant" but then there are some inconsistencies in our current implementation:
- default admin at start up of the platform creates an admin2 user at platform level
- default admin creates tenant A
- default admin logs out
- admin2 login => no tenant available in TenantSwitcher (no even the default tenant)
- admin can curl and access BE endpoint of all tenants

2 options:
- option1: we want admin user to have access to all tenant.s Each time, a platform user (with cap tenant settings) add a tenant all admin should have membership added for that new tenant
- option2: we dont want admin to access all tenants but admin can create tenant users. and can "see" all tenant.

cc @RomuDeuxfois @Seb-MIGUEL toughts?

openaev-api/src/main/java/io/openaev/config/TenantInterceptor.java
corinnekrych: fixed in [d93d995](https://github.com/OpenAEV-Platform/openaev/pull/5517/commits/d93d995239de9d11dde56de2d49e42ac228c2cf0) added a caffeine cache for membership

openaev-api/src/main/java/io/openaev/rest/user/UserApi.java
corinnekrych: fixed in [d93d995](https://github.com/OpenAEV-Platform/openaev/pull/5517/commits/d93d995239de9d11dde56de2d49e42ac228c2cf0)

openaev-api/src/main/java/io/openaev/config/TenantInterceptor.java
corinnekrych: Actually we ALSO have "isAdmin" when we create user at tenant level. 
When the "admin" tenant user log in he can see both Platform and Tenants setting. He can see all tenant but can not go inside tenant he had no membership.
=> I remember the discussion with JB I think we should go with option2. let me remove the check isAdmin!

corinnekrych: fixed in f1485d420

damgouj (approved):

PR #5527, opened 2026-04-20
[backend] fix(api): prevent IDOR on inject endpoints by validating parent ownership

RomuDeuxfois:
### Proposed changes

* Closed https://github.com/OpenAEV-Platform/filigran-private/issues/76

### DIFF
--- a/openaev-api/src/main/java/io/openaev/rest/inject/ScenarioInjectApi.java
+++ b/openaev-api/src/main/java/io/openaev/rest/inject/ScenarioInjectApi.java
@@ -7,7 +7,6 @@ import static io.openaev.utils.pagination.PaginationUtils.buildPaginationCriteri
 import io.openaev.aop.RBAC;
 import io.openaev.database.model.*;
 import io.openaev.database.repository.*;
-import io.openaev.rest.exception.ElementNotFoundException;
 import io.openaev.rest.helper.RestBehavior;
 import io.openaev.rest.inject.form.InjectAssistantInput;
 import io.openaev.rest.inject.form.InjectInput;
@@ -44,6 +43,8 @@ public class ScenarioInjectApi extends RestBehavior {
   private final InjectDuplicateService injectDuplicateService;
   private final ScenarioInjectService scenarioInjectService;
 
+  // -- READ --
+
   @GetMapping(SCENARIO_URI + "/{scenarioId}/injects/simple")
   @RBAC(
       resourceId = "#scenarioId",
@@ -78,6 +79,30 @@ public class ScenarioInjectApi extends RestBehavior {
         Inject.class,
         joinMap);
   }
+
+  @GetMapping(SCENARIO_URI + "/{scenarioId}/injects")
+  @RBAC(
+      resourceId = "#scenarioId",
+      actionPerformed = Action.READ,
+      resourceType = ResourceType.SCENARIO)
+  public Iterable<Inject> scenarioInjects(@PathVariable @NotBlank final String scenarioId) {
+    return this.injectRepository.findByScenarioId(scenarioId).stream()
+        .sorted(Inject.executionComparator)
+        .toList();
+  }
+
+  @GetMapping(SCENARIO_URI + "/{scenarioId}/injects/{injectId}")
+  @RBAC(
+      resourceId = "#scenarioId",
+      actionPerformed = Action.READ,
+      resourceType = ResourceType.SCENARIO)
+  public Inject scenarioInject(
+      @PathVariable @NotBlank final String scenarioId,
+      @PathVariable @NotBlank final String injectId) {
+    return scenarioInjectService.findInjectForScenario(scenarioId, injectId);
+  }
+
+  // -- CREATE --
 
   @PostMapping(SCENARIO_URI + "/{scenarioId}/injects")
   @RBAC(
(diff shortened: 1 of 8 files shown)


openaev-api/src/main/java/io/openaev/rest/inject/ScenarioInjectApi.java
```diff
@@ -79,6 +80,30 @@ public Iterable<InjectOutput> scenarioInjectsSimple(
...
+    return this.injectRepository.findByScenarioId(scenarioId).stream()
+        .sorted(Inject.executionComparator)
+        .toList();
+  }
+
+  @GetMapping(SCENARIO_URI + "/{scenarioId}/injects/{injectId}")
+  @RBAC(
+      resourceId = "#scenarioId",
+      actionPerformed = Action.READ,
+      resourceType = ResourceType.SCENARIO)
```
antoinemzs: We cover the gap effectively with gating the inject through the scenario, but what do you think about setting the RBAC on the actual resource we try to obtain, i.e. the inject itself?
```suggestion
  @RBAC(
      resourceId = "#injectId",
      actionPerformed = Action.READ,
      resourceType = ResourceType.INJECT)
```

openaev-api/src/main/java/io/openaev/rest/inject/ScenarioInjectApi.java
antoinemzs: Same thought about RBAC, on injectId rather than scenarioId?

antoinemzs: Same thought about RBAC, on injectId rather than scenarioId?

openaev-api/src/main/java/io/openaev/rest/inject/SimulationInjectApi.java
```diff
@@ -168,10 +165,7 @@ public Inject exerciseInject(@PathVariable String exerciseId, @PathVariable Stri
       resourceType = ResourceType.SIMULATION)
```
antoinemzs: Same thought about RBAC, on injectId rather than exerciseId?

openaev-api/src/main/java/io/openaev/rest/inject/SimulationInjectApi.java
```diff
@@ -158,7 +155,7 @@ public List<InjectResultOutput> exerciseInjectsResults(@PathVariable final Strin
       actionPerformed = Action.READ,
       resourceType = ResourceType.SIMULATION)
```
antoinemzs: Same thought about RBAC, on injectId rather than exerciseId?

openaev-api/src/main/java/io/openaev/rest/inject/SimulationInjectApi.java
```diff
@@ -181,13 +175,11 @@ public Iterable<Team> exerciseInjectTeams(
...
```
antoinemzs: Same thought about RBAC, on injectId rather than exerciseId?

openaev-api/src/main/java/io/openaev/rest/inject/SimulationInjectApi.java
antoinemzs: Same thought about RBAC, on injectId rather than scenarioId?

I note this is about writing, but privileges are transitive so WRITE on INJECT should validate the use case.

antoinemzs: Same thought about RBAC, on injectId rather than exerciseId?

openaev-api/src/main/java/io/openaev/rest/inject/SimulationInjectApi.java
```diff
@@ -306,10 +290,7 @@ public Inject updateInjectActivationForExercise(
...
```
antoinemzs: Same thought about RBAC, on injectId rather than exerciseId?

openaev-api/src/main/java/io/openaev/rest/inject/SimulationInjectApi.java
antoinemzs: Same thought about RBAC, on injectId rather than exerciseId?

antoinemzs: Same thought about RBAC, on injectId rather than exerciseId?

openaev-api/src/main/java/io/openaev/rest/inject/SimulationInjectApi.java
```diff
@@ -334,9 +315,18 @@ public Inject updateInjectTeams(
...
+  }
+
+  // -- DELETE --
+
+  @Transactional(rollbackFor = Exception.class)
+  @DeleteMapping(EXERCISE_URI + "/{exerciseId}/injects/{injectId}")
+  @RBAC(
+      resourceId = "#exerciseId",
+      actionPerformed = Action.WRITE,
+      resourceType = ResourceType.SIMULATION)
```
antoinemzs: Same thought about RBAC, on injectId rather than exerciseId?

openaev-api/src/main/java/io/openaev/rest/inject/service/ScenarioInjectService.java
antoinemzs, line 1: There should be better code coverage on this, if we don't improve the RBAC declarations, the whole security model improvement relies on the service methods correctly enforcing scenario or simulation gatekeeping.

The public service methods should be covered in both the happy day case (inject belongs to specified parent) and the opposite error case.

antoinemzs (changes requested):
Sorry about the load of similar comments, I wanted to flag all the occurrences, but they all relate to the same thought: I think we should consider narrowing the RBAC declarations on the actual resource we try to affect (i.e. the inject itself) in the changed endpoint methods; if we had in the first place, I think the issue would not have existed.

Other point relates to test coverage: new public service methods should be well covered to prove the parenting check enforcement.

Thanks!

RomuDeuxfois:
> Sorry about the load of similar comments, I wanted to flag all the occurrences, but they all relate to the same thought: I think we should consider narrowing the RBAC declarations on the actual resource we try to affect (i.e. the inject itself) in the changed endpoint methods; if we had in the first place, I think the issue would not have existed.
> 
> Other point relates to test coverage: new public service methods should be well covered to prove the parenting check enforcement.
> 
> Thanks!

From my point of view, we are not modifying an inject directly, but rather a scenario sub-resource. If a user has write permission on a scenario, then they should also have write permission on its injects, teams, challenges, and so on.

But, by reviewing the code, I came across `PermissionService.resolveTarget` and `PermissionService.RESOURCES_USING_PARENT_PERMISSION`. If we set the permission to `Inject/Write`, it gets resolved back to `Scenario/Write`, which brings us back to the RBAC behavior you propose.

So I can update all the RBAC declarations, but I wanted to highlight that the underlying logic is enforced in `PermissionService`, which intentionally relies on the parent resource for sub-resource permissions.

antoinemzs:
> But, by reviewing the code, I came across `PermissionService.resolveTarget` and `PermissionService.RESOURCES_USING_PARENT_PERMISSION`. If we set the permission to `Inject/Write`, it gets resolved back to `Scenario/Write`, which brings us back to the RBAC behavior you propose.

Ah yes I was referring to this and somehow assumed this was obvious (it wasn't). Yes the permissions on some resources are delegated to their logical parent, which is why I suggested we changed the declarations to benefit from the narrower permission scope.

RomuDeuxfois:
@antoinemzs ready for a new round

antoinemzs (approved):
thanks for fixing all the things, lgtm

PR #5528, opened 2026-04-20
[backend] fix(account): make password reset asynchronous (#0000)

antoinemzs:
### Proposed changes

* Async password reset request

### Testing Instructions

1. Use the reset password feature

### DIFF
--- a/openaev-api/src/test/java/io/openaev/rest/UserApiTest.java
+++ b/openaev-api/src/test/java/io/openaev/rest/UserApiTest.java
@@ -36,8 +36,10 @@ import io.openaev.utils.fixtures.composers.UserComposer;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import net.javacrumbs.jsonunit.core.Option;
+import org.awaitility.Awaitility;
 import org.junit.jupiter.api.*;
 import org.mockito.ArgumentCaptor;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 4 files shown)


openaev-api/src/test/java/io/openaev/rest/UserApiTest.java
```diff
@@ -297,7 +299,18 @@ void resetPassword() throws Exception {
 
       // -- ASSERT --
       ArgumentCaptor<List<User>> userCaptor = ArgumentCaptor.forClass(List.class);
-      verify(mailingService).sendEmail(anyString(), anyString(), userCaptor.capture());
+      // not ideal, but the actual reset happens in a background thread!
```
RomuDeuxfois, line 302: **nitpick**: We can remove @Async from integration test

@Configuration
@Profile("test")
public class SyncAsyncConfig implements AsyncConfigurer {
    @Override
    public Executor getAsyncExecutor() {
        // Execute @Async asynchronously
        return new SyncTaskExecutor();
    }
}

antoinemzs: I have a feeling this would create some surprises in testing if we magically change the code's behaviour globally.

We have adopted `Awaitility` as a support library for testing async behaviour (introduced by @Dimfacion and @johanah29  in InjectApiTest to cover the async processing of execution traces), as a step-up from blind "thread sleeps".

I'm confident the tests are robust as written: the Awaitility lib actually does not cause a delay because it unblocks as soon as the async condition clears.

RomuDeuxfois (approved):

RomuDeuxfois:
Test 🆗 
The workflow can be improved, but this it not the goal of this PR :)

PR #5531, opened 2026-04-21
[backend] chore(log): use the correct logging lib everywhere (#0000)

antoinemzs:
### Proposed changes

* Replace instances of lombok's `@Log` with chosen standard `@Slf4j`

RE: low test coverage, I'm only swapping the libs here, so that just means the original code region isn't well covered.


RomuDeuxfois:
@antoinemzs can you add a spotless gate for this ?

antoinemzs:
> @antoinemzs can you add a spotless gate for this ?

I don't think this is possible with spotless via maven, there are no custom rules extension points that I could find in the docs. https://github.com/diffplug/spotless/tree/main/plugin-maven

RomuDeuxfois:
> > @antoinemzs can you add a spotless gate for this ?
> 
> I don't think this is possible with spotless via maven, there are no custom rules extension points that I could find in the docs. https://github.com/diffplug/spotless/tree/main/plugin-maven

You are right, we need to use checkstyles or ArchUnit for this.

RomuDeuxfois (approved):

PR #5532, opened 2026-04-21
feat(multitenancy): transfer older rabbitmq messages to new queues (#3505)

### LINKED ISSUE #4864: Implement core multi tenancy

Dimfacion:
### Proposed changes

With the implementation of tenants we now create rabbitmq queues dedicated to each tenant with a name that includes the tenant id. This code is to transfert the messages that were in the old queues to their corresponding new queues. 

### Testing Instructions

1. Go to your local rabbitmq administration console (localhost:15672)
2. Create a queue with the old naming convention (for exemple openaev_execution_workflows-ready) and publish a message in that queue
3. Clean your database (there must be no tenant,  since the messages are transferred to the queues of the default tenant)
4. Launch the api and check if the messages are now in the corresponding queue (in this case openaev-2cffad3a-0001-4078-b0e2-ef74274022c3_execution_workflows-ready)


openaev-api/src/test/java/io/openaev/datapack/MigrationProcessorTest.java
```diff
@@ -25,7 +27,8 @@ public class DataPackProcessorTest extends IntegrationTest {
   @Test
   @DisplayName("Processor processes all known datapacks")
   public void processorProcessesAllKnownDatapacks() {
```
RomuDeuxfois, line 29: **todo**: Should be renamed to MigrationProcessorTest

openaev-model/src/main/java/io/openaev/service/RabbitmqService.java
```diff
@@ -327,6 +336,166 @@ public void checkHealth() throws IOException, TimeoutException {
...
+   * @param queueName the full queue name (already prefixed)
+   * @return a list of message bodies
+   */
+  public List<byte[]> drainQueue(String queueName) throws IOException, TimeoutException {
+    List<byte[]> messages = new java.util.ArrayList<>();
+    try (Connection connection = connectionFactory.newConnection();
+        Channel channel = connection.createChannel()) {
+      // Passive declare to check if the queue exists; throws IOException if not
+      try {
+        channel.queueDeclarePassive(queueName);
```
RomuDeuxfois: **todo**: check queue exists before drain

openaev-model/src/main/java/io/openaev/service/RabbitmqService.java
```diff
@@ -327,6 +336,166 @@ public void checkHealth() throws IOException, TimeoutException {
...
+      channel.exchangeDelete(exchangeName);
+      log.info("Deleted exchange '{}'.", exchangeName);
+    } catch (Exception e) {
+      log.debug("Could not delete exchange '{}' (may not exist): {}", exchangeName, e.getMessage());
+    }
+  }
+
+  // -- MANAGEMENT API --
+
+  private static final Pattern NAME_PATTERN = Pattern.compile("\"name\"\\s*:\\s*\"([^\"]+)\"");
```
RomuDeuxfois: **todo**: this regex can grab something with a property name not relativ to RabbitMQ names

Use something stronger instead

ObjectMapper mapper = new ObjectMapper();
JsonNode root = mapper.readTree(response.body());
for (JsonNode node : root) {
  String name = node.path("name").asText("");
  if (name.startsWith(prefix)) {
    result.add(name);
  }
}

openaev-model/src/main/java/io/openaev/service/RabbitmqService.java
```diff
@@ -327,6 +336,166 @@ public void checkHealth() throws IOException, TimeoutException {
...
+      String credentials = rabbitmqConfig.getUser() + ":" + rabbitmqConfig.getPass();
+      String authHeader =
+          "Basic "
+              + Base64.getEncoder().encodeToString(credentials.getBytes(StandardCharsets.UTF_8));
+
+      HttpRequest request =
+          HttpRequest.newBuilder().uri(uri).header("Authorization", authHeader).GET().build();
+
+      HttpResponse<String> response =
+          HttpClient.newHttpClient().send(request, HttpResponse.BodyHandlers.ofString());
```
RomuDeuxfois: **todo**: create one HttpClient instance  instead of recreate one each time

openaev-api/src/main/java/io/openaev/processor/core/JavaMigration.java
```diff
@@ -0,0 +1,44 @@
...
+
+import io.openaev.database.model.Tenant;
+import io.openaev.processor.MigrationProcessingResult;
+import io.openaev.service.DataPackService;
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.transaction.annotation.Transactional;
+
+@Slf4j
+public abstract class JavaMigration {
```
RomuDeuxfois: **todo**: Rename JavaMigration, as the name is misleading. Our first migration is not a Java migration, but a RabbitMQ migration.

Dimfacion: The idea is to have an abstract class to make migration that necessitate java code the same way we have the data processors working. Given that, is that still confusing ? Maybe BaseMigration would be better ?

RomuDeuxfois: RuntimeMigration maybe ? To be separate from FlywayMigration

openaev-model/src/main/java/io/openaev/service/RabbitmqService.java
Dimfacion: That's actually what we're doing here with queueDeclarePassive

openaev-model/src/main/java/io/openaev/service/RabbitmqService.java
```diff
@@ -327,6 +337,229 @@ public void checkHealth() throws IOException, TimeoutException {
...
+        Channel channel = connection.createChannel()) {
+      for (byte[] body : messages) {
+        channel.basicPublish(exchange, routingKey, null, body);
+      }
+      log.info(
+          "Published {} messages to exchange '{}' with routing key '{}'.",
+          messages.size(),
+          exchange,
+          routingKey);
+    }
```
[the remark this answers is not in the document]
Dimfacion: There shouldn't be any messages or only old ones so that's a risk we can accept

RomuDeuxfois (approved):

openaev-model/src/main/java/io/openaev/service/RabbitmqService.java
```diff
@@ -327,6 +337,229 @@ public void checkHealth() throws IOException, TimeoutException {
...
+      channel.exchangeDelete(exchangeName);
+      log.info("Deleted exchange '{}'.", exchangeName);
+    } catch (Exception e) {
+      log.debug("Could not delete exchange '{}' (may not exist): {}", exchangeName, e.getMessage());
+    }
+  }
+
+  // -- MANAGEMENT API --
+
+  private static final ObjectMapper MAPPER = new ObjectMapper();
```
RomuDeuxfois: **nitpick**: All the management API part should go to a dedicated file -> RabbitmqManagementClient
This will extracts a single responsibility: ask the broker via HTTP call.

openaev-model/src/main/java/io/openaev/service/RabbitmqService.java
```diff
@@ -327,6 +337,229 @@ public void checkHealth() throws IOException, TimeoutException {
...
+                  + encodedVhost
+                  + "?columns=name");
+
+      String credentials = rabbitmqConfig.getUser() + ":" + rabbitmqConfig.getPass();
+      String authHeader =
+          "Basic "
+              + Base64.getEncoder().encodeToString(credentials.getBytes(StandardCharsets.UTF_8));
+
+      HttpRequest request =
+          HttpRequest.newBuilder().uri(uri).header("Authorization", authHeader).GET().build();
```
RomuDeuxfois: **nitpick**: We could have a config class for this. This is the standard Spring pattern for infrastructure clients: declare the bean in config, inject it everywhere you need it. It keeps the timeout and connection settings in one place rather than hardcoded in the service.

package io.openaev.config;

import java.net.http.HttpClient;
import java.time.Duration;
import org.springframework.context.annotation.Bean;
import org.springframework.context.annotation.Configuration;

/** Produces the {@link HttpClient} bean used by {@link io.openaev.service.rabbitmq.RabbitmqManagementClient}. */
@Configuration
public class RabbitmqManagementClientConfig {

  @Bean
  public HttpClient rabbitmqManagementHttpClient() {
    return HttpClient.newBuilder()
        .connectTimeout(Duration.ofSeconds(30))
        .build();
  }
}

RomuDeuxfois:
Tests 🆗

PR #5545, opened 2026-04-21
[frontend] feat(multi-tenancy): add tenant identifier on settings (#4864)

RomuDeuxfois:
### Proposed changes

* add tenant identifier on settings


openaev-front/src/admin/components/settings/PlatformInfoPanel.tsx
```diff
@@ -8,10 +8,11 @@ import type { PlatformSettings } from '../../../utils/api-types';
 
 interface PlatformInfoPanelProps {
   settings: PlatformSettings;
-  children?: ReactNode;
+  header?: ReactNode;
```
corinnekrych: Suggestion: using the names `header` and `footer` here is not really a great fit since they imply a strict structural relationship to the whole panel (ie: a page has a hear with title and a footer), whereas in this case they are more like content slots at the top and bottom of a list.
```
<List>
  {topContent}
  {/* ...list items... */}
  {bottomContent}
</List>
```

corinnekrych:
Tested and display content properly

openaev-front/src/admin/components/settings/PlatformInfoPanel.tsx
RomuDeuxfois: I made the changes :)
Thank you

corinnekrych (approved):

PR #5555, opened 2026-04-22
[backend/frontend] fix(settings): split settings into public & private api

RomuDeuxfois:
### Proposed changes

* split settings into public & private api

Public API (not authenticated)
{EA72197E-E9C8-4109-9452-8960CCF3BC54}

Private API (authenticated)
Current API

### Related issues

* Closes [#ISSUE-NUMBER](https://github.com/OpenAEV-Platform/filigran-private/issues/84)

### DIFF
--- a/openaev-api/src/test/java/io/openaev/rest/PlatformSettingsApiTest.java
+++ b/openaev-api/src/test/java/io/openaev/rest/PlatformSettingsApiTest.java
@@ -30,11 +30,12 @@ class PlatformSettingsApiTest extends IntegrationTest {
           "platform_light_theme",
           "platform_dark_theme",
           "enabled_dev_features",
-          "platform_whitemark",
-          "platform_license");
+          "platform_whitemark");
 
   private static final List<String> PRIVATE_FIELDS =
       List.of(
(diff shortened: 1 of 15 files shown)


openaev-api/src/test/java/io/openaev/rest/PlatformSettingsApiTest.java
antoinemzs, line 1: nitpick: testing standard asks that we put `@DisplayName` on test classes (including nested ones) and test methods

RomuDeuxfois: I have the feeling to repeat myself with @DisplayName.. I already tell what I want to achieve with the given_that_should_that pattern.

antoinemzs: I'm just quoting the standard. It's true that it's prettier but I have a feeling we had this rule to compensate for tersely worded function names.

antoinemzs (approved):
Ok with the discussed caveats

PR #5569, opened 2026-04-23
[frontend/backend] fix(healthcheck): placeholder inject must remain disabled (#5424)

### LINKED ISSUE #5424: false "missing content" warning displayed on scenario inject page despite no missing content
## Description

On the **scenario inject page**, a "missing content" warning is incorrectly displayed even when no content is actually missing. The warning disappears and the inject status is correctly set to **Enabled** only after the user opens the update drawer and closes it without making any changes.

## Environment

1. OS  Mac OS 
2. OpenAEV version: pre-release

## Reproducible Steps

1. Open a scenario and navigate to the **Injects** page
2. Observe that a "missing content" warning is displayed on one or more injects, even though all required content is present
3. Open the **update drawer** for the affected inject (click the edit/update action)
4. Close the drawer **without making any changes**
5. Observe that the "missing content" warning disappears and the inject status is now correctly shown as **Enabled**

## Expected Output

The inject should display the correct status (**Enabled**) immediately on page load, without requiring the user to open and close the update drawer. No "missing content" warning should be shown when all required content is present.

## Actual Output

A false "missing content" warning is shown on the inject page even though no content is missing. The status is only corrected to **Enabled** after the update drawer is opened and closed (with no changes made)

## Additional information

This appears to be a frontend state/rendering issue where the inject's content validation is not correctly evaluated on initial page load. The drawer update action likely triggers a re-evaluation or re-fetch that resolves the incorrect state

## Screenshots 

https://github.com/user-attachments/assets/ac157c48-2522-4862-af21-60218d40b121

MarineLeM:
### Proposed changes

-  Injects called "placeholder" remain disabled and cannot have a "missing content" status
- Disabled injects should not be editable
- The scenario health check should not take into account the missing content of disabled injects

### Testing Instructions

1. Create a placeholder inject :
    - Create a scenario
    - Generate injects using the scenario assistant by selecting a TTP that has no injector contract linked to it
    - The scenario assistant could not match any injector contract to the selected TTP, so it will create a placeholder inject

2. Check the status of the placeholder inject — it should be disabled
3. Check that this inject is not editable
4. Verify the health check behavior — if the scenario contains only this placeholder inject, the health check information should not appear on the overview tab

### Related issues

* Closes #5424

### DIFF
--- a/openaev-api/src/main/java/io/openaev/healthcheck/utils/HealthCheckUtils.java
+++ b/openaev-api/src/main/java/io/openaev/healthcheck/utils/HealthCheckUtils.java
@@ -166,7 +166,9 @@ public class HealthCheckUtils {
   public List<HealthCheck> runMissingContentChecks(@NotNull final Scenario scenario) {
     List<HealthCheck> result = new ArrayList<>();
     boolean atLeastOneInjectIsNotReady =
-        scenario.getInjects().stream().anyMatch(inject -> !inject.isReady());
+        scenario.getInjects().stream()
+            .filter(Inject::isEnabled)
+            .anyMatch(inject -> !inject.isReady());
 
     if (atLeastOneInjectIsNotReady) {
       result.add(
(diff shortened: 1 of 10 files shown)


GaetanSantucci (approved):
test ok, code ok

PR #5584, opened 2026-04-23
[backend] fix(indexing): parameterable indexing size (#1)

Dimfacion:
### Proposed changes

* Make indexing batches size parametrable
*

### Related issues

* 
*

### Checklist

### DIFF
--- a/openaev-api/src/main/resources/application.properties
+++ b/openaev-api/src/main/resources/application.properties
@@ -238,6 +238,53 @@ logging.logback.rollingpolicy.max-history=7
 # EXECUTORS #
 #############
 
+# Executor Caldera
+executor.caldera.enable=false
+executor.caldera.id=2f9a0936-c327-4e95-b406-d161d32a2500
+executor.caldera.url=<caldera-url>
+executor.caldera.public-url=<caldera-public-url>
+executor.caldera.api-key=<caldera-api-key>
+
+# Executor Tanium
+executor.tanium.enable=false
+executor.tanium.id=fbcce3cc-9cd3-4fdb-b451-1c90f4ae035d
+executor.tanium.url=<tanium-url>
+executor.tanium.api-batch-execution-action-pagination=100
+executor.tanium.api-register-interval=1200
+executor.tanium.clean-implant-interval=8
+executor.tanium.api-key=<tanium-api-key>
+executor.tanium.computer-group-id=1
+executor.tanium.action-group-id=4
+executor.tanium.windows-package-id=
+executor.tanium.unix-package-id=
+
+# Executor CrowdStrike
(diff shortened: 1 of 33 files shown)


laugiov:
Thanks for turning this around so quickly after yesterday's incident. Clean refactor, and the switch from string concat to a named `:limit` param is a nice bonus for the PG prepared-statement cache.

A few small things below, none blocking, but worth a look. And just flagging for context : the heap dump from yesterday also pointed to a second path (`EndpointApi.getEndpointJobs`, ~30 MB retained through Hibernate lazy-collection init) that this PR doesn't cover. 
It si not a problem for this PR, just worth keeping visible so it doesn't get forgotten.

laugiov:
Any plan to add a small test for the parameterable behavior? 
Something like `findForIndexing(from, 10)` returning at most 10 rows would be enough to lock the contract down.

antoinemzs:
>And just flagging for context : the heap dump from yesterday also pointed to a second path (EndpointApi.getEndpointJobs, ~30 MB retained through Hibernate lazy-collection init) that this PR doesn't cover.

We have found a problem that would keep accruing jobs (= commands for the openaev agent to run) and cause ever increasing data volumes and delays when calling this route, linked to OpenAEV Agent. This will be addressed with overhauling the way agents deal with their auto updates (by not creating jobs at all for this purpose), and also change the endpoint's behaviour to consume the jobs polled from the endpoint rather than wait for an explicit delete (which causes even more traffic).

THis should all be referenced in upcoming work led by @damgouj

openaev-api/src/main/resources/application.properties
```diff
@@ -428,4 +428,7 @@ management.server.port=8080
 #############################
 # FEATURE UNDER DEVELOPMENT #
 #############################
-openaev.enabled-dev-features=
\ No newline at end of file
+openaev.enabled-dev-features=
+
+# Maximum number of records fetched per indexing batch from the database
+engine.indexing-batch-size=500
```
antoinemzs: if there is a default in code, why do we need a value in the properties file? Given it is loaded by default in the docker images, it could lead to surprises if we forget to update this place but have changed the in-code default.
wdyt?

Dimfacion: I think it's nice to have it in the properties to remember how it's spelled but yeah, I can set it as commented by default if that's fine with you ?

antoinemzs: seems reasonable, so yes please comment the value out

Dimfacion: Done

antoinemzs (approved):
looks good, thanks!

PR #5589, opened 2026-04-24
[frontend] feat(multi-tenant): strip detail segments on tenant switch (#4864)

corinnekrych:
… (#4864)

When a user views a detail page (e.g. `{tenant-id}/admin/scenarios/{scenario-id}`) and switches to another tenant, the page reloads keeping the entity ID (ie: {scenario-id}) in the URL.
The frontend URL becomes stale, showing an entity ID that doesn't exist in the new tenant. This results in a 404 error (once BE is fixed) and a confusing user experience.

### Proposed changes

* When switching tenant, strip the details part of the URL to make sure we land on the list of entity rather than the entity detail page

### Testing Instructions

1. log in as admin
2. create tenant A
3. create tenant B
4. in tenant A create sceanrio A
5. switch to tenant B in tenant switcher
and see how we lan in list of scanearios

### Related issues

* Closes #ISSUE-NUMBER
*

### Checklist

### Further comments


openaev-front/src/utils/url-helper.ts
```diff
@@ -13,7 +13,25 @@ export const TENANT_URI = '/api/tenants';
...
+ * lands on the parent list page rather than a detail page for a resource
+ * that may not exist in the target tenant.
+ *
+ * e.g. "/admin/scenarios/123e4567-e89b-12d3-a456-426614174000"
+ *        → "/admin/scenarios"
+ *      "/admin/scenarios/123e4567-e89b-12d3-a456-426614174000/injects"
+ *        → "/admin/scenarios"
+ *      "/admin/scenarios" → "/admin/scenarios" (unchanged)
+ */
+export const stripDetailSegments = (pathname: string): string => {
```
RomuDeuxfois, line 43: It seems a bit fragile to me. For example, if you’re on a list page and then switch tenant, but you don’t have access to that list in the new tenant, you’ll get an error as well.

Could we do something similar to what we do after login at first ?

corinnekrych: You mean loosing context and forward to home page? 
The goal here was to allow end user to switch between tenant keeping context. For ex if a tenant user browse  teams in tenantA and wants to check tenantB's teams.
In case the user has no read access to scenario, he will fallback to existing RBAC alert.

RomuDeuxfois (approved):

corinnekrych:

Issue #5039, opened 2026-03-02
[Chaining] Scope Definition


mergenhan:
As an OpenAEV user,
I want to define my scenario scope for my Chaining Scenario,
so that the scenario runs only within the scope that I configured.

PR #5595, opened 2026-04-24
[backend/frontend] feat(chaining): add variables to scope definition (#5039)

savacano28:
### Proposed changes

* Add variables to scope definition
* Update front: Scope definition

### Testing Instructions

1. Go to simulations
2. Create a chaining simulation
3. Add some variables into scope definition
4. Delete/recreate these variables (same behavoir from scenario/simualtion variables)
5. Change of tab and come back to scope definition, check variables

image
image
image

### Related issues

* Related #5039 

### Checklist

### Further comments
This pr is only fro CRUD Scope variables.  These variables will be after used in scope definition where we will use them to update the payloads.


openaev-api/src/main/java/io/openaev/api/chaining/WorkflowConfigurationMapper.java
```diff
@@ -44,4 +47,21 @@ private static WorkflowScopeRuleOutput toScopeRuleOutput(WorkflowScopeRule rule)
...
+
+  private static List<ScopeVariableOutput> toScopeVariableOutputList(
+      List<ScopeVariable> variables) {
+    if (variables == null || variables.isEmpty()) {
+      return List.of();
+    }
+    return variables.stream().map(WorkflowConfigurationMapper::toScopeVariableOutput).toList();
+  }
+
+  private static ScopeVariableOutput toScopeVariableOutput(ScopeVariable variable) {
```
heditar, line 59: Please use builder in dto

openaev-api/src/main/java/io/openaev/service/chaining/WorkflowService.java
```diff
@@ -257,6 +264,93 @@ private void copyScopeRules(Workflow source, Workflow target) {
...
+  }
+
+  private void updateScopeVariable(ScopeVariable existing, ScopeVariableInput input) {
+    existing.setKey(input.key());
+    existing.setType(input.type());
+    existing.setValue(input.value());
+    existing.setDescription(input.description());
+  }
+
+  private ScopeVariable buildScopeVariable(ScopeVariableInput input, Workflow workflow) {
```
heditar, line 344: Use toBuilder if possible

openaev-api/src/main/java/io/openaev/service/chaining/WorkflowService.java
```diff
@@ -257,6 +264,93 @@ private void copyScopeRules(Workflow source, Workflow target) {
...
+                .toList());
+  }
+
+  /**
+   * Reconciles the workflow's scope-variable collection against the provided inputs: removes
+   * variables not present in the input, adds new ones, and updates changed ones in-place.
+   *
+   * @return {@code true} if the collection was modified
+   */
+  private boolean applyScopeVariables(List<ScopeVariableInput> variableInputs, Workflow workflow) {
```
heditar, line 293: Why do we need to considerante the existing variables ?

heditar:
should we have format validation ? otherwise testing ok

openaev-api/src/main/java/io/openaev/service/chaining/WorkflowService.java
savacano28: We check if we dont have modify the existing values in that case we dont save them, but it is true that in the current UI we can not modify the variables, we can add or deleted.. but we could do it calling direclty to endpoint.

savacano28:
> should we have format validation ? otherwise testing ok

Currently, we don’t have a defined format for input arguments. What we’ll do here is fetch the existing argument payload and map it to predefined values in variables, matching the types and names of the fields/keys.

savacano28:
Hi @heditar, I’ve made the fixes. Thanks for the review! 🙂

openaev-front/src/admin/components/chaining/ScopeVariableCreateDialog.tsx
```diff
@@ -0,0 +1,156 @@
...
+      setError('scope_variable_key', {
+        type: 'required',
+        message: t('Key is required'),
+      });
+      return;
+    }
+    if (!type) {
+      setError('scope_variable_type', {
+        type: 'required',
+        message: t('Type is required'),
```
camrrx: Better to use zod for the validation form

camrrx (approved):

PR #5613, opened 2026-04-27
[backend] fix(injcts): fix delete inject statuses on simulation reset (#5608)

### LINKED ISSUE #5608: [Wis] Simulation reset does not update inject state (reset ineffective)
Description:
When resetting a simulation, the state of injects does not change, even though the reset action completes successfully.
The issue has been reproduced internally in a demo environment, indicating a confirmed product bug

Steps to Reproduce:
Run a simulation with active injects
Reset the simulation
Observe inject state before and after reset

Actual Behavior:
Inject states remain unchanged after reset
No difference between pre‑simulation and post‑reset state

Expected Behavior:
Reset should restore injects to their initial state
Inject lifecycle/state should be consistent after reset

Additional Context:
Customer reported multiple simulations with the same name, causing confusion during investigation
Correct simulation export was requested to avoid ambiguity

Impact:
Impacts exercise preparation and re‑runs
No workaround identified via reset

antoinemzs:
### Proposed changes

* Actually delete the inject statuses when simulation reset

### Testing Instructions

1. Start a simulation and let some injects complete (canceled state)
2. reset the simulation
3. Inject statuses are all deleted (back to draft status)
4. repeat with finished simulation

### Related issues

* Closes #5608 

### Checklist

### Further comments

### DIFF
--- a/openaev-api/src/main/java/io/openaev/rest/inject/service/InjectService.java
+++ b/openaev-api/src/main/java/io/openaev/rest/inject/service/InjectService.java
@@ -75,6 +75,8 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.Triple;
 import org.hibernate.Hibernate;
 import org.jetbrains.annotations.NotNull;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Lazy;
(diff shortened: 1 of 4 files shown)


camrrx (approved):
Test okk

PR #5622, opened 2026-04-28
[frontend] feat(chainig): add csv and manual input for scope rules (#5039)

savacano28:
### Proposed changes

* Add csv and manual input for scope rule in chaining scope definition

### Testing Instructions

1. Create chaining simualtion/scneario
2. Open drawer to add scope rules
3. Download csv file template
4. Upload csv file with some ips, subnest, domains
5. Save
6. Open again, update, delete

image
image
image

### Related issues

* Related #5039 

### Checklist

### Further comments


openaev-front/src/admin/components/chaining/scope-rules-csv.ts
```diff
@@ -0,0 +1,193 @@
...
+  if (!ip || !rawMask || !isIp(ip.trim())) {
+    return false;
+  }
+
+  const mask = Number(rawMask);
+  if (!Number.isInteger(mask)) {
+    return false;
+  }
+
+  const isIpv6 = ip.includes(':');
```
camrrx: the variable and the function have the same name, might be safer to call the local variable isIpv6Address for example?

openaev-front/src/admin/components/chaining/scope-rules-csv.ts
```diff
@@ -0,0 +1,193 @@
...
+export interface ScopeCsvParseResult {
+  valid: ScopeCsvRule[];
+  invalid: ScopeCsvInvalidRow[];
+}
+
+const IPV4_PART = String.raw`(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)`;
+const IPV4_REGEX = new RegExp(String.raw`^${IPV4_PART}(\.${IPV4_PART}){3}$`);
+const DOMAIN_REGEX = /^(?=.{1,253}$)(?!-)(?:[a-zA-Z0-9-]{1,63}\.)+[a-zA-Z]{2,63}$/;
+
+const normalizeType = (raw: string): ScopeCsvType | null => {
```
camrrx: normalizeType could be refactored using a map to make it easier :

const TYPE_MAP = {
  domain: 'DOMAIN',
  ip: 'IP',
  ipsubnet: 'IP_SUBNET',
  'ip_subnet': 'IP_SUBNET',
  'ip-subnet': 'IP_SUBNET',
};

openaev-front/src/admin/components/chaining/ScopeInventoryBox.tsx
```diff
@@ -0,0 +1,206 @@
...
+        alignItems: 'center',
+        justifyContent: 'space-between',
+      }}
+      >
+        <Typography variant="h4">
+          {`${listLabel} ${t('inventory')} (${totalSelected})`}
+        </Typography>
+        <div style={{
+          display: 'flex',
+          gap: theme.spacing(1),
```
camrrx, line 93: You can add some margin so the buttons aren’t stuck to the box :
           marginBottom: theme.spacing(1),

openaev-front/src/admin/components/chaining/scope-rules-csv.ts
```diff
@@ -0,0 +1,193 @@
...
+    return false;
+  }
+
+  return segments.every(segment => /^[0-9a-fA-F]{1,4}$/.test(segment));
+};
+
+const isIp = (value: string) => IPV4_REGEX.test(value) || isIpv6(value);
+
+const isIpSubnet = (value: string) => {
+  const [ip, rawMask] = value.split('/');
```
impolitepanda: This isn't a sufficient format check. An input like 10.0.0.0/24 is valid, as it should, but 10.0.0.0/24/32 would also pass, with the /32 being ignored. I believe we should trigger an error instead. Adding this check value.split('/').length === 2 would help guard against bad format like this

impolitepanda (changes requested):
I believe we've started adding tests in the front end as well ? If so, I didn't see any :)

openaev-front/src/admin/components/chaining/scope-rules-csv.ts
```diff
@@ -0,0 +1,193 @@
...
+  if (value === 'ip') {
+    return 'IP';
+  }
+  if (value === 'ipsubnet' || value === 'ip_subnet' || value === 'ip-subnet') {
+    return 'IP_SUBNET';
+  }
+  return null;
+};
+
+const isIpv6 = (value: string) => {
```
impolitepanda: I don't think this function covers ipv6 specifications. Maybe using a dedicated and well reviewed external library could be a good idea?

impolitepanda: For instance, there is a zone or area suffix, like %eth0 that could count. Also, an ipv6 can contain an ipv4, like ::ffff:192.168.0.1 -> this would fail

openaev-front/src/admin/components/chaining/ScopeInventoryBox.tsx
```diff
@@ -0,0 +1,206 @@
...
+        )}
+        {chips.map(chip => (
+          <Chip key={chip.key} label={chip.label} size="small" onDelete={chip.onDelete} />
+        ))}
+        <input
+          ref={inputRef}
+          value={inputValue}
+          onChange={e => setInputValue(e.target.value)}
+          onKeyDown={handleKeyDown}
+          placeholder={t('Type IPs, CIDRs or hostnames...')}
```
camrrx: If no asset are selected, it's weird to have the typography "no asset" and the placeholder "type ips..." so I propose to do this : 
`placeholder={chips.length === 0 && inputValue.length === 0 ? t('No asset selected. Add asset manually by typing IPs, CIDRs or hostnames or select some in the asset list.') : t('Type IPs, CIDRs or hostnames...')}`
and to delete the 
`<Typography variant="body2" sx={{ color: 'text.disabled' }}>
            {t('No asset selected. Add asset manually or select some in the asset list.')}
          </Typography>`

openaev-front/src/admin/components/chaining/ScopeInventoryBox.tsx
```diff
@@ -0,0 +1,206 @@
...
+            alignItems: 'center',
+            gap: theme.spacing(1),
+            borderColor: theme.palette.primary.main,
+          }}
+        >
+          <Typography
+            variant="body2"
+            sx={{
+              color: 'text.secondary',
+              mr: 1,
```
camrrx: you can delete mr: 1

openaev-front/src/admin/components/chaining/ScopeInventoryBox.tsx
```diff
@@ -0,0 +1,206 @@
...
+              whiteSpace: 'nowrap',
+              cursor: 'pointer',
+              textDecoration: 'underline',
+            }}
+            onClick={commitValues}
+          >
+            {t('Tap enter or click here to add:')}
+          </Typography>
+          {parsedValues.map((val, idx) => (
+            // eslint-disable-next-line react/no-array-index-key
```
camrrx: To delete?

openaev-front/src/admin/components/chaining/ScopeInventoryBox.tsx
```diff
@@ -0,0 +1,206 @@
...
+      )}
+
+      <Typography
+        variant="body2"
+        sx={{
+          color: 'text.disabled',
+          display: 'flex',
+          alignItems: 'center',
+          gap: 0.5,
+          mt: 1,
```
camrrx, line 217: theme for gap and mt

openaev-front/src/admin/components/chaining/scope-rules-csv.ts
savacano28: Yes, I will look for library then.

savacano28:
Hi @camrrx and @impolitepanda, I’ve made the fixes. Thanks for the review! 🙂

camrrx (approved):

impolitepanda (approved):

savacano28:
To consider: 
We could extend validation to support internationalized domain names (IDNs). The current regex only matches ASCII characters (a–z), so domains using other scripts are rejected. If we normalize them to their ASCII-compatible representation, for example, 例子.中国 : xn--fsqu00a.xn--fiqs8s => they would then match the existing regex.

PR #5625, opened 2026-04-28
[backend] feat(chaining): add timeout logic to chaining simulations (#4824)

impolitepanda:
### Proposed changes

* Add timeout logic to chaining simulations

### Testing Instructions

1. Create a simulation with a set timeout
2. Run simulation with steps lasting longer than set timeout
3. Check that steps and simulation are set to completed status when timeout is reached and that no other result is received after

### Related issues

* Relates to #4824 

### Checklist


openaev-model/src/main/java/io/openaev/database/repository/WorkflowRepository.java
```diff
@@ -30,5 +31,24 @@ public interface WorkflowRepository extends JpaRepository<Workflow, String> {
...
+  @Query(
+      value =
+          """
+        SELECT * FROM workflows
+        WHERE workflow_status = 'RUN'
+          AND workflow_timeout_enabled = true
+          AND workflow_timeout_seconds IS NOT NULL
+          AND workflow_created_at + (workflow_timeout_seconds || ' seconds')::interval <= now()
+        """,
+      nativeQuery = true)
```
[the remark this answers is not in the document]
impolitepanda: Added a partial index to fix this issue

openaev-api/src/main/java/io/openaev/service/chaining/StepService.java
```diff
@@ -169,6 +169,18 @@ private void startWorkflow(Workflow workflowRun, Workflow workflowTemplate)
   public Optional<Step> ready(Step nextStepTemplateToExecute, Workflow workflowRun, String input)
       throws ChainingException {
 
+    // Guard: ignore if workflow run has already ended (e.g. timeout).
+    // Unlike run() and handleExternalUpdateEvent(), we check the in-memory status here
+    // because ready() is always called within the same transaction that manages the workflow
+    // (startWorkflow, handleExternalUpdateEvent), so the in-memory state is guaranteed to be fresh.
+    if (workflowRun.getStatus() == WorkflowStatus.END) {
```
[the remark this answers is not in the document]
impolitepanda: updated the guard as suggested

antoinemzs:
@impolitepanda the migration numbering has changed to 5_XX, so you might want to check this before a full review

impolitepanda:
> @impolitepanda the migration numbering has changed to 5_XX, so you might want to check this before a full review

Don't know how I missed that, will correct ! thanks

openaev-api/src/main/java/io/openaev/service/chaining/WorkflowTimeoutService.java
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  /**
+   * Forces a workflow run to complete due to timeout expiration. Sets the workflow status to END,
+   * terminates all active steps (READY or RUN), and removes pending delay queue entries.
+   *
+   * @param workflowRun the running workflow to force-complete
+   */
+  @Transactional(rollbackFor = Exception.class)
+  public void forceCompleteWorkflow(Workflow workflowRun) {
```
heditar, line 47: nit: do we need another service just for this method? should it be in workflowservice ?

savacano28:
**Code**
OK

**Tests**
Simulation finishes when timeout is reached: OK
image
image
Simulation finishes correctly when no timeout is set or when the timeout is sufficiently long: OK

**Questions / Observations**
Is it expected that injections have the status “Executed” when the simulation timeout is reached?

My concern is that there may be some ambiguity around the simulation timeout behavior. Currently, we only define a timeout concept at the agent level, but now we also have a simulation-level timeout behavior, which may be inconsistent or unclear.

Should we consider introducing a dedicated status for injections (or for the simulation execution state), such as “Timed out”, “Interrupted”, or “Cancelled”, to more accurately reflect what actually happened?

Additionally, should this status also be propagated to the agents, so that their final state is consistent with the simulation outcome? @heditar @impolitepanda @mergenhan

openaev-api/src/main/java/io/openaev/service/chaining/WorkflowTimeoutService.java
```diff
@@ -0,0 +1,103 @@
...
+   */
+  private int stopActiveInjects(String simulationId) {
+    List<Inject> injects = injectService.findBySimulationId(simulationId);
+    int stoppedCount = 0;
+    for (Inject inject : injects) {
+      if (inject.getStatus().isPresent()
+          && ACTIVE_INJECT_STATUSES.contains(inject.getStatus().get().getName())) {
+        InjectStatus status = inject.getStatus().get();
+        status.addInfoTrace(
+            "Inject stopped due to simulation timeout.", ExecutionTraceAction.COMPLETE);
```
savacano28: To discuss: I’m not sure this is the correct status: “Complete”. Should it be “Blocked”, “Aborted”, or “Interrupted”?

savacano28 (approved):

openaev-api/src/main/java/io/openaev/service/chaining/WorkflowTimeoutService.java
impolitepanda: Discussed with product: we will merge as is this change, and we will create another task/bug during the next week to refactor this and have a more meaningful status

openaev-api/src/main/java/io/openaev/service/chaining/WorkflowTimeoutService.java
impolitepanda: I had to split it because of circular dependencies. Also, since your comment has been added, a lot has been added to this class, namely functions to manage injects and steps timeouts as well. 

As a whole, I feel like having more dedicated and smaller services will be more helpful than giant ones (like ExerciseService, for example) in the future, most importantly to avoid this issue, but also to reduce PR impacts and huge refacto

impolitepanda:
> **Code** OK
> 
> **Tests** Simulation finishes when timeout is reached: OK image image Simulation finishes correctly when no timeout is set or when the timeout is sufficiently long: OK
> 
> **Questions / Observations** Is it expected that injections have the status “Executed” when the simulation timeout is reached?
> 
> My concern is that there may be some ambiguity around the simulation timeout behavior. Currently, we only define a timeout concept at the agent level, but now we also have a simulation-level timeout behavior, which may be inconsistent or unclear.
> 
> Should we consider introducing a dedicated status for injections (or for the simulation execution state), such as “Timed out”, “Interrupted”, or “Cancelled”, to more accurately reflect what actually happened?
> 
> Additionally, should this status also be propagated to the agents, so that their final state is consistent with the simulation outcome? @heditar @impolitepanda @mergenhan

Decision made by Product regarding your question: as a first step, the status will be in error instead of a success. While we all understand that this is not an actual error, it's better, for now, than having it as a success. After MVP, new tasks to manage these statuses will be created as needed.

PR #5626, opened 2026-04-28
[backend/frontend] feat(multi-tenancy): have clear tenant access denied error

RomuDeuxfois:


openaev-front/src/utils/error/ErrorHandler.tsx
```diff
@@ -17,6 +17,8 @@ const ErrorHandler = () => {
         const messages = error?.errors?.children?.message?.errors;
         setEEFeatureDetectedInfo(Array.isArray(messages) ? messages.join(', ') : '');
         openDialog();
+      } else if (error.status === 403 && error.message === 'TENANT_ACCESS_DENIED') {
```
corinnekrych, line 20: for RBAC i think the approach was to thro 404 to not give to many authz information to end user. Do we really want to surface information on tenant membership on end user? We could log the information in lo trace and thro 404 for end user

RomuDeuxfois: I agree, but it was hard to debug and is not clear at all what's is going on.

corinnekrych: Agreed!
you're right
Best practice for multi-tenancy authz errors:

If the user is authenticated but not a member of the tenant, respond with 403 Forbidden and optionally display a message such as, “You are not a member of this tenant. Please contact your administrator to request access.”
Only use 404 if you intentionally want to hide the existence of the resource for security reasons (resource enumeration protection), typically in highly sensitive applications.

corinnekrych (approved):

PR #5630, opened 2026-04-28
[backend] feat(multi-tenancy): isolate tenant user logic (#4864)

RomuDeuxfois:


antoinemzs (approved):
Good job DRYing the code. It looks like the new tests is just an integration oriented rewrite of the former tests?

PR #5658, opened 2026-04-30
[frontend] fix(tests): fix flacky e2e tests

guillaumejparis:


laugiov:
Heads up: with fullyParallel: false + workers: 1, PR e2e runs go fully sequential. 
Right call to kill flakiness fast, but worth watching the run duration, if it gets painful the longer-term fix is making tests independent enough so we can reparallelize.

laugiov (approved):

PR #5694, opened 2026-05-05
[backend] fix(multi-tenancy): dev env with openaev_owner new user (#0000)

corinnekrych:
### Proposed changes

Flyway migration to use `openaev_owner` for running migration

To test it, in your `application-dev.properties` change:
```
spring.datasource.username=openaev_owner
spring.datasource.password=openaev
```

### Testing Instructions

1. Step-by-step how to test

### Related issues

* Closes #ISSUE-NUMBER
*

### Checklist

### Further comments


openaev-api/src/main/java/io/openaev/migration/V5_06__Enable_row_level_security.java
```diff
@@ -20,10 +17,17 @@
  * bypass — platform-level requests default to the default tenant UUID, and tenant creation
  * explicitly switches the connection's tenant via {@code SET app.current_tenant}.
  *
- * <p>Because superusers bypass RLS, this migration also creates a non-superuser role {@code
- * openaev_app} that the application adopts at runtime via {@code SET ROLE openaev_app} (configured
- * in {@code spring.datasource.hikari.connection-init-sql}). Flyway continues to run as the
- * superuser for DDL operations.
+ * <p>Prerequisites (run by DBA once):
```
Dimfacion, line 20: **Question :** Does that mean that we need to provide a clear documentation to the existing users that they need to run this before upgrading ?

corinnekrych: Yes, this is a pre-requisites

openaev-api/src/main/java/io/openaev/migration/V5_06__Enable_row_level_security.java
```diff
@@ -20,10 +17,17 @@
...
+ *
+ * <pre>
+ * CREATE ROLE openaev_app NOLOGIN NOSUPERUSER;
+ * GRANT openaev_app TO openaev_owner;
```
antoinemzs, line 24: I don't think `openaev_owner` will be created on most setups by the time this runs

openaev-dev/init-db.sql
```diff
@@ -0,0 +1,19 @@
...
+-- Creates the app role and owner user for RLS support
+
+-- 1. Create the app role that RLS will apply to
+CREATE ROLE openaev_app NOLOGIN NOSUPERUSER;
+
+-- 2. Create the migration/app user with enough privileges
+CREATE USER openaev_owner WITH PASSWORD 'openaev' NOSUPERUSER CREATEROLE;
+
+-- 3. Grant the app role to the owner (so SET ROLE works at runtime)
+GRANT openaev_app TO openaev_owner;
```
antoinemzs, line 11: Can you explain why this grant is necessary?

corinnekrych: This grant is necessary to make **SET ROLE openaev_app** work at runtime.

In PostgreSQL, SET ROLE only works if the current session user is a member of the target role. Without `GRANT openaev_app TO openaev_owner,` any attempt by the app to do SET ROLE openaev_app would fail with a "permission denied" error.
By granting openaev_app to openaev_owner, the application can switch into the openaev_app role at runtime, which ensures RLS policies are evaluated under the correct role context — enforcing tenant isolation as intended.

openaev_owner owns the tables (required to run DDL/RLS policies), but table ownership bypasses RLS at query time. The application therefore adopts openaev_app via SET ROLE — a non-owner role that RLS policies are enforced on. GRANT openaev_app TO openaev_owner makes that role switch possible.

openaev-api/src/main/java/io/openaev/migration/V5_06__Enable_row_level_security.java
corinnekrych: what do you mean? 
the user is created via the container bootstrap/seed `init-db.sql`
it works in dev env, tested locally

what i dont know how to fix - I'm stuck on that one - is how to make drone working as it is not using container to run `api-test`

antoinemzs (approved):

PR #5704, opened 2026-05-06
[frontend] fix: handle payload arguments correctly in updates (#4824)

savacano28:
### Proposed changes

* Fix update of payloads with arguments. This bug was introduced in chaining feature (typed arguments).

### Testing Instructions

1. Go threath arsenal page
2. Create payload with arguments
3. Update it
4. Re open and check values 

### Related issues

* Related #4824 

### Checklist

### Further comments

### DIFF
--- a/openaev-front/src/admin/components/threat_arsenal/ThreatArsenalActionForm.tsx
+++ b/openaev-front/src/admin/components/threat_arsenal/ThreatArsenalActionForm.tsx
@@ -2,7 +2,7 @@ import { zodResolver } from '@hookform/resolvers/zod';
 import { Box, Button } from '@mui/material';
 import { useTheme } from '@mui/material/styles';
 import { type SyntheticEvent, useEffect } from 'react';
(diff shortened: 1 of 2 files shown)


openaev-front/src/admin/components/threat_arsenal/ThreatArsenalActionForm.tsx
```diff
@@ -159,7 +159,7 @@ const ThreatArsenalActionForm = ({
 
   const methods = useForm<ThreatArsenalActionCreateCustomInput>({
     mode: 'onTouched',
-    resolver: zodResolver(schema),
+    resolver: zodResolver(schema) as never,
```
camrrx: Using "as never" here basically bypasses Zod’s type safety.
You could remove .nullable() and handle the null case directly in ThreatArsenalActionPopover, so that undefined is sent instead of null for the subtype

camrrx (approved):

Issue #109, opened 2021-11-24
Cyber range integration


SamuelHassine:
Please replace every line in curly brackets { like this } with appropriate answers, and remove this line.

## Problem to Solve

{ Please describe the problem you would like to solve. }

## Current Workaround

{ Please describe how you currently solve or work around this problem, given OpenEx's limitation. }

## Proposed Solution

{ Please describe the solution you would like OpenEx to provide, to solve the problem above. }

## Additional Information

{ Any additional information, including logs or screenshots if you have any. }

PR #5711, opened 2026-05-06
[backend] fix(multitenancy): fix incorrect minio migration (#109)

Dimfacion:
### Proposed changes

* Fix on migration in minio moving all the tenants excepts the default one into the default one
*

### Testing Instructions

1. Create a tenant
2. Upload a document
3. Reboot the backend
4. Check that the path of the document hasn't changed and it's still available

### Related issues

* Closes [#109](https://github.com/OpenAEV-Platform/filigran-private/issues/109)
*

### Checklist

### DIFF
--- a/openaev-model/src/main/java/io/openaev/driver/MinioDriver.java
+++ b/openaev-model/src/main/java/io/openaev/driver/MinioDriver.java
@@ -7,7 +7,8 @@ import io.openaev.config.MinioConfig;
 import io.openaev.config.S3Config;
 import io.openaev.database.model.Tenant;
 import io.openaev.database.repository.TenantRepository;
-import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.context.annotation.Bean;
@@ -72,8 +73,10 @@ public class MinioDriver {
   private void moveDefaultTenantFiles(MinioClient minioClient, String bucket) throws Exception {
     String defaultTenantPrefix = Tenant.DEFAULT_TENANT_UUID + "/";
 
-    List<String> tenants =
-        tenantRepository.findAll().stream().map(tenant -> tenant.getId() + "/").toList();
+    Set<String> tenants =
+        tenantRepository.findAll().stream()
+            .map(tenant -> tenant.getId() + "/")
+            .collect(Collectors.toSet());
 
     Iterable<Result<Item>> objects =
         minioClient.listObjects(ListObjectsArgs.builder().bucket(bucket).recursive(true).build());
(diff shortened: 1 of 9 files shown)


openaev-model/src/main/java/io/openaev/driver/MinioDriver.java
```diff
@@ -68,6 +72,9 @@ public MinioClient minioClient() throws Exception {
   private void moveDefaultTenantFiles(MinioClient minioClient, String bucket) throws Exception {
     String defaultTenantPrefix = Tenant.DEFAULT_TENANT_UUID + "/";
 
+    List<String> tenants =
```
RomuDeuxfois: **nitpick**: You can use a Set here

openaev-model/src/main/java/io/openaev/driver/MinioDriver.java
```diff
@@ -76,7 +83,7 @@ private void moveDefaultTenantFiles(MinioClient minioClient, String bucket) thro
       String objectName = item.objectName();
 
       // Skip files already under a tenant path
-      if (objectName.startsWith(defaultTenantPrefix)) {
+      if (tenants.stream().anyMatch(objectName::startsWith)) {
```
RomuDeuxfois, line 89: **nitpick**: Could be good to have a test for this use case, but since it's critical, I will approved the PR and let you decide if you have the time to add some tests.

RomuDeuxfois (approved):

openaev-model/src/main/java/io/openaev/driver/MinioDriver.java
Dimfacion: Yeah, I thought about it but since we don't have minio in our test env, it would mean mocking the calls to minio and at this point, I'm unsure if that still makes sense ... wdyt ?

openaev-model/src/main/java/io/openaev/driver/MinioDriver.java
Dimfacion: Done

openaev-model/src/main/java/io/openaev/driver/MinioDriver.java
RomuDeuxfois: MinIO is already running in CI, but we don’t currently have a dedicated MinIO instance for tests in our local environment.
We should plan to add one later.
Let's go like this for now.

EmilieFo17:
⚠️ **PRWatch — Missing test coverage**

This bug-fix PR does not appear to include any test changes. Please consider adding tests to cover the fix and help prevent regressions.

🤖 _This is an automated message from PRWatch._

openaev-model/src/main/java/io/openaev/driver/MinioDriver.java
Dimfacion: Right, I just looked at the docker-compose and assumed it was the same in CI. I'll add a minio test instance then in the docker-compose and add a test :)

Dimfacion: @RomuDeuxfois It should be good now. It induced changing the port for the tests though as letting the default 9000 port would have meant that tests run in local would impact the local dev minio instance

RomuDeuxfois (approved):

openaev-model/src/main/java/io/openaev/driver/MinioDriver.java
RomuDeuxfois: Love it, thanks Gael !

PR #5716, opened 2026-05-07
[backend] feat(chaining): add asset to injects based on scope allowlist/denylist (#4824)

heditar:
### Proposed changes
 
* In the chaining adding assets to creaqted inject based on allowlist/denylist

### Testing Instructions

1. Create a chained simulation and add assets in scope, play with the denylist by adding ip/ip subnet 
2. Launch the simulation and make sure they are executed on the right targets

### Related issues

* #4824 

### Checklist

### DIFF
--- a/openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
+++ b/openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
@@ -31,6 +31,7 @@ import io.openaev.rest.injector_contract.InjectorContractService;
 import io.openaev.rest.tag.TagService;
 import io.openaev.service.*;
 import io.openaev.service.chaining.ConditionService;
+import io.openaev.service.chaining.ScopeService;
 import io.openaev.service.chaining.StepService;
 import io.openaev.service.chaining.WorkflowStateService;
 import io.openaev.utils.ConditionUtils;
@@ -79,6 +80,7 @@ public class InjectExecutionStep implements ActionStep {
   private final AssetGroupService assetGroupService;
   private final ConditionService conditionService;
   private final WorkflowStateService workflowStateService;
+  private final ScopeService scopeService;
 
   private final InjectorContractRepository injectorContractRepository;
 
@@ -635,6 +637,8 @@ public class InjectExecutionStep implements ActionStep {
       // Modify payload arguments with inputs from step
       ObjectNode updatedContent = updateContentWithInputs(step, injectorContract.getContent());
       inject.setContent(updatedContent);
+
+      inject.setAssets(scopeService.getValidAssets(step.getWorkflow().getId()));
 
       return inject;
 
--- a/openaev-api/src/main/java/io/openaev/service/chaining/ScopeService.java
+++ b/openaev-api/src/main/java/io/openaev/service/chaining/ScopeService.java
@@ -1,0 +1,159 @@
+package io.openaev.service.chaining;
+
+import io.openaev.database.model.*;
+import io.openaev.database.repository.WorkflowScopeRuleRepository;
+import io.openaev.service.AssetGroupService;
+import io.openaev.service.AssetService;
+import io.openaev.utils.IpAddressUtils;
+import java.util.*;
+import java.util.stream.Collectors;
+import lombok.RequiredArgsConstructor;
+import org.springframework.stereotype.Service;
+
+@RequiredArgsConstructor
+@Service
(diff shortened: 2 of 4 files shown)


openaev-api/src/main/java/io/openaev/service/chaining/ScopeService.java
```diff
@@ -0,0 +1,159 @@
...
+    completeAllowedAssetList.addAll(allowedByAssetGroup);
+
+    // apply the denylist
+    if (completeAllowedAssetList.isEmpty() || denylistRules.isEmpty()) {
+      return List.copyOf(completeAllowedAssetList);
+    }
+
+    //  remove any asset matched by a denylist rule  and return
+    Set<String> deniedAssetIds = getDeniedAssetIds(completeAllowedAssetList, denylistRules);
+    return completeAllowedAssetList.stream()
```
MarineLeM, line 55: Nitpick : This code iterates over completeAllowedAssetList twice: once in getDeniedAssetIds to collect denied IDs, and again to filter them out. I think you can merge both to a single loop

MarineLeM (changes requested):

openaev-api/src/main/java/io/openaev/service/chaining/ScopeService.java
```diff
@@ -0,0 +1,159 @@
...
+
+        case IP_SUBNET ->
+            candidates.stream()
+                .filter(asset -> asset instanceof Endpoint)
+                .map(asset -> (Endpoint) asset)
+                .filter(endpoint -> endpointMatchesSubnet(endpoint, rule.getRuleValue()))
+                .map(Asset::getId)
+                .forEach(deniedAssetIds::add);
+
+        case DOMAIN ->
```
MarineLeM: Why is it called "domain" instead of "hostname"? The term domain is already used in the codebase to refer to the domain table, which creates confusion. hostname would be more accurate and unambiguous here, wdyt ?

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
```diff
@@ -636,6 +638,8 @@ private Inject getInjectFromDataStep(Step step) throws ChainingException {
       ObjectNode updatedContent = updateContentWithInputs(step, injectorContract.getContent());
       inject.setContent(updatedContent);
 
+      inject.setAssets(scopeService.getValidAssets(step.getWorkflow().getId()));
```
MarineLeM, line 641: Question : @guillaumejparis mentioned that allowList and denyList apply to the targeted asset, not the source asset — but based on this PR, the behavior seems to be the opposite. Can you clarify which is correct?

MarineLeM (dismissed):

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
heditar: they apply to both, deny list should block target or source

openaev-api/src/main/java/io/openaev/service/chaining/ScopeService.java
heditar: That makes sense I will update it, also there is a bug here I forgot to fix I will update it in the next commit

heditar: I removed this part from the PR  as it will require more details from PO to understand if it makes sense to have it here

openaev-api/src/main/java/io/openaev/service/chaining/ScopeService.java
heditar: I would prefer keep like that for readibility as I don't have concern about the performances in that case

openaev-api/src/main/java/io/openaev/service/chaining/ScopeService.java
heditar: Confirmed with Murat, not needed on this PR I will add another task to implement what we need for that

camrrx (commented):
TEST : 
- OK with asset in denylist ✅ 
- not working with asset group in denylist 👎   
- OK with subnet in denylist ✅

heditar:
> TEST :
> 
> * OK with asset in denylist ✅
> * not working with asset group in denylist 👎
> * OK with subnet in denylist ✅

Retested it, it works for me

camrrx (approved):

PR #5731, opened 2026-05-11
[backend] feat(chaining): check conditions to execute action (#4824)

camrrx:
### Proposed changes

* Remove AFTER/BEFORE condition types from ConditionType enum, frontend form, and API types : time-based scheduling is no longer used in the chaining engine.
* Implement FILTER conditions: resolves values from both local and global state

### Testing Instructions

1. Create chaining simulation
2. You can play around with the actions and events to test different combinations
Example : 
4. On the logic tab : add 2 actions and one event
2.1. First action : "whoami" who its output is a text
2.2. Second action : "echo #{input}" with text as input
2.3. Event : text="toto" (toto should match the actual output of the whoami action)

5. Launch the simulation
6. If the event evaluates to true, the linked action should be triggered.
7. If the event evaluates to false, the linked action should not be triggered.

### Related issues

* Link #4824 

### Checklist

### Further comments

### DIFF
--- a/openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
+++ b/openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
@@ -523,19 +523,6 @@ public class InjectExecutionStep implements ActionStep {
     StepsCreateInput.StepInput stepCreateInput = new StepsCreateInput.StepInput();
     stepCreateInput.setDataStep(input);
     stepCreateInput.setStepAction(StepActionClass.INJECT_EXECUTION);
-
-    if (input.getDependsDuration() != 0) {
-      ConditionCreateInput conditionCreateInput =
-          ConditionCreateInput.builder()
-              .temporaryId("0")
-              .type(ConditionType.AFTER)
-              .key(null)
-              .keyType(null)
-              .mappingType(null)
-              .value(String.valueOf(input.getDependsDuration()))
-              .build();
-      stepCreateInput.setConditions(List.of(conditionCreateInput));
-    }
     // TODO DEPEND ON
 
     return stepCreateInput;
@@ -579,6 +566,12 @@ public class InjectExecutionStep implements ActionStep {
     try {
       // GET INJECT FROM JSON
       Inject inject = om.readValue(step.getData(), Inject.class);
+
+      // Ensure dependsDuration has a default value — it is a legacy scheduling field
+      // that may not be present in the serialized step data but has a @NotNull constraint.
+      if (inject.getDependsDuration() == null) {
+        inject.setDependsDuration(0L);
+      }
 
       ObjectMapper mapper = new ObjectMapper();
       JsonNode root = mapper.readTree(step.getData());
--- a/openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
+++ b/openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
@@ -305,15 +305,16 @@ public class InjectExecutionStep implements ActionStep {
       for (Map<String, JsonElement> entry : output) {
         if (entry.containsKey("parsed")) {
           JsonObject parsed = entry.get("parsed").getAsJsonObject();
-          if (parsed.has("_children")) {
-            JsonObject children = parsed.getAsJsonObject("_children");
 
-            for (String key : children.keySet()) {
-              JsonArray valuesArray = children.getAsJsonObject(key).getAsJsonArray("_children");
-              for (JsonElement item : valuesArray) {
-                String val = item.getAsJsonObject().get("_value").getAsString();
-                // SyncState keys are usually Uppercase (e.g., "IP")
-                result.computeIfAbsent(key, k -> new ArrayList<>()).add(val);
+          for (String key : parsed.keySet()) {
+            JsonElement element = parsed.get(key);
+            if (element == null || !element.isJsonArray()) {
+              continue;
+            }
+            JsonArray valuesArray = element.getAsJsonArray();
+            for (JsonElement item : valuesArray) {
+              if (item.isJsonPrimitive()) {
+                result.computeIfAbsent(key, k -> new ArrayList<>()).add(item.getAsString());
               }
             }
           }
@@ -729,7 +730,7 @@ public class InjectExecutionStep implements ActionStep {
       if (trace.getStructuredOutput() != null) {
         log.info(
             "[Chaining] Trace has structuredOutput: {}", trace.getStructuredOutput().toString());
-        map.put("parsed", gson.toJsonTree(trace.getStructuredOutput()));
+        map.put("parsed", JsonParser.parseString(trace.getStructuredOutput().toString()));
       } else {
         log.info("[Chaining] Trace has NO structuredOutput, message: {}", trace.getMessage());
         try {
--- a/openaev-api/src/main/java/io/openaev/service/chaining/ConditionService.java
+++ b/openaev-api/src/main/java/io/openaev/service/chaining/ConditionService.java
@@ -447,94 +447,255 @@ public class ConditionService {
       return List.of(new ExecutionBatch(input, new ArrayList<>()));
     }
 
-    //    // TIME CONDITIONS
-    //    // TODO manage multi time condition (AND, OR: g C1 BEFORE OR C2 AFTER)
-    //    // Compute expected start time for the condition to be considered as valid
-    //     List<Condition> timeConditions =
-    //        conditionTemplate.stream().filter(this::isTimeCondition).toList();
-    //     for (Condition condition : timeConditions) {
-    //      Instant now = Instant.now();
-    //      Instant start = workflowRun.getWorkflowCreatedAt();
-    //      // TODO: can this happen ? Shouldn't it throw an exception instead?
-    //      if (start == null) start = now;
-    //      long value = Long.parseLong(condition.getValue());
-    //      Instant goal = start.plus(value, ChronoUnit.MILLIS);
-    //
-    //      if (isTimeConditionValid(condition, now, goal)) {
-    //        conditionsExecution.add(ConditionFactory.executionOf(condition, goal));
-    //        continue;
-    //      }
-    //      if (condition.getType().equals(ConditionType.AFTER)) {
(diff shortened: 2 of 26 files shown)


openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
```diff
@@ -521,19 +521,6 @@ public static StepsCreateInput.StepInput getInjectAsStepsCreateInput(InjectInput
     StepsCreateInput.StepInput stepCreateInput = new StepsCreateInput.StepInput();
     stepCreateInput.setDataStep(input);
     stepCreateInput.setStepAction(StepActionClass.INJECT_EXECUTION);
-
```
savacano28, line 526: 🙏🏼

EvaE-Filigran: The engine should be compatible with both time-based and event-based modes. This part, along with the AFTER and BEFORE conditions, was added to make the transition easier. Have we moved this code elsewhere or implemented another mechanism to preserve this compatibility?

camrrx: At the moment, we are not handling the time-based mode in the engine. The time-based logic is completely separate from it.
It felt easier to manage it later, once we implement the time-based mode in this new engine. However, I can add it back if it had already been decided that we started implement time-based in the new engine :)

openaev-api/src/main/java/io/openaev/utils/ConditionUtils.java
```diff
@@ -12,47 +11,14 @@
...
-
   /**
    * Checks whether the condition is a mapper condition.
    *
    * @param condition condition to evaluate
    * @return {@code true} if the condition type is MAPPER
    */
   public boolean isMapperCondition(Condition condition) {
-    return condition.getType() == ConditionType.MAPPER;
+    return condition.getType() != null && condition.getType() == ConditionType.MAPPER;
```
heditar: ConditionType.MAPPER.equals(condition.getType())

openaev-api/src/main/java/io/openaev/service/chaining/WorkflowStateService.java
```diff
@@ -230,6 +230,29 @@ private Map<String, List<String>> filterValuesMatchingConditions(
...
+  /**
+   * Checks whether a value matches any leaf condition in the condition tree, ignoring AND/OR
+   * logical grouping. This is used for propagation (deciding which values are relevant to an
+   * event), not for full evaluation (deciding if the event is fully satisfied).
+   *
+   * @param value the value to check
+   * @param node the condition tree node to inspect
+   * @return {@code true} if the value satisfies at least one leaf condition in the tree
+   */
+  private boolean matchesAnyLeafCondition(String value, Condition node) {
```
heditar: I think it is more. related to Condition than WorkflowState so should be in Condition service and any logic to evaluate a condition

heditar:
tested ok

impolitepanda:
Tested ok

openaev-api/src/main/java/io/openaev/service/chaining/ConditionService.java
```diff
@@ -447,94 +447,255 @@ public List<ConditionService.ExecutionBatch> checkCondition(
...
+    // FILTER CONDITIONS
+    // Evaluate filter conditions against the current input AND the finding state (global state).
+    // If another action produced a parameter needed by the filter, we look it up
+    // in the workflow state pool.
+    List<Condition> filterConditions =
+        conditionTemplate.stream().filter(conditionUtils::isFilterCondition).toList();
+
+    if (!filterConditions.isEmpty()) {
+      if (!evaluateFilterConditions(filterConditions, workflowRun, nextStepTemplateToExecute)) {
+        log.info(
```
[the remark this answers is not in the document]
impolitepanda: Copilot is correct here

openaev-api/src/main/java/io/openaev/service/chaining/ConditionService.java
```diff
@@ -447,94 +447,255 @@ public List<ConditionService.ExecutionBatch> checkCondition(
...
+   * @param dependOnConditions list of DEPEND_ON conditions to evaluate
+   * @param workflowRun the running workflow (provides the run ID to scope the check)
+   * @return {@code true} if all dependencies are satisfied
+   */
+  private boolean evaluateDependOnConditions(
+      List<Condition> dependOnConditions, Workflow workflowRun) {
+    for (Condition condition : dependOnConditions) {
+      String dependentStepTemplateId = condition.getValue();
+      if (dependentStepTemplateId == null || dependentStepTemplateId.isBlank()) {
+        log.info("DEPEND_ON condition has no step template ID value: {}", condition.getId());
```
impolitepanda: log level here should probably not be info. More debug. However, in what functional case could a depend on condition not have a step template ID ? If it's not supposed to happen ever, i would upgrade the log level to error, for example

openaev-api/src/main/java/io/openaev/service/chaining/ConditionService.java
```diff
@@ -447,94 +447,255 @@ public List<ConditionService.ExecutionBatch> checkCondition(
...
+   * @return {@code true} if all filter conditions are satisfied
+   */
+  private boolean evaluateFilterConditions(
+      List<Condition> filterConditions, Workflow workflowRun, Step stepTemplate) {
+
+    // Lazy-loaded state entries
+    WorkflowStateEntries[] globalEntries = {null};
+    WorkflowStateEntries[] localEntries = {null};
+
+    // Supplier to lazy-load context on first need
```
impolitepanda: I'm not sure I understand the reasonwe lazy load the context here. Could you explain ?

openaev-api/src/main/java/io/openaev/service/chaining/ConditionService.java
```diff
@@ -447,94 +447,255 @@ public List<ConditionService.ExecutionBatch> checkCondition(
...
+   * @param filterConditions list of root filter conditions to evaluate
+   * @param workflowRun the running workflow for accessing the state pools
+   * @param stepTemplate the step template for accessing local state
+   * @return {@code true} if all filter conditions are satisfied
+   */
+  private boolean evaluateFilterConditions(
+      List<Condition> filterConditions, Workflow workflowRun, Step stepTemplate) {
+
+    // Lazy-loaded state entries
+    WorkflowStateEntries[] globalEntries = {null};
```
impolitepanda: This seems weird to me so i'm probably not understanding this correctly. But if I understand correctly, we are using these arrays to store a state that is updated inside lambdas, correct? And this is due to the lazy loading below ? If that's the case, i'd prefer we use a supplier, as described in previous comment.

Or maybe I'm missing something?

openaev-api/src/main/java/io/openaev/service/chaining/ConditionService.java
impolitepanda: Ok, after more reading, I assume we are doing this for performances issues, to avoid loading the context if there are no leaves in the condition tree.

I don't know if that's something that will occur often. Any idea ?

If it does, then I see the merit in it but we still need to simplify it and remove the arrays. I'd suggest using a supplier to store and load the context as a whole.

impolitepanda (changes requested):

openaev-api/src/main/java/io/openaev/service/chaining/ConditionService.java
impolitepanda: Code snippet for the supplier if we end up using it:
```
    Supplier<WorkflowContext> contextSupplier =
        new Supplier<>() {
          private WorkflowContext cached;

          @Override
          public WorkflowContext get() {
            if (cached == null) {
              cached = fetchWorkflowContext(workflowRun, stepTemplate);
            }
            return cached;
          }
        };```

File: openaev-api/src/main/java/io/openaev/service/chaining/ConditionService.java (hunk 7 above)
Comment (impolitepanda) [reply openaev-api/src/main/java/io/openaev/service/chaining/ConditionService.java]: Comment can be resolved as the Supplier modification makes this code more understandable

**Review by impolitepanda — APPROVED** [review]

File: openaev-model/src/main/java/io/openaev/database/model/ConditionType.java [hunk 8]
```diff
@@ -13,8 +13,6 @@ public enum ConditionType {
   LTE,
   IN,
   NIN,
-  AFTER,
-  BEFORE,
   MAPPER,
```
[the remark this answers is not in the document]
camrrx: We’re going to handle the time condition later. For now, chaining does not take time conditions into account, as this is still managed by the previous engine.

openaev-api/src/test/java/io/openaev/service/chaining/ConditionServiceTest.java
```diff
@@ -320,6 +231,7 @@ void shouldDelegateToExtractInputsForStepExecution_withMapperConditionsOnly()
       List<Condition> conditions = List.of(filterTemplate, mapperTemplate);
 
       doReturn(conditions).when(conditionService).findAllConditionsByStepId(stepId);
+      when(filterTemplate.getType()).thenReturn(ConditionType.MAPPER);
```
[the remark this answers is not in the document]
camrrx: This test is interesting with the mapping condition

openaev-front/src/admin/components/chaining/logic/forms/CreateEventForm.tsx
```diff
@@ -154,8 +164,6 @@ const CreateEventForm: FunctionComponent<Props> = ({ onSubmit, handleClose }) =>
               <MenuItem value="LTE">{t('LTE')}</MenuItem>
               <MenuItem value="IN">{t('IN')}</MenuItem>
               <MenuItem value="NIN">{t('NIN')}</MenuItem>
-              <MenuItem value="AFTER">{t('AFTER')}</MenuItem>
-              <MenuItem value="BEFORE">{t('BEFORE')}</MenuItem>
               <MenuItem value="DEPEND_ON">{t('DEPEND_ON')}</MenuItem>
             </Select>
```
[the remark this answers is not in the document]
camrrx: We’re going to handle the time condition later. For now, chaining does not take time conditions into account, as this is still managed by the previous engine.

PR #5732, opened 2026-05-11
[backend] fix(threat-arsenal): payload creation take too long (#100)

gabriel-peze:
### Proposed changes

* Review the process to implement a basic data creation process to database instead of the java runtime process, to avoid time consuming and performance issues
* Payload creation should take less than 3 seconds, in local, from 5seconds to process to 260ms

### Testing Instructions

Everything should work as usual, only performance have been fixed.

1. Create a payload from the Threat Arsenal page
2. API should create the payload in less than 3seconds
3. Verify that your payload is created into injectors with payload value at true, on your tenant

### DIFF
--- a/openaev-model/src/main/java/io/openaev/database/repository/InjectorContractRepository.java
+++ b/openaev-model/src/main/java/io/openaev/database/repository/InjectorContractRepository.java
@@ -175,7 +175,7 @@ public interface InjectorContractRepository
    * and matching the {@code payloads} flag, a new row is inserted only if no association with the
    * given contract already exists (idempotent by design).
    *
-   * @param tenantId the identifier of the tenant whose injectors should be considered.
+   * @param injectorIds the identifier of the injectors who should be considered.
    * @param contractId the identifier of the injector contract to associate with the matching
    *     injectors.
    */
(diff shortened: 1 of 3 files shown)


openaev-model/src/main/java/io/openaev/database/repository/InjectorContractRepository.java
```diff
@@ -165,4 +165,36 @@ WHERE LOWER(vulnerability.vulnerability_external_id) IN (:externalIds)
...
+   *     injectors.
+   */
+  @Modifying(clearAutomatically = true, flushAutomatically = true)
+  @Query(
+      value =
+          """
+    INSERT INTO injectors_injector_contracts (injector_id, injector_contract_id, tenant_id)
+    SELECT i.injector_id, :contractId, i.tenant_id
+    FROM injectors i
+    WHERE i.injector_payloads = TRUE
```
MarineLeM: Since `injectorIds` is already retrieved in PayloadService, you can directly select injectors by their IDs instead of using `WHERE i.injector_payloads = TRUE AND i.tenant_id = :tenantId`.

MarineLeM (changes requested):

MarineLeM (approved):

PR #5778, opened 2026-05-13
[backend/frontend] fix(domains): make security domains widget display results (#108)

johanah29:
### Proposed changes

At the moment when you create a security domains widget, all the icons are grey. Also when trying to add a simulation as a filter during the creation, no proposition appears in the select since there is an error. 
I also spoted that the links in the enterprise edition agreement dialog were not obvious enough. I also changed that.

### Testing Instructions

1. Create a dashboard, then choose security domains widget
2. If you try to add simulations, you should have simulations in the select field
3. Your widget should have colors reflecting the results
4. Try with elasticsearch and opensearch engines

Note: Remember that this widget is based on results of inject expectations

### DIFF
--- a/openaev-model/src/main/java/io/openaev/service/OpenSearchService.java
+++ b/openaev-model/src/main/java/io/openaev/service/OpenSearchService.java
@@ -358,7 +358,6 @@ public class OpenSearchService implements EngineService {
     Query tenantFilter =
         BoolQuery.of(b -> b.should(matchesTenant, noTenantField).minimumShouldMatch("1")).toQuery();
     mainQuery.filter(tenantFilter);
-
     return mainQuery.must(mainMust).build().toQuery();
   }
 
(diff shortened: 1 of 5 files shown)


openaev-model/src/main/java/io/openaev/service/OpenSearchService.java
```diff
@@ -375,13 +374,17 @@ private Map<String, String> resolveIdsRepresentative(RawUserAuth user, List<Stri
...
     filter.setValues(ids);
+    filter.setMode(Filters.FilterMode.or);
     filterGroup.setFilters(List.of(filter));
     Query query = buildQuery(user, null, filterGroup, new HashMap<>(), new HashMap<>());
     try {
       SearchResponse<EsBase> response =
           openSearchClient.search(
               b -> b.index(engineConfig.getIndexPrefix() + "*").size(ids.size()).query(query),
               EsBase.class);
+      log.debug(
```
RomuDeuxfois: **todo**: remove this before merge

RomuDeuxfois (approved):

PR #5788, opened 2026-05-13
[frontend] chore: rollback apexcharts version

RomuDeuxfois:


.drone.yml
```diff
@@ -39,6 +39,7 @@ steps:
       - yarn install
       - yarn build
       - yarn check-ts
+      - yarn check-licenses
```
xfournet, line 42: isn't drone deprecated ? we should update GHA instead ?

RomuDeuxfois: Good point, I forgot that part

xfournet (approved):

PR #5799, opened 2026-05-15
feat(url-access-token): implement url access token management (#5544)

### LINKED ISSUE #5544: [player-access] Implement scoped access tokens for player links
## Context

Player access to exercise resources (communications, documents, challenges) is identified via a userId query parameter on public /api/player/** endpoints. 
This identifier acts as a permanent, non-expiring, non-revocable credential exposed in URLs.
This change introduces time-bounded, revocable, scoped access tokens to replace the current pattern. 
It is a hardening improvement with no user-facing behavior change, players still click a link from an email and land on their exercise resources.

## Use case

As a platform operator running an exercise, I need player access links to expire at the end of the exercise (plus a configurable margin) and to be revocable individually or in bulk, so that a leaked or forwarded link does not grant indefinite access to player data across past, present and future exercises.

## Current Workaround

None. There is no mechanism today to expire or revoke a player's access to exercise resources. 
Once the link is emitted, it remains valid for as long as the player exists in the system.

## Proposed Solution

Introduce opaque, short-lived, scoped access tokens backed by a dedicated table.

 * New table player_access_token (token_hash, user_id, exercise_id, expires_at, revoked_at, last_used_at, created_at)
 * Tokens generated at player creation within an exercise: 256-bit SecureRandom, stored SHA-256 hashed
 * Lifetime tied to the exercise end date + configurable margin
 * Scoped to a single exerciseId, a token from exercise A cannot be used on exercise B
 * First hit triggers a 302 redirect setting an HttpOnly; Secure; SameSite=Strict cookie, removing the token from the URL (prevents leakage via logs and Referer)
 * Bulk revocation hook on exercise closure
 * Admin endpoint for individual token revocation
 * Background purge of expired and revoked tokens after a configurable retention window

A JWT-based alternative was considered but rejected: it would not support instant revocation without maintaining a blocklist, and would not provide the audit trail needed for anomaly detection and bulk revocation.

## Additional Information

No user-facing UI change expected. Migration of existing active exercises is out of scope for this issue and will be handled separately if needed.

Available to pair on design and review the PR. Implementation assignment to be determined by EMs based on vertical roadmap (per alignment in Teams thread).

gabriel-peze:
### Proposed changes

* Implement the URL Access Token process to avoid sharing full url and user id data on mail.
* Control url availability by using revocable tokens

### Testing Instructions

1. Setup the feature flag "URL_ACCESS_TOKEN"
2. All received mail should display common url with a token to handle redirection and access, so run challenge or channel inject to receive a mail, or send a lessons learned mail.
3. Access to the provided url,you shouldn't have any user id in the url and everything should work as usual on the process
4. revoke your token
5. the url should not be accesssible anymore.
6. disable the feture flag, everything should come back as before

### Related issues

* Closes #5544


openaev-api/src/main/java/io/openaev/rest/lessons/ExerciseLessonsApi.java
```diff
@@ -383,17 +384,30 @@ public List<LessonsAnswer> exerciseLessonsAnswers(
...
+        .stream()
+        .filter(
+            lessonsCategory ->
+                userId.isEmpty()
+                    || lessonsCategory.getTeams().stream()
+                        .anyMatch(
+                            team ->
+                                team.getUsers().stream()
+                                    .anyMatch(user -> user.getId().equals(userId.get()))))
+        .toList();
```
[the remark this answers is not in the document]
gabriel-peze: Anonymous can't receive challenge or channel mail, neither lessons mail, and for a token the userId is mandatory, we will always get a userId here, except if it's call by an admin on preview

laugiov:
Hi Gabriel, nice work on this PR. The design matches the brainstorm as we discussed.

One question : the issue mentions "Bulk revocation hook on exercise closure" on top of the configurable margin. 
Today the auto-revocation on exercise closure (manual or automatic transition to FINISHED/CANCELED) is not implemented, only the reset path and the admin endpoints revoke. 
Is the 7-day margin enough for the use case, or do we want to also cover the "close + revoke immediately" scenario? If important, it's a small addition in changeExerciseStatus, InjectsExecutionJob and WorkflowTimeoutService. If not, happy to move it to a follow-up. 
cc @DeniseMonreale  for the product call.

DeniseMonreale:
thanks @laugiov for this detailed feedback 
Product is aware and @soumboussaha and @EvaE-Filigran will sync soon to give you an answer asap

openaev-api/src/main/java/io/openaev/api/url_access_token/UrlAccessTokenApi.java
```diff
@@ -0,0 +1,97 @@
...
+              .build();
+
+      return ResponseEntity.status(HttpStatus.FOUND)
+          .header(HttpHeaders.SET_COOKIE, cookie.toString())
+          .header(HttpHeaders.LOCATION, token.getUrl())
+          .build();
+    } catch (AccessDeniedException exception) {
+      return ResponseEntity.status(HttpStatus.FOUND)
+          .header(HttpHeaders.LOCATION, this.urlAccessTokenService.getErrorPageUrl())
+          .build();
```
antoinemzs: I think we need to either not do that or go all the way and customize the whitelabel error page https://www.baeldung.com/spring-boot-custom-error-page

openaev-api/src/main/java/io/openaev/api/url_access_token/UrlAccessTokenApi.java
```diff
@@ -0,0 +1,97 @@
...
+    urlAccessTokenService.revokeToken(tokenId);
+    return ResponseEntity.noContent().build();
+  }
+
+  @DeleteMapping("/exercise/{exerciseId}")
+  @LogExecutionTime
+  @AccessControl(skipRBAC = true)
+  @Operation(summary = "Revoke all URL access tokens for an exercise (admin only)")
+  public ResponseEntity<Void> revokeByExerciseId(@PathVariable("exerciseId") String exerciseId) {
+    ensureCurrentUserIsAdmin();
```
antoinemzs: Why no RBAC?

openaev-api/src/main/java/io/openaev/api/url_access_token/UrlAccessTokenApi.java
```diff
@@ -0,0 +1,97 @@
...
+          .build();
+    }
+  }
+
+  @DeleteMapping("/{tokenId}")
+  @LogExecutionTime
+  @AccessControl(skipRBAC = true)
+  @Operation(summary = "Revoke a URL access token by id (admin only)")
+  public ResponseEntity<Void> revokeByTokenId(@PathVariable("tokenId") String tokenId) {
+    ensureCurrentUserIsAdmin();
```
antoinemzs: Privileged actions can and should require auth and RBAC validation

openaev-model/src/main/java/io/openaev/database/model/UrlAccessToken.java
```diff
@@ -0,0 +1,80 @@
...
+  @JsonProperty("url_access_token_token_hash")
+  @NotBlank
+  private String tokenHash;
+
+  @Column(name = "url", nullable = false)
+  @JsonProperty("url_access_token_url")
+  @NotBlank
+  private String url;
+
+  @ManyToOne(fetch = FetchType.EAGER)
```
antoinemzs: We shouldn't use `FetchType.EAGER` unless there is a good reason. This would pull, for example, according to already existing eager fetches, all the user's groups with all their respective grants and respective users and this is a typical case of N+1 query.

https://www.baeldung.com/hibernate-lazy-eager-loading

openaev-api/src/main/java/io/openaev/config/AppSecurityConfig.java
```diff
@@ -104,6 +104,8 @@ public SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
                     .permitAll()
                     .requestMatchers("/api/login")
                     .permitAll()
+                    .requestMatchers("/api/url/access/**")
+                    .permitAll()
```
antoinemzs, line 116: Surely we need to except the privileged endpoints

openaev-api/src/main/java/io/openaev/aop/UrlAccessControlAspect.java
```diff
@@ -0,0 +1,128 @@
...
+    int userIdParamIndex = -1;
+
+    for (int i = 0; i < parameterNames.length; i++) {
+      if ("exerciseId".equals(parameterNames[i]) && args[i] instanceof String s) {
+        exerciseId = s;
+      }
+      if ("userId".equals(parameterNames[i])) {
+        userIdParamIndex = i;
+      }
+    }
```
antoinemzs: you may use SPEL to avoid explicit reflection.

https://docs.spring.io/spring-framework/reference/core/expressions.html

this allows you to pass some of the decorated method's arguments to the decorator (this class)
The SPEL pattern is more robust because the compiler detects when the decorated method's signature changes.

openaev-api/src/main/java/io/openaev/rest/document/DocumentApi.java
```diff
@@ -337,7 +337,7 @@ public void downloadDocument(@PathVariable String documentId, HttpServletRespons
       @PathVariable String injectorId) throws IOException {
     Injector injector =
         this.injectorRepository
-            .findByIdAndTenantId(injectorId, TenantContext.getCurrentTenant())
+            .findById(injectorId)
```
antoinemzs: question: please elaborate

openaev-api/src/main/java/io/openaev/rest/exercise/ExercisePlayerApi.java
antoinemzs, line 1: Some changes here look unrelated to this feature

openaev-api/src/main/java/io/openaev/scheduler/jobs/UrlAccessTokenPurgeJob.java
```diff
@@ -0,0 +1,37 @@
...
+import org.quartz.Job;
+import org.quartz.JobExecutionContext;
+import org.quartz.JobExecutionException;
+import org.springframework.stereotype.Component;
+
+/** Purges old expired or revoked URL access tokens. */
+@Component
+@RequiredArgsConstructor
+@Slf4j
+public class UrlAccessTokenPurgeJob implements Job {
```
antoinemzs, line 19: since we operate in a transaction (scoped in the service class), please introduce a `@DisallowConcurrentExecution` decorator to avoid a pile-up if the job is slower than expected
```suggestion
@Component
@RequiredArgsConstructor
@Slf4j
@DisallowConcurrentExecution
public class UrlAccessTokenPurgeJob implements Job {
```

openaev-api/src/test/java/io/openaev/aop/UrlAccessControlAspectTest.java
```diff
@@ -0,0 +1,323 @@
...
+      ProceedingJoinPoint joinPoint = mock(ProceedingJoinPoint.class);
+      when(joinPoint.proceed()).thenReturn("ok");
+
+      OpenAEVPrincipal principal = mock(OpenAEVPrincipal.class);
+      when(principal.getId()).thenReturn("user-1");
+      Authentication authentication = mock(Authentication.class);
+      when(authentication.getPrincipal()).thenReturn(principal);
+      SecurityContext context = SecurityContextHolder.createEmptyContext();
+      context.setAuthentication(authentication);
+      SecurityContextHolder.setContext(context);
```
antoinemzs, line 120: Could `@WithMockUser` work in this situation?

openaev-api/src/test/java/io/openaev/aop/UrlAccessControlAspectTest.java
```diff
@@ -0,0 +1,323 @@
...
+
+  // -- Feature flag disabled --
+
+  @Nested
+  @DisplayName("Feature flag disabled")
+  class FeatureFlagDisabled {
+
+    @Test
+    @DisplayName(
+        "given_feature_flag_disabled_when_url_access_control_is_applied_should_proceed_without_validation")
```
antoinemzs: FWIW display names are supposed to be "pretty" in that they are actual english sentences :)

openaev-api/src/test/java/io/openaev/aop/UrlAccessControlAspectTest.java
```diff
@@ -0,0 +1,323 @@
...
+      assertThatThrownBy(() -> aspect.validateUrlAccess(joinPoint))
+          .isInstanceOf(IllegalStateException.class)
+          .hasMessageContaining("no HTTP request");
+    }
+
+    @Test
+    @DisplayName("given_no_cookies_in_request_when_url_access_control_is_applied_should_return_401")
+    void given_no_cookies_in_request_when_url_access_control_is_applied_should_return_401() {
+      // -- Arrange --
+      when(previewFeatureService.isFeatureEnabled(PreviewFeature.URL_ACCESS_TOKEN))
```
antoinemzs: to avoid repeating the same setup many times, using `@Nested` classes like you did, but with specific `@BeforeEach` methods at various depths levels can help factorise this

openaev-api/src/main/java/io/openaev/aop/UrlAccessControlAspect.java
```diff
@@ -0,0 +1,128 @@
...
+      log.debug("UrlAccessControlAspect: token validation failed - {}", e.getMessage());
+      throw new ResponseStatusException(HttpStatus.UNAUTHORIZED, INVALID_TOKEN_MESSAGE);
+    }
+
+    // Inject the resolved userId into the method's Optional<String> userId argument
+    if (userIdParamIndex >= 0) {
+      Object[] newArgs = Arrays.copyOf(args, args.length);
+      newArgs[userIdParamIndex] = Optional.of(token.getUser().getId());
+      return joinPoint.proceed(newArgs);
+    }
```
antoinemzs: I don't understand this logic: so given a token tied to [`exercise_A`, `user_1`], if I call a decorated method with args [`exercise_A`, `user_2`] and pass that token, the call will be "fixed" with `user_1` as the user argument, which betrays the caller's intent.
I'd rather see the method's original nullable userId argument passed to the decorator (e.g. via SPEL) and validated against the token's metadata, throwing a 401 if necessary.

openaev-front/src/public/components/lessons/LessonsPlayer.jsx
```diff
@@ -66,7 +66,7 @@ const LessonsPlayer = (props) => {
   };
   const sortCategories = R.sortWith([R.ascend(R.prop('lessons_category_order'))]);
   const sortQuestions = R.sortWith([R.ascend(R.prop('lessons_question_order'))]);
-  const sortedCategories = sortCategories(R.filter(n => n.lessons_category_users.includes(source.finalUserId), lessonsCategories));
+  const sortedCategories = sortCategories(lessonsCategories);
```
antoinemzs, line 69: why do we remove this filter?

antoinemzs (changes requested):

openaev-api/src/test/java/io/openaev/aop/UrlAccessControlAspectTest.java
gabriel-peze: In this case this annotation doesn't allow me not implement the kind of context i wish, or there is something i don't know to allow me to do it :/
I tried again to use @WithMockUser but didn't succeed

openaev-front/src/public/components/lessons/LessonsPlayer.jsx
gabriel-peze: i remove this filter because this logic have been moved to the backend, in most of the cases, we won't have the finalUserId information, because it was the goal of this feature, to avoid passing the user uuid to the frontend, and that's why a moved the logic of filtering by user id to the backend, that's mean we don't need to filter anymore on the frontend, the list is already filtered.

openaev-api/src/main/java/io/openaev/api/url_access_token/UrlAccessTokenApi.java
gabriel-peze: It will affect all the returned errors from backend no ? :/
And do you know if this could handle the translations ?
From my point of view it looks like it will affect a lot of things :/

openaev-api/src/main/java/io/openaev/aop/UrlAccessControlAspect.java
gabriel-peze: You're right, i added a tests to verify if the userId is null before modifying it.

gabriel-peze:
Hello @antoinemzs 

Is there anything you want me to do on this topic, or it could be validated ?

antoinemzs (approved):

PR #5850, opened 2026-05-20
[backend] fix: replace raw JDBC connections with JdbcTemplate to prevent PostgreSQL deadlocks

### LINKED ISSUE #5914: [Bug] PostgreSQL deadlocks on concurrent inject execution callbacks
## Description

In SaaS Kubernetes deployments, concurrent agent callbacks on inject executions trigger PostgreSQL deadlocks, causing execution traces to fail silently.

## Root Cause

`ExecutionTraceRepositoryHelper` uses `dataSource.getConnection()` to open raw JDBC connections that bypass the Spring `@Transactional` context. When called from within a transactional method (e.g., `InjectExecutionService.handleInjectExecutionCallback()`), this creates two concurrent connections touching the same rows — a classic deadlock setup.

Contributing factors:
- `LockAspect` uses JVM-local Guava `Striped<Lock>` — no cross-pod protection in multi-pod deployments
- `EntityManager.merge()` + raw JDBC on the same tables within the same call chain

## Impact

- Deadlocks on `injects`, `injects_statuses`, and `execution_traces` tables

## Fix

Replace raw `dataSource.getConnection()` with `JdbcTemplate`, which uses `DataSourceUtils.getConnection()` and participates in the existing Spring transaction — eliminating the multi-connection deadlock.

RomuDeuxfois:
### SaaS Alert — PostgreSQL deadlocks

---

### The problem: mixed JPA Transaction + raw JDBC connections

`ExecutionTraceRepositoryHelper` uses `dataSource.getConnection()` to open **independent JDBC connections** that bypass the Spring `@Transactional` context. When called from within a `@Transactional` method (e.g., `InjectExecutionService.handleInjectExecutionCallback()`), this creates **two concurrent database connections touching the same rows** — a classic deadlock setup.

#### Deadlock scenario (two agents sending callbacks simultaneously)

```
Thread 1 (Agent A, Inject X)              Thread 2 (Agent B, Inject Y)
──────────────────────────                 ──────────────────────────
Conn A (JPA): LOCK injects row X          Conn A' (JPA): LOCK injects row Y
Conn A: LOCK injects_statuses X           Conn A': LOCK injects_statuses Y
Conn B (raw JDBC): INSERT traces          Conn B' (raw JDBC): INSERT traces
Conn C (raw JDBC): UPDATE injects X       Conn C' (raw JDBC): UPDATE injects Y
Conn A: FLUSH → UPDATE injects X          Conn A': FLUSH → UPDATE injects Y
  ↑ Conn A waits for Conn C                 ↑ Conn A' waits for Conn C'
     Conn C waits for Conn A                    Conn C' waits for Conn A'
     💀 DEADLOCK                                💀 DEADLOCK
```

#### Contributing factors

1. **`LockAspect` uses in-memory `Guava Striped<Lock>`** — these are JVM-local locks. In SaaS Kubernetes deployments with **multiple pods**, each pod has its own lock set → **zero cross-pod protection**
2. **`internal instances` also has CrashLoopBackoff issues** — rapid restart loops of collectors/injectors amplify concurrent callback patterns
3. **`EntityManager.merge()` + raw JDBC on same tables** — `InjectStatusService.updateInjectStatus()` calls `entityManager.merge(injectStatus)` (Connection A), then immediately calls `executionTraceRepositoryHelper.saveExecutionTrace()` which opens Connection B

---

### What changed: `ExecutionTraceRepositoryHelper`

**Before (deadlock-prone):**
```java
@Autowired private DataSource dataSource;

public String saveExecutionTrace(ExecutionTrace executionTrace) {
    try (Connection conn = dataSource.getConnection()) {  // ← NEW connection, outside transaction!
        try (PreparedStatement ps = conn.prepareStatement(INSERT_EXECUTION_TRACE)) {
            // ...
            ps.executeUpdate();
        }
    }
}
```

**After (transaction-safe):**
```java
private final JdbcTemplate jdbcTemplate;

public ExecutionTraceRepositoryHelper(JdbcTemplate jdbcTemplate) {
    this.jdbcTemplate = jdbcTemplate;
}

public String saveExecutionTrace(ExecutionTrace executionTrace) {
    jdbcTemplate.execute(INSERT_EXECUTION_TRACE, (PreparedStatement ps) -> {
        // ...
        // ps.getConnection() returns the transaction-bound connection
        ps.setArray(10, ps.getConnection().createArrayOf("text", ...));
        return ps.executeUpdate();
    });
}
```

### Why `JdbcTemplate` fixes the deadlock

`JdbcTemplate` internally uses `DataSourceUtils.getConnection(dataSource)` which:
1. **Checks if a Spring-managed transaction** already has a connection bound to the current thread
2. **If yes** → returns that **same connection** (participates in the `@Transactional` context)
3. **If no** → opens a new connection (standalone behavior, same as before)

This means all three SQL operations (`INSERT execution_traces`, `UPDATE injects_statuses`, `UPDATE injects`) now execute on the **same JDBC connection** as the JPA `@Transactional` method — eliminating the multi-connection deadlock.

### Changes to all three methods

| Method | Before | After |
|---|---|---|
| `saveExecutionTrace()` | `dataSource.getConnection()` + manual `PreparedStatement` | `jdbcTemplate.execute(sql, PreparedStatementCallback)` — uses callback form to access `ps.getConnection().createArrayOf()` for PostgreSQL `text[]` column |
| `updateInjectStatus()` | `dataSource.getConnection()` + manual `PreparedStatement` | `jdbcTemplate.update(sql, args...)` — simple parameterized update |
| `updateInjectUpdateDate()` | `dataSource.getConnection()` + manual `PreparedStatement` | `jdbcTemplate.update(sql, args...)` — simple parameterized update |

### Additional improvements

- **Constructor injection** instead of `@Autowired` field injection (Spring best practice)
- **`DataAccessException`** (unchecked, Spring-managed) instead of wrapping `SQLException` in `RuntimeException`
- **Javadoc updated** to explain why `JdbcTemplate` is used and the deadlock prevention rationale

---

## What this does NOT fix (future work)

1. **`LockAspect` is still JVM-local** — for true multi-pod safety, PostgreSQL advisory locks (`pg_advisory_xact_lock`) should replace or complement the Guava `Striped<Lock>`. This is a separate architectural change.
2. **No new tests added** — `ExecutionTraceRepositoryHelper` has no existing unit tests. A proper test would require a Spring integration test with a real PostgreSQL instance to verify transaction participation. This should be tracked as a follow-up.

---

## Testing

- [ ] CI passes (compile, unit tests, E2E, CodeQL)
- [ ] Manual verification: deploy to a staging namespace and trigger concurrent inject callbacks

## Risk Assessment

| Risk | Level | Mitigation |
|---|---|---|
| Behavioral change | 🟢 Low | `JdbcTemplate` with `DataSourceUtils` is the standard Spring pattern. The SQL statements are identical. |
| Performance impact | 🟢 None | Removing `getConnection()` overhead actually reduces connection pool pressure |
| Regression risk | 🟢 Low | Single file changed, no API surface modification, no schema change |
| Breaking change | 🟢 None | Internal implementation detail, no public API impact |


laugiov:
Nice job on the root cause analysis, the deadlock path is well explained and the JdbcTemplate switch makes sense.

I was looking at the CI run though and I'm not sure if it's related to the PR or not. There are 20 timeouts on `API Tests (rest-2)`, all in `InjectApiTest$FindingProcessingHandlingTest`, and they seem to sit on the path this PR touches. Same shard looks green on main from what I saw, so I was wondering if it could be linked.

I was also wondering about `InjectStatusService.updateInjectStatus` around line 196. The JDBC insert is between an `addTrace` on the JPA collection and an `entityManager.merge`. Before, the insert was on a separate autocommit connection so it was committed before the merge ran. Now both share the same tx, and the cascade has to handle a trace whose id was set after the insert. Maybe that's what's causing the `hasInjectStatusTrace` timeouts? 

Happy to dig further if useful.

laugiov (dismissed):

RomuDeuxfois:
@laugiov I tested it and it's ok, but I loose your approval with a force push..

laugiov (approved):

PR #5868, opened 2026-05-20
[backend] feat(signature): add signature stub processor (#5446)

### LINKED ISSUE #5446: [signature redesign]: Implement stub processor in backend
## Context
Just implement a stub processor with no behaviour in the backend so that we can create a ContractOutputElement of the correct type.

The stub processor should be linked to an enum option in `ContractOutputType` so that it is usable in `ContractOutputElement`, and importantly, available for use in `pyoaev`

Enum option: `ExpectationSignature`

## DoD
* There is a new OutputProcessor class for Signature
* There is a new enum option in ContractOutputType of name Signature

gabriel-peze:
### Proposed changes

* Add a Stub Signature Processor to allow pyoav to call the platform using this processor

### Testing Instructions

1. Run pyoaev using a signature processor, everything should work

### Related issues

* Closes #5446

### DIFF
--- a/openaev-model/src/main/java/io/openaev/database/model/ContractOutputType.java
+++ b/openaev-model/src/main/java/io/openaev/database/model/ContractOutputType.java
@@ -68,8 +68,8 @@ public enum ContractOutputType {
   @JsonProperty("asset")
   Asset("asset"),
 
-  @JsonProperty("signature")
-  Signature("signature");
+  @JsonProperty("expectation_signature")
+  ExpectationSignature("expectation_signature ");
 
   private final String label;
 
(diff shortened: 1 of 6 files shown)


openaev-model/src/main/java/io/openaev/database/model/ContractOutputType.java
```diff
@@ -66,7 +66,10 @@ public enum ContractOutputType {
 
   @Hidden
   @JsonProperty("asset")
-  Asset("asset");
+  Asset("asset"),
+
+  @JsonProperty("signature")
+  Signature("signature");
```
antoinemzs: Sorry the issue listed `ExpectationSignature` as the value, so that it's explicit

antoinemzs (dismissed):
Good stuff, please just adjust the enum option name thanks

antoinemzs (approved):

PR #5870, opened 2026-05-20
[backend] feat(service-account): creation of one user service by tenant and datapack. (#5769)

### LINKED ISSUE #5769: [Service Accounts] US.2-T.1 — Auto-provision service account on tenant creation with correct group/role/capabilities & use it on the OpenAEV agent
## 📖 User Story

**As a** platform operator,
**When** I create a tenant,
**Then** a service account is automatically provisioned and assigned with the correct group/role/capabilities (US.1),
**So that** OpenAEV agent technical integrations can operate immediately without manual account setup.

## 🎯 Acceptance Criteria

- [x] A **service account is automatically created** every time a new tenant is provisioned
- [x] The service account is **assigned the correct group** aligned with the tenant
- [x] The service account is **assigned the correct role** with appropriate capabilities (including `Agent Runtime Access` from US.1)
- [x] This service account is used for OpenAEV agent

AC1:
GIVEN an authorized account for installing OpenAEV Agent
WHEN I display the agents installation screen in OpenAEV
THEN I can see a Installation link to install command to copy/paste onto target systems with the related service account token
AND the generated install command contains the service account API token (not the current logged in admin account API

AC2:
GIVEN an inject
WHEN the run command is generated by the backend
THEN the API token of the respective related service account of each target agent is used in the generated execution command

EvaE-Filigran:
### Proposed changes
* Add `ServiceAccountPrivilegeService` to manage the lifecycle of the service account per tenant:
  * Creates (or reuses) a well-known role "Service integration" with `AGENT_RUNTIME_ACCESS` + `ACCESS_DOCUMENTS` capabilities
  * Creates (or updates) a well-known group "Service integration" bound to that role
  * Creates an internal user `service-{tenantId}@openaev.invalid` with a generated API token and assigns it to the group; if the user already exists without a token, a new token is generated and reused
* Add `V20260518_Service_Account` DataPack to bootstrap the service account automatically on tenant initialisation

* Reserved name protection:
  * Add ReservedNameValidator to block mutations on reserved role names, group names, and service-account email patterns
  * Split RoleService into public (createRole/updateRole) and internal (createRoleInternal/updateRoleInternal) methods — PrivilegeService uses the internal path to bypass the guard
  * Apply validation across TenantGroupService, UserService, and TenantUserService on all create/update/delete operations
image
image
image

### Testing Instructions
1. Start the application against a fresh tenant — verify the DataPack `V20260518_Service_Account` runs and creates:
   * A role named **"Service integration"** with capabilities `AGENT_RUNTIME_ACCESS` and `ACCESS_DOCUMENTS`
   * A group named **"Service integration"** linked to that role
   * A user `service-{tenantId}@openaev.invalid` with a valid API token and membership in that group
2. Re-run the DataPack (or restart) — verify it is idempotent: no duplicate role, group, or user is created
3. Manually delete the user token (set to `null`) and re-run — verify a new token is generated for the existing user
4. Run the tests: ServiceAccountPrivilegeServiceTest

2. Reserved name protection:
Try creating a role named "Service integration" or "Process STIX" — both should be blocked
* On the "Service integration" role, verify the following are blocked:
   * Edit name/capabilities
   * Delete

* On the "Service integration" group, verify the following are blocked:
   * Edit information
   * Add/remove a grant (permission)
   * Add/remove a user
   * Add/remove a role
   * Delete

* On any non-reserved group, verify the following are blocked:
  *Adding a user with email service-{tenantId}@openaev.invalid
  *Adding the "Service integration" role

### Related issues
* Closes #5769 

### Checklist

### Further comments
This PR complements the RBAC changes from the previous PR (introduction of `AGENT_RUNTIME_ACCESS`) by providing the bootstrapping layer that provisions a ready-to-use service account on every tenant.

The service account is intentionally minimal and non-interactive: it has no last name, is not admin, and its email follows a deterministic pattern (`service-{tenantId}@openaev.invalid`) to ensure uniqueness per tenant while remaining easily identifiable.

The token regeneration path (user exists but no token) handles cases where the account may have been partially provisioned in a previous failed run, avoiding orphaned users.

The `UnsupportedOperationException` thrown on duplicate group detection is a hard guard against data inconsistency — if two groups with the same name exist, the DataPack will fail fast and log the error rather than silently picking one.


openaev-api/src/main/java/io/openaev/datapack/packs/V20260518_Service_Account.java
```diff
@@ -0,0 +1,31 @@
...
+import io.openaev.datapack.DataPack;
+import io.openaev.service.DataPackService;
+import io.openaev.service.account.ServiceAccountPrivilegeService;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Component;
+
+@Component
+@Slf4j
+public class V20260518_Service_Account extends DataPack {
+  private final ServiceAccountPrivilegeService privilegeService;
```
antoinemzs, line 13: Datapack should be covered with auto tests to ensure it does what it is supposed to do 🙏 Please rely on precedents for pointers on how to do this

openaev-api/src/main/java/io/openaev/service/account/ServiceAccountPrivilegeService.java
```diff
@@ -0,0 +1,128 @@
...
+  public Optional<User> getUserServiceAccountByTenant(String tenantId) {
+    String email = SERVICE_EMAIL_PATTERN.formatted(tenantId);
+
+    return userService.findByEmailIgnoreCase(email);
+  }
+
+  private Role createWellKnownRole(String tenantId) {
+    List<Role> processRoles = roleService.findAll(tenantId);
+    Optional<Role> processRole =
+        processRoles.stream().filter(role -> role.getName().equals(SERVICE_ROLE_NAME)).findFirst();
```
antoinemzs: Role name is both not UNIQUE and editable, thus this might create a possible shift if someone tampers with the role name or creates one with the same name.

The concept of "well known" resides on a static, reserved constant value. In the previous implementation (created for the OCTI connector's use), we relied on an ID, which I understand is not sufficient in the multitenant paradigm.

We have an implementation for "well known" tags, which can be a source of inspiration: we proscribe the creation or renaming of tags to certain reserved tag names, but allow creating them in the backend for our reserved use.

https://github.com/OpenAEV-Platform/openaev/blob/412e99e8dce42f7682ac6e22320562ee831a6a02/openaev-model/src/main/java/io/openaev/database/model/TagRule.java#L27-L32

openaev-api/src/main/java/io/openaev/service/account/ServiceAccountPrivilegeService.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+    return processRole.get();
+  }
+
+  private Group createWellKnownGroupWithRole(Role role, String tenantId) {
+    List<Group> processGroups = tenantGroupService.findAllByTenantId(tenantId);
+
+    Optional<Group> groupByTenant =
+        processGroups.stream()
+            .filter(group -> group.getName().equals(SERVICE_GROUP_NAME))
```
antoinemzs: Same comment on well-known group, "by name" risks breaking if the groups are tampered with or duplicated.

openaev-api/src/test/java/io/openaev/service/account/ServiceAccountPrivilegeServiceTest.java
antoinemzs, line 1: Would like to see integration tests, we are currently not asserting that the user is created correctly by retrieving the object and comparing the attributes.

antoinemzs (changes requested):
Fragility in the "well-known" implementation using the entities' `name` attribute which can be edited or duplicated, this should probably follow the same pattern as for the well-known tags.

Please consider integration tests for both the new service and the datapack.

Thanks!

openaev-api/src/main/java/io/openaev/datapack/packs/V20260518_Service_Account.java
EvaE-Filigran: tests added

openaev-api/src/main/java/io/openaev/service/account/ServiceAccountPrivilegeService.java
```diff
@@ -0,0 +1,128 @@
...
+        groupByTenant
+            .map(
+                group ->
+                    tenantGroupService.updateGroupInfoWithRoles(
+                        group, input, new ArrayList<>(List.of(role))))
+            .or(
+                () ->
+                    Optional.of(
+                        tenantGroupService.createGroupWithRole(
+                            null, input, new ArrayList<>(List.of(role)))));
```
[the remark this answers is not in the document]
antoinemzs: Someone tell the bot that this is not true please :skull: 

This is what happens when passing `null`:
```java
  /**
   * Generates an ID for the entity, preserving any pre-assigned value.
   *
   * @param session the Hibernate session
   * @param owner the entity instance being persisted
   * @param currentValue the current ID value (unused)
   * @param eventType the event type triggering generation
   * @return the existing ID if set, otherwise a newly generated UUID string
   */
  @Override
  public Object generate(
      SharedSessionContractImplementor session,
      Object owner,
      Object currentValue,
      EventType eventType) {
    final String id;
    if (owner instanceof Base) {
      id = ((Base) owner).getId();
    } else {
      id = null;
    }

    return id != null ? id : UUID.randomUUID().toString();
  }
```

openaev-api/src/main/java/io/openaev/service/account/ServiceAccountPrivilegeService.java
```diff
@@ -0,0 +1,128 @@
...
+  private Role createWellKnownRole(String tenantId) {
+    List<Role> processRoles = roleService.findAll(tenantId);
+    Optional<Role> processRole =
+        processRoles.stream().filter(role -> role.getName().equals(SERVICE_ROLE_NAME)).findFirst();
+
+    if (processRole.isEmpty()) {
+      processRole =
+          Optional.of(
+              roleService.createRole(
+                  null, SERVICE_ROLE_NAME, SERVICE_ROLE_DESCRIPTION, SERVICE_ROLE_CAPABILITIES));
```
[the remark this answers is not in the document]
EvaE-Filigran: ID generated.

openaev-api/src/main/java/io/openaev/service/account/ServiceAccountPrivilegeService.java
```diff
@@ -0,0 +1,128 @@
...
+        userService.saveUser(user);
+      }
+    } else {
+      // No user exists — create one
+      User user =
+          userService.createInternalUser(email, SERVICE_FIRSTNAME, null, false, getNewTokenAgent());
+      user.setGroups(new ArrayList<>(List.of(group)));
+      tenantUserService.attachToTenant(user.getId(), tenantId);
+      userService.saveUser(user);
+    }
```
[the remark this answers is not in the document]
EvaE-Filigran: if the service account already exists with a valid token, we consider it to have already been provisioned correctly.

openaev-api/src/main/java/io/openaev/service/account/ServiceAccountPrivilegeService.java
```diff
@@ -0,0 +1,128 @@
...
+            user.getEmail());
+
+        applyAgentAttributes(user, email, group);
+
+        user.setTokens(
+            new ArrayList<>(List.of(userService.createUserToken(user, getNewTokenAgent()))));
+
+        tenantUserService.attachToTenant(user.getId(), tenantId);
+        userService.saveUser(user);
+      }
```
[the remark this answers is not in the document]
EvaE-Filigran: fix done

openaev-api/src/main/java/io/openaev/service/account/ServiceAccountPrivilegeService.java
```diff
@@ -0,0 +1,128 @@
...
+        processRoles.stream().filter(role -> role.getName().equals(SERVICE_ROLE_NAME)).findFirst();
+
+    if (processRole.isEmpty()) {
+      processRole =
+          Optional.of(
+              roleService.createRole(
+                  null, SERVICE_ROLE_NAME, SERVICE_ROLE_DESCRIPTION, SERVICE_ROLE_CAPABILITIES));
+    }
+    return processRole.get();
+  }
```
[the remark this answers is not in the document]
EvaE-Filigran: fix done

openaev-api/src/test/java/io/openaev/service/account/ServiceAccountPrivilegeServiceTest.java
```diff
@@ -0,0 +1,251 @@
...
+    when(userService.createInternalUser(any(), any(), any(), anyBoolean(), any()))
+        .thenReturn(mockUser);
+
+    // act
+    privilegeService.ensurePrivilegedUserExists(TENANT_ID);
+
+    // assert
+    verify(tenantGroupService).createGroupWithRole(isNull(), any(), any());
+    verify(tenantGroupService, never()).updateGroupInfoWithRoles(any(), any(), any());
+  }
```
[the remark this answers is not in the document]
EvaE-Filigran: id generated

openaev-api/src/test/java/io/openaev/service/account/ServiceAccountPrivilegeServiceTest.java
```diff
@@ -0,0 +1,251 @@
...
+    privilegeService.ensurePrivilegedUserExists(TENANT_ID);
+
+    // assert
+    verify(roleService)
+        .createRole(
+            isNull(),
+            eq(SERVICE_ROLE_NAME),
+            eq(SERVICE_ROLE_DESCRIPTION),
+            eq(SERVICE_ROLE_CAPABILITIES));
+  }
```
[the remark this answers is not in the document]
EvaE-Filigran: id generated

openaev-api/src/test/java/io/openaev/service/account/ServiceAccountPrivilegeServiceTest.java
```diff
@@ -0,0 +1,251 @@
...
+
+    // assert
+    verify(userService, never()).createInternalUser(any(), any(), any(), anyBoolean(), any());
+    verify(userService).createUserToken(eq(mockUser), anyString());
+    verify(tenantUserService).attachToTenant(any(), eq(TENANT_ID));
+    verify(userService).saveUser(mockUser);
+    assertThat(mockUser.getTokens()).hasSize(1);
+  }
+
+  @Test
```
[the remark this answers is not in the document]
EvaE-Filigran: set list token empty instead of null

openaev-api/src/main/java/io/openaev/service/account/ServiceAccountPrivilegeService.java
EvaE-Filigran: Done with ReservedNameValidator

openaev-api/src/main/java/io/openaev/service/account/ServiceAccountPrivilegeService.java
EvaE-Filigran: Done with ReservedNameValidator

openaev-api/src/test/java/io/openaev/service/account/ServiceAccountPrivilegeServiceTest.java
EvaE-Filigran: Done

openaev-api/src/test/java/io/openaev/datapack/packs/ServiceAccountPackTest.java
```diff
@@ -0,0 +1,111 @@
...
+import org.mockito.Mock;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.transaction.annotation.Transactional;
+
+@SpringBootTest
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@DisplayName("Service Account DataPack process tests")
+@Transactional
+public class ServiceAccountPackTest extends IntegrationTest {
```
RomuDeuxfois: **nitpick**: Seems easier for me to retrieve the test with V20260518_Service_AccountTest

openaev-api/src/main/java/io/openaev/service/account/ServiceAccountPrivilegeService.java
```diff
@@ -0,0 +1,122 @@
...
+
+  private void applyAgentAttributes(User user, String email, Group group) {
+    user.setFirstname(SERVICE_FIRSTNAME);
+    user.setLastname(null);
+    user.setEmail(email);
+    user.setAdmin(false);
+    user.setGroups(new ArrayList<>(List.of(group)));
+  }
+
+  private String getUUIDFromName(String name, String tenantId) {
```
RomuDeuxfois: **todo**: extract this to be reused by PrivilegeService line 119

openaev-api/src/main/java/io/openaev/service/account/ServiceAccountPrivilegeService.java
```diff
@@ -0,0 +1,122 @@
...
+      if (!userService.userHasToken(user.getId())) {
+        // Email-matched user exists but has no token — reuse and attach token
+        log.warn(
+            "User with email {} already exists, but no token found. Reusing existing user.",
+            user.getEmail());
+
+        applyAgentAttributes(user, email, group);
+
+        user.setTokens(
+            new ArrayList<>(List.of(userService.createUserToken(user, getNewTokenAgent()))));
```
RomuDeuxfois: **todo**: you can use createUserToken(user) direclty

openaev-api/src/main/java/io/openaev/service/account/ServiceAccountPrivilegeService.java
```diff
@@ -0,0 +1,122 @@
...
+    } else {
+      return tenantGroupService.createGroupWithRole(groupId, input, roles, tenantId);
+    }
+  }
+
+  private String getNewTokenAgent() {
+    return UUID.randomUUID().toString();
+  }
+
+  private void applyAgentAttributes(User user, String email, Group group) {
```
RomuDeuxfois: **todo**: extract this to be reused by PrivilegeService line 87

openaev-api/src/main/java/io/openaev/service/account/ServiceAccountPrivilegeService.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.UUID;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
+
+@Service
+@RequiredArgsConstructor
+@Slf4j
+public class ServiceAccountPrivilegeService {
```
RomuDeuxfois: **nitpick**: I added two comments about sharing code between `PrivilegeService` and this service.
They have the same behavior, so I think we can mutualize quite a bit. I’ll let you check what can be done easily.

openaev-api/src/main/java/io/openaev/service/RoleService.java
```diff
@@ -48,7 +49,22 @@ public Role createRole(
...
       String tenantId) {
+    ReservedNameValidator.validateRoleName(roleName);
+    return createRoleInternal(id, roleName, roleDescription, capabilities, tenantId);
+  }
+
+  /**
+   * Internal method for system-managed roles (e.g. service accounts). Bypasses reserved name
+   * validation.
+   */
+  public Role createRoleInternal(
```
RomuDeuxfois, line 60: **praise**: I like the idea to have one internal method and one for the API.
Should be great to have the same logic for user if you have the time to do it.

openaev-api/src/main/java/io/openaev/service/TenantGroupService.java
RomuDeuxfois: **nitpick**: could be rename to createInternalGroupWithRole

openaev-api/src/test/java/io/openaev/service/account/ServiceAccountPrivilegeServiceTest.java
```diff
@@ -0,0 +1,270 @@
+package io.openaev.service.account;
```
RomuDeuxfois, line 1: **question**: I am not sure to understand the goal of this test since you already cover the whole things in openaev-api/src/test/java/io/openaev/datapack/packs/ServiceAccountPackTest.java no ?

openaev-api/src/main/java/io/openaev/service/RoleService.java
```diff
@@ -99,12 +115,24 @@ public Role updateRole(
...
+      @NotBlank final String roleName,
+      @NotBlank final String roleDescription,
+      @NotNull final Set<Capability> capabilities) {
     Capability.validateForTenantRole(capabilities);
     String tenantId = TenantContext.getCurrentTenant();
     Role role =
         roleRepository
             .findByIdAndTenantId(roleId, tenantId)
             .orElseThrow(() -> new ElementNotFoundException("Role not found with id: " + roleId));
+    ReservedNameValidator.validateRoleName(role.getName());
```
RomuDeuxfois: **todo**: we can't let the validator here. It can trigger an error on ServiceAccountPrivilegeService.createWellKnownRole line 85, and I think this is not what we want.

openaev-api/src/main/java/io/openaev/service/UserService.java
```diff
@@ -265,6 +271,7 @@ public java.util.Set<Tag> resolveTags(List<String> tagIds) {
   @Transactional(rollbackFor = Exception.class)
   public void delete(String userId) {
     User existing = user(userId);
+    ReservedNameValidator.validateUserEmailPattern(existing.getEmail());
     if (existing == null) {
       throw new EntityNotFoundException("User not found: " + userId);
     }
```
RomuDeuxfois: **nitpick**; this is not part of your code, but this code is useless since we already throw line 273

openaev-api/src/main/java/io/openaev/service/UserService.java
```diff
@@ -212,7 +214,11 @@ public Page<UserOutput> search(SearchPaginationInput searchPaginationInput) {
 
   @Transactional(rollbackFor = Exception.class)
   public User updateUser(String userId, UserInput input) {
+    // Check if new email is reserved
+    ReservedNameValidator.validateUserEmailPattern(input.email());
     User existing = user(userId);
+    // Check if previews email is reserved
```
RomuDeuxfois:
```suggestion
    // Check if previous email is reserved
```

openaev-api/src/main/java/io/openaev/service/TenantGroupService.java
```diff
@@ -137,11 +144,15 @@ public Group updateGroupInfoWithRoles(
...
 
   public Group updateGroup(String groupId, TenantGroupCreateInput input) {
+    // Check if new name is reserved
+    ReservedNameValidator.validateGroupName(input.getName());
     String tenantId = TenantContext.getCurrentTenant();
     Group group =
         groupRepository
             .findByIdAndTenantId(groupId, tenantId)
             .orElseThrow(ElementNotFoundException::new);
+    // Check if previews name is reserved
```
RomuDeuxfois:
```suggestion
    // Check if previous name is reserved
```

openaev-api/src/main/java/io/openaev/service/account/ServiceAccountPrivilegeService.java
```diff
@@ -0,0 +1,122 @@
...
+  public static final String SERVICE_EMAIL_PATTERN = "service-%s@openaev.invalid";
+  private static final String SERVICE_FIRSTNAME = "discrete";
```
RomuDeuxfois: **todo**: I think is better to call it "Service"

openaev-api/src/main/java/io/openaev/service/account/ReservedNameValidator.java
```diff
@@ -0,0 +1,68 @@
...
+import io.openaev.rest.exception.BadRequestException;
+import java.util.List;
+import java.util.Set;
+import java.util.regex.Pattern;
+
+/**
+ * Validates that user-provided names do not conflict with system-reserved names used by service
+ * accounts.
+ */
+public final class ReservedNameValidator {
```
RomuDeuxfois: **suggestion**: having a new column is_system_managed direclty on DB can be easier to handle. What do you think ?

RomuDeuxfois:
Tests 🆗

RomuDeuxfois:
This not relativ to your code, but I try to remove the service account from the player interface and I have a TransientObjectException

Can you change the deletePlayer mechanism from this
  @DeleteMapping({PLAYER_URI + "/{userId}", TENANT_PLAYER_URI + "/{userId}"})
  @AccessControl(
      resourceId = "#userId",
      actionPerformed = Action.DELETE,
      resourceType = ResourceType.PLAYER)
  public void deletePlayer(@PathVariable String userId) {
    sessionManager.invalidateUserSession(userId);
    userRepository.deleteById(userId);
  }

  
  To this ?
    @DeleteMapping({PLAYER_URI + "/{userId}", TENANT_PLAYER_URI + "/{userId}"})
  @AccessControl(
      resourceId = "#userId",
      actionPerformed = Action.DELETE,
      resourceType = ResourceType.PLAYER)
  public void deletePlayer(@PathVariable String userId) {
    sessionManager.invalidateUserSession(userId);
    userRepository.deleteByIdNative(userId);
  }

thank you !

openaev-api/src/test/java/io/openaev/datapack/packs/ServiceAccountPackTest.java
EvaE-Filigran: done

openaev-api/src/main/java/io/openaev/service/account/ServiceAccountPrivilegeService.java
EvaE-Filigran: done

openaev-api/src/main/java/io/openaev/service/account/ServiceAccountPrivilegeService.java
EvaE-Filigran: Creation of an abstract class extended by `PrivilegeService` and `ServiceAccountPrivilegeService`.

openaev-api/src/main/java/io/openaev/service/RoleService.java
EvaE-Filigran: already done with : createInternalUser

openaev-api/src/main/java/io/openaev/service/TenantGroupService.java
EvaE-Filigran: done

openaev-api/src/test/java/io/openaev/service/account/ServiceAccountPrivilegeServiceTest.java
EvaE-Filigran: Both tests serve different purposes:
- ServiceAccountPrivilegeServiceTest is a unit test, it verifies the internal logic of ServiceAccountPrivilegeService in isolation (mocked dependencies), focusing on each branch: create vs reuse role, create vs update group, create user vs reuse vs skip. It also acts as a safety net if the constants (SERVICE_ROLE_NAME, SERVICE_EMAIL_PATTERN, etc.) are edited, the test will catch any regression at the method level.
- V20260518_Service_AccountTest is an integration test, it verifies the DataPack behaviour end-to-end against a real database: idempotency (no duplicates on re-run), error handling (DataPack not marked as processed if the service throws), and that the final state in DB is correct.

They are complementary: one tests how it works, the other tests that it works correctly as a whole.

openaev-api/src/main/java/io/openaev/service/RoleService.java
EvaE-Filigran: You are right, the fix has been done in chunk_3.

openaev-api/src/main/java/io/openaev/service/UserService.java
EvaE-Filigran: done

openaev-api/src/main/java/io/openaev/service/UserService.java
EvaE-Filigran: done

openaev-api/src/main/java/io/openaev/service/TenantGroupService.java
EvaE-Filigran: done

openaev-api/src/main/java/io/openaev/service/account/ServiceAccountPrivilegeService.java
EvaE-Filigran: done

openaev-api/src/main/java/io/openaev/service/account/ServiceAccountPrivilegeService.java
EvaE-Filigran: done

openaev-api/src/main/java/io/openaev/service/account/ServiceAccountPrivilegeService.java
EvaE-Filigran: change: SERVICE_FIRSTNAME = "discrete" -> SERVICE_FIRSTNAME = "service";

openaev-api/src/main/java/io/openaev/service/account/ReservedNameValidator.java
EvaE-Filigran: It's another architectural point of view and I considered it. However, the context here is an MVP, we need a dedicated token to communicate with external systems. Touching the data model at this stage, before we've aligned on the final design, felt like the wrong trade-off.
`is_system_managed` would also just be a transitional solution, not the target architecture. So I preferred to keep the data model untouched and isolate the logic in ReservedNameValidator instead, it protects the integrity of the system (reserved role names, group names, user email patterns).
In the longer term, I think the right architecture is a dedicated ServiceAccount entity with proper JWT token support, completely decoupled from the current user model. That would be a much cleaner design than patching it either way.
The current solution is intentionally transitional. I'd rather we align on the target design before investing further in either direction.

EvaE-Filigran:
> This not relativ to your code, but I try to remove the service account from the player interface and I have a TransientObjectException
> 
> Can you change the deletePlayer mechanism from this @DeleteMapping({PLAYER_URI + "/{userId}", TENANT_PLAYER_URI + "/{userId}"}) @accesscontrol( resourceId = "#userId", actionPerformed = Action.DELETE, resourceType = ResourceType.PLAYER) public void deletePlayer(@PathVariable String userId) { sessionManager.invalidateUserSession(userId); userRepository.deleteById(userId); }
> 
> To this ? @DeleteMapping({PLAYER_URI + "/{userId}", TENANT_PLAYER_URI + "/{userId}"}) @accesscontrol( resourceId = "#userId", actionPerformed = Action.DELETE, resourceType = ResourceType.PLAYER) public void deletePlayer(@PathVariable String userId) { sessionManager.invalidateUserSession(userId); userRepository.deleteByIdNative(userId); }
> 
> thank you !

This fix doesnt work I clean this code using userService.delete(userId);

openaev-api/src/test/java/io/openaev/datapack/packs/V20260518_Service_AccountTest.java
antoinemzs, line 1: Great, thanks for these!

FWIW you don't need to test whether the datapack processor works, you just need to ensure that after the specific datapack runs you have the expected data patterns in the database.

These tests should cover that the datapack will always produce the correct data even after it's been refactored because the internal API has changed for example.

antoinemzs (dismissed):
OK on my end thanks

openaev-api/src/main/java/io/openaev/service/RoleService.java
antoinemzs: Let's think about merging order then, if there is a known issue in this PR then it shouldn't be merged without a fix. WDYT?

RomuDeuxfois (dismissed):

openaev-api/src/main/java/io/openaev/service/RoleService.java
EvaE-Filigran: Cherry pick done

RomuDeuxfois (approved):

PR #5885, opened 2026-05-20
feat(audit-log): critical log write failure (#5793)

### LINKED ISSUE #5793: US.8 -> Critical Log Write Failure — Stack Halt
## Context

As a platform administrator, I want the OpenAEV stack to stop if a critical-level audit log fails to be written, so that the platform never operates in a state where critical security events are silently lost.

a19836:
As a platform administrator, I want the OpenAEV stack to stop if a critical-level audit log fails to be written, so that the platform never operates in a state where critical security events are silently lost.

### Proposed changes

* Add a new method called prepareLogFailure that simply calls the spring exit method

### Testing Instructions

1. Currently, since the search engine does not yet implement any engine throwing an exception, this method is being called every time the audit logging process runs, causing it to fail and stop the application.
2. To reproduce the issue, simply go to the platform and update something. For example, an organization name.

### Related issues

* #5793

### DIFF
--- a/openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
+++ b/openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
@@ -12,7 +12,6 @@ import io.openaev.service.LogService;
 import io.openaev.utils.log.LogUtils;
 import java.lang.annotation.Annotation;
 import java.util.UUID;
-import java.util.function.BiConsumer;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.aspectj.lang.JoinPoint;
@@ -133,30 +132,15 @@ public class AccessControlAuditLogAspect {
       JsonNode inputNode = getInputNode(joinPoint, eventScope);
       JsonNode signatureNode = getMethodSignature(joinPoint, inputNode);
 
-      BiConsumer<Boolean, Throwable> logCompletion =
-          (success, throwable) -> {
-            if (throwable != null || (success != null && !success)) {
-              log.warn(
-                  "[AUDIT] Failed to log access control event for {}.{}", resourceType, eventScope);
-              if (throwable != null) {
-                log.warn(LOG_ERROR_MSG, throwable);
-              }
-
-              auditLogger.prepareLogFailure();
-            }
-          };
-
-      auditLogger
-          .logAccessControlEvent(
-              eventScope,
-              eventStatus,
-              resourceType,
-              resourceId,
-              inputNode,
-              outputNode,
-              signatureNode,
-              logUUID)
-          .whenComplete(logCompletion);
+      auditLogger.logAccessControlEvent(
+          eventScope,
+          eventStatus,
+          resourceType,
+          resourceId,
+          inputNode,
+          outputNode,
+          signatureNode,
+          logUUID);
     } catch (Exception ex) {
       log.warn(LOG_ERROR_MSG, ex);
     }
--- a/openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
+++ b/openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
@@ -141,6 +141,8 @@ public class AccessControlAuditLogAspect {
               if (throwable != null) {
                 log.warn(LOG_ERROR_MSG, throwable);
               }
+
+              accessControlAuditLogger.prepareLogFailure();
             }
           };
 
--- a/openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
+++ b/openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
@@ -155,7 +155,6 @@ public class AccessControlAuditLogAspect {
               signatureNode,
               logUUID)
           .whenComplete(logCompletion);
-      ;
     } catch (Exception ex) {
       log.warn(LOG_ERROR_MSG, ex);
     }
--- a/openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
+++ b/openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
@@ -12,6 +12,7 @@ import io.openaev.service.LogService;
 import io.openaev.utils.log.LogUtils;
 import java.lang.annotation.Annotation;
 import java.util.UUID;
+import java.util.function.BiConsumer;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.aspectj.lang.JoinPoint;
@@ -132,15 +133,29 @@ public class AccessControlAuditLogAspect {
       JsonNode inputNode = getInputNode(joinPoint, eventScope);
       JsonNode signatureNode = getMethodSignature(joinPoint, inputNode);
 
-      auditLogger.logAccessControlEvent(
-          eventScope,
-          eventStatus,
-          resourceType,
-          resourceId,
-          inputNode,
-          outputNode,
-          signatureNode,
-          logUUID);
+      BiConsumer<Boolean, Throwable> logCompletion =
+          (success, throwable) -> {
+            if (throwable != null || (success != null && !success)) {
+              log.warn(
+                  "[AUDIT] Failed to log access control event for {}.{}", resourceType, eventScope);
+              if (throwable != null) {
+                log.warn(LOG_ERROR_MSG, throwable);
+              }
+            }
+          };
+
+      auditLogger
+          .logAccessControlEvent(
+              eventScope,
+              eventStatus,
+              resourceType,
+              resourceId,
+              inputNode,
+              outputNode,
+              signatureNode,
+              logUUID)
+          .whenComplete(logCompletion);
+      ;
     } catch (Exception ex) {
       log.warn(LOG_ERROR_MSG, ex);
     }
--- a/openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
+++ b/openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
@@ -142,7 +142,7 @@ public class AccessControlAuditLogAspect {
                 log.warn(LOG_ERROR_MSG, throwable);
               }
 
-              accessControlAuditLogger.prepareLogFailure();
+              auditLogger.prepareLogFailure();
             }
           };
 
--- a/openaev-api/src/main/java/io/openaev/aop/audit_log/AuditLogger.java
+++ b/openaev-api/src/main/java/io/openaev/aop/audit_log/AuditLogger.java
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 import io.openaev.aop.AccessControl;
 import io.openaev.aop.AccessControlAspect;
 import io.openaev.config.ThreadPoolTaskLoggerConfig;
+import io.openaev.config.audit_log.AuditLogProperties;
 import io.openaev.database.model.Action;
 import io.openaev.database.model.ResourceType;
 import io.openaev.service.LogService;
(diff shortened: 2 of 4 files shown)


openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
```diff
@@ -112,17 +112,19 @@ public Object auditAround(ProceedingJoinPoint joinPoint, AccessControl accessCon
...
+                eventStatus,
+                resourceType,
+                resourceId,
+                inputNode,
+                null,
+                entityName,
+                null,
+                sourceId,
+                logUUID)
+            .thenRun(accessControlAuditLogger::prepareLogFailure);
```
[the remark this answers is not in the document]
a19836: N/A anymore

openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogger.java
```diff
@@ -253,6 +271,10 @@ public CompletableFuture<Boolean> logMutationEventIfDifferentSnapshots(
       log.warn("[AUDIT] Audit logging failed (non-blocking): {}", e.getMessage(), e);
     }
 
+    if (!status) {
+      prepareLogFailure();
+    }
```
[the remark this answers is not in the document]
a19836: N/A

openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogger.java
```diff
@@ -219,6 +223,20 @@ else if ("update".equals(eventScope)
...
   }
 
+  @Async("accessControlAuditLoggerExecutor")
+  public void prepareLogFailure() {
+    if (stw) {
+      log.error("[AUDIT] Stop-the-world triggered — shutting down application.");
+
+      int exitCode = SpringApplication.exit(context, () -> 1);
+      System.exit(exitCode);
+
```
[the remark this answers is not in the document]
a19836: That's is on purpose. If in the future we decide to change the behaviour at least we already have everything centralized in this method

openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogger.java
```diff
@@ -219,6 +223,20 @@ else if ("update".equals(eventScope)
...
+      // TODO AUDIT: this is not working and app is still running.
```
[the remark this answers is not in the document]
a19836: N/A anymore. fixed in the new commits

openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogger.java
```diff
@@ -219,6 +223,20 @@ else if ("update".equals(eventScope)
...
+    } else {
+      log.warn("[AUDIT] Audit logging failed, but continuing without blocking the operation.");
+    }
+  }
+
   /** Wraps the audit service call in try/catch — audit must never break the business flow. */
```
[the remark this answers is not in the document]
a19836: done

openaev-api/src/main/java/io/openaev/aop/audit_log/AuditLogger.java
```diff
@@ -46,7 +52,26 @@ public boolean isAuditUnauthorizedLoggingValid() {
...
 
-  /** Wraps the audit service call in try/catch — audit must never break the business flow. */
+  public void prepareLogFailure() {
+    // Halt the application when stop-the-world is enabled and a log failure occurs.
+    if (stw) {
+      log.error("[AUDIT] Stop-the-world triggered — shutting down application.");
+
+      try {
+        int exitCode = SpringApplication.exit(context, () -> 1);
+        System.exit(exitCode);
```
RomuDeuxfois: **todo**: please don't use this and keep the Spring mechanism only.
Here you are bypassing the Spring shutdown hooks (closing @Async, cosing DB, ect)

openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
```diff
@@ -141,6 +141,8 @@ private void logAccessControlEvent(
               if (throwable != null) {
                 log.warn(LOG_ERROR_MSG, throwable);
               }
+
+              auditLogger.prepareLogFailure();
```
RomuDeuxfois: **todo**: better to let the AuditLogger service handle that part.

openaev-api/src/main/resources/application.properties
```diff
@@ -237,13 +237,13 @@ logging.logback.rollingpolicy.file-name-pattern=${LOG_FILE}.-%d{yyyy-MM-dd}.%i
...
 logging.logback.rollingpolicy.max-history=7
 
-
 ################
 # AUDIT LOGGING #
 ################
 # Active transports (comma-separated): console, file, engine
 # Audit logging is enabled when at least one transport is configured; leave empty to disable.
 openaev.audit-logs.transports=
+openaev.audit-logs.stop-the-world=false
```
RomuDeuxfois: **nitpick**: I readed that it's a JVM word
Maybe it's better to use openaev.audit-logs.halt-on-failure=false to not add some confusion around this.
But, is up to you :)

openaev-api/src/main/java/io/openaev/aop/audit_log/AuditLogger.java
```diff
@@ -113,6 +149,10 @@ public CompletableFuture<Boolean> logAccessControlEvent(
       log.warn("[AUDIT] Audit logging failed (non-blocking): {}", e.getMessage(), e);
     }
 
+    if (!status) {
+      prepareLogFailure();
```
RomuDeuxfois, line 167: **question**: Do we should call prepareLogFailure in the catch clause also ?

openaev-api/src/main/java/io/openaev/aop/audit_log/AuditLogger.java
```diff
@@ -79,9 +107,17 @@ public CompletableFuture<Boolean> logAuthEvent(
...
+    }
+
     return CompletableFuture.completedFuture(status);
   }
 
+  /**
+   * Log Mutation events Wraps the audit service call in try/catch — non-blocking by default, but
+   * may terminate the process when stop-the-world audit failure handling is enabled.
+   */
   @Async("taskLoggerExecutor")
```
Dimfacion, line 133: **thoughts :** I don't think we should keep the async with the stack halt in case of failure because if we do have a failure, we can't rollback the transaction anymore since we're async meaning that a change is made without being logged

openaev-api/src/main/resources/application.properties
a19836: Good point. Done in the next commit

openaev-api/src/main/java/io/openaev/aop/audit_log/AuditLogger.java
a19836: prepareLogFailure's code is already inside of a try and catch statement, so there is no need for it. 
Additionally the status might be incorrect, but no exception was raised.

openaev-api/src/main/java/io/openaev/aop/audit_log/AuditLogger.java
```diff
@@ -30,8 +33,11 @@
 @Slf4j
 public class AuditLogger {
 
-  private final AuditRequestValidator auditRequestValidator;
+  @Value("${openaev.audit-logs.stop-the-world:false}")
+  private boolean stw;
```
Dimfacion: **todo :** I think this property might be better in AuditLogProperties.java

a19836: good point. Done in the next commit.

openaev-api/src/main/java/io/openaev/aop/audit_log/AuditLogger.java
a19836: done in  next commit.

openaev-api/src/main/java/io/openaev/aop/audit_log/AuditLogger.java
a19836: The only scenario not accounted for is if the disk space becomes full and logging is no longer possible. In that situation, within prepareLogFailure, the exception is caught, but when we attempt to log the exception to disk, it results in another exception.
Additionally, if this situation occurs, we will have other major issues to address.
I can add try and catch statements around the all the logger code, but it will make the code look ugly and it doesn't make sense either.

openaev-api/src/main/java/io/openaev/aop/audit_log/AuditLogger.java
a19836: This is a classic chicken-and-egg problem. I considered this as well, but the issue is we can't have it both ways. If we don't make it asynchronous, we introduce latency in the request. On the other hand, if we make it asynchronous, we can't roll back the request if the logging fails. Future requests will stop working, but this one would still go through. I think I recorded this scenario on the Notion page, but I'm not sure.
In any case, I think this was discussed earlier, and the product team confirmed that this was an acceptable risk.

Dimfacion: IMO, the latency was an issue when we had the whole mechanism to get the data from the database before the call. Since it was removed, I think it's reasonable to not be async anymore. Like you said, we should discuss this with product

a19836: We also have the latency of the search engine, connection timeout...

a19836: Waiting for Soumaya's answer...
Regardless of her decision, this task should be handled in a separate ticket and PR. Adding new behaviors at this stage may cause further issues. The current async process has been tested and works well. My suggestion is to keep it as it is for now and, if we need to make changes, address them in a later phase or through a different PR. Otherwise, we might risk missing today's deadline.

openaev-api/src/main/java/io/openaev/aop/audit_log/AuditLogger.java
```diff
@@ -46,7 +50,26 @@ public boolean isAuditUnauthorizedLoggingValid() {
...
+  public void prepareLogFailure() {
+    // Halt the application when Halt-on-failure is enabled and a log failure occurs.
+    try {
+      if (auditLogProperties.isHaltOnFailure()) {
+        log.error("[AUDIT] Halt-on-failure triggered — shutting down application.");
+
+        int exitCode = SpringApplication.exit(context, () -> 1);
+        log.error("[AUDIT] Spring shutdown initiated with exit code {}.", exitCode);
+      } else {
+        log.warn("[AUDIT] Halt-on-failure disabled, application continue running...");
```
RomuDeuxfois: **question**: why having warn here ? Is that a problem to have disabled the halt on failure feature ? Info or debug is maybe better.

openaev-api/src/main/java/io/openaev/aop/audit_log/AuditLogger.java
```diff
@@ -46,7 +50,26 @@ public boolean isAuditUnauthorizedLoggingValid() {
     return auditRequestValidator.validUnauthorized();
   }
 
-  /** Wraps the audit service call in try/catch — audit must never break the business flow. */
+  public void prepareLogFailure() {
```
RomuDeuxfois, line 55: **nitpick**: can be good to use AtomicBoolean here because we are on an async thread. So, we have multiple failed at the same time and want to stop spring several time.
With an AtomicBoolean, you can be sure we will stop Spring just once.

openaev-api/src/main/java/io/openaev/aop/audit_log/AuditLogger.java
a19836: yup. It makes sense. Done in the next commit

openaev-api/src/main/java/io/openaev/aop/audit_log/AuditLogger.java
Dimfacion: This PR does not work because it was actually the right way to close the application as documented here : https://docs.spring.io/spring-boot/docs/1.5.17.RELEASE/reference/html/boot-features-spring-application.html#boot-features-application-exit

a19836: re-adding System.exit

Dimfacion (approved):
Todo later : 
- discuss whether or not we keep logging async
- add tests on this feature

PR #5891, opened 2026-05-21
[backend] chore(deps): pom upgrades (#5925)

### LINKED ISSUE #5925: Backend dependency maintenance

antoinemzs:
Close https://github.com/OpenAEV-Platform/openaev/issues/5925

### DIFF
--- a/openaev-front/src/utils/api-types.d.ts
+++ b/openaev-front/src/utils/api-types.d.ts
@@ -14,6 +14,7 @@ type UtilRequiredKeys<T, K extends keyof T> = Omit<T, K> & Required<Pick<T, K>>;
 
 export interface Agent {
   agent_active?: boolean;
+  /** @default "" */
   agent_asset: string;
   /** @format date-time */
   agent_cleared_at?: string;
@@ -22,13 +23,16 @@ export interface Agent {
   agent_deployment_mode: "service" | "session";
   /** @minLength 1 */
   agent_executed_by_user: string;
+  /** @default "" */
   agent_executor?: string;
   agent_external_reference?: string;
   /** @minLength 1 */
   agent_id: string;
+  /** @default "" */
   agent_inject?: string;
   /** @format date-time */
   agent_last_seen?: string;
+  /** @default "" */
   agent_parent?: string;
   agent_privilege: "admin" | "standard";
   agent_process_name?: string;
@@ -52,36 +56,65 @@ export interface AgentCallOutput {
 }
 
 export interface AgentExecutorOutput {
-  /** Agent executor id */
+  /**
+   * Agent executor id
+   * @default ""
+   */
   executor_id?: string;
-  /** Agent executor name */
+  /**
+   * Agent executor name
+   * @default ""
+   */
   executor_name?: string;
-  /** Agent executor type */
+  /**
+   * Agent executor type
+   * @default ""
+   */
   executor_type?: string;
 }
 
 export interface AgentOutput {
-  /** Indicates whether the endpoint is active. The endpoint is considered active if it was seen in the last 3 minutes. */
+  /**
+   * Indicates whether the endpoint is active. The endpoint is considered active if it was seen in the last 3 minutes.
+   * @default false
+   */
   agent_active?: boolean;
-  /** Agent deployment mode */
+  /**
+   * Agent deployment mode
+   * @default ""
+   */
   agent_deployment_mode?: "service" | "session";
-  /** The user who executed the agent */
+  /**
+   * The user who executed the agent
+   * @default ""
+   */
   agent_executed_by_user?: string;
-  /** Agent executor */
+  /**
+   * Agent executor
+   * @default ""
+   */
   agent_executor?: AgentExecutorOutput;
   /**
    * Agent id
    * @minLength 1
+   * @default ""
    */
   agent_id: string;
   /**
    * Instant when agent was last seen
    * @format date-time
+   * @default ""
    */
   agent_last_seen?: string;
-  /** Agent privilege */
+  /**
+   * Agent privilege
+   * @default ""
+   */
   agent_privilege?: "admin" | "standard";
-  /** The version of the agent */
+  /**
+   * The version of the agent
+   * @default ""
+   */
   agent_version?: string;
 }
 
@@ -129,11 +162,13 @@ export interface AggregatedFindingOutput {
   /**
    * Asset groups linked to endpoints
    * @uniqueItems true
+   * @default ""
    */
   finding_asset_groups?: AssetGroupSimple[];
   /**
    * Endpoint linked to finding
    * @uniqueItems true
+   * @default ""
    */
   finding_assets: EndpointSimple[];
   /** @format date-time */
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -61,7 +61,34 @@
...
+                <groupId>org.bouncycastle</groupId>
+                <artifactId>bc-jdk18on-bom</artifactId>
+                <version>1.84</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+            <dependency>
+                <groupId>tools.jackson.core</groupId>
+                <artifactId>jackson-core</artifactId>
+                <version>3.1.3</version>
```
RomuDeuxfois: **question**: Which dependency needs this ?
It's a major upgrade.

pom.xml
```diff
@@ -61,7 +61,34 @@
...
+            </dependency>
+            <dependency>
+                <groupId>io.netty</groupId>
+                <artifactId>netty-bom</artifactId>
+                <version>4.1.133.Final</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+            <dependency>
+                <groupId>org.bouncycastle</groupId>
```
RomuDeuxfois, line 83: **todo**: We already have this in openaev-api>pm.xml

pom.xml
antoinemzs: it's for the elastic client

note this override is for the `tools.jackson.core` group while the spring dependency is under the `com.fasterxml.jackson.core` group and is resolved as such:
```xml
<dependency>
  <groupId>com.fasterxml.jackson.core</groupId>
  <artifactId>jackson-core</artifactId>
  <version>2.21.2</version>
</dependency>
```

The elastic client required vesion `3.0.0` (vulnerable)

pom.xml
antoinemzs: It's resolving a single dependency directly, but others were transitive and resolved as a lower version. Here I'm setting a mandatory minimum version for all bouncy castle dependencies (a "fixed" version is resolved as `>= version` in maven).

I've put it there so that this requirement is globally applied.

RomuDeuxfois: Ok, so please remove the version from the openaev-api pom.xml file

pom.xml
```diff
@@ -61,7 +61,34 @@
             <url>https://repository.jboss.org/nexus/content/groups/public/</url>
         </repository>
     </repositories>
-
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.apache.commons</groupId>
+                <artifactId>commons-lang3</artifactId>
+                <version>3.20.0</version> <!-- Force version for lack of update in commons-collections4 -->
```
RomuDeuxfois: **todo**: please follow the convention we have and put every version on the properties tag

RomuDeuxfois:
@antoinemzs CI is green.
I’m removing `@Nullable` from the back-end, since I remember we initially added it to align with `api-types` and Zod.

openaev-front/src/utils/api-types.d.ts
```diff
@@ -2343,9 +2796,15 @@ export interface EngineSortField {
...
+   * Pagination to set (optional)
+   * @default ""
+   */
+  pagination?: null;
+  /**
+   * Parameters to set
+   * @default ""
+   */
   parameters?: Record<string, string>;
 }
```
antoinemzs, line 2350: @RomuDeuxfois @corinnekrych @guillaumejparis how do we feel about this? It does not seem used throughout the frontend though.

openaev-front/src/utils/api-types.d.ts
```diff
@@ -2951,7 +3433,7 @@ export interface Executor {
   executor_created_at: string;
   executor_doc?: string;
   /** @minLength 1 */
-  executor_id: string;
+  executor_id: string | null;
```
antoinemzs: TODO: check

openaev-front/src/utils/api-types.d.ts
```diff
@@ -1069,7 +1259,7 @@ export interface Collector {
   collector_created_at: string;
   collector_external?: boolean;
   /** @minLength 1 */
-  collector_id: string;
+  collector_id: string | null;
```
antoinemzs: TODO: check

openaev-front/src/utils/api-types.d.ts
```diff
@@ -4233,7 +4925,7 @@ export interface Injector {
   injector_executor_commands?: Record<string, string>;
   injector_external?: boolean;
   /** @minLength 1 */
-  injector_id: string;
+  injector_id: string | null;
```
antoinemzs: TODO: check

openaev-front/src/utils/api-types.d.ts
```diff
@@ -4248,19 +4940,22 @@ export interface InjectorContract {
...
   injector_contract_attack_patterns?: string[];
   /** @minLength 1 */
   injector_contract_content: string;
   /** @format date-time */
   injector_contract_created_at: string;
   injector_contract_custom?: boolean;
+  /** @default "" */
   injector_contract_domains?: string[];
-  injector_contract_external_id?: string;
+  injector_contract_external_id?: string | null;
```
antoinemzs: TODO: check

openaev-front/src/utils/api-types.d.ts
```diff
@@ -184,28 +222,28 @@ export interface AiGenericTextInput {
...
   /** @format int32 */
-  ai_paragraphs?: number;
-  ai_recipient?: string;
-  ai_sender?: string;
-  ai_tone?: string;
+  ai_paragraphs?: number | null;
+  ai_recipient?: string | null;
+  ai_sender?: string | null;
+  ai_tone?: string | null;
 }
```
antoinemzs, line 209: TODO; check ai types

antoinemzs (commented):
A few places where the type becomes a union with `null`, maybe harmless, will double check.

A place where the type becomes just `null` but does not seem used in the frontend? Please can I have a second opinion @guillaumejparis @corinnekrych @RomuDeuxfois this looks related to the "related entities" drawer on dashboards

antoinemzs:
The changes look related to how the openapi spec is emitted after the POM upgrades, I'm tracking this.

openaev-front/src/utils/api-types.d.ts
RomuDeuxfois: Hmm, I hadn’t seen all of that.

From my point of view, we have two possible options:

1. Keep the `@Nullable` annotations on the back-end side, so we can perform sanity checks in the methods that use this object and avoid NPEs.
   However, this would also mean aligning the entire front-end accordingly, which could take some time.

2. Remove the `@Nullable` annotations on the back-end side, because a Java object without explicit constraints can technically be nullable anyway. This would allow us to keep our API types as they were before, while keeping in mind on the back-end side that “no constraint” means “can be null.”

I personally prefer the second option.

antoinemzs: Looks similar to https://github.com/springdoc/springdoc-openapi/pull/3245#discussion_r2932723941

antoinemzs:
I've rolled back:

* springdoc-openapi: 2.8.17 -> 2.8.16
* swagger-annotations-jakarta: 2.2.50 -> 2.2.43

both these libs pull swagger-core > 2.2.43 which has a bug with propagating nullable annotations which can result in specifying an `object` type attribute as just `null`, see https://github.com/swagger-api/swagger-core/issues/5115

damgouj (approved):

PR #5902, opened 2026-05-22
[frontend] fix(front): use 'eq' operator for domain filter in Threat Library (#5901)

### LINKED ISSUE #5901: Bug: Default filter uses "contains" instead of "equals" in Threat Library
## Description

When navigating the Threat Library, the displayed counts are inconsistent due to the default filter operator being set to `contains` instead of `equals`.
Also AND operator returns empty results: When selecting multiple domains via the icon bar, filters are automatically combined with OR. Switching to AND returns zero results, even when items matching both domains should exist.
Both issues impact the reliability of the Threat Library navigation and filtering experience.

## Steps to reproduce

SCENARIO : `contains` instead of `equals`.
1. Navigate to the Threat Library
2. Click on a domain (e.g. "Email-Infiltration")
3. Observe the inject count for related domains (e.g. "Endpoint")
4. Click on "Endpoint" → the counts displayed do not match expectations
5. Compare the results using `contains` vs `equals` filter → different results returned

SCENARIO : AND/OR filter logic with multiple domains
Navigate to the Threat Library
Click on a first domain (e.g., "Email-Infiltration")
Click on a second domain (e.g., "Endpoint")
Observe that both domain filters are combined with the OR operator by default
Switch the filter logic to AND
Observe that zero results are returned, even though items belonging to both domains should exist

## Expected behavior

SCENARIO : `contains` instead of `equals`.

- The `equals` filter returns the correct, expected results
- The default filter should use `equals` as the operator

SCENARIO : AND/OR filter logic with multiple domains

When multiple domains are selected the filter logic have to be set to AND, results should return injects that are tagged with all selected domains
If no injects genuinely match all selected domains, the empty state should be clearly communicated  but the filter logic itself should function correctly for OR and AND

## Actual behavior
SCENARIO : `contains` instead of `equals`.
- The default filter uses `contains`, which returns inconsistent/incorrect counts
- `contains` does not appear to be a valid or expected operator in this context

SCENARIO : AND/OR filter logic with multiple domains
Switching from OR to AND with multiple domain filters returns INCONSINSTENT results regardless of the data

## Root cause (initial analysis)

The default filter operator is set to `contains` instead of `equals`. The `equals` operator works as expected — the default should be changed accordingly.

## Impact

Inconsistent count display when navigating between domains in the Threat Library. Reported by VP Engineering.

RomuDeuxfois:
## Summary

Fixes #5901 — The domain icon filter in the Threat Library was using `contains` as the default operator when clicking on a domain. This caused inconsistent inject counts when navigating between domains.

## Root cause

In `useDomainIconFilter.ts`, when a new domain filter is created via `handleAddFilterWithEmptyValue`, the operator was set to `contains` instead of `eq`.

The `contains` operator performs a partial/substring match, while `eq` performs an exact match on the domain value. This discrepancy caused incorrect counts to be displayed.

## Fix

Changed the filter operator from `contains` to `eq` in the `handleDomainClick` callback (line 75).

This is consistent with other navigation filters in the codebase (e.g. `craftedDocumentFilter` in `FilterUtils.tsx` uses `eq`).

## Changes

- `openaev-front/src/admin/components/common/domains/useDomainIconFilter.ts`: `operator: 'contains'` → `operator: 'eq'`

## Testing

1. Navigate to the Threat Library
2. Click on a domain (e.g. "Email-Infiltration")
3. Click on a related domain (e.g. "Endpoint")
4. Verify the inject counts are now consistent and correct

### DIFF
--- a/openaev-framework/src/main/java/io/openaev/utils/FilterUtilsJpa.java
+++ b/openaev-framework/src/main/java/io/openaev/utils/FilterUtilsJpa.java
@@ -255,7 +255,7 @@ public final class FilterUtilsJpa {
       case lt -> (paths, texts) -> lessThanTexts((Expression<Instant>) paths, cb, texts);
       case lte -> (paths, texts) -> lessThanOrEqualTexts((Expression<Instant>) paths, cb, texts);
       case not_eq -> (paths, texts) -> notEqualsTexts((Expression<String>) paths, cb, texts, type);
-      default -> (paths, texts) -> equalsTexts((Expression<String>) paths, cb, texts, type);
+      default -> (paths, texts) -> equalsTexts((Expression<String>) paths, cb, texts, type, mode);
     };
   }
 }
--- a/openaev-framework/src/main/java/io/openaev/utils/OperationUtilsJpa.java
+++ b/openaev-framework/src/main/java/io/openaev/utils/OperationUtilsJpa.java
@@ -224,6 +224,15 @@ public final class OperationUtilsJpa {
 
   public static Predicate equalsTexts(
       Expression<String> paths, CriteriaBuilder cb, List<String> texts, Class<?> type) {
+    return equalsTexts(paths, cb, texts, type, FilterMode.or);
+  }
+
+  public static Predicate equalsTexts(
+      Expression<String> paths,
+      CriteriaBuilder cb,
+      List<String> texts,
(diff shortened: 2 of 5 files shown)


gabriel-peze (approved):

gabriel-peze:
Tested OK.

PR #5903, opened 2026-05-22
[backend] fix(multi-tenancy): manage jobs globally to execute injects in each tenant (#4864)

### LINKED ISSUE #5904: [multi-tenancy] Can't execute an inject in a tenant created
## Description

A lot of jobs are used to launch/schedule an atomic testing, scenario and scenario. They don't work on a tenant created.

## Reproducible Steps

1. Create a tenant
2. Create a payload
3. Launch an atomic testing
4. Traces aren't created on OpenAEV
5. Same for scenarios and simulations, they are not executed if they are scheduled

## Expected Output

I can launch an inject as on the default tenant

## Actual Output

I can't launch an inject if I am on a tenant created

damgouj:
### Proposed changes

* Fix jobs to manage them globally and not with tenant
* PreUpdate removed in TenantListener because useless, all the objects tenant related have the "updatable=false" attribute
* Tests will be done in the coming weeks

### Testing Instructions

1. Create a tenant
2. Launch a scheduled/now scenario, a scheduled/now simulation, an atomic testing with a technical inject
3. Check everything is working fine
4. Do the same for the default tenant

### Checklist


openaev-api/src/main/java/io/openaev/scheduler/jobs/InjectsExecutionJob.java
```diff
@@ -447,6 +452,8 @@ public void execute(JobExecutionContext jobExecutionContext) throws JobExecution
   }
 
   private void handleInjectExpectationCollectStatus() {
+    // Disable tenant filter — called from InjectsExecutionJob which runs cross-tenant
```
corinnekrych, line 455: @damgouj because of the session scope / not being part of a transaction we end up to have to put this disable Filter in several start of method...
Proposal: create an annotation / Aspect @CrossTenantJob that we can use to annote Job class that need to run cross tenant 
proposal (just drafted / not yet tested in this branch [here : poc-crosstenatjob-aspect](https://github.com/OpenAEV-Platform/openaev/compare/issue/5888...poc-crosstenatjob-aspect?expand=1))

> Note: we can do a separate ticket fro a cleaner / reusable approach

corinnekrych:
Tested locally:
1/ create tenant A
2/ install Agent sur tenant A as admin
3/ create an atomic testing send email
check the email is sent 👍 
Screenshot 2026-05-22 at 13 38 23

corinnekrych (dismissed):

johanah29 (approved):

Issue #5480, opened 2026-04-16
[Audit Logging][Collection: CRUD] Core AOP aspect + AuditLogService refactoring


corinnekrych:
📖 **Technical Design**: [Audit Logging — Platform-Level Log Forwarding (No UI)](https://www.notion.so/filigran/Audit-Logging-Phase-1-Platform-Level-Log-Forwarding-No-UI-3198fce17f2a81e68f37c1a67f445486?t=3418fce17f2a800ca64000a9dd59d6fb)
🔗 **PoC branch PR**: [#5221](https://github.com/OpenAEV-Platform/openaev/pull/5221)
🎯 **Epic**: [#5241](https://github.com/OpenAEV-Platform/openaev/issues/5241)

---

## Description

Refactor the core AOP audit aspect and AuditLogService from the PoC into production-quality code. The PoC `AuditLogAspect` (~835 lines) and `AuditLogService` (~373 lines) need to be decomposed into focused utility classes:

- **ChildResourceDetector** — Detect when the return type differs from the parent resource, reclassify event scope/type/id, handle child deletion (HTTP DELETE), handle bulk child creation (`Collection<Base>`).
- **AuditDiffEngine** — Semantic diff between old entity snapshot and input DTO, JPA relation flattening, numeric/null normalization.
- **AutomatedRequestFilter** — Identify and skip agent/machine traffic (User-Agent patterns, URI prefix checks).
- **AuditEventClassifier** — Map `Action` → `event_scope`, determine `event_access` from `ResourceType`, build human-readable message.

The `AuditLogAspect.auditAround()` becomes a lean orchestrator calling these utilities.

## Definition of Done

- [ ] AOP aspect intercepts all `@AccessControl`-annotated endpoints and emits structured audit events
- [ ] Utility classes are individually testable with unit tests
- [ ] Unit tests for each utility class (ChildResourceDetector, AuditDiffEngine, AutomatedRequestFilter, AuditEventClassifier)
- [ ] Integration test: perform a CRUD operation and verify an audit event is emitted with correct fields
- [ ] `mvn spotless:check` passes

PR #5919, opened 2026-05-24
[backend] feat(audit-logging): improv code (#5480)

RomuDeuxfois:


openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
```diff
@@ -61,8 +58,7 @@ public Object auditAround(ProceedingJoinPoint joinPoint, AccessControl accessCon
       action = accessControl.actionPerformed();
       isActive =
           accessControlAuditLogger.isAuditLoggingEnabled()
-              && accessControlAuditLogger.isAuditLoggingValid(action)
-              && previewFeatureService.isFeatureEnabled(PreviewFeature.AUDIT_LOG);
+              && accessControlAuditLogger.isAuditLoggingValid(action);
```
Dimfacion, line 62: **question :** Not sure why it's been removed tbh. I know we'll check again the feature flag later down the line in the service layer but don't you think it's a good thing to just proceed with the call if we can to avoid adding unnecessary latency ?

openaev-model/src/main/java/io/openaev/driver/EngineObjectMapperFactory.java
```diff
@@ -0,0 +1,29 @@
...
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
+import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
+
+/**
+ * Factory for the shared {@link ObjectMapper} used by search-engine drivers (Elasticsearch,
+ * OpenSearch) and any component that needs identical serialization settings.
+ */
+public class EngineObjectMapperFactory {
```
Dimfacion, line 13: **praise :** good call creating this factory

openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
RomuDeuxfois: The enabled logic is already handled in the `LogService` method. If that logic changes later, duplicating the code increases the risk of missing an update in one place.
So I think it’s better to keep everything handled in a single method.

I’m not sure the latency impact will be significant, since the methods are very simple.

Dimfacion: They are simple but it's around the accesscontrol annotation which is everywhere so it will be done for all the calls to the API

RomuDeuxfois: So maybe a @ConditionalOnProperty direclty on the Aspect ?

Dimfacion: Didn't thought of it but yeah, I think it would be great

RomuDeuxfois: I made an update

openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
```diff
@@ -39,13 +37,13 @@
  */
 @Aspect
 @Component
-@Order(Ordered.LOWEST_PRECEDENCE)
+@ConditionalOnProperty(name = "openaev.audit-logs.enabled", havingValue = "true")
```
Dimfacion:
```suggestion
@ConditionalOnProperty(name = "openaev.audit-logs.service.enabled", havingValue = "true")
```
Or maybe you want openaev.audit-logs.enabled here but then you forgot the change on the application.properties

Dimfacion (approved):

openaev-api/src/main/java/io/openaev/utils/ResourceManagerUtils.java
```diff
@@ -1,7 +1,25 @@
 package io.openaev.utils;
```
a19836, line 1: I believe we no longer need this class. Currently, the only method used in LogService is ResourceManagerUtils.extractNameFromSnapshot, and in LogUtils, it's EntityManagerUtils.buildAdministrationResourceTypes. It may be possible to remove these two classes and move the respective methods into LogUtils and LogService.

openaev-model/src/main/java/io/openaev/driver/ElasticDriver.java
a19836, line 308: Just so u guys know, this parallel stream with sequential iteration is causing startup deadlock, where Spring's MetricsRepositoryMethodInvocationListener acquires a ReentrantLock when recording repository call metrics. When multiple ForkJoinPool threads all hit a repository method  (indexingStatusRepository.findByType) at the same time during startup, they contend on that same lock and deadlock — none can proceed and the application never finishes booting. Iterating sequentially eliminates the contention entirely at a negligible cost: the number  of ES models is small and the bottleneck is network I/O, not CPU parallelism.

To fix this we must remove the .parallel() statement.  We just need to replace parallel model initialization with sequential iteration. This avoids startup-time lock contention/deadlock during repository calls while building indexes.
I already fixed this in the branch issue/5483, since I couldn't start my app.

But basically here is what's happening:
The app is sometimes hanging after migration during Elasticsearch initialization.
Thread dump showed lock contention around MetricsRepositoryMethodInvocationListener triggered from ElasticDriver.elasticClient() while iterating models in parallel.

openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogger.java
```diff
@@ -23,6 +24,7 @@
  * <p>Phase 1: delegates to {@link LogService} for console-only output.
  */
 @Component
+@ConditionalOnProperty(name = "openaev.audit-logs.service.enabled", havingValue = "true")
 @RequiredArgsConstructor
 @Slf4j
 public class AccessControlAuditLogger {
```
a19836, line 30: Just a quick note about this class: in the branch issue/5483, I've renamed it to AuditLogger because the previous name is no longer relevant. This class is now a generic AuditLogger used by AccessControlAuditLogAspect (based on the AccessControl annotation), as well as by OpemSamlConfig, AppSecurityConfig, UserApi, and other classes unrelated to AOP or the AccessControl annotation. Therefore, the name needed to be updated to something more intuitive, like AuditLogger. After the PR 5807 (issue/5483) be merged, it won't make sense anymore to have this name. 
If you wish, we can move the renaming part to this branch. Whatever you think better. For me both approaches are correct.

a19836 (commented):
reviewing 5919

PR #5921, opened 2026-05-24
[backend] feat(audit-logging): avoid fetching user to improv performance (#5480)

RomuDeuxfois:


openaev-api/src/main/java/io/openaev/service/LogService.java
```diff
@@ -268,19 +265,16 @@ private void populateUserMetadata(LogEvent doc) {
     LogEvent.UserMetadata meta = new LogEvent.UserMetadata();
     boolean hasData = false;
 
-    // User email — denormalized for display
+    // User email — extracted from SecurityContext (no DB hit)
     try {
-      String userId = doc.getUserId();
-      if (userId != null) {
-        User user = userService.user(userId);
```
a19836, line 275: **We don’t need any of these improvements** because userRepository is already called earlier in AccessControlAspect, and EntityManager caches objects by default.

More specifically, the AccessControlAspect.methodRBACVerification method, which is called before this flow, calls userService.currentUser(). This then calls userRepository, caching the user object.

Later, when we call the userService.user method in LogService, it also calls userRepository, which will return the cached object. This is already very fast, so there is no need for further improvements.

By introducing all these changes across multiple classes, we may also introduce new risks and future issues. I would not advise proceeding with this PR.

I only see logic in validating these improvements if `skipRBAC` is true in some cases where the `LogService` also runs, or if there are places where the `AccessControlAspect` is not called but the `LogService` is. 
**In this case, then this PR looks good to me.**

a19836 (commented):
Is this really necessary?

openaev-api/src/main/java/io/openaev/service/LogService.java
RomuDeuxfois: Ok, we can close this PR.
Thanks for the clarification.

PR #5923, opened 2026-05-24
[backend] feat(audit-logging): improv properties (#5480)

RomuDeuxfois:

PR #5924, opened 2026-05-24
[backend] feat(audit-logging): remove async duplication (#5480)

RomuDeuxfois:


gabriel-peze:
To test : 
- Set openaev.audit-logs.engine.enabled and openaev.audit-logs.service.enabled to true in properties, and enable  AUDIT_LOG flag
- Create a Scenario, a Simulation or an Atomic Test
- Every HTTP calls others than GET should log

gabriel-peze:
Tested OK.
image

gabriel-peze (dismissed):

a19836:
I think we should brainstorm first and think carefully about going forward with this PR, because I see some risks with this approach, this is:
In this PR, you removed the async processing from the Transport objects by taking out the async from the send method in all AuditLogTransportUtils classes and updating AuditLogTransportDispatcherUtils.dispatch to run the Transport synchronously. 
This was how it was originally implemented, but we later changed it to async so each transport could operate independently. With your current approach, if the search engine runs first and gets blocked, the other transports won't execute, which could disrupt the order of logs during concurrent requests. 

Our previous implementation ensures the correct order is maintained. From a performance standpoint, I believe we should keep this as async, as it was before. 
The goal of our approach was not just to make all logs asynchronous to avoid blocking current requests, but also to ensure each transport operates independently and asynchronously for improved speed and maintain consistency.
I know it may look a bit unusual to have async logic inside another async flow (async duplication), but that was intentional.

But this is just my perspective. If you guys are comfortable with the risk, please proceed.

RomuDeuxfois:
Thanks @a19836 for the review.
You're right that parallelism matters here, but let me clarify what the actual risks are.

**On transport isolation**

The sequential loop in dispatchInternal() does preserve isolation between transports via individual try/catch blocks, if OpenSearch throws an exception, Console and File still execute. So on that specific point, isolation is not broken.

However, you're correct that temporal coupling is a real issue: if OpenSearch is slow (e.g. 30s timeout), Console and File transports will wait for it to finish before executing. In the previous async implementation, all three transports ran in parallel so they were fully independent. This PR does regress on that specific behavior.

**On blocking the HTTP request**

This is not a concern here, AccessControlAuditLogger is already annotated with @Async("accessControlAuditLoggerExecutor") at the caller level, which means the entire dispatch runs in a dedicated thread pool, fully decoupled from the HTTP request. The user's request is never blocked regardless of what happens inside dispatch().

**On ordering guarantees**

The async approach actually provides no ordering guarantee, concurrent threads indexing into OpenSearch have no guaranteed arrival order. The sequential approach gives a predictable execution order within a single dispatch, which is arguably better for audit logs.

a19836:
Would it be logical to add ordering to the transports so that the search index is the last one to execute?

RomuDeuxfois:
Good point, I made the changes here -> https://github.com/OpenAEV-Platform/openaev/pull/5924/changes/5792faf9618b0cd1b085f5f328292a5ac32da156
What do you think ?

openaev-api/src/main/java/io/openaev/utils/log/transport/AuditEngineLogTransport.java
```diff
@@ -3,71 +3,64 @@
...
-@ConditionalOnProperty(name = "openaev.audit-logs.opensearch.enabled", havingValue = "true")
+@ConditionalOnProperty(name = "openaev.audit-logs.engine.enabled", havingValue = "true")
 @RequiredArgsConstructor
 @Slf4j
-public class AuditESLogTransportUtils implements AuditLogTransportUtils {
+public class AuditEngineLogTransport implements AuditLogTransportUtils {
 
   @Getter
-  @Value("${openaev.audit-logs.opensearch.enabled:false}")
+  @Value("${openaev.audit-logs.engine.enabled:false}")
```
corinnekrych, line 27: Do we want to ship Engine transport in phase 1? in you pevious [PR](https://github.com/OpenAEV-Platform/openaev/pull/5923) this properties was removed

RomuDeuxfois: I don't remove the whole mechanism, I only remove it from the default config.

corinnekrych (dismissed):
Much simplier.

This PR make me wonder: with the priority: file first, console and then engine, I wonder in case one of them fails for example file (issue with disk apce, rotation etc..) but engine, console are successfull should we still stop the paltform?

RomuDeuxfois:
@corinnekrych that's a good question
Can we open a discussion for this on the US8 ?

a19836:
Your changes look good to me now and they correctly address the concern we previously had regarding log ordering for Console and File transports. It’s one less item I have on my to-do list. :)
However, I would rather use the `@Order` annotation on each class instead of the `priority()` method. I believe this keeps the code simpler, since the `List<AuditLogTransportUtils> transports` in `AuditLogTransportDispatcherUtils` would already be automatically ordered by default, removing the need for ordering logic in the constructor. (I'll comment this directly in the code)

From a behavior and concurrency perspective, I think the current approach is also moving in the right direction.

For example, if a user updates an object and then immediately deletes it, the logs should now appear in the correct order because Console and File transports execute quickly and without significant delay. In the previous fully concurrent approach, this behavior often appeared correct for these transports in practice, since they were executed immediately and independently without waiting on slower transports, but ordering was still dependent on concurrency and timing and was not strictly guaranteed. However, similar timing-dependent behavior can still exist with the current implementation, although in a more constrained and predictable way due to reduced concurrency complexity.
 
There is also another theoretical edge case in the future, if we introduce additional transports with higher latency. For example, if Transport A blocks or hangs before Transport B executes, then under concurrent requests it could still be possible for a “delete” event to be logged before a previous “update” event on Transport B, even if the delete occurs a few seconds later, meaning the delete could appear before the update in the logs. Although timestamps would still be available, at first glance this could look confusing. In practice, this would only matter if strict chronological ordering becomes a hard requirement. But this was also an issue in the previous implementation, and a queue-based architecture had already been considered as a way to fully address it in the future if needed.

Still, I think your current approach is a good tradeoff for now:
- less complex,
- easier to reason about,
- easier to debug,
- lower risk of thread exhaustion/explosion,
- lower risk of resource saturation due to deeply nested or excessive async duplications and execution chains,
- lower risk of hidden concurrency issues.

So overall, I think we can move forward with these changes or with the @Order annotation suggestion.

openaev-api/src/main/java/io/openaev/utils/log/dispatcher/AuditLogTransportDispatcherUtils.java
```diff
@@ -2,43 +2,52 @@
...
-            CompletableFuture
-                ::join); // only return true if all results from all send methods are true.
+  private boolean dispatchInternal(
+      BiFunction<AuditLogTransportUtils, Object, Boolean> sendFn, Object level) {
+    boolean allSucceeded = true;
+    for (AuditLogTransportUtils transport : transports) {
+      if (!transport.isEnabled()) {
+        continue;
+      }
+      try {
```
a19836, line 39: Although it’s always good to have try-catch blocks, note that all send methods in the existing transports already handle their own exceptions internally. However, keeping this extra protection is still beneficial for the future in case someone adds a new transport and forgets to implement proper error handling. So I think it’s perfectly fine to leave it as it is.

openaev-api/src/main/java/io/openaev/utils/log/dispatcher/AuditLogTransportDispatcherUtils.java
```diff
@@ -2,43 +2,52 @@
...
-  public boolean dispatch(LogEvent event, Object level) {
-    List<CompletableFuture<Boolean>> futures =
-        transports.stream()
-            .filter(AuditLogTransportUtils::isEnabled)
-            .map(transport -> transport.send(event, level))
-            .toList();
+  private final List<AuditLogTransportUtils> transports;
 
-    CompletableFuture.allOf(futures.toArray(new CompletableFuture[0])).join();
+  public AuditLogTransportDispatcherUtils(List<AuditLogTransportUtils> transports) {
```
a19836, line 17: Suggestion: if you add the @Order annotation to the Transport classes, remove the constructor

a19836:
@corinnekrych: Regarding your question, I remember asking Soumaya about this in the past, and her understanding was that if any of them fails, it should be considered critical and the entire platform should stop. However, we should probably confirm this with her again.

corinnekrych:
> @corinnekrych: Regarding your question, I remember asking Soumaya about this in the past, and her understanding was that if any of them fails, it should be considered critical and the entire platform should stop. However, we should probably confirm this with her again.

Yeap let's refine during our refinement session. For phase1, it's mainly file that is used once OS is introduced maybe we can define a source of truth and one critical path (OS/DB) and the other transport can resync when ive again

corinnekrych (approved):

a19836 (commented):
revieing 5924

openaev-api/src/main/java/io/openaev/utils/log/transport/AuditConsoleLogTransportUtils.java
```diff
@@ -3,15 +3,14 @@
...
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Value;
-import org.springframework.scheduling.annotation.Async;
 import org.springframework.stereotype.Component;
 
+/** Audit log transport that writes events to the application console (stdout via SLF4J). */
 @Component
 @RequiredArgsConstructor
 @Slf4j
```
a19836, line 16: Suggestion: add the @Order(1) annotation instead and remove priority method.

openaev-api/src/main/java/io/openaev/utils/log/transport/AuditEngineLogTransport.java
```diff
@@ -3,71 +3,64 @@
...
- * <p>This service is only active when {@code openaev.audit-logs.opensearch.enabled=true}. It
- * receives a fully-populated {@link LogEvent} and indexes it asynchronously to avoid blocking the
- * API response.
+ * <p>This transport is only active when {@code openaev.audit-logs.engine.enabled=true}.
  */
 @Component
-@ConditionalOnProperty(name = "openaev.audit-logs.opensearch.enabled", havingValue = "true")
+@ConditionalOnProperty(name = "openaev.audit-logs.engine.enabled", havingValue = "true")
 @RequiredArgsConstructor
 @Slf4j
```
a19836, line 23: Suggestion: add the @Order(3) annotation instead and remove priority method.

openaev-api/src/main/java/io/openaev/utils/log/transport/AuditFileLogTransportUtils.java
```diff
@@ -3,14 +3,12 @@
...
 import java.util.logging.Level;
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Value;
-import org.springframework.scheduling.annotation.Async;
 import org.springframework.stereotype.Component;
 
 @Component
```
a19836, line 14: Suggestion: add the @Order(2) annotation instead and remove priority method.

openaev-api/src/main/java/io/openaev/utils/log/transport/AuditLogTransportUtils.java
```diff
@@ -1,12 +1,19 @@
...
+  boolean send(String message, Object level);
 
-  CompletableFuture<Boolean> send(LogEvent event, Object level);
+  boolean send(LogEvent event, Object level);
+
+  /**
+   * Execution priority among audit transports. Lower values execute first. This is scoped to audit
+   * transports only (not Spring-wide ordering).
+   */
+  default int priority() {
```
a19836, line 16: If you end up using the @order annotation, then remove the priority method from the interface

a19836 (commented):
redoing my comments bc were in the wrong place

PR #5929, opened 2026-05-25
[backend] feat(audit-log): set correct and redacted user metadata (#5526)

### LINKED ISSUE #5526: [Audit Logging][Collection: CRUD] user_metadata changes

---

## Description
* Email should be redacted in `user_metadata`, User_id will be enough to track action of a user, and email is PII data that we should not log.
See existing:
```json
  "user_metadata" : {
    "user_email" : "admin@openaev.io",
    "user_agent" : "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/146.0.0.0 Safari/537.36",
    "ip" : "0:0:0:0:0:0:0:1"
  },
```

* When deployed to staging, ensure the client IP captures the real client IP (not loopback) –handle X-Forwarded-For / proxy scenarios when deployed behind load balancers.
similar to what is done in OCTI.
The implementation already matches the OCTI pattern from the plan. The ip field gets the real client IP (first entry from X-Forwarded-For if present), 
and the full X-Forwarded-For header is preserved separately. 
The loopback issue (0:0:0:0:0:0:0:1) you see in the example JSON only occurs in local dev where there's no proxy — that's expected behavior since getRemoteAddr() returns localhost when connecting directly
```json
  "user_metadata": {
    "user-agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/26.4 Safari/605.1.15",
    "x-forwarded-for": "136.226.166.100",
    "ip": "136.226.166.100"
  },
```

## Definition of Done

- [ ] user redacted
- [ ] e2e tests

a19836:
Set client IP and redact user email

### Proposed changes

* LogService should include client IP and hash the logged user email

### Testing Instructions

1. Update something in the platform.
2. Go to your console. You should see a json payload with your changes and client ip and user email hashed.

### Related issues

* #5526

### DIFF
--- a/openaev-api/src/main/java/io/openaev/service/LogService.java
+++ b/openaev-api/src/main/java/io/openaev/service/LogService.java
@@ -139,13 +139,13 @@ public class LogService {
       if (input != null) {
         // Redacted input
         input = objectNormalizationUtils.normalize(input);
-        input = ObjectRedactionUtils.redact(input, entityTypeName);
+        input = ObjectRedactionUtils.redact(input, resourceType);
         ctx.put("input", toContextValue(input));
       }
 
       if (output != null) {
         output = objectNormalizationUtils.normalize(output);
-        output = ObjectRedactionUtils.redact(output, entityTypeName);
+        output = ObjectRedactionUtils.redact(output, resourceType);
         ctx.put("output", toContextValue(output));
       }
 
--- a/openaev-api/src/main/java/io/openaev/service/LogService.java
+++ b/openaev-api/src/main/java/io/openaev/service/LogService.java
@@ -184,14 +184,16 @@ public class LogService {
     doc.setEventScope(eventScope);
 
     // Request context
-    LogEvent.RequestMetadata requestData = new LogEvent.RequestMetadata();
+    LogEvent.RequestMetadata meta = new LogEvent.RequestMetadata();
 
     ThreadPoolTaskLoggerConfig.ThreadRequestContextHolder.RequestContextData requestContextData =
         ThreadPoolTaskLoggerConfig.ThreadRequestContextHolder.getRequestContextData();
     String url = requestContextData != null ? requestContextData.url() : null;
+    String method = requestContextData != null ? requestContextData.method() : null;
 
-    requestData.setUrl(url);
-    doc.setRequestMetadata(requestData);
+    meta.setUrl(url);
+    meta.setMethod(method);
+    doc.setRequestMetadata(meta);
 
     // Tenant context
     doc.setTenantId(TenantContext.getCurrentTenant());
--- a/openaev-api/src/main/java/io/openaev/service/LogService.java
+++ b/openaev-api/src/main/java/io/openaev/service/LogService.java
@@ -139,7 +139,7 @@ public class LogService {
       if (input != null) {
         // Redacted input
         input = objectNormalizationUtils.normalize(input);
-        input = ObjectRedactionUtils.redact(input, null);
+        input = ObjectRedactionUtils.redact(input, entityTypeName);
         ctx.put("input", toContextValue(input));
       }
 
--- a/openaev-api/src/main/java/io/openaev/service/LogService.java
+++ b/openaev-api/src/main/java/io/openaev/service/LogService.java
@@ -149,7 +149,7 @@ public class LogService {
         ctx.put("output", toContextValue(output));
       }
 
-      doc.getRequestData().put("signature", signatureNode);
+      doc.getRequestMetadata().setSignature(signatureNode);
 
       ctx.put("message", message);
       doc.setContextData(ctx);
@@ -184,14 +184,14 @@ public class LogService {
     doc.setEventScope(eventScope);
 
     // Request context
-    Map<String, Object> requestData = new LinkedHashMap<>();
+    LogEvent.RequestMetadata requestData = new LogEvent.RequestMetadata();
 
     ThreadPoolTaskLoggerConfig.ThreadRequestContextHolder.RequestContextData requestContextData =
         ThreadPoolTaskLoggerConfig.ThreadRequestContextHolder.getRequestContextData();
     String url = requestContextData != null ? requestContextData.url() : null;
 
-    requestData.put("url", url);
-    doc.setRequestData(requestData);
+    requestData.setUrl(url);
+    doc.setRequestMetadata(requestData);
 
     // Tenant context
     doc.setTenantId(TenantContext.getCurrentTenant());
--- a/openaev-api/src/main/java/io/openaev/service/LogService.java
+++ b/openaev-api/src/main/java/io/openaev/service/LogService.java
@@ -1,5 +1,7 @@
 package io.openaev.service;
 
+import static io.openaev.helper.CryptoHelper.hashWithSHA256;
+
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.openaev.config.OpenAEVPrincipal;
@@ -225,7 +227,7 @@ public class LogService {
       if (userId != null) {
         User user = userService.user(userId);
         if (user != null && user.getEmail() != null) {
-          String email = ObjectRedactionUtils.hash(user.getEmail());
+          String email = hashWithSHA256(user.getEmail());
           meta.setUserEmail(email);
           hasData = true;
         }
--- a/openaev-api/src/main/java/io/openaev/utils/object/ObjectRedactionUtils.java
+++ b/openaev-api/src/main/java/io/openaev/utils/object/ObjectRedactionUtils.java
@@ -1,18 +1,20 @@
 package io.openaev.utils.object;
 
-import static io.openaev.helper.CryptoHelper.hashWithSHA256;
-
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import java.util.Locale;
 import java.util.Set;
 
 public class ObjectRedactionUtils {
+
+  private ObjectRedactionUtils() {}
 
   private static final String REDACTED = "*** Redacted ***";
 
   /** Fields whose values are replaced with {@link #REDACTED} before logging. */
   private static final Set<String> SENSITIVE_FIELDS =
-      Set.of("password", "token", "secret", "newpassword", "apikey", "credential");
+          Set.of("password", "token", "secret", "apikey", "api_key", "credential");
 
   /** Fields redacted only when the entity type is User (PII protection). */
   private static final Set<String> USER_PII_FIELDS = Set.of("name", "user_email");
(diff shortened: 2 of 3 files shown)


openaev-api/src/main/java/io/openaev/service/LogService.java
```diff
@@ -141,30 +144,45 @@ public boolean logRequestEvent(
...
       Map<String, Object> ctx = new LinkedHashMap<>();
 
       ctx.put("entity_type", entityTypeName);
 
+      // Normalize and redacted input
       if (input != null) {
-        // Redacted input
         input = objectNormalizationUtils.normalize(input);
         input = ObjectRedactionUtils.redact(input, null);
         ctx.put("input", toContextValue(input));
```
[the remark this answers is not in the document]
a19836: fix in the next commit

openaev-api/src/main/java/io/openaev/service/LogService.java
```diff
@@ -312,6 +323,34 @@ private void populateUserMetadata(LogEvent doc) {
...
+
+    if (url != null) {
+      meta.setUrl(url);
+      hasData = true;
+    }
+
+    if (method != null) {
+      meta.setMethod(method);
+      hasData = true;
+    }
```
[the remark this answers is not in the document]
a19836: Soumaya confirmed through Teams chat, this is ok.

openaev-model/src/main/java/io/openaev/engine/model/log/LogEvent.java
```diff
@@ -117,4 +119,22 @@ public static class UserMetadata {
...
+
+    /** HTTP Url. */
+    private String url;
+
+    /** HTTP Method. */
+    private String method;
+
+    /** Called method. */
+    private JsonNode signature;
+  }
```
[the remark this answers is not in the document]
a19836: But the ticket https://github.com/OpenAEV-Platform/openaev/issues/5526 says: "every audit event to include **request metadata** (protocol, port, request URL/API endpoint, HTTP method)". It says "request metadata". That's why I put all these info inside of this new nested class, because it doesn't make sense to put it inside of the user metadata class, since this is related to the request and not to the user.

openaev-model/src/main/java/io/openaev/engine/model/log/LogEvent.java
```diff
@@ -73,6 +74,10 @@ public class LogEvent {
   @JsonProperty("user_metadata")
   private UserMetadata userMetadata;
 
+  /** Nested request metadata (protocol, port, request URL/API endpoint, HTTP method). */
+  @JsonProperty("request_metadata")
+  private RequestMetadata requestMetadata;
+
```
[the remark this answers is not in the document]
a19836: Soumaya saw a log sample and said it was ok. But we should re-confirm this part.

openaev-api/src/main/java/io/openaev/utils/object/ObjectRedactionUtils.java
```diff
@@ -25,21 +27,53 @@ public static JsonNode redact(JsonNode node, String entityTypeName) {
...
+
+    if (node.isObject()) {
+      ObjectNode copy = ((ObjectNode) node).deepCopy();
+      copy.properties()
+          .forEach(
+              entry -> {
+                String key = entry.getKey();
+                String fieldName = key.toLowerCase(Locale.ROOT);
+                boolean containsSensitiveToken =
+                    SENSITIVE_FIELDS.stream().anyMatch(fieldName::contains);
```
RomuDeuxfois: **question**: are we sure we want to have contains instead of equals ?
field "api_key_rotation_date" will be redacted for example

openaev-api/src/main/java/io/openaev/utils/object/ObjectRedactionUtils.java
```diff
@@ -1,18 +1,20 @@
...
 
 public class ObjectRedactionUtils {
 
+  private ObjectRedactionUtils() {}
+
   private static final String REDACTED = "*** Redacted ***";
 
   /** Fields whose values are replaced with {@link #REDACTED} before logging. */
   private static final Set<String> SENSITIVE_FIELDS =
-      Set.of("password", "token", "secret", "newpassword", "apikey", "credential");
```
RomuDeuxfois, line 15: **question**: why we don't want newpassword anymore ?

openaev-model/src/main/java/io/openaev/engine/model/log/LogEvent.java
```diff
@@ -117,4 +119,22 @@ public static class UserMetadata {
     /** Resolved remote IP address (remoteAddr or X-Real-IP). */
     private String ip;
   }
+
+  /** User metadata: request-level information about the actor. */
```
RomuDeuxfois: **todo**: wrong javadoc

openaev-api/src/main/java/io/openaev/utils/object/ObjectRedactionUtils.java
```diff
@@ -25,21 +27,53 @@ public static JsonNode redact(JsonNode node, String entityTypeName) {
...
+    return normalized.equals("user") || normalized.endsWith(" user");
+  }
+
+  private static JsonNode redactNode(JsonNode node, boolean isUserEntity) {
+    if (node == null || node.isNull()) {
+      return node;
+    }
+
+    if (node.isObject()) {
+      ObjectNode copy = ((ObjectNode) node).deepCopy();
```
RomuDeuxfois: **todo**: Casting 'node' to 'ObjectNode' is redundant

openaev-api/src/main/java/io/openaev/utils/object/ObjectRedactionUtils.java
```diff
@@ -25,21 +27,53 @@ public static JsonNode redact(JsonNode node, String entityTypeName) {
...
+                  copy.put(key, REDACTED);
+                } else {
+                  copy.set(key, redactNode(entry.getValue(), isUserEntity));
+                }
+              });
+      return copy;
+    }
+
+    if (node.isArray()) {
+      ArrayNode copy = ((ArrayNode) node).deepCopy();
```
RomuDeuxfois: **todo**: Casting 'node' to 'ArrayNode' is redundant

openaev-api/src/main/java/io/openaev/utils/object/ObjectRedactionUtils.java
```diff
@@ -25,21 +27,53 @@ public static JsonNode redact(JsonNode node, String entityTypeName) {
...
+  /** Returns true for user-scoped entity labels like "User", "Platform User" or "Tenant User". */
+  private static boolean isUserEntityType(String entityTypeName) {
+    if (entityTypeName == null || entityTypeName.isBlank()) {
+      return false;
+    }
+    String normalized = entityTypeName.trim().toLowerCase(Locale.ROOT);
+    return normalized.equals("user") || normalized.endsWith(" user");
+  }
+
+  private static JsonNode redactNode(JsonNode node, boolean isUserEntity) {
```
RomuDeuxfois, line 49: **suggestion**: I think we can do something better here
instead of performing a deep copy upfront (which duplicates every node regardless of depth), we can build a new tree incrementally during recursion

private static JsonNode redactNode(JsonNode node, boolean isUserEntity) {
    if (node == null || node.isNull()) {
      return node;
    }

    if (node.isObject()) {
      ObjectNode result = ((ObjectNode) node).objectNode();
      node.fields()
          .forEachRemaining(
              entry -> {
                String key = entry.getKey();
                String fieldName = key.toLowerCase(Locale.ROOT);
                boolean containsSensitiveToken =
                    SENSITIVE_FIELDS.stream().anyMatch(fieldName::contains);

                if (containsSensitiveToken
                    || (isUserEntity && USER_PII_FIELDS.contains(fieldName))) {
                  result.put(key, REDACTED);
                } else {
                  result.set(key, redactNode(entry.getValue(), isUserEntity));
                }
              });
      return result;
    }

    if (node.isArray()) {
      ArrayNode result = ((ArrayNode) node).arrayNode(node.size());
      for (JsonNode element : node) {
        result.add(redactNode(element, isUserEntity));
      }
      return result;
    }

    // Scalar nodes are immutable; returning as-is preserves value and avoids unnecessary copies.
    return node;
  }

openaev-api/src/main/java/io/openaev/utils/object/ObjectRedactionUtils.java
```diff
@@ -25,21 +27,53 @@ public static JsonNode redact(JsonNode node, String entityTypeName) {
...
 
-  public static String hash(String value) {
-    return hashWithSHA256(value);
+  /** Returns true for user-scoped entity labels like "User", "Platform User" or "Tenant User". */
+  private static boolean isUserEntityType(String entityTypeName) {
+    if (entityTypeName == null || entityTypeName.isBlank()) {
+      return false;
+    }
+    String normalized = entityTypeName.trim().toLowerCase(Locale.ROOT);
+    return normalized.equals("user") || normalized.endsWith(" user");
```
RomuDeuxfois: **todo**: could be better to use the direct class name

RomuDeuxfois: Oh maybe use the resource type
// ObjectRedactionUtils.java
private static final Set<ResourceType> USER_ENTITY_TYPES =
    Set.of(ResourceType.USER, ResourceType.PLATFORM_USER);

public static JsonNode redact(JsonNode node, ResourceType resourceType) {
    if (node == null || node.isNull()) {
        return node;
    }
    boolean isUserEntity = resourceType != null && USER_ENTITY_TYPES.contains(resourceType);
    return redactNode(node, isUserEntity);
}

openaev-api/src/main/java/io/openaev/utils/object/ObjectRedactionUtils.java
a19836: We do, but now we check whether the attribute contains any of these strings - which means that something like `newPassword` would still match the string `password`.

openaev-api/src/main/java/io/openaev/utils/object/ObjectRedactionUtils.java
a19836: Changed in the next commit.

openaev-api/src/main/java/io/openaev/utils/object/ObjectRedactionUtils.java
a19836: changed in the next commit

openaev-api/src/main/java/io/openaev/utils/object/ObjectRedactionUtils.java
a19836: That's a good point. I'll change the logic so we use regex instead; we'll have regex for both redacted fields and allowed fields.

openaev-api/src/main/java/io/openaev/utils/object/ObjectRedactionUtils.java
a19836: fixed in the next commit

openaev-api/src/main/java/io/openaev/utils/object/ObjectRedactionUtils.java
a19836: fixed in the next commit

RomuDeuxfois (dismissed):

RomuDeuxfois (approved):

PR #5945, opened 2026-05-26
[backend] feat(audit-log): collect unauthorized 403 access attempts (#5481)

### LINKED ISSUE #5481: [Audit Logging][Collection: CRUD] Collect unauthorized (403) access attempts

---

## Description

Extend audit logging to capture RBAC denials (403) for visibility into unauthorized access attempts.

1. Add `@AfterThrowing` advice that captures access-denied exceptions from `@AccessControl` checks and logs them as `event_scope="unauthorized"` / `event_status="error"`.
2. Only log actual RBAC denials — NOT Spring Security's `authenticationEntryPoint` (401) which fires for every unauthenticated request including frontend polling/SSE reconnections.
3. Remove the configurable aspect for READ events from the PoC — always log UNAUTHORIZED events. Do not log successful READ events.

## Definition of Done

- [ ] 403 RBAC denials produce audit events with `event_scope="unauthorized"` and `event_status="error"`
- [ ] 401 unauthenticated requests are NOT logged
- [ ] Successful READ operations are NOT logged
- [ ] Integration test: call a protected endpoint without permission → verify audit event with correct fields
- [ ] Integration test: successful GET request → verify NO audit event is produced

a19836:
Extend audit logging to capture RBAC denials (403) for visibility into unauthorized access attempts.
1. Add @AfterThrowing advice that captures access-denied exceptions from @AccessControl checks and logs them as event_scope="unauthorized" / event_status="error".
2. Only log actual RBAC denials — NOT Spring Security's authenticationEntryPoint (401) which fires for every unauthenticated request including frontend polling/SSE reconnections.
3. Remove the configurable aspect for READ events from the PoC — always log UNAUTHORIZED events. Do not log successful READ events.

### Proposed changes

* Update AccessControlAspect so that before sending a ResponseStatusException, it calls a method from AccessControlAuditLogAspect to check if the response is a 403 for access control requests.
* If AccessControlAuditLog is enabled, log this exception.

### Testing Instructions

1. Create a new user with Observer role and then access the url: http://localhost:3001/api/tenants/2cffad3a-0001-4078-b0e2-ef74274022c3/tenant-settings/home-dashboard

### Related issues

* #5481


openaev-api/src/main/java/io/openaev/aop/AccessControlAspect.java
```diff
@@ -124,8 +127,18 @@ public void methodRBACVerification(JoinPoint joinPoint, AccessControl accessCont
           resourceId,
           accessControl.resourceType(),
           accessControl.actionPerformed());
-      throw new ResponseStatusException(
-          HttpStatus.FORBIDDEN, "Access denied for user: " + principal.getEmail()) {};
+
+      var ex =
```
RomuDeuxfois: **todo**: We have a coupling issue here
AccessControlAspect had a direct dependency on AccessControlAuditLogAspect via ObjectProvider<AccessControlAuditLogAspect>. On RBAC denial, it manually called auditDeniedAccess(...) before throwing the exception. This mixed two distinct concerns, access control enforcement and audit event production, in the same class...

Please removed the coupling.

RomuDeuxfois (changes requested):

openaev-api/src/main/java/io/openaev/aop/AccessControlAspect.java
a19836: ok. Trying to find another solution through AOP...

openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
```diff
@@ -146,14 +152,62 @@ public Object auditAround(ProceedingJoinPoint joinPoint, AccessControl accessCon
...
+    }
+
+    try {
+      String logUUID = UUID.randomUUID().toString();
+      ResourceType resourceType = accessControl.resourceType();
+      String resourceId = resolveResourceId(joinPoint, accessControl);
+      String eventScope = "unauthorized";
+      JsonNode inputNode = getInputNode(joinPoint, eventScope);
+      JsonNode signatureNode = getMethodSignature(joinPoint, inputNode);
+      JsonNode errorNode = buildErrorNode(null, exception);
```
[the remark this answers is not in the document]
a19836: This is already taken cared by the LogService.

openaev-api/src/main/java/io/openaev/aop/AccessControlAspect.java
a19836: done in the last commit.

openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
```diff
@@ -146,14 +169,76 @@ public Object auditAround(ProceedingJoinPoint joinPoint, AccessControl accessCon
...
+              errorNode,
+              signatureNode,
+              logUUID)
+          .whenComplete(logCompletion);
+    } catch (Exception ex) {
+      log.warn(logErrorMsg, ex);
+    }
+  }
+
+  private AccessControl resolveAccessControlAnnotation(JoinPoint joinPoint) {
```
RomuDeuxfois, line 157: **question**: Why are you doing this by code instead of letting Spring doing it ?

openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
```diff
@@ -275,4 +360,15 @@ public String resolveResourceId(ProceedingJoinPoint joinPoint, AccessControl acc
...
+  private boolean isRbacDeniedException(Throwable exception) {
+    try {
+      if (exception instanceof ResponseStatusException responseStatusException) {
+        return HttpStatus.FORBIDDEN.equals(responseStatusException.getStatusCode());
+      }
+    } catch (Exception e) {
+    }
+
+    return false;
+  }
```
RomuDeuxfois, line 309: **todo**: can be simplify
  private boolean isRbacDeniedException(Throwable exception) {
    return exception instanceof ResponseStatusException rse
            && HttpStatus.FORBIDDEN.equals(rse.getStatusCode());
  }

openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
```diff
@@ -146,14 +169,76 @@ public Object auditAround(ProceedingJoinPoint joinPoint, AccessControl accessCon
...
           .whenComplete(logCompletion);
     } catch (Exception ex) {
-      log.warn("Error during audit logging", ex);
+      log.warn(logErrorMsg, ex);
     }
 
     return result;
   }
 
+  private void logUnauthorizedDeniedEvent(
```
RomuDeuxfois: **question**: why not having this logUnauthorizedDeniedEvent method inside AccessControlAuditLogger like the others ?

a19836: This is not easy to explain in writing, but here it is:
The `logUnauthorizedDeniedEvent` method must be inside a try-catch statement and should only be executed when it is active. This avoids duplicated code.

We are currently calling this method in two different places:
* Line 83: this happens when the user gets a permission denial request and the audit log is inactive for a specific action (like a READ action). Note that the inactive check we perform before may also throw an exception, so we check it again here by calling `logUnauthorizedDeniedEvent`.
* Line 131: this happens when the user gets a permission denial request and the audit log is active for the corresponding action.

In both cases, the main exception must be returned to the user, so this code must always be inside a try-catch statement.

However, I will change this class because I realized there is other duplicated code that we can move into another method.

openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
a19836: Because now the `auditAround` method can only receive the `JoinPoint` parameter; otherwise it will not be triggered every time `AccessControlAspect.methodRBACVerification` throws a `ResponseStatusException`.

Passing the `AccessControl` parameter to the `auditAround` method was causing runtime issues in Spring when binding the `JoinPointMatch` during exception handling.

This approach avoids the fragility of AspectJ annotation argument binding, and it works consistently even with proxy vs. interface or implementation method resolution.

openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
a19836: Yes, it's simpler that way, but we still need to have this inside of a try-catch statement.

openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
RomuDeuxfois: But the annotation binding happens at pointcut matching time (before the advice body runs), not during proceed(). So when AccessControlAspect.methodRBACVerification throws a ResponseStatusException inside proceed(), the annotation is already bound and the exception will be catch by your catch block as expected.

openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
```diff
@@ -61,99 +72,132 @@ public Object auditAround(ProceedingJoinPoint joinPoint, AccessControl accessCon
...
-                  finalResourceType,
-                  finalEventScope);
-              if (throwable != null) {
-                log.warn("Error during audit logging", throwable);
-              }
-            }
-          };
-    } catch (Throwable ex) {
-      log.warn("Error during audit logging", ex);
+      try {
```
RomuDeuxfois: **todo**: can you simplify your logic here ?

// if !accessControlAuditLogger.isAuditLoggingEnabled()
// proceed

// if not
//  try
// proceed 
// catch check if accessControlAuditLogger.isAuditUnauthorizedLoggingValid & log
// then
// check accessControlAuditLogger.isAuditLoggingValid & log

openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
a19836: You’re right that binding happens before proceed(). Our issue is that binding itself failed at runtime (JoinPointMatch not bound), so the advice never entered. We removed annotation-parameter binding to avoid that fragile step and resolve the annotation reflectively inside the advice, which has been stable

openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
a19836: Not sure if I understood correctly, but let me explain this better:
1. This `joinPoint.execute()` method should only get executed, if isActive==false. But is this is false, it must be executed!
- IsActive==true means that `isAuditLoggingEnabled` and `isAuditLoggingValid`. 
- `isAuditLoggingValid` means that the request action must be valid (available actions: create, write, delete, read, research, etc...).

This means that this 1st `joinPoint.execute()` method will only be executed in 2 situations:
- If the action is different than CREATE, WRITE, DELETE, LAUNCH, DUPLICATE, such as a READ or RESEARCH.
- Or if we got an exception before when trying to init the `isActive` and `action` variables.

2. Then the `joinPoint.execute()` gets executed and this method throws an exception. In the catch statements:
- If this exception is a non-access permission (`ResponseStatusException`), then we need to log this event. The responsible method for doing this is the `logUnauthorizedDeniedEvent`.
- In any case we must always throw the joinPoint exception.

3. Let's say that `isActive` is true. In that case, the code continues and `joinPoint.execute()` is called too. If an exception occurs, then step 2 should run again, with a few slight changes.

a19836: Humm... I think I know what you meant. Let me try to improve this code. Since now we don't have the diffs anymore, maybe we can combine this, as you were suggesting... Let me try...

RomuDeuxfois (approved):

Issue #5691, opened 2026-05-05
phase 1: platform audit event hooks


RomuDeuxfois:
## Context

As a platform administrator, I want all user actions (authentication, CRUD operations, simulation execution, RBAC changes) to be captured via platform-level hooks so that every auditable event is intercepted consistently at the infrastructure level.

**Acceptance criteria:** 

- GIVEN a user performs any auditable action (login, create, update, delete, execute simulation, change RBAC role)
- WHEN the action is processed by the platform
- THEN an audit event is emitted containing: timestamp, user ID, action type, target entity, source IP, and result (success/failure)
- GIVEN a user logs out, WHEN the logout event is emitted, THEN it includes the same user_agent, ip, and x_forwarded_for metadata as the corresponding login event (no field omission)

# Checklist

### MUST have DIFF (before & after values captured)

These are platform-critical events — changes that affect user capabilities, access, or platform security configuration.

- User modified — before/after field values (e.g. email, status, role)
- User disabled — before/after user status
- User deleted — last known state captured
- Role assigned to user — before/after role list
- Role revoked from user — before/after role list
- Permission / capability changed on a user — before/after values
- User added to a group — before/after group membership
- User removed from a group — before/after group membership
- Admin security configuration changed (e.g. group config, role config, capability management) — before/after config values

### does NOT need a DIFF (event logged, no before/after required)

These are either metadata enrichments, read events, or content-level changes — granularity not required for Phase 1.

- User created — log the creation event and new state only (no "before" exists)
- Login event — log timestamp, user, IP, user agent, outcome
- Logout event — log timestamp, user, IP, user agent
- Failed authentication attempt — log timestamp, user, IP, outcome
- Export action — log the event and what was exported, no diff
- Request metadata captured (protocol, port, URL/API endpoint) — enrichment only
- Client IP / X-Forwarded-For — enrichment only
- Changes to inject content — no diff required in Phase
- Changes to scenario content — no diff required in Phase
- Changes to payload content — no diff required in Phase
- when a resource is deleted, the only information we need to log is the corresponding resource ID.
- any other request.

a19836:
According with Sairam Jetty's spreadsheet, we also nee to log the request URL, protocol, port...
Here it is the info in his spreadsheet: "Add request metadata: protocol, port number, request URL/API endpoint."

PR #5951, opened 2026-05-27
[backend] feat(audit-logging): enterprise edition feature (#5691)

RomuDeuxfois:


openaev-api/src/main/java/io/openaev/service/LogService.java
```diff
@@ -58,14 +61,38 @@ public enum AuditLogType {
...
+
+    return switch (logType) {
+      case AUDIT -> {
+        boolean isAuditConfigured =
+            auditLogsEnabled && previewFeatureService.isFeatureEnabled(PreviewFeature.AUDIT_LOG);
+        if (!isAuditConfigured) {
+          yield false;
+        }
+
+        boolean isEeActive =
```
a19836: Maybe is a stupid suggestion: Would it be a good idea to place this inside a try-catch statement, just in case something goes wrong? If an error occurs, we can set isEeActive to false.

RomuDeuxfois: Done

a19836 (approved):
It looks good. I tested with and without a license, and the behaviour is the correct one!

PR #5952, opened 2026-05-27
[frontend] fix(inject): scroll on inject list and not the inject detail (#5934)

### LINKED ISSUE #5934: Scrollbar missing on inject list when inject detail drawer is open
Description
Context:
In the inject list creation page (Scenarios, Simulations, Atomic Tests), when a user clicks on an inject, a detail drawer opens on the right side of the screen. Previously, the left-side inject list retained its scrollbar, allowing users to continue browsing injects while consulting the details of the selected one.

Current behavior:
The scrollbar on the left-side inject list disappears when the detail drawer is open. Users can no longer scroll through the inject list while viewing inject details.

Expected behavior:
The scrollbar should remain visible and functional on the inject list (left side) when the detail drawer is open (right side), enabling a side-by-side browse-and-inspect workflow.

Steps to reproduce
Go to a Scenario / Simulation / Atomic test with multiple injects (enough to require scrolling)
Scroll through the inject list — scrollbar works normally ✅
Click on any inject to open the detail drawer on the right
Try to scroll the inject list on the left side
Result: Scrollbar is gone, the list is not scrollable ❌
Impact
Usability regression users must close the drawer to scroll the list, find another inject, and reopen the drawer. This breaks the natural side-by-side workflow that was previously available.

Affected pages
Scenario inject list
Simulation inject list
Atomic test inject list

camrrx:
### Proposed changes

* Update the CreateInject drawer layout: the IconBar is now fixed at the top, and the content below is split into two independently scrollable panels : the inject list on the left (which takes full width when no inject is selected) and the selected inject detail on the right.

### Testing Instructions

1. Go on scenario > inject tabs 
2. Click on "+" to add an inject
3. You should have the inject list and if you click on an inject the content is split into two panels : the list and the inject details : you can scroll the two panels independently

### Related issues

* Closes #5934

### DIFF
--- a/openaev-front/src/admin/components/common/injects/CreateInject.tsx
+++ b/openaev-front/src/admin/components/common/injects/CreateInject.tsx
@@ -407,7 +407,7 @@ const CreateInject: FunctionComponent<Props> = ({
       <Grid style={{
         display: 'flex',
         flexDirection: 'column',
-        height: 'calc(100vh - 100px)',
+        height: 'calc(100vh - 60px)',
         overflow: 'hidden',
       }}
       >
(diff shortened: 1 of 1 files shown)


gabriel-peze (changes requested):
Hello,

The code is ok for me, but i have issues when i test : 
- there is a space at the bottom of the page that shouldn't exist
- when i scroll on the inject list, i lost the header, including the bulk bar and filters, maybe it's normal @DeniseMonreale ?

Image

gabriel-peze (approved):
Tested OK.

PR #5988, opened 2026-05-29
ci: add GitHub Actions workflow to move closed issues

RomuDeuxfois:


.github/workflows/move-closed-issue-to-milestone.yml
```diff
@@ -0,0 +1,29 @@
+name: Move Closed Issue to Milestone
```
RomuDeuxfois, line 1: We were going with something that moves all bug/task issues directly into the `0. Candidate` milestone, and for now, features will still have to be handled manually, or we’ll need to rely on the status in the project to see whether it’s an FF item or not.

guillaumejparis (approved):

PR #6003, opened 2026-06-01
feat(audit-log): log changes on sensitive entities (#5482)

### LINKED ISSUE #5482: [Audit Logging][Collection: CRUD] Role CRUD should log Capabilities

---

## Description

When creating, updating, or deleting a Role, the audit event should include the list of Capabilities assigned to that Role in the `input` field. This provides visibility into changes to user permissions.

Capability names should be human-readable (e.g. `"MANAGE_ASSESSMENT"`) — not enum ordinals.

## Definition of Done

- [ ] Role create event `input` includes the list of capability names
- [ ] Role update event shows capability diff (`input` vs `old_value`)
- [ ] Role delete event includes the role name and capabilities in context
- [ ] Integration test: create a role with capabilities → verify audit event `input` contains capability names
- [ ] Integration test: update role capabilities → verify diff shows added/removed capabilities

Dimfacion:
### Proposed changes

* Piggy backing on the ModelBaseListener to store changes on specific objects
*

### Testing Instructions

1. Create/Modify/Delete a group or a role
2. Check that the changes were logged

### Related issues

* Closes #5482 
*

### Checklist

### Further comments

Since we're a bit short on time, we decided to move the delete to JPA ones so they are properly listened by the ModelBaseListener. This is a temporary fix as seen with @RomuDeuxfois 

We're trying to get close to OCTI format : 
`
{
  "event_type": "modify",
  "entity_type": "User",
  "entity_id": "user--123",
  "actor_id": "admin--456",
  "actor_name": "****",
  "timestamp": "2026-05-29T10:42:00Z",
  "source_ip": "192.168.1.12",
  "user_agent": "Mozilla/5.0...",
  "request_id": "req-abc",
  "changes": [
    {
      "field": "email",
      "old_value": "old@example.com",
      "new_value": "new@example.com"
    },
    {
      "field": "role",
      "old_value": "USER",
      "new_value": "ADMIN"
    }
  ]
}
`


openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
```diff
@@ -77,10 +82,13 @@ public Object auditAround(ProceedingJoinPoint joinPoint) throws Throwable {
     }
 
     Object result = null;
+    JsonNode entityDiffsNode = null;
 
     try {
       result = joinPoint.proceed();
     } catch (Throwable ex) {
+      // Capture diffs on the servlet thread before any async handoff.
+      entityDiffsNode = captureEntityDiffs();
```
a19836: You should only init this var inside of each if condition inside of the try statement, this is, at line 95 and 103

openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
```diff
@@ -105,13 +115,17 @@ public Object auditAround(ProceedingJoinPoint joinPoint) throws Throwable {
       throw ex;
     }
 
+    // Capture diffs on the servlet thread before any async handoff.
+    entityDiffsNode = captureEntityDiffs();
```
a19836: move this to line 123 inside of the condition

openaev-model/src/main/java/io/openaev/database/audit/ModelBaseListener.java
```diff
@@ -117,4 +225,79 @@ void postRemove(Object base) {
...
+        mapper.valueToTree(entity), new TypeReference<Map<String, Object>>() {});
+  }
+
+  /**
+   * Computes a field-level change list between two snapshots in audit-friendly format.
+   *
+   * @return a list of changes {@code [{field, old_value, new_value}]} containing only changed
+   *     fields
+   */
+  private static List<EntityDiffContext.Change> computeChanges(
```
a19836: Improvement: It would be beneficial if this method included options for maximum depth level, maximum event size in bytes, and truncation rules. Additionally, having allowlist and denylist settings for each entity type would help, as schema-based normalization or redaction is preferable to generic recursion.

openaev-model/src/main/java/io/openaev/database/audit/ModelBaseListener.java
```diff
@@ -117,4 +225,79 @@ void postRemove(Object base) {
...
+      }
+    }
+    return changes;
+  }
+
+  /**
+   * Normalizes a snapshot value for equality comparison. Lists (e.g., user IDs, role IDs) are
+   * sorted to avoid false positives caused by insertion-order differences.
+   */
+  private static String normalizeForComparison(Object val) {
```
a19836: Improvement: It would be beneficial if this method included options for maximum depth level, maximum event size in bytes, and truncation rules. Additionally, having allowlist and denylist settings for each entity type would help, as schema-based normalization or redaction is preferable to generic recursion.

openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
Dimfacion: Good call, doing

openaev-model/src/main/java/io/openaev/database/audit/EntityDiffContext.java
```diff
@@ -0,0 +1,178 @@
...
+   * until the controller-level audit aspect consumes them.
+   */
+  public static void clearAfterTransactionCompletion() {
+    if (hasRequestContext()) {
+      RequestAttributes attrs = requestAttributes();
+      attrs.removeAttribute(REQUEST_ATTR_BEFORE_SNAPSHOTS, RequestAttributes.SCOPE_REQUEST);
+      attrs.removeAttribute(REQUEST_ATTR_CLEANUP_REGISTERED, RequestAttributes.SCOPE_REQUEST);
+    }
+
+    BEFORE_SNAPSHOTS_TL.get().clear();
```
a19836: My AI Agent:
When a request context is active, data is stored in request attributes, not in the ThreadLocals. Clearing the ThreadLocals is a no-op in that case but is misleading. More importantly, clearAfterTransactionCompletion() has the same pattern: it clears ThreadLocals unconditionally but intentionally preserves REQUEST_ATTR_ENTITY_DIFFS. This asymmetry means if someone switches from request-scoped to ThreadLocal storage mid-execution (e.g., in a background job that transitions), the cleanup semantics are inconsistent.
Suggestion: Add an else branch to clear() so ThreadLocals are only cleared when there is no request context:
`if (hasRequestContext()) {
  // clear request attrs
} else {
  BEFORE_SNAPSHOTS_TL.get().clear();
  DIFFS_TL.get().clear();
  CLEANUP_REGISTERED_TL.set(false);
}`

openaev-model/src/main/java/io/openaev/database/audit/ModelBaseListener.java
Dimfacion: Given that we're only auditing changes on roles and groups for now and that they are small entities, would it be fine to let it as is for now and put that on our list of improvement to do later ?

a19836: ok. done.

openaev-model/src/main/java/io/openaev/database/audit/ModelBaseListener.java
```diff
@@ -117,4 +234,79 @@ void postRemove(Object base) {
...
+        mapper.valueToTree(entity), new TypeReference<Map<String, Object>>() {});
+  }
+
+  /**
+   * Computes a field-level change list between two snapshots in audit-friendly format.
+   *
+   * @return a list of changes {@code [{field, old_value, new_value}]} containing only changed
+   *     fields
+   */
+  private static List<EntityDiffContext.Change> computeChanges(
```
a19836: Suggestion: Why not using a Comparator instead?

openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
```diff
@@ -153,13 +168,44 @@ private void logAccessControlEvent(
...
+   *
+   * <p>Output format is an array of entries with an explicit {@code id} field:
+   *
+   * <pre>
+   * [
+   *   {"id":"...","entity_type":"...","operation":"update","changes":[...]}
+   * ]
+   * </pre>
+   */
+  private JsonNode captureEntityDiffs() {
```
a19836: Instead of simply getting the diff, I think the Diff should be done in the API layer and not in the modelBaseListener. The ModelBaseListener should only store the snapshot before and after.
Also the diff should move to AuditLogger to run async, meaning that we should have another method that logs the diffs.

openaev-model/src/main/java/io/openaev/database/audit/ModelBaseListener.java
```diff
@@ -73,6 +122,55 @@ void postPersist(Object base) {
...
+   */
+  @PreUpdate
+  void preUpdateForDiff(Object base) {
+    if (!shouldCaptureAuditDiff(base)) return;
+    Base instance = (Base) base;
+    try {
+      Map<String, Object> before = EntityDiffContext.getBefore(instance.getId());
+      if (before == null) return;
+      Map<String, Object> after = buildSnapshot(instance);
+      List<EntityDiffContext.Change> changes = computeChanges(before, after);
```
a19836: I'm inclined to suggest that we perform this diff in the AuditLogger class and only store the new snapshot here.
Note that this code runs synchronously, adding latency to every database request. If the diff is handled in the AuditLogger, it can run asynchronously and avoid introducing any additional request latency.

openaev-model/src/main/java/io/openaev/database/audit/ModelBaseListener.java
```diff
@@ -73,6 +122,55 @@ void postPersist(Object base) {
...
     BaseEvent event = new BaseEvent(DATA_PERSIST, instance, mapper);
     appPublisher.publishEvent(event);
+
+    if (!shouldCaptureAuditDiff(base)) return;
+    try {
+      registerCleanupIfNeeded();
+      Map<String, Object> snapshot = buildSnapshot(instance);
+      List<EntityDiffContext.Change> changes =
+          snapshot.entrySet().stream()
+              .map(e -> new EntityDiffContext.Change(e.getKey(), null, e.getValue()))
```
a19836: I'm inclined to suggest that we perform this diff in the AuditLogger class and only store the new snapshot here.
Note that this code runs synchronously, adding latency to every database request. If the diff is handled in the AuditLogger, it can run asynchronously and avoid introducing any additional request latency.

openaev-model/src/main/java/io/openaev/database/audit/ModelBaseListener.java
```diff
@@ -92,15 +190,34 @@ void postUpdate(Object base) {
...
     Base instance = (Base) base;
     appPublisher.publishEvent(new BaseEvent(DATA_DELETE, instance, mapper));
+
+    if (!shouldCaptureAuditDiff(base)) return;
+    try {
+      registerCleanupIfNeeded();
+      Map<String, Object> snapshot = buildSnapshot(instance);
+      List<EntityDiffContext.Change> changes =
+          snapshot.entrySet().stream()
+              .map(e -> new EntityDiffContext.Change(e.getKey(), e.getValue(), null))
```
a19836: I'm inclined to suggest that we perform this diff in the AuditLogger class and only store the new snapshot here.
Note that this code runs synchronously, adding latency to every database request. If the diff is handled in the AuditLogger, it can run asynchronously and avoid introducing any additional request latency.

a19836 (changes requested):
Although I also agree with this approach, we are sharing the `EntityDiffContext` across both the DB Model and API layers. This may violate the separation of concerns and is architecturally incorrect, since the API should communicate with the DB Model only through the Service layer, never directly.

openaev-api/src/main/java/io/openaev/opencti/connectors/service/LegacyOpenCTIConnectorMigration.java
```diff
@@ -59,7 +59,9 @@ public void deleteLegacyConnectorIfExists(String newEmail) {
           legacyEmail,
           newEmail);
       userService.delete(legacyUser.get().getId());
-      groupRepository.deleteByIdNative(PROCESS_STIX_GROUP_ID);
+      if (groupRepository.existsById(PROCESS_STIX_GROUP_ID)) {
```
a19836, line 62: Shouldn't this be surrounded by a Try-catch statement, just in case someone mistakes and adds a wrong value to PROCESS_STIX_GROUP_ID?

openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
```diff
@@ -152,13 +165,30 @@ private void logAccessControlEvent(
...
     }
   }
 
+  /**
+   * Captures all pending entity snapshots from {@link EntityDiffContext}. Must be called on the
+   * servlet thread before any async handoff, since {@link EntityDiffContext} is request-scoped.
+   */
+  private Map<String, EntityDiffContext.EntitySnapshot> captureEntitySnapshots() {
+    try {
+      Map<String, EntityDiffContext.EntitySnapshot> snapshots =
```
a19836: The local variable snapshots is assigned and immediately used in the return. Minor readability improvement, change it to `var snapshots = EntityDiffContext.consumeAllSnapshots();`

openaev-model/src/main/java/io/openaev/database/audit/EntityDiffContext.java
```diff
@@ -0,0 +1,162 @@
...
+      }
+      Map<String, EntitySnapshot> created = new LinkedHashMap<>();
+      attrs.setAttribute(REQUEST_ATTR_ENTITY_SNAPSHOTS, created, RequestAttributes.SCOPE_REQUEST);
+      return created;
+    }
+    return SNAPSHOTS_TL.get();
+  }
+
+  private static boolean hasRequestContext() {
+    return RequestContextHolder.getRequestAttributes() != null;
```
a19836, line 147: Maybe you should also use ThreadRequestContextHolder when no request attributes are available. This could be useful for future methods that run asynchronously. You can check an example in HttpReqRespUtils.java

openaev-model/src/main/java/io/openaev/database/audit/EntityDiffContext.java
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  private static RequestAttributes requestAttributes() {
+    return RequestContextHolder.getRequestAttributes();
```
a19836, line 151: same than https://github.com/OpenAEV-Platform/openaev/pull/6003/changes#r3349425653

openaev-api/src/main/java/io/openaev/aop/audit_log/AuditLogger.java
```diff
@@ -165,6 +176,89 @@ public CompletableFuture<Boolean> logAccessControlEvent(
     return CompletableFuture.completedFuture(status);
   }
 
+  // -- Diff computation (runs async on taskLoggerExecutor) --
```
a19836: It might be better to place this code in a utils class named ObjectDiffUtils, so the AuditLogger class remains uncluttered. In the future, we’ll be adding features like maximum depth level, maximum event size in bytes, truncation rules, allowlist and denylist settings, and cpu and memory usage validators, which will significantly expand this class, so this concern should be in the ObjectDiffUtils and ObjectNormalizationUtils. I'll send you through Teams the previous classes that we had related with the normalization and diff.

a19836 (dismissed):
I tested by changing the role capabilities, and it looks good. I just suggest a few improvements, described below and in my other comments.

Suggestion: Rather than sharing the EntityDiffContext, we could publish a Spring ApplicationEvent, such as EntitySnapshotEvent, which can then be handled by the API layer (AccessControlAuditLogAspect and AuditLogger) or a new EntitySnapshotEventListener in openaev-api. This approach follows the same pattern used for BaseEvent and IndexEvent, and helps maintain separation of concerns between layers.

a19836:
During my testing, I noticed that when updating "asset rules," "parameters," "taxonomy tags," "Policies," "Organizations," and "Tenants," these entities were not being logged correctly. Specifically, there was no difference recorded; I could only see the updated payload, not the previous one.

Dimfacion:
> During my testing, I noticed that when updating "asset rules," "parameters," "taxonomy tags," "Policies," "Organizations," and "Tenants," these entities were not being logged correctly. Specifically, there was no difference recorded; I could only see the updated payload, not the previous one.

Yes, only the group and role are annotated to have their diff visible as specified

openaev-api/src/main/java/io/openaev/opencti/connectors/service/LegacyOpenCTIConnectorMigration.java
Dimfacion: Sorry, I'm not sure what you mean. The existsById is specifically to avoid try to delete something that does not exists anymore

openaev-model/src/main/java/io/openaev/database/audit/EntityDiffContext.java
Dimfacion: I'm not sure this would make sense here since we're specifically in a case where no request exists (for instance, a job that would update role or groups - a case that does not exist for now)

openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
Dimfacion: I don't think we can since we need the temp variable to check if empty or not

openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
```diff
@@ -77,25 +79,32 @@ public Object auditAround(ProceedingJoinPoint joinPoint) throws Throwable {
...
             String eventScope = LogUtils.getEventScope(Action.UNAUTHORIZED);
             String eventStatus = LogUtils.getEventStatus(EventStatus.ERROR);
             JsonNode errorNode = buildErrorNode(null, ex);
 
-            logAccessControlEvent(joinPoint, accessControl, eventScope, eventStatus, errorNode);
+            logAccessControlEvent(
+                joinPoint, accessControl, eventScope, eventStatus, errorNode, snapshots);
           } else if (isActionActive) {
+            // Capture snapshots on the servlet thread before any async handoff.
+            snapshots = captureEntitySnapshots();
```
impolitepanda, line 100: Call can be mutualized with line 91 if put under if (isActive)

openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
impolitepanda: IMO, var shouldn't be used for types using generics like maps. It reduce readability for me.

There is a way to skip the inner variable, though, but I'm not sure I would advise it, as it's less readable (IMO, again):
`return Optional.of(EntityDiffContext.consumeAllSnapshots())
               .filter(m -> !m.isEmpty())
               .orElse(null);`

It's also less efficient.

On my end, I would rather recommand using an optional as a return value, instead of a null value if empty as it eliminates any NPE problem. But that's preference and not really important in this very specific case

openaev-api/src/main/java/io/openaev/opencti/connectors/service/LegacyOpenCTIConnectorMigration.java
impolitepanda: JPA's deleteById already performs a findById first. Not sure the exists check is necessary. Unless it's just to avoid having to handle the EmptyResultDataAccessException exception ?

openaev-api/src/main/java/io/openaev/utils/log/LogUtils.java
```diff
@@ -145,6 +145,13 @@ public static String buildAuthLogMessage(String eventScope, String eventStatus,
     return message;
   }
 
+  public static String buildSessionExpiredLogMessage(
```
impolitepanda, line 155: nitpick: Javadoc missing on public function. I'd argue even more important as it's a utility class

openaev-model/src/main/java/io/openaev/database/audit/EntityDiffContext.java
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  public static void clear() {
+    if (hasRequestContext()) {
+      RequestAttributes attrs = requestAttributes();
+      attrs.removeAttribute(REQUEST_ATTR_BEFORE_SNAPSHOTS, RequestAttributes.SCOPE_REQUEST);
+      attrs.removeAttribute(REQUEST_ATTR_ENTITY_SNAPSHOTS, RequestAttributes.SCOPE_REQUEST);
+      attrs.removeAttribute(REQUEST_ATTR_CLEANUP_REGISTERED, RequestAttributes.SCOPE_REQUEST);
+    } else {
+      BEFORE_SNAPSHOTS_TL.get().clear();
```
impolitepanda: I'm not sure if it's better or not, but I usually use the remove() method instead of get().clear(). What do you think ?

openaev-model/src/main/java/io/openaev/database/audit/EntityDiffContext.java
impolitepanda: I agree with @Dimfacion . I don't think your suggestion applies here. EntityDiffContext already handles async paths intentionally using the ThreadLocal fallback. Propagating request attributes into child threads would mean multiple threads sharing the same snapshot maps, which are unsynchronized LinkedHashMaps, introducing race conditions on audit data.

ThreadRequestContextHolder makes sense when you need to read request metadata (user identity, correlation ID, headers) in async threads. But this class is stateful and write-heavy, so sharing context across threads would probabbly cause bugs

openaev-model/src/main/java/io/openaev/database/audit/ModelBaseListener.java
```diff
@@ -49,8 +63,19 @@ public class ModelBaseListener {
...
+  private AuditLogProperties auditLogProperties;
   private ApplicationEventPublisher appPublisher;
 
+  /**
+   * Sets the audit log properties for determining whether diff tracking is enabled.
+   *
+   * @param auditLogProperties the audit log configuration properties
+   */
+  @Autowired
+  public void setAuditLogProperties(AuditLogProperties auditLogProperties) {
```
impolitepanda: use @RequireArgsConstructor instead, maybe ? (limiting boiler plate code like getter and setter is bettr IMO)

openaev-model/src/main/java/io/openaev/database/audit/ModelBaseListener.java
```diff
@@ -73,6 +121,49 @@ void postPersist(Object base) {
...
+    if (!shouldCaptureAuditDiff(base)) return;
+    try {
+      registerCleanupIfNeeded();
+      Map<String, Object> snapshot = buildSnapshot(instance);
+      EntityDiffContext.storeSnapshot(
+          instance.getId(),
+          new EntityDiffContext.EntitySnapshot(
+              instance.getClass().getSimpleName(), "create", null, snapshot));
+    } catch (Exception e) {
+      log.debug(
```
impolitepanda: Should this be a debug level log ? Not an error ?

openaev-model/src/main/java/io/openaev/database/audit/ModelBaseListener.java
```diff
@@ -61,10 +86,33 @@ public void setApplicationEventPublisher(ApplicationEventPublisher applicationEv
...
+  @PostLoad
+  void postLoad(Object base) {
+    if (!shouldCaptureAuditDiff(base)) return;
+    Base instance = (Base) base;
+    if (instance.getId() == null) return;
+    try {
+      registerCleanupIfNeeded();
+      EntityDiffContext.storeBefore(instance.getId(), buildSnapshot(instance));
+    } catch (Exception e) {
+      log.debug(
```
impolitepanda: Should this be a debug level log ? Not an error ?

openaev-model/src/main/java/io/openaev/database/audit/ModelBaseListener.java
```diff
@@ -73,6 +121,49 @@ void postPersist(Object base) {
...
+    try {
+      Map<String, Object> before = EntityDiffContext.getBefore(instance.getId());
+      if (before == null) return;
+      Map<String, Object> after = buildSnapshot(instance);
+      EntityDiffContext.storeSnapshot(
+          instance.getId(),
+          new EntityDiffContext.EntitySnapshot(
+              instance.getClass().getSimpleName(), "update", before, after));
+    } catch (Exception e) {
+      log.debug(
```
impolitepanda: Should this be a debug level log ? Not an error ?

openaev-model/src/main/java/io/openaev/database/audit/ModelBaseListener.java
```diff
@@ -92,15 +183,31 @@ void postUpdate(Object base) {
...
+    if (!shouldCaptureAuditDiff(base)) return;
+    try {
+      registerCleanupIfNeeded();
+      Map<String, Object> snapshot = buildSnapshot(instance);
+      EntityDiffContext.storeSnapshot(
+          instance.getId(),
+          new EntityDiffContext.EntitySnapshot(
+              instance.getClass().getSimpleName(), "delete", snapshot, null));
+    } catch (Exception e) {
+      log.debug(
```
impolitepanda: Should this be a debug level log ? Not an error ?

openaev-api/src/main/java/io/openaev/opencti/connectors/service/LegacyOpenCTIConnectorMigration.java
Dimfacion: Yes, since this is a migration, if the ID does not exists, there is just nothing to do and an exception wouldn't really make sense

openaev-model/src/main/java/io/openaev/database/audit/EntityDiffContext.java
Dimfacion: You're right, that should be better

impolitepanda:
Test ok

openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
Dimfacion: As seen together, that shouldn't be needed

impolitepanda (dismissed):

openaev-model/src/main/java/io/openaev/database/audit/EntityDiffContext.java
Dimfacion: See discussion here : https://github.com/OpenAEV-Platform/openaev/pull/6003/changes#r3349425653

openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
Dimfacion: I made a change as the fact that the map is empty is already accounted for when it's used meaning that we don't need the test and can remove the temp var

impolitepanda (approved):

PR #6004, opened 2026-06-01
[backend] fix(settings): add unique constraint on parameters to prevent duplicate keys

### LINKED ISSUE #6005: [backend] Add unique constraint on parameters table to prevent duplicate keys
## Description

The `parameters` table currently allows duplicate `parameter_key` entries, which can lead to inconsistent settings behavior.

## Expected behavior

A `parameter_key` should be unique:
- Per tenant (when `tenant_id IS NOT NULL`)
- Globally for platform-scoped settings (when `tenant_id IS NULL`)

## Fix

Add a Flyway migration that:
1. Cleans up existing duplicate rows
2. Adds a partial unique index for tenant-scoped settings `(parameter_key, tenant_id)`
3. Adds a partial unique index for platform-scoped settings `(parameter_key)` where `tenant_id IS NULL`

## Related

PR: #6004

RomuDeuxfois:


antoinemzs (approved):
good spot

PR #6012, opened 2026-06-02
fix(auth): remove false-positive error logs in group mapping (#5979)

### LINKED ISSUE #5979: SAML provisioning does not assign tenant to user on first login
## Description

When a user authenticates via SAML SSO, no row is written to `users_tenants`, resulting in a white screen after login.

## Current workaround

Manually inserting the tenant ID for all users in the `users_tenants` table.

## Steps to reproduce

1. Configure SAML SSO provider with `OPENAEV_PROVIDER_AUR_TENANT_ID` set
2. User authenticates via SSO for the first time
3. Authentication succeeds but no `users_tenants` row is created
4. User sees a white/blank screen

RomuDeuxfois:
## Problem

`UserMappingService.mapCurrentUserWithGroup()` uses a nested loop (mappings × token groups) and logs an `ERROR` for every non-matching pair. With 4 mappings and 2 token groups, this produces **6 false-positive ERROR logs per login** — even though group assignment works correctly.

This has been reported by NAB and causes alert noise in production monitoring.

## Root Cause

The `else` branch (line 53-54) fires on every `(mapping, tokenGroup)` combination that doesn't match, instead of only firing when a token group has **zero matches** across all mappings.

```java
// BEFORE — nested loop with wrong error placement
for (GroupMapping mapping : groupMappings) {
    for (String role : groupsFromToken) {
        if (idpGroup.equals(role)) {
            // ✅ match
        } else {
            log.error("No corresponding group found for group %s", role); // ❌ fires on every non-matching PAIR
        }
    }
}
```

## Fix

- **Remove the nested loop** — use `contains()` instead of iterating every token group per mapping
- **Move the "no match" log** to after all mappings are processed, and only for token groups that truly have no mapping
- **Downgrade to `log.debug()`** — unmapped token groups are expected behavior (users often belong to more IDP groups than are configured)
- **Improve log messages** — use structured `{}` placeholders instead of `String.format()`, and include both the IDP group and user group in error messages for easier debugging

## Impact

- **Functional**: None — group assignment logic is unchanged
- **Logs**: Eliminates N×M false-positive ERROR logs per SSO login
- **Monitoring**: Reduces alert noise for customers with multiple group mappings

## Testing

Existing tests in `UserMappingServiceTest` continue to pass — the fix does not change assignment behavior, only logging.

Closes #5979


camrrx (approved):

RomuDeuxfois:
@copilot handle build failed thanks

PR #6024, opened 2026-06-02
fix(executor-agent): fix relation agent executor (#5985)

### LINKED ISSUE #5985: New executors are not linked to their agent.
## Description

When connecting and starting an executor for the first time, it is not linked to its agents (`id_executor = null` in the `agent` table). As a result, the executor icon is replaced by "Invalid" in both the Endpoint and Agent views:
**Endpoint** (executor icon → "Unknown/Inconnu") → **Agent** (executor icon → ""Unknown/Inconnu").

If the executor is stopped and restarted, the associations are correctly re-established.

Bug found on a feature branch instance.

## Environment

1. OS: Mac OS
2. OpenAEV version: 2.260529.0
3. OpenAEV client: frontend
4. No feature flag activated
5. Other environment details:
## Reproducible Steps

1. Deploy a fresh OpenAEV instance on a feature branch or localy.
2. Connect an executor for the first time. Integrations / Executors
3. Start the executor.
4. Navigate to the Endpoint view — observe the executor icon is replaced by "Unknown/Inconnu".
5. Navigate to the Agent view — observe the same "Unknown/Inconnu" icon.
6. Check the `agent` table in the database — `id_executor` is `null`.
7. Run an atomic test — observe the error in the logs.

## Expected Output

On first connection and start, the executor should be correctly linked to its agents (`id_executor` populated in the `agent` table), and the executor icon should be displayed properly in both the Endpoint and Agent views. Atomic tests should execute without errors.

## Actual Output

The executor is not linked to its agents on first start (`id_executor = null`). The executor icon is replaced by "Unknown/Inconnu" in both the Endpoint and Agent views. Stopping and restarting the executor resolves the issue and associations are correctly created.

When running an atomic test, the following error is thrown:

```
ERROR Cannot find the executor for the agent nt authority\system from the asset oaev-test-win-06
```

## Additional information

- The issue affects **all executors**: PaloAlto, CrowdStrike, SentinelOne, and Tanium.
- The issue only occurs on the **first** connection/start of an executor.
- A restart (stop + start) of the executor is a working workaround.
- The error when running an atomic test confirms that the executor-agent link is missing at runtime.
- Suspected cause: the agent registration or association logic may not be triggered properly during the initial executor startup sequence.

## Screenshots (optional)

Image

Image

Image

EvaE-Filigran:
### Proposed changes

* Fix agents being persisted with `agent_executor = NULL` when created by external executor integrations (PaloAltoCortex, CrowdStrike, SentinelOne, Tanium)
* Ensure the `tenantId` read-only field is populated on newly created `Executor` entities so Hibernate can resolve the composite `@JoinColumnsOrFormulas` join

### Root cause

The `Agent → Executor` relationship uses a **composite join**:

```java
@JoinColumnsOrFormulas({
    @JoinColumnOrFormula(
        column = @JoinColumn(name = "agent_executor", referencedColumnName = "executor_id")),
    @JoinColumnOrFormula(
        formula = @JoinFormula(value = "tenant_id", referencedColumnName = "tenant_id"))
})
private Executor executor;
```

Hibernate resolves this by matching the agent's `tenant_id` with the executor's `tenant_id` **and** `agent_executor` with `executor_id`.

The `Executor` entity has two tenant-related fields:
```java
@ManyToOne
@JoinColumn(name = "tenant_id", updatable = false, nullable = false)
private Tenant tenant;           // ← the relation (used for write)

@Column(name = "tenant_id", insertable = false, updatable = false)
private String tenantId;         // ← read-only scalar (used by @JoinFormula resolution)
```

When creating a new executor, only `tenant` was set but **not** `tenantId`. After `save()`, the `tenantId` field remained `null` in memory until the entity was re-read from DB. When the scheduler passed this executor object to `agent.setExecutor(executor)`, Hibernate couldn't match the composite join (`referencedColumnName = "tenant_id"` was null) → `agent_executor = NULL`.

### Fix

Set `tenantId` explicitly when creating a new executor, so the in-memory entity is fully consistent:

```java
Tenant tenant = new Tenant(TenantContext.getCurrentTenant());
executor.setTenant(tenant);
executor.setTenantId(tenant.getId());  // ← ensures composite join resolves immediately
```

### Testing Instructions

1. Configure a PaloAltoCortex (or any external) executor integration
2. Wait for the first scheduler tick to sync agents
3. Verify: `SELECT agent_id, agent_executor FROM agents WHERE agent_executor IS NULL` — should return no results for agents linked to external executors
4. Verify on the **first** sync (not only after the second tick)

### Related issues

* Closes #5985 

### Checklist

### Further comments

One-line fix. The `tenantId` field is mapped as `insertable = false, updatable = false` (read-only from DB perspective), but its **in-memory value** is used by Hibernate to resolve `@JoinFormula(value = "tenant_id", referencedColumnName = "tenant_id")` when persisting an `Agent`. Without it set, the composite FK is unresolvable → NULL.

### DIFF
--- a/openaev-api/src/main/java/io/openaev/executors/ExecutorService.java
+++ b/openaev-api/src/main/java/io/openaev/executors/ExecutorService.java
@@ -167,9 +167,11 @@ public class ExecutorService extends AbstractConnectorService<Executor, Executor
     Executor executor =
         executorRepository.findByIdAndTenantId(id, TenantContext.getCurrentTenant()).orElse(null);
     if (executor == null) {
+      Tenant tenant = new Tenant(TenantContext.getCurrentTenant());
       executor = new Executor();
       executor.setId(id);
-      executor.setTenant(new Tenant(TenantContext.getCurrentTenant()));
+      executor.setTenant(tenant);
+      executor.setTenantId(tenant.getId());
     }
 
     executor.setName(name);
--- a/openaev-api/src/test/java/io/openaev/executors/ExecutorServiceTest.java
+++ b/openaev-api/src/test/java/io/openaev/executors/ExecutorServiceTest.java
@@ -1,0 +1,136 @@
+package io.openaev.executors;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.*;
+
+import io.openaev.context.TenantContext;
+import io.openaev.database.model.Executor;
+import io.openaev.database.repository.ExecutorRepository;
+import java.util.Optional;
+import org.junit.jupiter.api.*;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.ArgumentCaptor;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+class ExecutorServiceTest {
+
+  @Mock private ExecutorRepository executorRepository;
+  @InjectMocks private ExecutorService executorService;
+
+  @BeforeEach
+  void setUp() {
+    TenantContext.setCurrentTenant("tenant-001");
+  }
+
+  @AfterEach
+  void tearDown() {
+    TenantContext.clearCurrentTenant();
+  }
+
+  @Nested
+  @DisplayName("register - Composite join fix (executor_id + tenant_id)")
+  class Register {
+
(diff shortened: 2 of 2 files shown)


corinnekrych (changes requested):
I've tested your PR on default teant it does work I can register multiple executors.
when i create a new tenant, the starter pack create the build in executor.
However when i try to register a new executor on my new tenant => nothing happen

Let's sync tomorrow

corinnekrych (approved):

PR #6034, opened 2026-06-03
feat(audit-log): adding more user fields to be redacted (#6032)

### LINKED ISSUE #6032: US.5 PII-Configurable User Identity
As a platform administrator, I want to the user PIIS to be either hashed or redacted so that I can comply with data minimisation requirements (GDPR, internal PII policies).

a19836:
As a platform administrator, I want to the user PIIS to be either hashed or redacted so that I can comply with data minimisation requirements (GDPR, internal PII policies).

### Proposed changes

* Adding more fields to ObjectRedactionUtils.java file

### Testing Instructions

1. Change some user fields and check if in the console logs the sensitive fields were redacted.

### Related issues

* #6032 

### Checklist

### Further comments

Instead of redacting this fields, we should just remove (not include) them in the logs. This way, we avoid adding unnecessary dummy data and prevent overloading the logs.

### DIFF
--- a/openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
+++ b/openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
@@ -88,7 +88,6 @@ public class AccessControlAuditLogAspect {
             String eventStatus = LogUtils.getEventStatus(EventStatus.ERROR);
             JsonNode errorNode = buildErrorNode(null, ex);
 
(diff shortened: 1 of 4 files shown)


Dimfacion (changes requested):
This works for the users but not for the players. I think we should add ResourceType.PLAYER in USER_ENTITY_TYPES

Dimfacion (changes requested):
There is still a small place where a PII (the first name) is visible, it's in the message explaining what was done. For instance, after creating a new player, I can see "message" : "creates Player `Gael`"

Dimfacion (approved):

PR #6035, opened 2026-06-03
feat(chaining): updated permission logic to handle grants on chained Scenario (#4824)

heditar:
### Proposed changes

* Updating the PermissionService to take the opermission from the parent Simulation/scenario when a workflow/step/condition is passed
* Updating some APis to have correct type

### Testing Instructions

1. Create a chained workflow with one action and one event
2. Go to the db and note the id of the workflow, condition and step
3.  Create a user and a group, and only give the permission to the chained scenario using a grant
4. Log with the new user and call the apis for workflow/step/condition using the id you got earlier, you should have access with your new user

### Related issues

* #4824 
*

### Checklist


openaev-api/src/main/java/io/openaev/service/PermissionService.java
```diff
@@ -240,10 +249,35 @@ private Target resolveTarget(
       Action parentAction = (action == Action.READ) ? Action.READ : Action.WRITE;
       return new Target(
           evaluation.getParentResourceId(), evaluation.getParentResourceType(), parentAction);
+    } else if (resourceType == ResourceType.WORKFLOW) {
+      Workflow workflow = workflowService.findById(resourceId);
+      // delete/write on a workflow is treated as a write on its parent (simulation/scenario)
+      Action parentAction = (action == Action.READ) ? Action.READ : Action.WRITE;
+      return resolveWorkflowTarget(workflow, parentAction);
```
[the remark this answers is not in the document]
heditar: That is intented

openaev-api/src/main/java/io/openaev/service/PermissionService.java
```diff
@@ -240,10 +249,35 @@ private Target resolveTarget(
...
+    } else if (resourceType == ResourceType.STEP) {
+      Step step = stepService.findById(resourceId);
+      Action parentAction = (action == Action.READ) ? Action.READ : Action.WRITE;
+      return resolveWorkflowTarget(step.getWorkflow(), parentAction);
```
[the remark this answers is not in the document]
heditar: That is intented

openaev-api/src/test/java/io/openaev/service/PermissionServiceTest.java
```diff
@@ -385,9 +417,11 @@ public void test_hasPermission_update_workflow_step_WHEN_has_manage_assessment()
   }
 
   @Test
-  public void test_hasPermission_delete_workflow_step_WHEN_has_delete_assessment() {
+  public void test_hasPermission_delete_workflow_step_WHEN_has_manage_assessment() {
+    // DELETE on workflow/step/condition maps to WRITE on the parent (simulation/scenario),
+    // so MANAGE_ASSESSMENT (which grants WRITE on simulation/scenario) is required.
     User user = getUser(USER_ID, false);
-    user.setGroups(List.of(getGroup(Capability.DELETE_ASSESSMENT)));
+    user.setGroups(List.of(getGroup(Capability.MANAGE_ASSESSMENT)));
```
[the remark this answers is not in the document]
heditar: That is intented, delete on worklow is an update on a scenario/simulation

openaev-api/src/main/java/io/openaev/api/chaining/StepApi.java
```diff
@@ -70,15 +70,21 @@ public StepOutput createStep(@Valid @RequestBody StepInput input) throws Chainin
...
     return toOutput(stepService.findStepTemplateById(stepId));
   }
 
   @Operation(summary = "List step templates by workflow")
   @ApiResponses({@ApiResponse(responseCode = "200", description = "Step templates retrieved")})
-  @AccessControl(actionPerformed = Action.READ, resourceType = ResourceType.SIMULATION_OR_SCENARIO)
+  @AccessControl(
+      resourceId = "#workflowId",
+      actionPerformed = Action.READ,
+      resourceType = ResourceType.WORKFLOW)
```
savacano28, line 87: I'm not sure if the resource type is workflow because we are using StepService and returning steps.

openaev-api/src/main/java/io/openaev/api/chaining/ConditionApi.java
```diff
@@ -71,7 +74,10 @@ public EventOutput findById(@PathVariable String conditionId) {
       summary = "Get condition trees by workflow",
       description = "Lists all root conditions for a given workflow")
   @ApiResponses({@ApiResponse(responseCode = "200", description = "Condition trees retrieved")})
-  @AccessControl(actionPerformed = Action.READ, resourceType = ResourceType.SIMULATION_OR_SCENARIO)
+  @AccessControl(
+      resourceId = "#workflowId",
+      actionPerformed = Action.READ,
+      resourceType = ResourceType.WORKFLOW)
   @GetMapping(params = "workflow_id")
```
savacano28, line 81: I'm not sure if the resource type is workflow because we are querying ConditionService and returning conditions.

openaev-api/src/main/java/io/openaev/service/PermissionService.java
```diff
@@ -50,14 +53,20 @@ public class PermissionService {
           ResourceType.NOTIFICATION_RULE,
           ResourceType.INJECTOR_CONTRACT,
           ResourceType.OBJECTIVE,
-          ResourceType.EVALUATION);
+          ResourceType.EVALUATION,
+          ResourceType.WORKFLOW,
+          ResourceType.STEP,
```
savacano28, line 58: I'm not sure if we ultimately need Step and Condition as a resource, knowing that Step and Condition will only exist within a workflow and cannot be shared between workflows. I think that just the workflow should be enough.

savacano28:
@heditar Maybe could you check also ChainingAPI? we have findAll with Scenario_Or_Simulation ressource. Is that still valid? ty!
Thanks for the changes!

openaev-api/src/main/java/io/openaev/api/chaining/StepApi.java
heditar: We are actually reading a workflow as we get all the step for a workflow

openaev-api/src/main/java/io/openaev/api/chaining/ConditionApi.java
heditar: The parameter passed is the workflow id and the permission should be applied based on this workflow

openaev-api/src/main/java/io/openaev/service/PermissionService.java
heditar: I agree, we would need some work on the api and the path, there is a TODO  in the code. I think if possible we should remove that. I would take the task to review our api path and propose a new solution. If possible I would like the apis to only use scneario/simulation id

heditar:
> @heditar Maybe could you check also ChainingAPI? we have findAll with Scenario_Or_Simulation ressource. Is that still valid? ty! Thanks for the changes!

Still valid but there is another stpe to do for RBAC to filter these kind of API, we have another issue for it

openaev-api/src/main/java/io/openaev/service/chaining/WorkflowService.java
```diff
@@ -67,6 +67,13 @@ public Workflow getWorkflowByIdAndStatus(
                         + workflowId));
   }
 
+  public Workflow findById(@NotBlank final String workflowId) {
```
savacano28, line 70: Question: We should use tenantId here?

openaev-api/src/main/java/io/openaev/service/PermissionService.java
savacano28: We have an issue for that?

savacano28:
Tests:

* Workflow/step/conditions: **OK**
* Access ✔️ 
* Update✔️ 
* Launch✔️

savacano28 (dismissed):

openaev-api/src/main/java/io/openaev/service/chaining/WorkflowService.java
heditar: yes that should be done in another pr

openaev-api/src/main/java/io/openaev/service/PermissionService.java
heditar: no but will be created this week

savacano28:
Tests:
* I have created a group, user and I have granted with my Scenario/Simulation and I can get correctly: Access, Manage, Launch but I have some errors for endpoints/asset-groups (403), maybe here we should allow access to endpoints/assets/asset-groups related to these simulations/scenarios. 
image

If we create a role with these capabilities: assets, I dont have 403, but I m not sure if this is the correct way to grant the simulations/scenarios?

heditar:
> Tests:
> 
> * I have created a group, user and I have granted with my Scenario/Simulation and I can get correctly: Access, Manage, Launch but I have some errors for endpoints/asset-groups (403), maybe here we should allow access to endpoints/assets/asset-groups related to these simulations/scenarios.
> 
> image
> If we create a role with these capabilities: assets, I dont have 403, but I m not sure if this is the correct way to grant the simulations/scenarios?

that is intended, we should use existing api with Scenario/Simulation id in the path

savacano28 (approved):

Issue #5718, opened 2026-05-07
test(multi-tenancy): tenant isolation integration test


corinnekrych:
# Tenant Isolation Test Coverage

This sis the list of API test that need a Tenant isolation test

- [x] `ScenarioApiTest` (on branch `issue/multitenant-groups-roles-settings-isolation`)
- [x] `TenantGroupApiTest` (on branch `issue/multitenant-groups-roles-settings-isolation`)
- [x] `ExerciseApiTest` — SIMULATION resource type
- [x] `AssetGroupApiTest` — ACCESS_ASSETS / MANAGE_ASSETS
- [x] `AtomicTestingApiTest` — ATOMIC_TESTING resource type
- [x] `PayloadApiTest` — PAYLOAD resource type
- [ ] `ScenarioInjectApiTest` — INJECT (parent=SCENARIO)
- [ ] `SimulationInjectApiTest` — INJECT (parent=SIMULATION)
- [x] `EndpointApiTest` — ACCESS_ASSETS / MANAGE_ASSETS -> DAMIEN
- [x] `TeamApiTest` — open resource (TEAM in RESOURCES_OPEN) -> JOHANAH
- [x] `CustomDashboardApiTest` -> DAMIEN
- [x] `CustomDashboardWidgetApiTest` -> DAMIEN
- [x] `FindingApiTest` -> Damien
- [x] `NotificationRuleApiTest` -> JOHANAH
- [x] `TagRuleApiTest` -> JOHANAH
- [x] `CollectorApiTest` -> Damien
- [x] `ConnectorInstanceApiTest` -> Damien
- [x] `ExecutorApiTest` -> Damien
- [x] `InjectorApiTest` -> Damien
- [ ] `ScenarioTeamApiTest`
- [ ] `ScenarioSimulationApiTest`
- [ ] `ScenarioImportApiTest`
- [ ] `ExerciseTeamApiTest`
- [x] `ReportApiTest` -> JOHANAH
- [x] `DashboardApiTest` -> DAMIEN
- [ ] `ChallengeApiTest` -> JOHANAH
- [ ] `ChannelApiTest` -> JOHANAH
- [ ] `MapperApiTest` -> JOHANAH
- [ ] `VariableApiTest` -> JOHANAH
- [ ] `DomainApiTest` -> JOHANAH
- [ ] `InjectExpectationTraceApiTest`
- [ ] `ExpectationApiTest`
[...]

PR #6036, opened 2026-06-03
test(multi-tenancy): mapper API for multi-tenancy (#5718)

johanah29:
### Proposed changes
Add tenant isolation test cases for mappers.

### Testing Instructions


corinnekrych (changes requested):
@johanah29 similar to the work done in your other PR: will it be possible to move the TenantContext outside the Service layer? For that you need to pass it as param, check who's calling (if API level move the TenantContext usage here, if background job take the tenatnId from the stored object's tenantId column)

corinnekrych (approved):

PR #6042, opened 2026-06-03
test(multi-tenancy): variable API for multi-tenancy (#5718)

johanah29:
### Proposed changes
Add tenant isolation test cases for variables.

### Testing Instructions


openaev-model/src/main/java/io/openaev/database/repository/VariableRepository.java
```diff
@@ -11,6 +13,10 @@
 public interface VariableRepository
     extends CrudRepository<Variable, String>, JpaSpecificationExecutor<Variable> {
 
+  Optional<Variable> findByIdAndExerciseId(@NotNull String variableId, @NotNull String exerciseId);
```
corinnekrych, line 16: he new methods improve on findById by scoping through exerciseId/scenarioId, but consider making tenant isolation explicit by adding a tenantId parameter (e.g., findByIdAndExerciseIdAndTenantId) to avoid relying on implicit filter activation and to add a defense-in-depth layer.

corinnekrych (dismissed):
would you mind revisit the query to be more explicit?

RomuDeuxfois (approved):

PR #6086, opened 2026-06-04
test(multi-tenancy): player api for multi-tenancy (#5718)

johanah29:
### Proposed changes
Add tenant isolation test cases for players.

### Testing Instructions


openaev-api/src/main/java/io/openaev/rest/user/PlayerService.java
```diff
@@ -115,7 +115,9 @@ public User createPlayer(@Valid @RequestBody PlayerInput input) {
       String userId = existingUser.get().getId();
       tenantUserService.attachToTenant(userId, TenantContext.getCurrentTenant());
       // Reload user after @Modifying queries cleared the persistence context
-      return userRepository.findById(userId).orElseThrow();
+      return userRepository
+          .findByIdAndTenantId(userId, TenantContext.getCurrentTenant())
```
corinnekrych: would it be possible to move TenantContext.getCurrentTenant() out of Service layer

corinnekrych (changes requested):

PR #6096, opened 2026-06-05
test(multi-tenancy): report api for multi-tenancy (#5718)

johanah29:
### Proposed changes
Add tenant isolation test cases for reports.

### Testing Instructions

### DIFF
--- a/openaev-api/src/main/java/io/openaev/rest/report/service/ReportService.java
+++ b/openaev-api/src/main/java/io/openaev/rest/report/service/ReportService.java
@@ -2,6 +2,7 @@ package io.openaev.rest.report.service;
 
 import static java.time.Instant.now;
 
+import io.openaev.context.TenantContext;
 import io.openaev.database.model.*;
 import io.openaev.database.model.Report;
 import io.openaev.database.model.ReportInformation;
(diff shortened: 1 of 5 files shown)


openaev-api/src/main/java/io/openaev/rest/report/service/ReportService.java
```diff
@@ -39,7 +42,7 @@ public Report report(@NotNull final UUID reportId) {
   public Report reportFromSimulation(
       @NotBlank final String simulationId, @NotNull final UUID reportId) {
     return this.reportRepository
-        .findByIdAndExercise_Id(reportId, simulationId)
+        .findByIdAndExercise_IdAndTenantId(reportId, simulationId, TenantContext.getCurrentTenant())
```
corinnekrych: could TenantContext.getCurrentTenant() be out of service layer?

openaev-model/src/main/java/io/openaev/database/repository/ReportRepository.java
```diff
@@ -17,11 +18,22 @@ public interface ReportRepository
   @NotNull
   Optional<Report> findById(@NotNull final UUID id);
 
+  @Query("SELECT r FROM Report r JOIN r.exercise e WHERE r.id = :id AND e.tenant.id = :tenantId")
```
corinnekrych, line 21: why do we need native query  here?

johanah29: Report entity does not have a tenant_id column. @Query avoids having a PropertyReferenceException that would be raised by spring data when searching for tenantId on Report

RomuDeuxfois (approved):

PR #6097, opened 2026-06-05
test(audit_log): e2e testing for login and logout successfully with audit (#6092)

### LINKED ISSUE #6092: test(audit-logging): add e2e test for the login action

---

## Description

End-to-end integration test covering the frontend login:

1. With Playwright
2. Execute a login action
3. Then logout
4. Check if login action was logged in the backend-api's console (console transport must be active)
Note: Because we don't have yet the audit-log search API, we need to perform this check manually.

## Definition of Done

- [ ] was login and logout actions logged into console

a19836:
## Description

End-to-end integration test covering the frontend login:

1. With Playwright
2. Execute a login action
3. Then logout
4. Check **manually** if login action was logged in the backend-api's console (console transport must be active)
Note: Because we don't have yet the audit-log search API, we need to perform this **check manually.**

### Proposed changes

Create a test file to comply with the requirements in the related issue and above description.

### Testing Instructions

1. Run one of the following commands: 
`yarn test:e2e tests_e2e/tests/aop/audit_log/login-audit.spec.ts --project="Google Chrome" --no-deps`
 or
 `yarn playwright test tests_e2e/tests/aop/audit_log/login-audit.spec.ts --project="Google Chrome" --no-deps`
 or
 `npx playwright test tests_e2e/tests/aop/audit_log/login-audit.spec.ts --project="Google Chrome" --no-deps`

### Related issues

* #6092

## Definition of Done

- [x] was login and logout actions logged into console

### DIFF
--- a/openaev-front/tests_e2e/tests/aop/audit_log/login-audit.spec.ts
+++ b/openaev-front/tests_e2e/tests/aop/audit_log/login-audit.spec.ts
@@ -1,0 +1,58 @@
+import { expect } from '@playwright/test';
+
+import { test } from '../../../fixtures/baseFixtures';
+import LoginPage from '../../../model/login.page';
+import appUrl, { tenantUrl } from '../../../utils/url';
+
+test.describe('Authentication flow', () => {
+  test.use({
+    storageState: {
+      cookies: [],
+      origins: [],
+    },
+  });
+
+  test('should login and logout successfully', async ({ page }) => {
+    // -- ARRANGE --
+    const loginPage = new LoginPage(page);
+    const username = process.env.E2E_USERNAME ?? 'admin@openaev.io';
+    const password = process.env.E2E_PASSWORD ?? 'admin';
+
+    await page.goto(appUrl());
+    await expect(loginPage.getLoginPage()).toBeVisible();
+
+    // -- ACT --
+    await loginPage.getLoginInput().fill(username);
+    await loginPage.getPasswordInput().fill(password);
+    await loginPage.getSignInButton().click();
+
+    // Ensure we are on an admin page where the top bar account menu is rendered.
+    await page.goto(tenantUrl('/admin'));
+    await expect(page).toHaveURL(/\/admin(?!\/login)/);
+
+    // Trigger CsrfFilter once to ensure XSRF-TOKEN cookie exists, then logout with matching header.
+    await page.evaluate(async () => {
+      await fetch('/api/scenarios/search', {
+        method: 'POST',
+        credentials: 'include',
+        headers: { 'Content-Type': 'application/json' },
+        body: JSON.stringify({}),
+      }).catch(() => {});
+
+      const tokenCookie = document.cookie
+        .split('; ')
+        .find(cookie => cookie.startsWith('XSRF-TOKEN='));
+      const token = tokenCookie ? decodeURIComponent(tokenCookie.split('=')[1]) : '';
+
+      await fetch('/logout', {
+        method: 'POST',
+        credentials: 'include',
+        headers: token ? { 'X-XSRF-TOKEN': token } : {},
+      });
+    });
+    await page.goto(appUrl());
+
+    // -- ASSERT --
+    await expect(loginPage.getSignInButton()).toBeVisible();
+  });
+});
--- a/openaev-front/tests_e2e/tests/aop/audit_log/login-audit.spec.ts
+++ b/openaev-front/tests_e2e/tests/aop/audit_log/login-audit.spec.ts
@@ -15,7 +15,7 @@ test.describe('Authentication flow', () => {
   test('should login and logout successfully', async ({ page }) => {
     test.info().annotations.push({
       type: 'manual',
-      description: 'Verify backend-api console contains audit log entries for login and logout (console transport enabled).',
+      description: 'Please verify if backend-api console contains audit log entries for login and logout (console transport enabled).',
     });
 
     // -- ARRANGE --
--- a/openaev-front/tests_e2e/tests/aop/audit_log/login-audit.spec.ts
+++ b/openaev-front/tests_e2e/tests/aop/audit_log/login-audit.spec.ts
@@ -13,6 +13,11 @@ test.describe('Authentication flow', () => {
   });
 
   test('should login and logout successfully', async ({ page }) => {
+    test.info().annotations.push({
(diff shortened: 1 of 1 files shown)


openaev-front/tests_e2e/tests/aop/audit_log/login-audit.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+    await expect(loginPage.getLoginPage()).toBeVisible();
+
+    // -- ACT --
+    await loginPage.getLoginInput().fill(username);
+    await loginPage.getPasswordInput().fill(password);
+    await loginPage.getSignInButton().click();
+
+    // Ensure we are on an admin page where the top bar account menu is rendered.
+    await page.goto(tenantUrl('/admin'));
+    await expect(page).toHaveURL(/\/admin(?!\/login)/);
```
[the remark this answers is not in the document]
a19836: done in the next commit

openaev-front/tests_e2e/tests/aop/audit_log/login-audit.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+      origins: [],
+    },
+  });
+
+  test('should login and logout successfully', async ({ page }) => {
+    // -- ARRANGE --
+    const loginPage = new LoginPage(page);
+    const username = process.env.E2E_USERNAME ?? 'admin@openaev.io';
+    const password = process.env.E2E_PASSWORD ?? 'admin';
+
```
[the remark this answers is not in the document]
a19836: done in next commit

openaev-front/tests_e2e/tests/aop/audit_log/login-audit.spec.ts
```diff
@@ -0,0 +1,63 @@
+import { expect } from '@playwright/test';
+
+import { test } from '../../../fixtures/baseFixtures';
```
RomuDeuxfois, line 3: **question**: I am not sure to understand where are you testing the audit part here

a19836: As described in step 4 of this PR, the goal is to review the backend API logs. You need to simulate login and logout requests from the frontend, then check the API logs to confirm that these events are recorded. This is a straightforward test.

RomuDeuxfois: Yes, but I don't see in your code when are you checking the API logs is here ?

Dimfacion: Was this discussed directly ? Because I'm not sure to understand the relevance of this test if we can't look at the audit logs, especially since there is already login.spec.ts that does test the login page (more succinctly though but still)

a19836: This was discussed during one of our calls when everyone was present, and Corinne mentioned that we needed two frontend tests: one simple test, like a simple CRUD action, and another more complex one, such as scenario or simulation creation with injects. The plan was to use Playwright to simulate user behavior from the frontend and check **manually** if audit-logging was functioning. Since we already had e2e tests in the backend, we also needed two tests with Playwright in the frontend. The process involved simulating user actions and then **manually** checking the logs in the backend. I confirmed this twice during the call and emphasized that it was only a simulation in the frontend, with **manual** log verification in the backend — not an automated process. We all agreed on this, including both of you.

a19836: I didn't use the login.spec.ts file because this test does more things, this is: login action, then lists scenarios and then logout. This way, we can cover 3 audit-loggings: one for the login, other for list scenarios and another for the logout action.

a19836: Perhaps it would be a good idea to rename this file to something more intuitive?

Dimfacion: Sorry, I must have misunderstood when we agreed then because I feel like this defeats the point of CI if we have to check manually if the audit logs have been written.
Also, even if we do a search and a logout, why not enrich the existing test ?

I did a small PR here to demonstrate how we could get the audit logs from playwright, even without the feature of search on the audit logs : https://github.com/OpenAEV-Platform/openaev/pull/6256 It's working in local but might need adjustments to work during CI.
Could you have a look and see if that could maybe be interesting here ?

a19836: While I think your PR is a good idea - adding an endpoint to access backend logs in the test environment - the reason why I didn't do this before, was because I feel this approach is a workaround for something that wasn't fully planned from the start. I'm not sure if we should make backend changes just because some frontend tests require it. 

Playwright tests are meant to simulate user actions, or at least that's my understanding of why we're using them.
Given this, I think we should reconsider, as this solution doesn't seem necessary at this stage.
These tests will be more relevant once the audit-log search feature is implemented in the backend, allowing us to check the logs properly. This PR was also originally implemented to be updated later to use the search endpoint, which isn't available yet.
So, my question is: 
**Would it make more sense to postpone this PR until the audit-log search feature is ready, rather than adding new functionality to the backend API?**

Dimfacion: While I agree that this doesn't make sense to do this test until we have a proper way to check the audit log was created when doing those actions, I want to point out that what I added is merely activating a subfeature of spring actuator - a lib that is already present but deactivated on prod environment.

a19836: How would you both like to move forward?
**Option A:** Use a new endpoint in the backend-api only for testing.
**Option B**: Wait until audit-log search feature be available.

Option A or B?

Votes:
A: 
B: 1 vote (JP)

corinnekrych (changes requested):
The new test duplicates the login steps (fill username/password, click sign in) instead of reusing the existing login() utility from tests_e2e/utils/login.ts. It would be cleaner to call await login(page) from that shared utility and then proceed to the logout + audit verification steps

also the test should check the audit log file and expect audit entries to match user action login/logout

PR #6124, opened 2026-06-07
fix: centralize drawer content spacing (#3278)

### LINKED ISSUE #3278: Revising spacing on the Drawer component
Following feedback from product testing 1.17.0 on a spacing issue in the target header of the assistant scenario, we realized that in several places in obas, hard spacing had been added to the Drawer component ( openbas-front/src/components/common/Drawer.tsx ).

In order to homogenize the whole, we need to make sure that the spaces are on the component instead of in hard copy.
To do this, you need to check all the places where the component is used

ded-furby:
## Related issue
- Fixes #3278

## What changed
- Added explicit Drawer API props to control spacing semantics: `disableContainerPadding` and `containerMaxHeight`.
- Moved hardcoded per-call spacing (`padding: 0`, `maxHeight: '100%'`) into Drawer prop usage for ATT&CK drawers.
- Replaced inline `containerStyle` overrides in:
  - `SelectTTPsDrawer`
  - `PaginationComponentV2`
  - `ExercisePopover`

## Why
This keeps spacing defaults in `Drawer` and reduces duplicated inline overrides across usage sites.

## Validation
- `npm run check-ts` (attempted) could not run: `tsc` is not available in this environment (`sh: tsc: command not found`).
- `npm run check` script is not defined in `openaev-front/package.json` (attempted and received missing script error).

### DIFF
--- a/openaev-front/src/components/common/Drawer.tsx
+++ b/openaev-front/src/components/common/Drawer.tsx
@@ -155,7 +155,7 @@ const Drawer: FunctionComponent<DrawerProps> = ({
       </div>
       <div style={{
         ...(disableContainerPadding ? {} : { padding: '10px 20px 20px 20px' }),
-        ...(containerMaxHeight ? { maxHeight: containerMaxHeight } : {}),
+        ...(containerMaxHeight !== undefined && containerMaxHeight !== null ? { maxHeight: containerMaxHeight } : {}),
         ...containerStyle,
       }}
       >
--- a/openaev-front/src/components/common/Drawer.tsx
+++ b/openaev-front/src/components/common/Drawer.tsx
@@ -58,6 +58,8 @@ interface DrawerProps {
   PaperProps?: PaperProps;
   disableEnforceFocus?: boolean;
   containerStyle?: CSSProperties;
+  disableContainerPadding?: boolean;
+  containerMaxHeight?: CSSProperties['maxHeight'];
 }
 
 const Drawer: FunctionComponent<DrawerProps> = ({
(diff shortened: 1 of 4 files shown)


MarineLeM:
Hi @ded-furby  👋

Thank you so much for your contribution and your interest in our project, we really appreciate it! 🙏

However, after reviewing this PR, it looks like the implementation doesn't fully address the related issue. Let me clarify what needs to be done:

The goal of this issue is to audit all drawers across the project and ensure they follow a consistent pattern:

- Replace direct Material UI Drawer usage — any component calling the MUI Drawer directly should be updated to use our shared component instead: openaev-front/src/components/common/Drawer.tsx
- Remove custom padding overrides — once that's done, we need to make sure none of the drawers override the padding already defined inside openaev-front/src/components/common/Drawer.tsx. 
This can happen in two ways:
      -   Custom padding set directly on the drawer
      -   A marginTop set on the first element inside the drawer
 (I'm not sure we actually have any of those, so this might be a quick check!)
 
 Would you be up for updating the PR with these changes? Don't hesitate to reach out if anything is unclear, happy to help! 😊

openaev-front/src/components/common/Drawer.tsx
```diff
@@ -150,7 +154,8 @@ const Drawer: FunctionComponent<DrawerProps> = ({
         </div>
       </div>
       <div style={{
-        padding: '10px 20px 20px 20px',
+        ...(disableContainerPadding ? {} : { padding: '10px 20px 20px 20px' }),
+        ...(containerMaxHeight !== undefined && containerMaxHeight !== null ? { maxHeight: containerMaxHeight } : {}),
         ...containerStyle,
```
MarineLeM, line 159: Thanks for the effort here! 🙂 
I think disableContainerPadding and containerMaxHeight are redundant with the existing containerStyle prop — since it's spread last, it will actually override those two props anyway, making them unreliable. Beyond that, keeping the interface simpler is always better.

PR #6249, opened 2026-06-15
fix(multi-tenancy): disable switching to newly created tenants (#4864)

johanah29:
### Proposed changes
Disable the automatic switch to a newly created tenants.

### Testing Instructions

1. Create a new tenant
2. Confirm that you are not automatically redirected to that tenant
3. Switch between tenants manually to chack it still works properly

### DIFF
--- a/openaev-front/src/__tests__/utils/hooks/UseTenant.test.tsx
+++ b/openaev-front/src/__tests__/utils/hooks/UseTenant.test.tsx
@@ -388,7 +388,7 @@ describe('useTenant', () => {
       expect(mockFetchUserTenants).toHaveBeenCalledTimes(2);
     });
 
-    it('given_reloadWithPreferredTenantId_should_navigateToPreferredTenant', async () => {
+    it('given_reloadCalled_should_keepCurrentUrlTenant', async () => {
       // Arrange — URL points to ALPHA so setTenant is called during init
       mockExtractTenantFromUrl.mockReturnValue(TENANT_ALPHA.tenant_id);
       mockTenantsResponse([TENANT_ALPHA, TENANT_BETA]);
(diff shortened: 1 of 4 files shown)


openaev-front/src/utils/hooks/useAuth.ts
```diff
@@ -9,7 +9,7 @@ export interface UserContextType {
   userTenants: TenantOutput[] | undefined;
   currentUserTenant: TenantOutput | null;
   switchUserTenant: (tenantId: string) => Promise<void>;
-  reloadUserTenants: (newCurrentTenantId?: string) => Promise<void>;
+  reloadUserTenants: () => Promise<void>;
```
a19836: Please keep this method as before, with the parameter `newCurrentTenantId?: string`, as it is already optional. The page only redirects to a specific tenant if a parameter is provided; otherwise, it refreshes the tenants page without redirecting to a new tenant. Ideally, this function should work as follows:
- If a valid newCurrentTenantId is given, the page should redirect to that tenant.
- If no newCurrentTenantId is provided or it is invalid, the page should simply refresh.

openaev-front/src/utils/hooks/useAuth.ts
```diff
@@ -19,7 +19,7 @@ const defaultContext: UserContextType = {
   userTenants: undefined,
   currentUserTenant: null,
   switchUserTenant: async (_tenantId: string) => {},
-  reloadUserTenants: async (_newCurrentTenantId?: string) => {},
+  reloadUserTenants: async () => {},
```
a19836: same than https://github.com/OpenAEV-Platform/openaev/pull/6249/changes#r3414080603

openaev-front/src/utils/hooks/useTenant.ts
```diff
@@ -69,7 +69,7 @@ const useTenant = (me: User | undefined, logged: unknown) => {
     return true;
   }, [setTenant, location]);
 
-  const loadUserTenants = useCallback(async (newCurrentTenantId?: string) => {
+  const loadUserTenants = useCallback(async () => {
```
a19836: same than https://github.com/OpenAEV-Platform/openaev/pull/6249/changes#r3414080603

openaev-front/src/utils/hooks/useTenant.ts
```diff
@@ -80,9 +80,6 @@ const useTenant = (me: User | undefined, logged: unknown) => {
         setUserTenants(tenants);
 
         // If a preferred tenant is requested, switch to it
-        if (newCurrentTenantId && navigateToTenant(newCurrentTenantId, tenants)) {
```
a19836, line 83: Do not remove this line.
This is related with https://github.com/OpenAEV-Platform/openaev/pull/6249/changes#r3414080603

openaev-front/src/utils/hooks/useTenant.ts
```diff
@@ -105,8 +102,7 @@ const useTenant = (me: User | undefined, logged: unknown) => {
 
   useEffect(() => {
     if (me && logged) {
-      const urlTenantId = extractTenantFromUrl() ?? undefined;
-      loadUserTenants(urlTenantId);
```
a19836, line 109: Suggestion: I think it's best to keep the urlTenantId as it was before to ensure the page refreshes correctly, but I'm not completely certain. Please check if the behavior is correct with the urlTenantId, and if it is, keep this code unchanged.

openaev-front/src/admin/components/platform/tenants/tenant/TenantCreate.tsx
```diff
@@ -37,7 +37,7 @@ const TenantCreate: FunctionComponent<Props> = ({ onCreate }) => {
 
         const createdTenant = result.entities.tenants[result.result];
         onCreate(createdTenant);
-        await reloadUserTenants(createdTenant.tenant_id);
+        await reloadUserTenants();
```
a19836, line 40: same than https://github.com/OpenAEV-Platform/openaev/pull/6249/changes#r3414080603

openaev-front/src/__tests__/utils/hooks/UseTenant.test.tsx
```diff
@@ -400,44 +400,15 @@ describe('useTenant', () => {
         expect(result.current.currentUserTenant?.tenant_id).toBe(TENANT_ALPHA.tenant_id);
       });
 
-      // Arrange — reload with a preferred tenant
+      // Arrange — reload with an updated list (auto-switch is disabled)
       mockTenantsResponse([TENANT_ALPHA, TENANT_BETA, TENANT_GAMMA]);
 
       // Act
       await act(async () => {
-        await result.current.reloadUserTenants(TENANT_GAMMA.tenant_id);
```
a19836, line 408: Regarding my review https://github.com/OpenAEV-Platform/openaev/pull/6249/changes#r3414080603, I think it's best to keep the original tests and add a new test case that doesn't include the tenant id. In other words, we should have tests for the refresh functionality both with and without the tenant id.

a19836 (changes requested):
Please avoid removing the tenantId parameter from the reloadUserTenants method. This method is useful for future purposes and functions correctly with or without tenantId.

a19836 (approved):
The code appears to be correct, and based on my tests, it functions as expected.

PR #6254, opened 2026-06-15
fix(simulation): resolve infinite hang during deletion (#5886)

savacano28:
### Proposed changes

* Refact method deletebyId for simulations
* Add Delete on for foreign key exercise in exercise_email_reply_to

### Testing Instructions

1. Connect via SSO
2. Create a new tenant
3. Navigate to the tenant
4. Create a scenario and set up the required agent/player configuration
5. Execute the simulation
6. Immediately cancel the simulation
7. Once the simulation is in a canceled state, attempt to delete it
8. Observe the UI response
9. Refresh the page

### Related issues

Related to #5886 

### Checklist

### Further comments

### DIFF
--- a/openaev-api/src/main/java/io/openaev/migration/V5_20__Add_delete_cascade_exercise_email_reply_to.java
+++ b/openaev-api/src/main/java/io/openaev/migration/V5_20__Add_delete_cascade_exercise_email_reply_to.java
@@ -15,22 +15,20 @@ public class V5_20__Add_delete_cascade_exercise_email_reply_to extends BaseJavaM
           """
           DO $$
           BEGIN
+            -- Drop existing constraint if present
             IF EXISTS (
               SELECT 1 FROM pg_constraint
               WHERE conname = 'fk_exercise_id'
+                AND conrelid = 'exercise_mails_reply_to'::regclass
             ) THEN
               ALTER TABLE exercise_mails_reply_to DROP CONSTRAINT fk_exercise_id;
             END IF;
-          END $$;
-          """);
 
-      statement.execute(
-          """
-          DO $$
-          BEGIN
+            -- Re-create with ON DELETE CASCADE if not already present
             IF NOT EXISTS (
               SELECT 1 FROM pg_constraint
               WHERE conname = 'fk_exercise_id'
+                AND conrelid = 'exercise_mails_reply_to'::regclass
             ) THEN
               ALTER TABLE exercise_mails_reply_to
                 ADD CONSTRAINT fk_exercise_id
--- a/openaev-api/src/main/java/io/openaev/migration/V5_20__Add_delete_cascade_exercise_email_reply_to.java
+++ b/openaev-api/src/main/java/io/openaev/migration/V5_20__Add_delete_cascade_exercise_email_reply_to.java
@@ -1,0 +1,22 @@
+package io.openaev.migration;
+
+import java.sql.Statement;
+import org.flywaydb.core.api.migration.BaseJavaMigration;
+import org.flywaydb.core.api.migration.Context;
+import org.springframework.stereotype.Component;
+
+@Component
+public class V5_20__Add_delete_cascade_exercise_email_reply_to extends BaseJavaMigration {
+
+  @Override
+  public void migrate(Context context) throws Exception {
+    try (Statement statement = context.getConnection().createStatement()) {
+      statement.execute(
+          "ALTER TABLE exercise_mails_reply_to DROP CONSTRAINT IF EXISTS fk_exercise_id");
+      statement.execute(
+          "ALTER TABLE exercise_mails_reply_to "
+              + "ADD CONSTRAINT fk_exercise_id "
+              + "FOREIGN KEY (exercise_id) REFERENCES exercises(exercise_id) ON DELETE CASCADE");
+    }
+  }
+}
--- a/openaev-api/src/main/java/io/openaev/migration/V5_20__Add_delete_cascade_exercise_email_reply_to.java
+++ b/openaev-api/src/main/java/io/openaev/migration/V5_20__Add_delete_cascade_exercise_email_reply_to.java
@@ -12,11 +12,32 @@ public class V5_20__Add_delete_cascade_exercise_email_reply_to extends BaseJavaM
   public void migrate(Context context) throws Exception {
     try (Statement statement = context.getConnection().createStatement()) {
       statement.execute(
-          "ALTER TABLE exercise_mails_reply_to DROP CONSTRAINT IF EXISTS fk_exercise_id");
+          """
+          DO $$
+          BEGIN
+            IF EXISTS (
+              SELECT 1 FROM pg_constraint
+              WHERE conname = 'fk_exercise_id'
+            ) THEN
+              ALTER TABLE exercise_mails_reply_to DROP CONSTRAINT fk_exercise_id;
+            END IF;
+          END $$;
+          """);
+
       statement.execute(
-          "ALTER TABLE exercise_mails_reply_to "
-              + "ADD CONSTRAINT fk_exercise_id "
-              + "FOREIGN KEY (exercise_id) REFERENCES exercises(exercise_id) ON DELETE CASCADE");
+          """
+          DO $$
+          BEGIN
+            IF NOT EXISTS (
+              SELECT 1 FROM pg_constraint
+              WHERE conname = 'fk_exercise_id'
+            ) THEN
+              ALTER TABLE exercise_mails_reply_to
+                ADD CONSTRAINT fk_exercise_id
+                FOREIGN KEY (exercise_id) REFERENCES exercises(exercise_id) ON DELETE CASCADE;
+            END IF;
+          END $$;
+          """);
     }
   }
 }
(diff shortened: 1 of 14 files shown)


openaev-model/src/main/java/io/openaev/database/repository/AgentRepository.java
```diff
@@ -41,7 +40,6 @@ Optional<Agent> findByAssetExecutorIdUserDeploymentAndPrivilege(
   List<Agent> findByExternalReferenceAndTenantId(String externalReference, String tenantId);
 
   @Modifying
-  @Query(value = "DELETE FROM agents agent where agent.agent_id = :agentId;", nativeQuery = true)
-  @Transactional
-  void deleteByAgentId(String agentId);
+  @Query(value = "DELETE FROM agents WHERE agent_id = :agentId", nativeQuery = true)
+  void deleteByAgentId(@Param("agentId") String agentId);
```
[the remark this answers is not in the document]
savacano28: This is intended behavior, ignore.

openaev-model/src/main/java/io/openaev/database/repository/ExerciseRepository.java
```diff
@@ -29,6 +29,10 @@ public interface ExerciseRepository
 
   boolean existsByIdAndTenantId(@NotNull String id, @NotNull String tenantId);
 
+  @Modifying
+  @Query(value = "DELETE FROM exercises WHERE exercise_id = :simulationId", nativeQuery = true)
+  void deleteById(@Param("simulationId") String simulationId);
```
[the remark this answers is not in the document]
savacano28: This is intended behavior, ignore.
Tenant validation for simulation is done in existsByIdAndTenantId.

openaev-api/src/main/java/io/openaev/migration/V5_20__Add_delete_cascade_exercise_email_reply_to.java
```diff
@@ -0,0 +1,22 @@
+package io.openaev.migration;
+
+import java.sql.Statement;
+import org.flywaydb.core.api.migration.BaseJavaMigration;
+import org.flywaydb.core.api.migration.Context;
+import org.springframework.stereotype.Component;
+
+@Component
+public class V5_20__Add_delete_cascade_exercise_email_reply_to extends BaseJavaMigration {
```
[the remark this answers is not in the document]
savacano28: This is intended behavior, ignore.

openaev-api/src/main/java/io/openaev/rest/exercise/service/ExerciseService.java
```diff
@@ -547,6 +547,18 @@ public Exercise saveSimulation(Exercise simulation) {
...
+   * Delete a simulation
+   *
+   * @param simulationId ID of the simulation to delete
+   */
+  @Transactional
+  public void delete(String simulationId) {
+    existsByIdAndTenantId(simulationId);
+    exerciseRepository.deleteById(simulationId);
+    log.info("Simulation {} deleted by user {}", simulationId, currentUser().getId());
+  }
```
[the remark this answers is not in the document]
savacano28: Code updated.

openaev-model/src/main/java/io/openaev/database/repository/ExerciseRepository.java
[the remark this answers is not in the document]
savacano28: This is intended behavior, ignore.

openaev-api/src/main/java/io/openaev/service/AgentService.java
```diff
@@ -62,6 +62,7 @@ public List<Agent> saveAllAgents(List<Agent> agents) {
     return agentsSaved;
   }
 
+  @Transactional
```
a19836: Why using Transaction annotation here? The method code is only calling agentRepository.
Although I would not change this bc is outside of scope of this issue. (same than https://github.com/OpenAEV-Platform/openaev/pull/6254/changes#r3419804922)

openaev-api/src/main/java/io/openaev/rest/exercise/service/ExerciseService.java
```diff
@@ -552,8 +552,11 @@ public Exercise saveSimulation(Exercise simulation) {
    *
    * @param simulationId ID of the simulation to delete
    */
+  @Transactional
```
a19836: Why is a transaction being called here? This method calls two other methods:
- existsByIdAndTenantId: which simply queries the database to check if the simulation exists, and throws an exception if it doesn't.
- exerciseRepository.deleteById: which executes a single delete query in the database.
I don't see any reason to use a transaction here, but correct me if I'm wrong.

openaev-api/src/main/java/io/openaev/rest/exercise/service/ExerciseService.java
```diff
@@ -552,8 +552,11 @@ public Exercise saveSimulation(Exercise simulation) {
...
   public void deleteById(String simulationId) {
+    existsByIdAndTenantId(simulationId);
     exerciseRepository.deleteById(simulationId);
```
a19836, line 557: By calling only deleteById in this service, we expect that all dependencies and related tables will be deleted through CASCADE. I'm concerned there may be other tables without the delete cascade option enabled. We already found that the exercise_mails_reply_to table was missing this cascade option, and there could be others we've overlooked. (e.g., exercises_teams, exercises_tags, exercise_team_users, grants, pauses, injects, url_access_tokens, comchecks, objectives, articles, variables, lessons_categories)

**Suggestion**: Before deleting, make sure to call **changeExerciseStatus** to avoid any missed cases.
**Suggestion**: To check other related tables you can run `SELECT conname, conrelid::regclass FROM pg_constraint WHERE confrelid = 'exercises'::regclass AND contype = 'f' `to enumerate them all.
**Suggestion**: It might be helpful to run additional tests to ensure that all scenarios are accounted for.

openaev-model/src/main/java/io/openaev/database/repository/AgentRepository.java
```diff
@@ -41,7 +40,6 @@ Optional<Agent> findByAssetExecutorIdUserDeploymentAndPrivilege(
   List<Agent> findByExternalReferenceAndTenantId(String externalReference, String tenantId);
 
   @Modifying
-  @Query(value = "DELETE FROM agents agent where agent.agent_id = :agentId;", nativeQuery = true)
-  @Transactional
-  void deleteByAgentId(String agentId);
+  @Query(value = "DELETE FROM agents WHERE agent_id = :agentId", nativeQuery = true)
```
a19836: While I understand why you made this change, I would prefer to keep it as it was, since it's outside the scope of this issue and we might unintentionally introduce new problems.

openaev-model/src/main/java/io/openaev/database/repository/ExerciseRepository.java
```diff
@@ -29,6 +29,10 @@ public interface ExerciseRepository
 
   boolean existsByIdAndTenantId(@NotNull String id, @NotNull String tenantId);
 
+  @Modifying
+  @Query(value = "DELETE FROM exercises WHERE exercise_id = :simulationId", nativeQuery = true)
```
a19836, line 33: While I understand the reasoning behind your approach, using a native DELETE bypasses Hibernate's entity lifecycle events like @PreRemove, @PostRemove, orphan removal, and cascade. As a result, audit-logging would not occur. Therefore, we should avoid using native queries whenever possible.

The likely reason the original code failed is that the Exercise JPA entity had associations with other entities that lacked the delete cascade option, which led to a constraint violation. However, calling changeExerciseStatus should resolve this issue.
Can you please try this approach, with the previous code?

openaev-api/src/test/java/io/openaev/rest/exercise/service/ExerciseServiceUnitTest.java
```diff
@@ -5,6 +5,8 @@
 import static org.mockito.ArgumentMatchers.*;
```
a19836, line 5: update tests according with my suggestions

openaev-api/src/test/java/io/openaev/rest/exercise/ExerciseApiTest.java
```diff
@@ -596,6 +596,64 @@ void replacingTeamsShouldPersistNewTeamListInDatabase() throws Exception {
     }
```
a19836, line 596: update tests according with my suggestions

a19836 (changes requested):
Ping me if you need some help

openaev-api/src/main/java/io/openaev/rest/exercise/service/ExerciseService.java
savacano28: Tested but still I have the same issue

openaev-model/src/main/java/io/openaev/database/repository/ExerciseRepository.java
savacano28: I have tried other approaches but am still hitting the same issue. Seen with Gael, the current native queries are not blocking the audit logs.

a19836: Yes, it is true. This is not ideal to have, but if it's the only option, then that's the reality.

a19836 (approved):

PR #6229, opened 2026-06-12
chore(backend): @Transactional alignements and attaching context (#6212)

xfournet:
### Proposed changes

* Align and enforce
  * every API method annotated with `@Transactional`
  * use only Spring `@Transactional`
  * `@Transactional` are permitted only for `openaev-api`
  * web annotation are permitted only for `openaev-api`
  * `@Repository` are permitted only for `openaev-model`

### Testing Instructions

1. Step-by-step how to test
2. Environment or config notes

### Related issues

* Closes #6212

### Checklist

- [x] I consider the submitted work as finished
- [ ] I tested the code for its functionality
- [ ] I wrote test cases for the relevant uses case
- [ ] I added/update the relevant documentation (either on github or on notion)
- [ ] Where necessary I refactored code to improve the overall quality
- [ ] For bug fix -> I implemented a test that covers the bug

PR #6255, opened 2026-06-15
feat(multi-tenancy): tenant transaction scope and SQL filtering foundations (#6212)

laugiov:
## What this is

The v2 tenant isolation. The tenant scope lives on the transaction, and a SQL rewriter filters every tenant table against it. The mechanism is behind an allowlist that is empty by default, so for almost every table merging this changes nothing in production.

The one exception is `import_mappers`: this PR switches it fully onto v2 (its v1 filter removed, the table added to the allowlist), so it ships as the first table isolated entirely by the new mechanism rather than as dormant code.

The old v1 isolation (the `@Filter` driven by the `TenantContext` thread-local) is untouched and still runs for every other table. v2 sits next to it, dormant for those tables.

## What's in here

The mechanism is live infrastructure, but dormant for a table until that table is in the allowlist:

- `TxCtx`, the scope contract. Two states, no "all tenants" wildcard: a restricted list of tenant ids, or no scope (fail-closed, denies everything).
- `can_access_tenant(...)`, the function the filter calls (Flyway `V5_26`, fail-closed, `PARALLEL SAFE`).
- An aspect that writes the scope into the transaction at begin with `set_config(..., true)`, so it is transaction-local and never leaks. The order is pinned explicitly (lock, then transaction, then scope), and a nested transaction cannot redefine a scope that is already set.
- `TenantStatementInspector`, the rewriter: SELECT (joins, sub-queries, CTEs, unions), UPDATE and `UPDATE ... FROM`, DELETE and `DELETE ... USING`, INSERT (with tenant validation on the write side). Anything it cannot safely filter on an active table is refused, not let through.
- The HTTP binding: a resolver builds the `TxCtx` from the request (path tenant wins, else the `X-Tenant-Ids` header, else your full set of rights). The selector is intersected with your actual rights, and asking for a tenant you are not a member of is a 403, before RBAC. Rights are membership, not the admin flag.
- The write policy: a write needs a single tenant in scope, else 400.
- An ArchUnit rule that fails the build on any new raw JDBC in prod code, so nothing slips past the rewriter.

## The first table: `import_mappers` (activated here)

I took one table all the way through and switched it on in this PR. `import_mappers` is HTTP-only (nothing but HTTP touches it), which is why it goes first.

What "fully on v2" means here:

- its v1 `@Filter` is removed (v1 and v2 must not both filter the same table: v1 forces a single tenant, v2 wants the request scope, and on the header route the two together returned ~0 rows);
- `import_mappers` is added to `openaev.tenant.active-tables`;
- a guard test fails the build if it is ever dropped from the allowlist, since with v1 gone that would leave the table with no isolation at all.

The whole `MapperApi` surface is scoped, with real two-tenant tests against Postgres (no mocks on the isolation path): reads, update and delete, duplicate, create and import, plus the mapper lookup in the scenario and exercise import endpoints. There are non-admin tests (a foreign tenant is refused at the binding; a non-admin in two tenants only sees the one in the path), and the header-route test is now enabled (it was parked while v1 still filtered the table).

Behaviour change at merge: create and import now require a single-tenant scope (else 400). The frontend already tenant-path-scopes every API call, so its mapper create and import keep a single tenant, and no other caller (backend, connectors) creates mappers, so nothing breaks. The api suite is green.

## What's left, and how

Coordination, roughly in this order.

**1. Confirm the aspect order.** The three `@Order` values move the global transaction-advisor precedence, which is your call (transaction layer). They are committed and the tests pin the behaviour either way, so it is a yes/no on keeping them, not work.

**2. The next tables.** Same recipe as `import_mappers`, but first every path touching the table has to carry a `TxCtx`. The HTTP paths do; the background ones (jobs, event handlers, indexing, startup seeding) do not yet, and an active table with no scope is fail-closed (zero rows). So those tables wait on the Track A work to carry the scope off the request thread. `import_mappers` went first only because nothing but HTTP touches it.

**3. Drop v1.** Once enough tables are on v2, remove the `TenantContext` thread-local and the filter aspect (overlaps #6229). End state, not a per-table step.


openaev-api/src/main/java/io/openaev/rest/mapper/MapperApi.java
```diff
@@ -83,7 +86,9 @@ public Page<RawPaginationImportMapper> getImportMapper(
       resourceId = "#mapperId",
       actionPerformed = Action.READ,
       resourceType = ResourceType.MAPPER)
-  public ImportMapper getImportMapperById(@PathVariable String mapperId) {
+  // TxCtx is resolved from the request and applied by the transaction aspect; it scopes this read
+  // to the caller's tenants. The handler does not use it directly.
+  public ImportMapper getImportMapperById(TxCtx ctx, @PathVariable String mapperId) {
```
corinnekrych, line 91: why do we need to pass it as a param?

laugiov: After our meeting, I checked this point. The TxCtx param is how the tenant scope reaches the transaction: TenantScopeTransactionAspect reads it from the method args and sets app.current_tenants, which can_access_tenant then uses to filter the SQL. The read handler doesn't touch it, but if I drop it the aspect sets nothing, so once import_mappers is active the query returns zero rows. It's a param and not a ThreadLocal so a non-HTTP caller (a background job) can pass its own scope later.

openaev-api/src/test/java/io/openaev/architecture/TenantNonOrmAccessArchTest.java
```diff
@@ -0,0 +1,55 @@
...
+          .and()
+          .doNotHaveFullyQualifiedName(
+              "io.openaev.database.helper.InjectExpectationRepositoryHelper")
+          .should()
+          .dependOnClassesThat(
+              // Spring JDBC query API by package, so the interfaces (JdbcOperations,
+              // NamedParameterJdbcOperations) are caught too, not just the implementations.
+              resideInAPackage("org.springframework.jdbc.core..")
+                  // Raw JDBC types; Statement also covers Prepared/CallableStatement.
+                  .or(assignableTo(Connection.class))
```
corinnekrych, line 41: it could be good to align to preprocessor rules in `openaev-annotation-preprocessor` + (if needed explicit suppression annotation like @AllowRawJdbc)

openaev-api/src/main/java/io/openaev/rest/mapper/MapperApi.java
corinnekrych: ok it needs to be well documented and understood...
To me this is the weak point: the safety mechanism depends on an “unused-looking” parameter that can be deleted by a refactor/cleanup.
Long run once all code base aligned owe can make it robust, pair it with guardrails: compile-time rule that any @Transactional method on active-table paths must declare TxCtx, plus explicit marker annotation (e.g. @RequiresTxScope) and tests that fail if scope is missing. 

Meanwhile we can add an ArchUnit test:
```
package io.openaev.architecture;

import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.methods;

import com.tngtech.archunit.base.DescribedPredicate;
import com.tngtech.archunit.core.domain.JavaMethod;
import com.tngtech.archunit.core.importer.ImportOption;
import com.tngtech.archunit.junit.AnalyzeClasses;
import com.tngtech.archunit.junit.ArchTest;
import com.tngtech.archunit.lang.ArchCondition;
import com.tngtech.archunit.lang.ArchRule;
import com.tngtech.archunit.lang.ConditionEvents;
import com.tngtech.archunit.lang.SimpleConditionEvent;
import io.openaev.context.TxCtx;
import java.util.Set;
import java.util.stream.Collectors;

/**
 * Guards v2 tenant-scoped entrypoints against accidental signature drift.
 *
 * <p>For migrated entrypoints, {@link TxCtx} must stay explicit on the transactional boundary so the
 * transaction scope aspect can bind tenant scope for downstream repository calls.
 */
@AnalyzeClasses(packages = "io.openaev", importOptions = ImportOption.DoNotIncludeTests.class)
class TenantScopedEntrypointsTxCtxArchTest {

  private static final Set<String> TX_SCOPED_ENTRYPOINTS =
      Set.of(
          "io.openaev.rest.mapper.MapperApi#getImportMapper",
          "io.openaev.rest.mapper.MapperApi#getImportMapperById",
          "io.openaev.rest.mapper.MapperApi#createImportMapper",
          "io.openaev.rest.mapper.MapperApi#exportMappers",
          "io.openaev.rest.mapper.MapperApi#importMappers",
          "io.openaev.rest.mapper.MapperApi#duplicateMapper",
          "io.openaev.rest.mapper.MapperApi#updateImportMapper",
          "io.openaev.rest.mapper.MapperApi#deleteImportMapper",
          "io.openaev.rest.scenario.ScenarioImportApi#dryRunImportXLSFile",
          "io.openaev.rest.scenario.ScenarioImportApi#validateImportXLSFile",
          "io.openaev.rest.exercise.ExerciseImportApi#dryRunImportXLSFile",
          "io.openaev.rest.exercise.ExerciseImportApi#validateImportXLSFile");

  @ArchTest
  static final ArchRule tx_scoped_entrypoints_must_declare_tx_ctx =
      methods()
          .that(
              new DescribedPredicate<>("are v2 tenant-scoped entrypoints") {
                @Override
                public boolean test(JavaMethod method) {
                  return TX_SCOPED_ENTRYPOINTS.contains(
                      method.getOwner().getName() + "#" + method.getName());
                }
              })
          .should(
              new ArchCondition<>("declare TxCtx parameter") {
                @Override
                public void check(JavaMethod method, ConditionEvents events) {
                  boolean hasTxCtx =
                      method.getRawParameterTypes().stream()
                          .anyMatch(type -> TxCtx.class.getName().equals(type.getName()));
                  if (!hasTxCtx) {
                    String parameterList =
                        method.getRawParameterTypes().stream()
                            .map(javaClass -> javaClass.getName())
                            .collect(Collectors.joining(", "));
                    events.add(
                        SimpleConditionEvent.violated(
                            method,
                            "Entrypoint "
                                + method.getOwner().getName()
                                + "#"
                                + method.getName()
                                + "("
                                + parameterList
                                + ") must declare TxCtx to activate tenant v2 transaction scope"));
                  }
                }
              });
}

```

openaev-api/src/test/java/io/openaev/service/MapperServiceTest.java
```diff
@@ -383,13 +387,86 @@ void given_mappersInput_should_appendImportedSuffix_whenImportMappers() {
...
     // Act
-    mapperService.importMappers(List.of(input));
+    mapperService.importMappers("tenant-test", List.of(input));
 
     // Assert
     ArgumentCaptor<List<ImportMapper>> captor = ArgumentCaptor.forClass(List.class);
     verify(importMapperRepository).saveAll(captor.capture());
     assertEquals(1, captor.getValue().size());
     assertTrue(captor.getValue().get(0).getName().endsWith(Constants.IMPORTED_OBJECT_NAME_SUFFIX));
+    assertEquals("tenant-test", captor.getValue().get(0).getTenant().getId());
```
corinnekrych, line 397: @laugiov in parallell fo v1 approach we have this open tickets https://github.com/OpenAEV-Platform/openaev/issues/5718 with a bunch of Pr to add API level test for tenant isolation, we have a skills defined to create those in TENANT_ISOLATION.md
@johanah29 did one PR [here](https://github.com/OpenAEV-Platform/openaev/pull/6036/changes#diff-763bc970c154b693dc1a8ba7f1de53ca49c53d5829aff11f8698d25fe52c2f6eR316) to add the test, we can merge the PR and do a rebase afterward

corinnekrych (commented):
Let's improve the API level integration test we can improve the SKILLs to genrate those test and aa tenant scope coverage

openaev-api/src/main/java/io/openaev/rest/mapper/MapperApi.java
laugiov: agree,  I'll add your ArchUnit test now as a stopgap and a fix on @RequiresTxScope marker plus a preprocessor rule, derived from the code. I'll also add a test that proves the fail-closed behavior when the scope is missing.

openaev-api/src/test/java/io/openaev/architecture/TenantNonOrmAccessArchTest.java
laugiov: Good idea, one source of truth. I'll align the rule with openaev-annotation-preprocessor and keep ArchUnit only for what the compile-time check cannot see. I'll add @AllowRawJdbc for the rare legitimate cases, with a required justification so it stays an exception.

openaev-model/src/main/java/io/openaev/database/helper/ExecutionTraceRepositoryHelper.java
```diff
@@ -26,6 +27,7 @@
  *
  * @see ExecutionTrace
  */
+@AllowRawJdbc(reason = "writes execution_traces and injects_statuses, both non-tenant tables")
```
corinnekrych, line 30: I think this is wrong the table need tenatn scoped 
https://github.com/OpenAEV-Platform/openaev/issues/6394 to track it
Also for the perf comment maybe we can explore if we really need a jdbc call here
https://github.com/OpenAEV-Platform/openaev/issues/6395

openaev-model/src/main/java/io/openaev/database/helper/InjectExpectationRepositoryHelper.java
```diff
@@ -17,6 +18,7 @@
  * @see io.openaev.database.model.InjectExpectation
  * @see io.openaev.database.model.InjectExpectationSignature
  */
+@AllowRawJdbc(reason = "writes injects_expectations, a non-tenant table")
```
corinnekrych: @laugiov I can not find any caller for this method, got the feeling this is dead code... @Dimfacion checking git blama you're the last one that chanegd this class. is it still in use?

openaev-api/src/test/java/io/openaev/rest/scenario/ScenarioImportApiTenantIsolationTest.java
```diff
@@ -0,0 +1,111 @@
+package io.openaev.rest.scenario;
```
corinnekrych, line 1: the convention (written in skills is more to use a nested class inside the *ApiTest class

openaev-model/src/main/java/io/openaev/database/helper/InjectExpectationRepositoryHelper.java
Dimfacion: It was when I wrote it. It looks like this commit removed it : https://github.com/OpenAEV-Platform/openaev/commit/b4e323f96f96ebdaaa3d69f812ef58f9f54d0e5e

corinnekrych: @laugiov let's remove it. not used anymore

corinnekrych (approved):
We're good to go, just clean up unused code.

PR #6296, opened 2026-06-18
fix(icon): address review comments on image endpoints and file service (#4864)

RomuDeuxfois:
Several bugs introduced in the icon management rework: a compile error from a missing brace, a duplicate MinIO lookup, unreliable stream sizing, and a frontend URL mismatch between collector UUID and collector type.

### Backend

- **`FileService.getCatalogConnectorImage()`**: Added missing closing `}` — method body was swallowing the `getPlatformImage` private method, causing a compile failure.
- **`FileService.getPlatformImage()`**: Fallback was calling `getFilePathForTenant(currentTenant, ...)` — identical to the first call. Fixed to fall back to `DEFAULT_TENANT_UUID` for built-in platform assets.
- **`FileService.getExecutorBannerImage()`**: Removed stale `@param isExternal` from Javadoc.
- **`MinioService`**: Replaced `data.available()` (unreliable — may return 0) with `size=-1, partSize=10MB` in both `uploadStreamInTenantPath` and `uploadStreamInPlatformPath`.
- **`InjectorApi` / `CollectorApi`**: Fixed `@Operation(summary)` from "by id" → "by type"; wrapped `InputStream` in `try-with-resources` to prevent connection leak.
- **`CollectorApi`**: Added `GET /api/collectors/id/{collectorId}/image` (+ tenant variant) — resolves image by UUID→type lookup, replacing the removed `getCollectorImageFromId` in `DocumentApi`.

### Frontend

- **`InjectExpectationResultList.tsx`**: `expectationResult.sourceId` is a collector UUID (set via `collector.getId()` in `ExpectationResultBuilder`), not a type. Switched to the new `/api/collectors/id/{id}/image` endpoint.
- **`ConnectorPage.tsx`**: `logoUrl(connector?.type)` could produce `/api/injectors/undefined/image`. Guarded to only call `logoUrl` when `connector.type` is defined.

### Tests

- **`CatalogConnectorApiTest`**: Updated to upload via `uploadCatalogLogo()` (platform path) and assert against the non-tenant-scoped route `/api/images/catalog/connectors/logos/`.
- **`InjectorApiTest`** / **`CollectorApiTest`**: Added `@Nested` image test groups covering 200 (existing image) and 404 (missing image/unknown id) for the new endpoints.

### DIFF
--- a/openaev-api/src/main/java/io/openaev/integration/impl/executors/openaev/OpenAEVExecutorIntegrationFactory.java
+++ b/openaev-api/src/main/java/io/openaev/integration/impl/executors/openaev/OpenAEVExecutorIntegrationFactory.java
@@ -9,7 +9,6 @@ import io.openaev.executors.ExecutorService;
 import io.openaev.integration.BuiltinIntegrationFactory;
 import io.openaev.integration.ComponentRequestEngine;
 import io.openaev.integration.Integration;
-import io.openaev.rest.exception.ElementNotFoundException;
 import io.openaev.service.account.ServiceAccountPrivilegeService;
 import io.openaev.service.catalog_connectors.CatalogConnectorService;
 import io.openaev.service.connector_instances.ConnectorInstanceService;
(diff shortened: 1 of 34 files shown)


openaev-api/src/main/java/io/openaev/integration/impl/executors/openaev/OpenAEVExecutorIntegrationFactory.java
```diff
@@ -76,23 +75,19 @@ public Integration spawn(ConnectorInstance instance) {
 
   @Override
   public void registerConnectorForTenant(String tenantId) throws Exception {
-    try {
```
corinnekrych, line 79: Good catch!

corinnekrych (approved):
Simplier to All icons in a common place

PR #6299, opened 2026-06-18
refactor: optimize SQL queries in engine sync for better perf

guillaumejparis:
### Proposed changes

*
*

### Testing Instructions

1. Step-by-step how to test

### Related issues

* Closes #ISSUE-NUMBER
*

### Checklist

### Further comments


laugiov:
Could you add a before/after EXPLAIN (ANALYZE, BUFFERS) on the PR? Not a blocker, just a suggestion. 

On an empty dev database both versions look instant, so the numbers only mean something on prod-sized data with a recent :from.

It would be good to make this a habit on "perf" PRs in general, so the gain is backed by real numbers and not only by reasoning.

For this case you can run the native query directly, with :from set to a recent value and :limit to the job's value:


EXPLAIN (ANALYZE, BUFFERS)
WITH changed_assets AS ( ... )    -- the full query from this PR
SELECT * FROM endpoint_data ed
ORDER BY ed.endpoint_updated_at ASC LIMIT 1000;   -- :from = now() - interval '15 seconds'

Then we just check we get an Index Scan on the new indexes instead of a Seq Scan, and compare the timing with the old query.

openaev-model/src/main/java/io/openaev/database/repository/EndpointRepository.java
```diff
@@ -120,30 +120,58 @@ List<Object[]> findAllByNameLinkedToFindingsWithContext(
...
               + "LEFT JOIN exercises e ON i.inject_exercise = e.exercise_id "
               + "LEFT JOIN scenarios s ON i.inject_scenario = s.scenario_id "
-              + "WHERE a.asset_type = '"
-              + AssetType.Values.ENDPOINT_TYPE
-              + "' "
               + "GROUP BY a.asset_id"
               + ") "
               + "SELECT * FROM endpoint_data ed "
-              + "WHERE ed.endpoint_updated_at > :from "
               + "ORDER BY ed.endpoint_updated_at ASC LIMIT :limit;",
```
laugiov, line 175: We should add non-regression tests for these four queries. Today the only indexing test is InjectHandlerTest.FetchInjectsForIndexing, which covers the Inject query but not these.

This kind of rewrite fails silently: if the changed-set logic misses a case, a document just never gets re-indexed, and we only find out weeks later in prod. A test catches it early.

The case to cover: an entity whose only change is on a linked row after :from (a linked inject, contract or finding) must still come back from findForIndexing. You can use InjectHandlerTest as a model, it already runs on a real Postgres.

openaev-model/src/main/java/io/openaev/database/repository/EndpointRepository.java
```diff
@@ -120,30 +120,58 @@ List<Object[]> findAllByNameLinkedToFindingsWithContext(
...
+              + "JOIN assets a ON fa.asset_id = a.asset_id AND a.asset_type = '"
+              + AssetType.Values.ENDPOINT_TYPE
+              + "' WHERE f.finding_updated_at > :from"
+              + "), "
+              + "endpoint_data AS ("
               + "SELECT a.asset_id, a.asset_type, a.asset_name, a.asset_external_reference, "
               + "a.endpoint_ips, a.endpoint_hostname, a.endpoint_platform, a.endpoint_arch, "
               + "a.endpoint_mac_addresses, a.endpoint_seen_ip, a.asset_created_at, a.endpoint_is_eol, a.asset_description, a.tenant_id, "
-              + "GREATEST(a.asset_updated_at, max(i.inject_updated_at), max(e.exercise_updated_at), max(s.scenario_updated_at), max(f.finding_updated_at)) as endpoint_updated_at, "
+              + "COALESCE(GREATEST(a.asset_updated_at, max(i.inject_updated_at), max(e.exercise_updated_at), max(s.scenario_updated_at), max(f.finding_updated_at)), a.asset_updated_at) as endpoint_updated_at, "
```
laugiov: Can you remove this COALESCE? (on each 4 queries) It never does anything. 

Every row here comes from the JOIN changed_*, so at least one of the timestamps feeding the GREATEST is always > :from, so non-null. And GREATEST only returns null when all of its arguments are null. So the fallback can never trigger.

So this:

COALESCE(GREATEST(a.asset_updated_at, ...), a.asset_updated_at)
can just be:

GREATEST(a.asset_updated_at, ...)
which also lines up with the Inject query, where it is not wrapped. 

Same change applies in the three other queries.

openaev-model/src/main/java/io/openaev/database/repository/EndpointRepository.java
guillaumejparis: Adding regression test

guillaumejparis:
With EXPLAIN on different sync requests against a basic SaaS DB (with no real problems), I'm seeing an improvement in the order of 50x faster

laugiov (approved):

PR #6311, opened 2026-06-19
feat(multi-tenancy): remove feature-flag (#4864)

RomuDeuxfois:


openaev-api/src/main/resources/application-ci.properties
```diff
@@ -3,7 +3,7 @@
 # (those use src/test/resources/application.properties instead).
 
 # Enable feature flags required by the E2E test suite.
-openaev.enabled-dev-features=MULTI_TENANCY
+openaev.enabled-dev-features=
```
corinnekrych: Suggestion not directly linked to the PR, but i think we should replace with `*` as more and more if we want to test e2e new feature with playwright we'll need to be able to access them early.

corinnekrych (dismissed):
🎉 🚢

corinnekrych (approved):

Issue #163, opened 2022-07-24
Replace SpringFox with SpringDoc


richard-julien:
## Problem to Solve

SpringFox is old and throw an exception at startup when parsing the API.

SpringDoc is a good replacement for the swagger api generation.

PR #6313, opened 2026-06-19
fix(audit-log): remove constant register from agent (#163)

Dimfacion:
### Proposed changes

* Disable log of the register of the agent if there are no changes with the previous state
*

### Testing Instructions

1. Have an agent running
2. Check that the agent is registering if there are changes but never otherwise

### Related issues

* Closes OpenAEV-Platform/filigran-private#163
*

### Checklist

### DIFF
--- a/openaev-api/src/main/java/io/openaev/service/EndpointService.java
+++ b/openaev-api/src/main/java/io/openaev/service/EndpointService.java
@@ -18,8 +18,10 @@ import static java.time.Instant.now;
 import static java.util.Optional.ofNullable;
 import static java.util.stream.Collectors.toList;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
 import io.openaev.config.OpenAEVConfig;
 import io.openaev.config.cache.LicenseCacheManager;
+import io.openaev.database.audit.AuditLoggedService;
 import io.openaev.database.model.*;
 import io.openaev.database.repository.*;
 import io.openaev.database.specification.AssetAgentJobSpecification;
@@ -62,7 +64,7 @@ import org.springframework.transaction.annotation.Transactional;
 @RequiredArgsConstructor
 @Service
 @Slf4j
-public class EndpointService {
+public class EndpointService implements AuditLoggedService {
 
   private static final String ASSET_GROUP_FILTER = "assetGroups";
 
@@ -91,6 +93,7 @@ public class EndpointService {
   @Resource private OpenAEVConfig openAEVConfig;
 
   private final EndpointMapper endpointMapper;
+  private final ObjectMapper objectMapper;
 
   /** Cache of raw install/upgrade script templates keyed by platform/script name. */
   private final Map<String, String> agentScriptTemplateCache = new ConcurrentHashMap<>();
@@ -728,11 +731,20 @@ public class EndpointService {
 
   private Agent updateExistingAgent(Agent agent, AgentRegisterInput input) {
     Endpoint endpoint = (Endpoint) agent.getAsset();
+    // Capture significant state before mutation
+    Map<String, Object> before = endpoint.significantState(objectMapper);
+
     setUpdatedEndpointAttributes(endpoint, input);
     addSourceTagToEndpoint(endpoint, input);
     updateEndpoint(endpoint);
     setUpdatedAgentAttributes(agent, input, endpoint);
-    return agentService.createOrUpdateAgent(agent);
+    Agent saved = agentService.createOrUpdateAgent(agent);
+
+    // Suppress audit logging for heartbeat-only updates (no significant endpoint change)
+    Endpoint after = (Endpoint) saved.getAsset();
+    suppressAuditIfUnchanged(before, after.significantState(objectMapper));
+
+    return saved;
   }
 
   private Agent updateExistingEndpointAndCreateAgent(Endpoint endpoint, AgentRegisterInput input) {
--- a/openaev-api/src/main/java/io/openaev/service/EndpointService.java
+++ b/openaev-api/src/main/java/io/openaev/service/EndpointService.java
@@ -725,8 +725,9 @@ public class EndpointService implements AuditLoggedService {
   private Agent updateExistingEndpointAndManageAgent(Endpoint endpoint, AgentRegisterInput input) {
     setUpdatedEndpointAttributes(endpoint, input);
     addSourceTagToEndpoint(endpoint, input);
+    Agent agent = createOrUpdateAgent(endpoint, input);
     updateEndpoint(endpoint);
-    return createOrUpdateAgent(endpoint, input);
+    return agent;
   }
 
   private Agent updateExistingAgent(Agent agent, AgentRegisterInput input) {
@@ -736,25 +737,24 @@ public class EndpointService implements AuditLoggedService {
 
     setUpdatedEndpointAttributes(endpoint, input);
     addSourceTagToEndpoint(endpoint, input);
-    updateEndpoint(endpoint);
     setUpdatedAgentAttributes(agent, input, endpoint);
-    Agent saved = agentService.createOrUpdateAgent(agent);
+    updateEndpoint(endpoint);
 
     // Suppress audit logging for heartbeat-only updates (no significant endpoint change)
-    Endpoint after = (Endpoint) saved.getAsset();
-    suppressAuditIfUnchanged(before, after.significantState(objectMapper));
+    suppressAuditIfUnchanged(before, endpoint.significantState(objectMapper));
 
-    return saved;
+    return agent;
   }
 
   private Agent updateExistingEndpointAndCreateAgent(Endpoint endpoint, AgentRegisterInput input) {
     setUpdatedEndpointAttributes(endpoint, input);
     addSourceTagToEndpoint(endpoint, input);
-    updateEndpoint(endpoint);
     Agent agent = new Agent();
     setNewAgentAttributes(input, agent);
     setUpdatedAgentAttributes(agent, input, endpoint);
-    return agentService.createOrUpdateAgent(agent);
+    endpoint.getAgents().add(agent);
+    updateEndpoint(endpoint);
+    return agent;
   }
 
   private Agent createOrUpdateAgent(Endpoint endpoint, AgentRegisterInput input) {
@@ -777,7 +777,10 @@ public class EndpointService implements AuditLoggedService {
       setNewAgentAttributes(input, agent);
     }
     setUpdatedAgentAttributes(agent, input, endpoint);
-    return agentService.createOrUpdateAgent(agent);
+    if (!endpoint.getAgents().contains(agent)) {
+      endpoint.getAgents().add(agent);
+    }
+    return agent;
   }
 
   private void setUpdatedEndpointAttributes(Endpoint endpoint, AgentRegisterInput input) {
@@ -813,12 +816,13 @@ public class EndpointService implements AuditLoggedService {
     endpoint.setSeenIp(input.getSeenIp());
     endpoint.setMacAddresses(input.getMacAddresses());
     endpoint.setTenant(input.getExecutor().getTenant());
+    Agent agent = new Agent();
+    setNewAgentAttributes(input, agent);
+    setUpdatedAgentAttributes(agent, input, endpoint);
+    endpoint.getAgents().add(agent);
     createEndpoint(endpoint);
     addSourceTagToEndpoint(endpoint, input);
-    Agent agent = new Agent();
-    setUpdatedAgentAttributes(agent, input, endpoint);
-    setNewAgentAttributes(input, agent);
-    return agentService.createOrUpdateAgent(agent);
+    return agent;
   }
 
   private void setNewAgentAttributes(AgentRegisterInput input, Agent agent) {
--- a/openaev-model/src/main/java/io/openaev/annotation/AuditDiffTracked.java
+++ b/openaev-model/src/main/java/io/openaev/annotation/AuditDiffTracked.java
@@ -1,5 +1,6 @@
 package io.openaev.annotation;
 
+import io.openaev.database.audit.AuditLogContext;
 import java.lang.annotation.*;
 
 /**
@@ -10,7 +11,7 @@ import java.lang.annotation.*;
  * <ul>
  *   <li>Capture a JSON snapshot of the entity at {@code @PostLoad} time (the "before" state).
  *   <li>Compute a field-level diff at {@code @PreUpdate} time and record it in {@link
- *       io.openaev.database.audit.EntityDiffContext}.
+ *       AuditLogContext}.
  * </ul>
  *
  * <p>The diff context is consumed by the audit aspect ({@code AccessControlAuditLogAspect})
--- a/openaev-model/src/main/java/io/openaev/database/audit/AuditLogContext.java
+++ b/openaev-model/src/main/java/io/openaev/database/audit/AuditLogContext.java
@@ -17,11 +17,12 @@ import org.springframework.web.context.request.RequestContextHolder;
  *   <li>Otherwise, a ThreadLocal fallback is used for non-web execution paths.
  * </ul>
  */
-public final class EntityDiffContext {
+public final class AuditLogContext {
 
   private static final String REQUEST_ATTR_BEFORE_SNAPSHOTS = "openaev.audit.beforeSnapshots";
   private static final String REQUEST_ATTR_ENTITY_SNAPSHOTS = "openaev.audit.entitySnapshots";
   private static final String REQUEST_ATTR_CLEANUP_REGISTERED = "openaev.audit.cleanupRegistered";
+  private static final String REQUEST_ATTR_ENABLED = "openaev.audit.enabled";
 
   private static final ThreadLocal<Map<String, Map<String, Object>>> BEFORE_SNAPSHOTS_TL =
       ThreadLocal.withInitial(LinkedHashMap::new);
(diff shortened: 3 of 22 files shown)


openaev-model/src/main/java/io/openaev/database/audit/AuditStateCapturable.java
```diff
@@ -0,0 +1,73 @@
...
+  /**
+   * Returns a map representing this entity's significant state.
+   *
+   * <p>Serializes the entity via {@code objectMapper.convertValue(this, Map)} and removes fields
+   * annotated with {@link AuditStateIgnore}. New fields are automatically captured.
+   *
+   * @param objectMapper the ObjectMapper used to serialize the entity
+   * @return a map of significant field names to their current values
+   */
+  default Map<String, Object> significantState(ObjectMapper objectMapper) {
```
RomuDeuxfois, line 50: **question**; What do you for @ManyToOne, @ManyToMany ect properties ?

openaev-model/src/main/java/io/openaev/database/audit/AuditStateCapturable.java
```diff
@@ -0,0 +1,73 @@
...
+  /**
+   * Resolves the set of JSON property names to exclude for the given class. The result is cached
+   * per class so reflection is only performed once.
+   */
+  private static Set<String> resolveIgnoredFields(Class<?> clazz) {
+    return IGNORED_FIELDS_CACHE.computeIfAbsent(clazz, AuditStateCapturable::scanIgnoredFields);
+  }
+
+  /** Walks the class hierarchy and collects JSON keys of fields annotated with @AuditDiffIgnore. */
+  private static Set<String> scanIgnoredFields(Class<?> clazz) {
```
RomuDeuxfois: **nitpick**: We have some utility method that can help on FieldUtils

openaev-model/src/main/java/io/openaev/database/model/Asset.java
```diff
@@ -103,6 +104,7 @@ public void setTags(Set<Tag> tags) {
   @CreationTimestamp
   private Instant createdAt = now();
 
+  @AuditStateIgnore
```
RomuDeuxfois, line 182: **question**: Wha't the goal to use this annotation without implementing the AuditStateCapturable interface ?

openaev-model/src/main/java/io/openaev/annotation/AuditDiffTracked.java
```diff
@@ -1,5 +1,6 @@
 package io.openaev.annotation;
 
+import io.openaev.database.audit.AuditLogContext;
```
RomuDeuxfois, line 3: Wow, import is needed to write a comment. I didn't know that.

Dimfacion: It's not needed but I don't like having the full name in the comment x) Do you feel it would be better to remove it ?

Dimfacion: I mean, it's not needed for openaev to compile but the IDE will yell at you

openaev-model/src/main/java/io/openaev/database/model/Endpoint.java
```diff
@@ -200,4 +205,26 @@ public Endpoint(String id, String type, String name, PLATFORM_TYPE platform) {
...
+  /**
+   * Returns the significant state of this endpoint for audit comparison.
+   *
+   * <p>Serializes the entire entity via Jackson, then strips non-significant fields (annotated with
+   * {@link AuditStateIgnore}) and replaces the raw agents list with each agent's own {@link
+   * Agent#significantState}. Any new field added to the entity is automatically included without
+   * code changes.
+   */
+  @Override
+  public Map<String, Object> significantState(ObjectMapper objectMapper) {
```
RomuDeuxfois: **refacto**: I am pretty this can be done dirclty in the interface.

openaev-model/src/main/java/io/openaev/database/model/Asset.java
Dimfacion: Endpoint extends Asset so it's needed to ignore this property when using significantState() on an Endpoint

openaev-model/src/main/java/io/openaev/database/audit/AuditStateCapturable.java
Dimfacion: objectMapper.convertValue will use the jackson configuration linked to those fields so for instance if the field has the @JsonSerialize(using = MonoIdSerializer.class), we will replace it with an id

openaev-model/src/main/java/io/openaev/database/audit/AuditStateCapturable.java
Dimfacion: Good to know !

openaev-model/src/main/java/io/openaev/database/model/Endpoint.java
Dimfacion: I was afraid it would lead to an over engineered method but with the helpers of FieldUtils, it simplifies things. Thanks !

RomuDeuxfois:
@Dimfacion I have one concern with the current AuditStateCapturable approach: objectMapper.convertValue(entity, Map) serializes the raw JPA entity, ~bypassing MonoIdSerializer and MultiModelSerializer~ (not true, confirm with Gaël).
This means we can end up with deep object graphs in the snapshot, exactly the kind of thing we want to avoid in audit logs. This is actually the same concern as on the API side: we never expose raw JPA entities directly, we always go through Output DTOs for perf and graph issue. The audit snapshot should follow the same rule.

Could be good also to set-up some rules for next audit logging objects

| Relation type | Pattern | Example |
|---|---|---|
| Large collection, read-only reference | `Set<String>` IDs in the DTO | `ScenarioOutput.exercises` |
| Child entity without its own API endpoint | Full Output object | `EndpointOutput.agents` → `List<AgentOutput>` |
| Child entity with its own API endpoint | ID only | `InjectOutput` → `exercise_id` |
| Complex embedded object | ⚠️ Avoid — extract a dedicated Output DTO | `InjectorContract` in `InjectOutput` |
| Parent reference from a child DTO | ID only — never the full parent object | `AgentOutput.asset_id` |

Dimfacion:
As seen together @RomuDeuxfois, I've added a filter to avoid serializing some fields if they are with the correct annotation. That should make comparisons slimmer. Let me know what you think :)

openaev-model/src/main/java/io/openaev/database/model/Asset.java
```diff
@@ -103,6 +104,7 @@ public void setTags(Set<Tag> tags) {
   @CreationTimestamp
   private Instant createdAt = now();
```
RomuDeuxfois, line 180: **todo**: This should be AuditStateIgnore like updatedAt. What you think ?

openaev-model/src/main/java/io/openaev/database/model/Asset.java
RomuDeuxfois, line 171: **Question:**
I’m wondering whether we should also mark this kind of property with `@AuditStateIgnore`.

`@JsonIgnore` currently excludes it automatically, but if someone removes that annotation later, the audit behavior could be impacted unintentionally.

What do you think?

openaev-model/src/main/java/io/openaev/database/model/Agent.java
RomuDeuxfois, line 69: **todo**: Should be @AuditStateIgnore no ?

openaev-model/src/main/java/io/openaev/database/model/Agent.java
```diff
@@ -142,6 +147,7 @@ public boolean isActive() {
   @NotNull
   private Instant createdAt = now();
```
RomuDeuxfois, line 150: **todo**: Should be @AuditStateIgnore like updateAt

openaev-model/src/main/java/io/openaev/database/model/Agent.java
RomuDeuxfois, line 161: **todo**: Seems to be used only for Caldera, can be also annotate @AuditStateIgnore

openaev-model/src/main/java/io/openaev/database/model/Agent.java
```diff
@@ -142,6 +147,7 @@ public boolean isActive() {
...
 
+  @AuditStateIgnore
```
RomuDeuxfois, line 152: **nitpick**: this can be enforced direclty in the AuditStatePropertyFilter

public class AuditStatePropertyFilter extends SimpleBeanPropertyFilter {

  private static final Set<String> ALWAYS_IGNORED = Set.of("createdAt", "updatedAt");

private static boolean isAuditStateValid(BeanPropertyWriter writer) {
    return !ALWAYS_IGNORED.contains(writer.getName())
        && writer.getAnnotation(AuditStateIgnore.class) == null
  }
}

openaev-model/src/main/java/io/openaev/database/audit/AuditLogContext.java
```diff
@@ -32,10 +33,36 @@ public final class EntityDiffContext {
   private static final ThreadLocal<Boolean> CLEANUP_REGISTERED_TL =
       ThreadLocal.withInitial(() -> false);
 
-  private EntityDiffContext() {}
+  private static final ThreadLocal<Boolean> ENABLED_TL = ThreadLocal.withInitial(() -> true);
```
RomuDeuxfois: **nitpick:** I would prefer not to add a new `ThreadLocal` property, since we have had a lot of issues with `ThreadLocal` in the past.

As I see it, we have two cases:

1. **From an API call**: everything can be handled through request attributes.
2. **From a service call**: at the service level, we can explicitly call the audit log service and decide whether the diff is significant or not. And if not, not call the audit log service.

What do you think?

openaev-api/src/main/java/io/openaev/service/EndpointService.java
```diff
@@ -691,11 +694,20 @@ private Agent updateExistingEndpointAndManageAgent(Endpoint endpoint, AgentRegis
...
     setUpdatedEndpointAttributes(endpoint, input);
     addSourceTagToEndpoint(endpoint, input);
     updateEndpoint(endpoint);
     setUpdatedAgentAttributes(agent, input, endpoint);
-    return agentService.createOrUpdateAgent(agent);
+    Agent saved = agentService.createOrUpdateAgent(agent);
+
+    // Suppress audit logging for heartbeat-only updates (no significant endpoint change)
+    Endpoint after = (Endpoint) saved.getAsset();
+    suppressAuditIfUnchanged(before, after.significantState(objectMapper));
```
RomuDeuxfois: **nitpick**: Explicit call, no magic. I’m starting to prefer this kind of approach.

openaev-model/src/main/java/io/openaev/database/audit/AuditLogContext.java
Dimfacion: The thread local logic is used as a backup if we're not in an API call so I guess it's true that we could remove it since it shouldn't be happening outside of it

openaev-model/src/main/java/io/openaev/database/model/Asset.java
Dimfacion: Good call

openaev-model/src/main/java/io/openaev/database/model/Agent.java
Dimfacion: I'd rather have a declarative approach (let's ban magic) so that a dev wouldn't pull their hair out wondering why it's not logged even though the update date is changing. Wdyt ?

openaev-model/src/main/java/io/openaev/database/model/Agent.java
Dimfacion: The created date shouldn't move between two comparison but you're right that since it shouldn't move anyway, we can speed up the comparison by ignoring it

Dimfacion:
As seen together @RomuDeuxfois, I've added a filter to avoid serializing some fields if they are with the correct annotation. That should make comparisons slimmer. Let me know what you think :)

openaev-api/src/main/java/io/openaev/service/EndpointService.java
RomuDeuxfois: **todo**: We have an issue here.
This is the log I have

2026-06-29T15:58:19.771+02:00 [AUDIT] {
  "id" : "199ffd03-b36a-41c3-bc74-1452d406c81e",
  "timestamp" : "2026-06-29T13:58:19.638054Z",
  "entity_type" : "Activity",
  "created_at" : "2026-06-29T13:58:19.638054Z",
  "event_type" : "mutation",
  "event_status" : "success",
  "event_access" : "extended",
  "event_scope" : "create",
  "user_id" : "a9d14910-34b6-4679-aa11-262a12ccf2cc",
  "tenant_id" : "2cffad3a-0001-4078-b0e2-ef74274022c3",
  "user_metadata" : {
    "ip" : "0:0:0:0:0:0:0:1",
    "user_email" : "9241fc76f00e9b502f4eeba27ff863f56e65c97cbb712f3589bec863118a6a3e",
    "user_agent" : "openaev-agent/2.260623.0",
    "x_forwarded_for" : null,
    "session_id" : "5E488BE6B04B730CF1F3DBBF2C9091CA"
  },
  "request_metadata" : {
    "url" : "http://localhost:8080/api/tenants/2cffad3a-0001-4078-b0e2-ef74274022c3/endpoints/register",
    "method" : "POST",
    "signature" : {
      "method" : "io.openaev.rest.asset.endpoint.EndpointApi.upsertEndpoint",
      "parameters" : {
        "input" : "@RequestBody"
      }
    }
  },
  "context_data" : {
    "entity_type" : "Agent",
    "input" : {
      "seenIp" : "0:0:0:0:0:0:0:1",
      "elevated" : true,
      "asset_name" : "LAPT-FILI-0067",
      "asset_external_reference" : "6acc8c2a54dac178cc73c2eed06d23ab3ea8bb3f2bda6178d802a8ddbe3a1f62",
      "endpoint_platform" : "Windows",
      "endpoint_arch" : "x86_64",
      "endpoint_ips" : [ "fe80::8070:739f:2d82:c37f", "2a02:8428:ae01:6101:fde4:68f8:3c63:5383", "2a02:8428:ae01:6101:38a1:f493:f5eb:dd5b", "fe80::d3c3:2bb9:fdde:34db", "192.168.1.219", "fe80::93d6:8689:7d65:aba0", "fe80::d033:45f2:208f:466f", "fe80::b316:d785:547b:18d", "fe80::dee3:70e6:7b20:e16", "172.18.144.1", "fe80::13d9:ffb0:3d33:95fa", "172.17.192.1" ],
      "endpoint_hostname" : "LAPT-FILI-0067",
      "endpoint_agent_version" : "2.260623.0",
      "endpoint_mac_addresses" : "e56843542f05c938ed043df1b9a00b5753e14db71103b25e7b1ed046a3b64b20",
      "agent_is_elevated" : true,
      "agent_executed_by_user" : "azuread\\romualdlemesle",
      "agent_installation_mode" : "session-user",
      "agent_installation_directory" : "C:\\Users\\RomualdLemesle\\.openaev\\OAEVAgent-Session-Administrator-azureadromualdlemesle",
      "agent_service_name" : "OAEVAgent-Session"
    },
    "output" : {
      "listened" : true,
      "asset_id" : "551fd5c5-ed22-495e-94c4-002cc0fb0981",
      "asset_type" : "Endpoint",
      "asset_name" : "LAPT-FILI",
      "asset_tags" : [ "3d7efede-19f0-4489-a50f-58e53fa43563" ],
      "asset_created_at" : "2026-06-29T13:21:35.935649Z",
      "asset_updated_at" : "2026-06-29T13:58:19.626383Z",
      "endpoint_ips" : [ "fe80::b316:d785:547b:18d", "192.168.1.219", "fe80::d033:45f2:208f:466f", "fe80::13d9:ffb0:3d33:95fa", "172.17.192.1", "fe80::8070:739f:2d82:c37f", "fe80::93d6:8689:7d65:aba0", "2a02:8428:ae01:6101:38a1:f493:f5eb:dd5b", "172.18.144.1", "fe80::dee3:70e6:7b20:e16", "2a02:8428:ae01:6101:fde4:68f8:3c63:5383", "fe80::d3c3:2bb9:fdde:34db" ],
      "endpoint_seen_ip" : "0:0:0:0:0:0:0:1",
      "endpoint_hostname" : "lapt-fili-0067",
      "endpoint_platform" : "Windows",
      "endpoint_arch" : "x86_64",
      "endpoint_mac_addresses" : "d2056229107e23ab6f9d93c837b20609529245bde7acf9e5977c7ef4b5530129",
      "asset_agents" : [ {
        "listened" : true,
        "agent_id" : "c9d562ed-3583-42bf-a7a0-320141519911",
        "agent_asset" : "551fd5c5-ed22-495e-94c4-002cc0fb0981",
        "agent_privilege" : "admin",
        "agent_deployment_mode" : "session",
        "agent_executed_by_user" : "azuread\\romualdlemes",
        "agent_executor" : "2f9a0936-c327-4e95-b406-d161d32a2501",
        "agent_version" : "2.260623.0",
        "agent_external_reference" : "6acc8c2a54dac178cc73c2eed06d23ab3ea8bb3f2bda6178d802a8ddbe3a1f62",
        "agent_last_seen" : "2026-06-29T13:56:18.776376Z",
        "agent_created_at" : "2026-06-29T13:21:35.948156Z",
        "agent_updated_at" : "2026-06-29T13:21:35.948156Z",
        "agent_cleared_at" : "2026-06-29T13:21:35.948156Z",
        "agent_active" : true
      } ]
    },
    "message" : "creates Agent"
  }
}

I have "entity_type" : "Agent", input is an agent but output is an Endpoint. That's weird no ?

Dimfacion: That's the way the API is designed for this endpoint. I agree that this is weird but it's out of scope IMO :
`
  @PostMapping({ENDPOINT_URI + "/register", TENANT_ENDPOINT_URI + "/register"})
  @AccessControl(actionPerformed = Action.CREATE, resourceType = ResourceType.AGENT)
  public Endpoint upsertEndpoint(@Valid @RequestBody final EndpointRegisterInput input)
      throws IOException {
    input.setSeenIp(HttpReqRespUtils.getClientIpAddressIfServletRequestExist());
    return this.endpointService.register(input, TenantContext.getCurrentTenant());
  }
`
  
  As you can see, the endpoint is returning an Endpoint but the resourceType is Agent

RomuDeuxfois: Can you check if it's possible to align that ? It's really weird when I read the log.
Also, I add a new agent and my output don't contains it.
{2A9C8625-BFFA-4779-AE8F-B0E4D0DDB1BF}

Dimfacion: @RomuDeuxfois As seen together, I made a few changes to the way the agent and endpoint are registered

RomuDeuxfois (approved):

Issue #6332, opened 2026-06-22
fix(api): in CE, all APIs are always responding 403 denied while using an admin bearer token


SamuelHassine:
## Description

Maybe be related to #6331.

fix(api): in CE, all APIs are always responding 403 denied 

Request GET http://localhost:3001/api/endpoints
Bearer token admin

=> "Forbidden",
    "trace": "io.openaev.aop.AccessControlAspect$2: 403 FORBIDDEN \"Access denied for user: admin@filigran.io\"\

RomuDeuxfois:
Maybe link this to this issue:
[https://github.com/OpenAEV-Platform/openaev/issues/6331](https://github.com/OpenAEV-Platform/openaev/issues/6331)

I reproduced it with the following steps: I started from EE with Tenant A, deleted Tenant A, then switched to CE with the default tenant. The agent from Tenant A then tried to register again, but failed.

PR #6342, opened 2026-06-22
fix(multi-tenancy): remove prevent data leak and let fallback on default tenant (#6331)

### LINKED ISSUE #6331: fix(login): in community edition, login with an admin account triggers access denied and full empty menu (needs hard refresh)
## Description

fix(login): in community edition, login with an admin account triggers access denied and full empty menu (needs hard refresh)

1. Create an admin account (in group admin, role admin)
2. Login with the admin account

=> Left menu is empty, error message "Access denied"
=> Hard refresh => everything is back to normal

RomuDeuxfois:


SamuelHassine:
Closing in favor of #6344, which fixes the same CE capability regression (this issue #6331, and #6332) while keeping the multi-tenancy data-leak guard intact.

The difference: this PR removes the `scopedGroups()` early-return and falls back to the default tenant unconditionally, which also grants default-tenant capabilities to tenant-less platform operations in EE (the "remove prevent data leak" in the title). #6344 instead populates `TenantContext` to the default tenant ONLY in Community Edition (EE license inactive), via the `TenantInterceptor`, so CE is fixed while #6039's EE platform/tenant scoping is preserved (no cross-scope leak).

#6344 also fixes a second, related bug this PR does not touch: bearer-token API requests get a 403 once the platform returns a session cookie (CSRF is re-enabled when the client replays the cookie) - see #6343. That is what breaks Postman / httpx / requests and the external integration.

Thanks @RomuDeuxfois - pinpointing `scopedGroups()` as the culprit is exactly what led to the root cause.

Issue #6276, opened 2026-06-17
[Multi-tenancy] refactor(integrations): make all built-ins follow the self-register pattern


corinnekrych:
## Goal
Remove `BuiltinTenantRegistrable` loop from `ManagerFactory.createManager()` by converting the two standalone collectors to proper `IntegrationFactory` + `Integration` pairs — the same pattern used by `EmailInjector`, `OpenaevInjector`, etc.

## Current Design (problematic)

```
ManagerFactory.createManager(tenantId)
  ├── for (BuiltinTenantRegistrable) → registerForTenant(tenantId)  ← TO REMOVE
  │     ├── EmailInjectorIntegrationFactory (also an IntegrationFactory — redundant call)
  │     ├── OpenaevInjectorIntegrationFactory (also an IntegrationFactory — redundant call)
  │     ├── ExpectationsExpirationManagerJob          ← standalone, NOT an Integration
  │     └── ExpectationsVulnerabilityManagerCollector  ← standalone, NOT an Integration
  │
  └── new Manager(tenantId, factories)
        └── monitorIntegrations()
              └── for factory: findRelatedInstances → sync → spawnedIntegrations
```

**Problem:** `ExpectationsExpirationManagerJob` and `ExpectationsVulnerabilityManagerCollector` are special-cased outside the Integration framework. They don't follow the factory→spawn→innerStart lifecycle.

## Target Design

Both collectors become proper integration triads (Factory + Integration + existing service):

```
ManagerFactory.createManager(tenantId)
  └── new Manager(tenantId, factories)  ← ALL factories, no separate loop
        └── monitorIntegrations()
              └── for factory: findRelatedInstances → sync → innerStart
                    ├── EmailInjectorIntegrationFactory
                    ├── ...existing injectors/executors...
                    ├── ExpectationsExpirationManagerIntegrationFactory  ← NEW
                    └── ExpectationsVulnerabilityManagerIntegrationFactory  ← NEW
```

## That's the full plan. The key insight is:
* Convert both standalone collectors into proper IntegrationFactory + Integration pairs (like EmailInjector already does)
* Move registration into BuiltinIntegrationFactory.sync() so it self-registers on first discovery
* Move scheduling into Integration.innerStart() so the periodic work is tenant-scoped and lifecycle-managed (start/stop/restart)
* Delete the BuiltinTenantRegistrable interface and the loop in ManagerFactory

[...]

PR #6374, opened 2026-06-24
perf(integrations): ManagerIntegrationsSyncJob.execute is slow under multi-tenancy (#6263)

### LINKED ISSUE #6263: [Multi-Tenancy] perf(integrations): ManagerIntegrationsSyncJob.execute is slow under multi-tenancy
## Description

`ManagerIntegrationsSyncJob.execute` job runs every 15s. simce we have introduce multi-tenancy the job queries all instances per factory per tenant

### Root Causes
1. **Full scan every tick** — `findRelatedInstances()` queries all instances per factory per tenant, even if nothing changed
2. **Re-initialise all integrations every tick** — `integration.initialise()` runs on every spawned integration (DB read + potential write), not just new/changed ones
3. **Sequential tenant processing** — no parallelism, one slow tenant blocks all others
4. Is **for... component.registerForTenant(tenantId);** necessary ?

## AC
1/ Let's add monitoring 
- Per-tenant `WARN` log when > 500ms
- OTel histograms: `manager_sync_job_duration_ms`, `manager_sync_tenant_duration_ms`
- OTel counters: `manager_sync_tenants_processed_total`, `manager_sync_errors_total`

2/ Explore next potential steps - here some AI proposals
1. **Skip unchanged integrations** — only call `initialise()` on integrations with dirty/changed state
2. **Incremental discovery** — use a cursor/timestamp to avoid full `findRelatedInstances()` scan
3. **Chunk tenants per tick** — process N tenants per run with a rotating cursor
4. **Parallelize** — bounded thread pool for tenant processing (careful with `TenantContext`)

damgouj:
### Proposed changes

* Improvements for the manager integrations sync job (async for each tenant instead of sync)

### Testing Instructions

* Check everything (catalog) is working as before

### Related issues

* Closes #6263 

### Checklist

### DIFF
--- a/openaev-api/src/main/java/io/openaev/config/ThreadPoolTaskSchedulerConfig.java
+++ b/openaev-api/src/main/java/io/openaev/config/ThreadPoolTaskSchedulerConfig.java
@@ -3,6 +3,7 @@ package io.openaev.config;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ThreadPoolExecutor;
 import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
@@ -11,6 +12,13 @@ import org.springframework.scheduling.concurrent.ThreadPoolTaskScheduler;
 @Configuration
 @Slf4j
 public class ThreadPoolTaskSchedulerConfig {
+
+  @Value("${spring.datasource.hikari.maximum-pool-size:20}")
+  private int dbPoolSize;
+
+  // optional manual override per environment
+  @Value("${manager.integrations.concurrency:#{null}}")
+  private Integer concurrencyOverride;
 
   @Bean
   public ThreadPoolTaskScheduler threadPoolTaskScheduler() {
@@ -36,6 +44,25 @@ public class ThreadPoolTaskSchedulerConfig {
     // overloading the tab of a user connected when having a lot of events
     executor.setRejectedExecutionHandler(new ThreadPoolExecutor.DiscardOldestPolicy());
 
+    executor.initialize();
+    return executor;
+  }
+
+  /** Dedicated executor for manager integrations tenant sync dispatches. */
+  @Bean(name = "managerIntegrationsExecutor")
+  public Executor managerIntegrationsExecutor() {
+    // Bottleneck is the shared DB connection pool, not CPU.
+    // Default to ~40% of the pool so the rest of the app keeps connections available.
+    int concurrency =
+        concurrencyOverride != null ? concurrencyOverride : Math.max(1, dbPoolSize * 2 / 5);
+    ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
+    executor.setCorePoolSize(concurrency);
+    executor.setMaxPoolSize(concurrency);
+    executor.setQueueCapacity(0);
+    executor.setKeepAliveSeconds(30);
+    executor.setAllowCoreThreadTimeOut(true);
+    executor.setThreadNamePrefix("ManagerIntegrations-");
+    executor.setRejectedExecutionHandler(new ThreadPoolExecutor.CallerRunsPolicy());
     executor.initialize();
     return executor;
   }
--- a/openaev-api/src/main/java/io/openaev/config/ThreadPoolTaskSchedulerConfig.java
+++ b/openaev-api/src/main/java/io/openaev/config/ThreadPoolTaskSchedulerConfig.java
@@ -16,7 +16,14 @@ public class ThreadPoolTaskSchedulerConfig {
   @Value("${spring.datasource.hikari.maximum-pool-size:20}")
   private int dbPoolSize;
 
-  // optional manual override per environment
+  // Optional per-environment override for the managerIntegrationsExecutor thread-pool size.
+  // When unset (the default), concurrency is auto-tuned to ~40% of the Hikari pool so that
+  // tenant-sync threads do not starve the rest of the application.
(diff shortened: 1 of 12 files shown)


openaev-api/src/main/java/io/openaev/config/ThreadPoolTaskSchedulerConfig.java
```diff
@@ -12,6 +13,13 @@
 @Slf4j
 public class ThreadPoolTaskSchedulerConfig {
 
+  @Value("${spring.datasource.hikari.maximum-pool-size:20}")
+  private int dbPoolSize;
+
+  // optional manual override per environment
```
Dimfacion: **nitpick :** maybe add a comment here to explain what this property does ?

Dimfacion:
Should we also increase the test coverage of ManagerIntegrationsSyncJob.java during this PR ?

openaev-api/src/main/java/io/openaev/scheduler/jobs/ManagerIntegrationsSyncJob.java
antoinemzs, line 1: Could it be simpler to pass a tenant parameter to this class, and provision as many jobs as there are tenants in `io.openaev.scheduler.PlatformJobDefinitions` ?

the JobBUilder class accepts data objects in its JobDataMap via e.g. https://www.quartz-scheduler.org/api/2.5.x/org/quartz/JobBuilder.html#usingJobData(java.lang.String,java.lang.String) (more overloads available)

we would then provision a separate trigger for each tenant-specific job instance

we would benefit from the quartz thread pool and also prevent concurrency for monitoring integrations within the same tenant with the already set `@DisallowConcurrentExecution` annotation (this rule obeys the `jobKey` parameter, si this would have to bear the tenant id or some other discriminator).

damgouj: I thought it was part of this issue https://github.com/OpenAEV-Platform/openaev/issues/6276 ?
With your solution, if there are 50 tenants, 50 jobs will be added. I tried to manage the number of "jobs" by sizing the async threads that the platform can manage.
But available to discuss of it if you want with you.

openaev-api/src/main/java/io/openaev/scheduler/jobs/ManagerIntegrationsSyncJob.java
```diff
@@ -29,30 +38,56 @@ public void execute(JobExecutionContext jobExecutionContext) throws JobExecution
...
+        } catch (RuntimeException e) {
+          runningTenantSyncs.remove(tenantId);
+          throw new JobExecutionException(e);
         }
       }
       long jobDuration = System.currentTimeMillis() - jobStart;
       if (jobDuration > EXECUTION_TIME_THRESHOLD) {
         log.warn(
-            "==> ManagerIntegrationsSyncJob.execute took {} ms (>500ms threshold)", jobDuration);
+            "==> ManagerIntegrationsSyncJob.execute took {} ms (threshold {} ms)",
```
corinnekrych: neat: but remove "==>' to align with other log

openaev-api/src/main/java/io/openaev/scheduler/jobs/ManagerIntegrationsSyncJob.java
corinnekrych: I think Damien's approach is good fit for our need. It reuses existing  pattern in code base and enrich it with the controls of parallelism via a bounded executor sized to the DB connection pool — a sensible choice to avoid DB contention. Let's go for it and monitor with warn log if job execution at tenant level will take mroe soem.
Tested locally with a bunch of tenants (40+) and it works well.

corinnekrych (approved):

openaev-api/src/main/java/io/openaev/scheduler/jobs/ManagerIntegrationsSyncJob.java
antoinemzs: You can configure the quartz thread pool

My concern is duplicating and bypassing a solution that we have in place with our in house solution and increase the maintenance load and code surface.

antoinemzs: > I thought it was part of this issue #6276 ?

IMO this is orthogonal, the concern here is to parallelise the background tenant monitoring (instead of serialising the monitoring per tenant); I'm suggesting a pattern that integrates with our accepted background job implementation (quartz), implementing another background job synchronisation engine within another seems riskier

laugiov: I think we should separate two horizons here, because they answer different questions.

Short term. @damgouj 's approach is the right one to fix the perf issue now. It targets the actual constraint, the DB connection pool, with a bounded executor sized to it; per-tenant overlap is now guarded by runningTenantSyncs; and it's a contained, tested change. I'm for merging it as the fix, once the remaining review points are wrapped up (a test for the overlap guard, the executor sizing, and the metrics from the AC).

Medium / long term. @antoinemzs 's point is the right one, and it lines up with the direction I would to take multi-tenancy. We're moving toward a model where every unit of work carries its tenant and the database rejects everything else (fail-closed, no default wildcard). In that model, "one Quartz job per tenant" is the most natural expression: each job is a unit of work scoped to its tenant, with no need for a "read all tenants" call in the hot path, and it reuses the Quartz infrastructure we're already making tenant-aware (self-proxy + PlatformJobDefinitions) instead of maintaining a second concurrency engine alongside it. So it's less surface area in the long run, not more.

So I see @damgouj 's PR as transitional: it unblocks the perf problem, but the target shape for background jobs (how they carry the tenant and handle their parallelism) is a decision for the multi-tenancy rework, not for this PR. I need to align it with @xfournet, so I'd rather not freeze the underlying architecture here.

IMO : let's move forward with @damgouj 's proposal for the perf now, and we'll pick up the "per-tenant jobs, @antoinemzs -style" topic in the rework if this is the way we want to go.

PR #6375, opened 2026-06-24
fix(asset): auto-fill seen_ip on agentless endpoint creation (#3723)

### LINKED ISSUE #3723: fix: update agentless IP values
## Context

When I add an agentless, the IP address fill the local_ip, if I made a mistake and use the seen_ip on injectors like nmap/nuclei that will not work and it's logical.

To avoid that and make it more easy for new comers I'd like to fill both local_ip & seen_ip with the IP entered by user.

Image

Image

Seb-MIGUEL:
Closes #3723

## Problem
When creating an agentless endpoint, `seen_ip` was never set while `local_ip` (from `ips[0]`) was filled correctly. This caused injectors using `seen_ip` targeting (nmap, nuclei) to silently fail.

## Root Cause
`EndpointService.createEndpoint(EndpointInput)` called `endpoint.setIps()` but never `endpoint.setSeenIp()`. The agent registration flow (`createNewEndpointAndAgent`) already set `seenIp` correctly — only the agentless path was missing it.

## Fix
- **createEndpoint**: auto-fill `seenIp` from `ips[0]` after IP sanitization
- **upsertEndpoint**: fill `seenIp` only when currently null (preserves intentionally different values)

## Tests
5 regression tests in `@Nested Bug #3723`:
- createEndpoint with IPs → seenIp set to first IP
- createEndpoint with null/empty IPs → seenIp stays null
- upsertEndpoint with null seenIp → filled from IPs
- upsertEndpoint with existing seenIp → not overwritten


openaev-api/src/main/java/io/openaev/service/EndpointService.java
```diff
@@ -128,7 +128,11 @@ public Endpoint createEndpoint(@NotNull final Endpoint endpoint) {
   public Endpoint createEndpoint(@NotNull final EndpointInput input) {
     Endpoint endpoint = new Endpoint();
     endpoint.setUpdateAttributes(input);
-    endpoint.setIps(EndpointMapper.setIps(input.getIps()));
+    String[] ips = EndpointMapper.setIps(input.getIps());
```
RomuDeuxfois, line 131: **question**: What about updateEndpoint ? Could be usefull to change the getter from the Endpoint model direclty to handle it more globally. What do you think ?

Seb-MIGUEL: Good catch! I have added seenIp handling to updateEndpoint(String, EndpointInput, String) as well in the latest push.

Regarding the model-level approach (getter/setter on Endpoint): I considered it but decided against it because Base.setUpdateAttributes() uses BeanUtils.copyProperties(input, this) which copies raw IPs from the DTO before EndpointMapper.setIps() sanitizes them. A setter override would set seenIp to an unsanitized value during that copy phase.

The service-level approach keeps the logic explicit and ensures we always use sanitized IPs (after EndpointMapper.setIps() has filtered, lowercased, and deduplicated).

openaev-api/src/test/java/io/openaev/service/endpoint/EndpointServiceTest.java
```diff
@@ -110,6 +110,136 @@ void shouldCreateEndpointFromInput_withIpsAndMacAndTags() {
...
+    @Test
+    @DisplayName("upsertEndpoint should fill seenIp when updating endpoint with null seenIp")
+    void given_existingEndpointWithNullSeenIp_should_fillSeenIpOnUpsert() {
+      // Arrange
+      Endpoint existing = new Endpoint();
+      existing.setId("ep-existing");
+      existing.setSeenIp(null);
+      existing.setTags(Collections.emptySet());
+
+      when(endpointRepository.findByExternalReference("ext-ref-1")).thenReturn(List.of(existing));
```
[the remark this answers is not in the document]
Seb-MIGUEL: Fixed: added TenantContext.getCurrentTenant() as second argument to findByExternalReference() in the latest push (c6d3e50).

openaev-api/src/test/java/io/openaev/service/endpoint/EndpointServiceTest.java
```diff
@@ -110,6 +110,136 @@ void shouldCreateEndpointFromInput_withIpsAndMacAndTags() {
...
+      EndpointInput input = new EndpointInput();
+      input.setName("updated-endpoint-2");
+      input.setPlatform(Endpoint.PLATFORM_TYPE.Linux);
+      input.setArch(Endpoint.PLATFORM_ARCH.x86_64);
+      input.setExternalReference("ext-ref-2");
+      input.setIps(new String[] {"172.16.0.2"});
+      input.setTagIds(List.of());
+
+      // Act
+      Endpoint result = endpointService.upsertEndpoint(input);
```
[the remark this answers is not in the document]
Seb-MIGUEL: Fixed: added TenantContext.getCurrentTenant() as second argument to upsertEndpoint() in the latest push (c6d3e50).

openaev-api/src/test/java/io/openaev/service/endpoint/EndpointServiceTest.java
```diff
@@ -110,6 +110,136 @@ void shouldCreateEndpointFromInput_withIpsAndMacAndTags() {
...
+      EndpointInput input = new EndpointInput();
+      input.setName("updated-endpoint");
+      input.setPlatform(Endpoint.PLATFORM_TYPE.Linux);
+      input.setArch(Endpoint.PLATFORM_ARCH.x86_64);
+      input.setExternalReference("ext-ref-1");
+      input.setIps(new String[] {"172.16.0.1"});
+      input.setTagIds(List.of());
+
+      // Act
+      Endpoint result = endpointService.upsertEndpoint(input);
```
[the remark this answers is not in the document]
Seb-MIGUEL: Fixed: added TenantContext.getCurrentTenant() as second argument to upsertEndpoint() in the latest push (c6d3e50).

openaev-api/src/test/java/io/openaev/service/endpoint/EndpointServiceTest.java
```diff
@@ -110,6 +110,136 @@ void shouldCreateEndpointFromInput_withIpsAndMacAndTags() {
...
+    @Test
+    @DisplayName("upsertEndpoint should NOT overwrite existing seenIp")
+    void given_existingEndpointWithSeenIp_should_notOverwriteOnUpsert() {
+      // Arrange
+      Endpoint existing = new Endpoint();
+      existing.setId("ep-existing");
+      existing.setSeenIp("10.0.0.99");
+      existing.setTags(Collections.emptySet());
+
+      when(endpointRepository.findByExternalReference("ext-ref-2")).thenReturn(List.of(existing));
```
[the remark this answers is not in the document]
Seb-MIGUEL: Fixed: added TenantContext.getCurrentTenant() as second argument to findByExternalReference() in the latest push (c6d3e50).

openaev-api/src/main/java/io/openaev/service/EndpointService.java
RomuDeuxfois: Ok, I understand the issue. Can we have something more generic in the service in this case ? Like a sanity check. And do we want to verify that seen Ip is always fullfilled on creation & updat on the DB ?

Seb-MIGUEL: Done! Extracted a private `ensureSeenIp(Endpoint)` helper that centralizes the sanity check (f69a561). It is now called in all 3 paths: createEndpoint, updateEndpoint, and upsertEndpoint.

Regarding a DB-level constraint: a NOT NULL on `endpoint_seen_ip` would be too strict since endpoints can legitimately have no IPs (and thus no seenIp). However, we could add a data migration to backfill existing endpoints that have `ips` but null `seenIp` — want me to add that?

RomuDeuxfois: @Seb-MIGUEL what do you think ?

Seb-MIGUEL: Hey @RomuDeuxfois, I replied above but maybe you missed it — I extracted a private `ensureSeenIp(Endpoint)` helper that centralizes the sanity check (see commit f69a561, now updated in d042c5e). It is called in all 3 paths: createEndpoint, updateEndpoint, and upsertEndpoint.

For the DB-level guarantee: a NOT NULL constraint on `endpoint_seen_ip` would be too strict since endpoints can legitimately have no IPs (and thus no seenIp). But we could add a data migration to backfill existing endpoints that have `ips` set but null `seenIp` — happy to add that if you think it is worth it.

Seb-MIGUEL: Good point — yes, I think we should add a data migration to backfill existing endpoints. Something like:

```sql
UPDATE endpoints SET endpoint_seen_ip = endpoint_ips[1]
WHERE endpoint_seen_ip IS NULL AND endpoint_ips IS NOT NULL AND array_length(endpoint_ips, 1) > 0;
```

This ensures existing data is also retrocompatible. I will add it to this PR.

RomuDeuxfois: Ok go for it.

.github/bugs/BUG-3723.md
```diff
@@ -0,0 +1,123 @@
+# Bug Analysis: #3723 — Update agentless IP values
```
RomuDeuxfois: Why adding this ? What's the usage behind ? having a markdown to explain a bugfix seems overkill to me.

openaev-api/src/main/java/io/openaev/migration/V5_26__Backfill_endpoint_seen_ip.java
```diff
@@ -0,0 +1,26 @@
+package io.openaev.migration;
+
+import java.sql.Statement;
+import org.flywaydb.core.api.migration.BaseJavaMigration;
+import org.flywaydb.core.api.migration.Context;
+import org.springframework.stereotype.Component;
+
+/**
+ * Backfills {@code endpoint_seen_ip} for existing endpoints that have IPs but a null seen_ip. This
```
RomuDeuxfois, line 9: Not sure it's good to link to a specific issue when we explain things. Issue can be rewrite on github.

openaev-api/src/test/java/io/openaev/service/endpoint/EndpointServiceTest.java
```diff
@@ -110,6 +110,143 @@ void shouldCreateEndpointFromInput_withIpsAndMacAndTags() {
     }
   }
 
+  @Nested
+  @DisplayName("Bug #3723: agentless seen_ip should be auto-filled")
```
RomuDeuxfois: **todo**: Can be better to modify existing tests than adding new one. What do you think ?

Seb-MIGUEL: Done (25928002d). Removed the separate @Nested AgentlessSeenIpBug3723 class and integrated seenIp assertions into existing test groups:

- `CreateEndpoint.shouldCreateEndpointFromInput_withIpsAndMacAndTags`: added seenIp assertion + IPs in input
- `CreateEndpoint.shouldLeaveSeenIpNull_whenNoIpsProvided`: new (edge case, no equivalent existed)
- `CreateEndpoint.shouldLeaveSeenIpNull_whenIpsArrayIsEmpty`: new (edge case, no equivalent existed)
- `UpdateEndpoint.shouldUpdateEndpointFromInput`: added IPs + seenIp assertion
- `UpdateEndpoint.shouldFillSeenIpOnUpsert_whenSeenIpIsNull`: new (upsert, no equivalent existed)
- `UpdateEndpoint.shouldNotOverwriteExistingSeenIpOnUpsert`: new (upsert, no equivalent existed)

.github/bugs/BUG-3723.md
Seb-MIGUEL: You're right, it was part of an automated bug analysis workflow I was experimenting with, but it's definitely overkill for this kind of fix. I'll remove it from the PR.

openaev-api/src/main/java/io/openaev/migration/V5_26__Backfill_endpoint_seen_ip.java
Seb-MIGUEL: Good point — I'll make the Javadoc description standalone without referencing the issue number, so it stays meaningful even if the issue changes.

openaev-api/src/test/java/io/openaev/service/endpoint/EndpointServiceTest.java
```diff
@@ -105,9 +105,54 @@ void shouldCreateEndpointFromInput_withIpsAndMacAndTags() {
       // -------- Assert --------
       assertNotNull(result);
       assertTrue(result.isEoL());
+      assertThat(result.getSeenIp()).isEqualTo("10.0.0.1");
```
RomuDeuxfois, line 108: **todo**: This tests is enough. Don't add a lot of testing for just one little changes.

Seb-MIGUEL: Agreed — removed the 4 extra edge case tests (null IPs, empty IPs, upsert fill, upsert preserve). Kept only the seenIp assertions in the existing create and update happy path tests (9006b5a).

RomuDeuxfois (approved):

Issue #6379, opened 2026-06-24
feat(debug): Carry the trace id across RabbitMQ and Quartz to trace a simulation end to end


laugiov:
## Context

The debug mode (#6378) correlates work within one request thread and across Spring `@Async`, but
the trace id stops at the RabbitMQ and Quartz boundaries. The simulation engine runs behind
those two boundaries, so today we can trace a request up to the queue or the job, but not through
the work it triggers. This is the piece that turns a per-request SQL/JFR diagnostic into real
end-to-end tracing.

## Goal

Propagate the trace context across both boundaries so one simulation can be followed end to end:
the request that creates it, the message it publishes, the consumer, and the Quartz job run.

## Approach

- RabbitMQ: put `traceId`/`spanId` in the message headers on publish, and restore them (start a
  child span) on consume. Micrometer Tracing has an instrumentation pattern for this.
- Quartz: a `JobListener` that opens a Micrometer observation per `execute()`, so each job run
  gets its own span. Link it to the triggering request with a span link where that relation is
  known.

## Notes

- This is not debug-only. Decide whether context propagation should be always on or gated by the
  debug flag.
- The same masking rules apply to anything logged on these paths.

## Acceptance

- [ ] A request that creates a simulation, the RabbitMQ message, the consumer, and the Quartz
      job execution all share or link one trace, shown by a test
- [ ] No measurable cost when the debug mode is off
- [ ] The end-to-end path is documented

PR #6380, opened 2026-06-24
feat(debug): global debug mode (correlation ids, SQL detail, JFR profiling) (#6378)

### LINKED ISSUE #6378: feat(debug) Global debug mode: correlation ids, SQL detail, and JFR profiling
## What this adds

A global, instance-wide debug mode, off by default, behind one flag. Three parts:

- Correlation id in every log line (Micrometer Tracing). All log lines of one request share the
  same trace id.
- SQL detail via datasource-proxy: each statement (ORM and native) is logged with its time and
  its parameters, masked.
- JVM profiling via the JDK-built-in Java Flight Recorder: a bounded local `.jfr` file, no
  server, no agent.

Plus a per-request ORM summary that flags N+1 and chatty requests.

## Safety

- Off by default. Near-zero cost when off (no proxy on the query path).
- Production barrier: it does not start in production unless `allow-in-production=true` is set.
- Optional `auto-disable-after` timer so a forgotten activation stops on its own.
- Mandatory masking of secrets and PII, with a deny-by-default option (`mask-all-parameters`).
- Single profiler at a time (JFR; yields to Pyroscope if its agent is enabled).
- Bounded SQL log file and bounded JFR dump retention. Clear failure on a read-only filesystem.

## Known limitation

The trace id stays in-process. It does not cross the RabbitMQ or Quartz boundaries yet, so a
simulation is not traced end to end. Tracked in #6379 .

## Acceptance

- [ ] Off by default, no behaviour change on a standard install
- [ ] One request: all log lines and SQL lines share one trace id
- [ ] Secrets and PII never appear in the logs (tests)
- [ ] Production barrier and auto-disable work (tests)
- [ ] JFR file count and size stay bounded (tests)
- [ ] Operator doc: how to turn it on and off, the writable volume needed

laugiov:
# Add a global debug mode (correlation ids, SQL detail, JFR profiling)

Closes #6378

## What this adds

A global, instance-wide debug mode to diagnose bugs and performance issues on on-prem installations.
It is delivered in the product (application code, no infra) and is fully disabled by default. It has
three parts, all behind a single flag:

- **Correlation id in every log line** via Micrometer Tracing. All log lines of a request or job
  share the same `traceId`. No tracing backend is required; ids go to the normal log output.
- **SQL detail** via datasource-proxy: every statement (ORM and native) is logged with its execution
  time and parameters, on the dedicated `io.openaev.debug.sql` logger.
- **JVM profiling** via the JDK-built-in Java Flight Recorder: a bounded recording is started, dumped
  on a timer and flushed on shutdown to a known directory. No extra dependency, no agent.

The toggle scope is global. Per-request scoping is out of scope for this POC.

## Known limitation

The trace id stays in-process and across Spring `@Async`, but does not cross the RabbitMQ or Quartz
boundaries, so a simulation is not traced end to end yet. That propagation is the follow-up tracked in
#6379.

## ⚠️ Needs reviewer input: should tracing run on OpenTelemetry?

Please look at this point and help decide. The owners' decision is to unify on **OpenTelemetry**, but
that change is **blocked** today and I did not force it. Here is the situation in plain terms.

**Two different things share one toolbox.** "Observability" has two parts:

- **Metrics** = numbers on dashboards (requests/min, DB load, ...). OpenAEV already sends these to the
  Filigran telemetry server. This is the existing, working setup.
- **Tracing** = following one request end to end. That is the `traceId` this PR adds.

Both can run on the same library, **OpenTelemetry** (OTel). The decision was: put both on OTel, one
stack.

**Why it is blocked (verified, not theory).** OpenAEV configures OTel manually for metrics only:
`io.openaev.telemetry.OpenTelemetryConfig` exposes one `OpenTelemetry` bean with a meter provider but
**no tracer provider** (and a `noop` fallback when the telemetry endpoint is unreachable). Spring
Boot's tracing auto-config creates its `OpenTelemetry` only `@ConditionalOnMissingBean` -- so it sees
the project's bean, backs off, and the tracer is never wired. Tested by swapping the dependency to
`micrometer-tracing-bridge-otel`: a real request then logged an **empty** trace id. So the PR keeps
the self-contained **Brave** bridge, which produces correct trace ids today.

**What it would take to unblock (small, but in sensitive code).** Give the metrics `OpenTelemetry`
bean a tracer provider too (a few dozen lines), and make sure tracing does **not** get switched off by
the metrics `noop` fallback (the trace id in logs must not depend on the telemetry server being up).
The change is small but lives in the metrics code that feeds production dashboards, so it should be
done/validated by the **metrics owners**, not forced from this debug feature. Once done, switching the
debug mode from Brave to OTel is a one-line dependency change.

**Ask for the reviewer (metrics/Platform):** is it safe to extend the metrics `OpenTelemetry` bean
with a tracer provider (decoupled from the `noop` fallback), so we can run on a single OTel stack? If
yes, that is a small follow-up; if not, we keep Brave. Full analysis in `ADR-0002-tracing-stack.md`
(Status: Accepted decision, implementation blocked).

## How to enable / disable

Off by default. Enable (preferably via env var so it is easy to revert):

```bash
OPENAEV_DEBUG_ENABLED=true
```

or `openaev.debug.enabled=true` in `application.properties`. Disable by removing the flag and
restarting. While active, a loud warning banner is logged at startup and repeated on a fixed
interval. See `docs/docs/administration/debug-mode.md` for the full configuration reference and the
writable volume needed in containers.

## Safety

- **Off by default, near-zero overhead when off.** No datasource proxy on the query path, no
  parameter capture, no tracer, no extra files. Proven by a structural test (the datasource is the
  plain pool, not a proxy, when the flag is off).
- **Mandatory masking.** Secrets and personal data in SQL parameters and fields are masked before
  they reach the logs (key based + value based, both configurable). Tests assert known secret/PII
  patterns never appear.
- **No new attack surface.** No new port, no new endpoint.
- **Production barrier.** Disabled by default; in production (no `dev`/`test`/`ci` profile) it refuses
  to start unless `openaev.debug.allow-in-production=true` is also set, with an optional
  `auto-disable-after` timer and a loud repeated banner.
- **Deny-by-default masking option.** `masking.mask-all-parameters=true` masks every value, keeping
  only column name and type.
- **Bounded when on.** The SQL log is a rotated file (off the console). JFR recordings are bounded in
  size/age/duration and their periodic dumps are bounded by a retention pass (`jfr.max-dump-files`,
  `jfr.max-total-dump-size`). On a read-only filesystem JFR fails loudly and the SQL log falls back to
  the console; the app keeps running.
- **Single profiler.** When the Pyroscope agent is enabled, JFR does not start (hard-enforced).

## Changed after the review

- JFR dump retention added (the periodic dumps no longer accumulate without bound).
- Real production barrier (`DebugEnabledCondition` + `allow-in-production`) and `auto-disable-after`,
  replacing the previous advisory-only guardrail.
- Deny-by-default masking (`mask-all-parameters`).
- Bounded regex scan (8 KB) with mask-before-truncate, instead of truncate-before-regex (which would
  have leaked a prefix of a long secret).
- Single-profiler enforcement (JFR yields to Pyroscope).
- Doc/report accuracy fixes (the "files never grow unbounded" and "env-var as a security control"
  claims were corrected; the RabbitMQ/Quartz limit is now stated up front).

## Owners' decisions applied

- **Tracing stack (decided: OpenTelemetry) -- BLOCKED, kept Brave.** See the "Needs reviewer input"
  section above for the full explanation and the question for the metrics owners. In short: the
  project's metrics-only `OpenTelemetry` bean prevents Boot from wiring the tracer, so the OTel bridge
  produced no trace id (verified); the PR keeps Brave until the metrics bean is extended.
- **Profilers: keep both, by deployment.** JFR for on-prem, Pyroscope for cloud; the single-profiler
  lock (JFR yields to Pyroscope) stays. Documented.
- **Tenant tag compatible with both mechanisms.** New `DebugTenantSource` reads the v2 request
  selector (path `{tenantId}` / `X-Tenant-Ids` header) and falls back to v1 `TenantContext`. There is
  no ambient accessor for the resolved v2 `TxCtx` (it is argument-passed by design), so the resolver
  reads the same request inputs the v2 resolver uses, isolated in one class for a trivial future
  switch.
- **mask-all also masks statement-text literals.** In `mask-all-parameters`, single-quoted literals in
  the statement text are blanked too (bounded), so an inline secret in a native query does not leak.
- **Debug-on in the cloud: accepted, bounded risk.** No new barrier; operator guide documents the
  incident settings (`auto-disable-after`, masking, multi-tenant aggregation note). Formal ISO/SOC
  validation deferred to security.

## Dependencies added

- `io.micrometer:micrometer-tracing-bridge-brave` (Apache-2.0, version from the Spring Boot BOM)
- `net.ttddyy:datasource-proxy:1.10` (Apache-2.0, pinned)

JFR uses the JDK; no dependency added. Details in `docs/debug-mode-poc/DEPENDENCIES.md`.

## Notes for reviewers

- The project is on Spring Boot 3.5.15 / Java 21; all versions are aligned on that BOM.
- Operator reference (committed): `docs/docs/administration/debug-mode.md`.
- The ADR (p6spy vs datasource-proxy), the findings report (security coverage, container impact,
  RabbitMQ/Quartz async-propagation analysis) and the dependency/license list are kept as separate
  review artifacts and are not committed in this PR.

## Example output

One request, `GET /api/scenarios/sc-42`. Console keeps the app logs and the ORM summary; the
verbose per-statement SQL goes to the rotated file `openaev-debug-sql.log`.

Console — every line carries the request's `trace` and `tenant`:

INFO  [trace=6a3c4dea tenant=0e7c2f1a-tenant-acme] ...ScenarioApi : Loading scenario sc-42
INFO  [trace=6a3c4dea tenant=0e7c2f1a-tenant-acme] ...ScenarioApi : Scenario sc-42 loaded
WARN  [trace=6a3c4dea tenant=0e7c2f1a-tenant-acme] ...debug.orm   : ORM GET /api/scenarios/sc-42: 13 queries (2 distinct), 7ms
  N+1 SUSPECTED: 'select team_name from teams where team_id = ?' executed 12x (6ms total)

SQL file `openaev-debug-sql.log` — one line per statement, masked:

... INFO [trace=6a3c4dea... tenant=...acme] sql success=true time=1ms statement=insert into users (user_id, user_email, user_password) values (?, ?, ?) params=[{user_id=u-1, user_email=***MASKED***, user_password=***MASKED***}]
... INFO [trace=6a3c4dea... tenant=...acme] sql success=true time=5ms statement=select team_name from teams where team_id = ? params=[{team_id=team-0}]
... the same SELECT 11 more times, one per team -- the N+1 the summary flagged

Reading it:
- `trace=` correlation id shared by every line of the request (full 32-hex in the logs); filter on it.
- `tenant=` the tenant the request targets (default tenant when not tenant-scoped).
- `time=` JDBC execution time per statement.
- `statement=` the real SQL sent to PostgreSQL, with `?` placeholders.
- `params=[{column=value}]` bound params by column; `user_email` (PII) and `user_password` (sensitive
  column) are `***MASKED***`; `user_id`/`team_id` are shown.
- The `ORM ...` line: one summary per request, flagging the N+1 (same SELECT once per team).


openaev-api/pom.xml
```diff
@@ -286,6 +288,34 @@
...
+            <version>${datasource-proxy.version}</version>
+        </dependency>
+        <!-- Must match the version Spring Data JPA uses (declared provided/optional there),
+             otherwise its native-query handling breaks at startup. -->
+        <dependency>
+            <groupId>com.github.jsqlparser</groupId>
+            <artifactId>jsqlparser</artifactId>
+            <version>5.2</version>
+        </dependency>
+>>>>>>> Stashed changes
```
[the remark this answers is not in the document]
laugiov: fixed

docs/docs/administration/debug-mode.md
```diff
@@ -0,0 +1,271 @@
...
+on-prem installations. It produces a rich, correlated execution trace and stays fully disabled by
+default.
+
+When it is off, the installation behaves exactly as before: no datasource proxy on the query path,
+no parameter capture, no extra per-request work and no extra files written.
+
+## What it produces
+
+- **Correlation id and tenant in every log line.** Each request carries a `traceId` (and `spanId`)
+  and, for tenant-scoped requests, the `tenant` in the MDC, so all the lines emitted while handling it
```
antoinemzs: Define MDC?

docs/docs/administration/debug-mode.md
antoinemzs, line 1: Excellent doc thanks

openaev-api/src/main/java/io/openaev/debug/DebugProperties.java
```diff
@@ -0,0 +1,130 @@
...
+  private boolean allowInProduction = false;
+
+  /** Auto-disable the verbose tracing after this duration. {@code 0} = never. */
+  private Duration autoDisableAfter = Duration.ZERO;
+
+  /** How often the "debug mode is active" warning repeats. */
+  private Duration warningInterval = Duration.ofMinutes(5);
+
+  /** Writable directory for the JFR recordings and the rotated SQL log file. */
+  private String outputDir = "./logs/debug";
```
antoinemzs, line 35: Maybe worth liaising with platform if we want to make this persistent, or it risks being deleted every time we recycle the container.

Similarly, let's think about recommending making this persistent in the demo compose file (repo `docker`)

antoinemzs (approved):
Some nitpicks but also a logging format change that cannot be disabled (should only exist in debug=true mode)

Trace when everything off (large empty brackets)
```
2026-06-29T09:38:47.286+02:00  INFO 106352 --- [OpenAEV API] [  restartedMain] [                                                 ] io.openaev.App                           : Starting App using Java 21.0.11 with PID 106352 (/home/antoinemzs/filigran/openbas/openaev-api/target/classes started by antoinemzs in /home/antoinemzs/filigran/openbas)
```

When enabled, the brackets hold a correlation id
```
2026-06-29T09:47:21.539+02:00  INFO 116789 --- [OpenAEV API] [0.0-8080-exec-1] [6a42230910f449788694b1176553beca-09a186ba46edc482] io.openaev.debug.orm                     : ORM GET /api/xtm-composer/d9491550-1ffc-4070-8291-334f2f327919/connector-instances: 8 queries (8 distinct), 6ms
```

docs/docs/administration/debug-mode.md
laugiov: Mapped Diagnostic Context, i will define it in the documentation ;)

laugiov: Done. MDC is now spelled out on first use: "the MDC (Mapped Diagnostic Context, the per-thread key/value bag the logging library attaches to every line)".

openaev-api/src/main/java/io/openaev/debug/DebugProperties.java
laugiov: Good point. The output dir is configurable (openaev.debug.output-dir) but defaults to ./logs/debug, which sits on the container's ephemeral layer and is wiped on recycle. I added an explicit note in the docs: point it at a persistent volume so the SQL logs and JFR dumps survive a restart (the container example already uses a named volume). 
For the two follow-ups: I'll liaise with platform on whether to make it persistent by default.

PR #6396, opened 2026-06-26
chore(ci): improve ci speed and reliability (#7117)

### LINKED ISSUE #7117: ci: speed up pipeline via caching, image handoff, and test sharding
## Context

`chore/improve_ci` (#6396) rebuilds the Core/Nightly CI pipeline. Pipeline wall clock (checkout → required gate) dropped from **~26 min to ~9-12 min** on `pull_request` runs, with Nightly kept at full browser/variant coverage. This issue records the requirements the branch satisfies, so they can be checked off and referenced independently of the PR diff.

Baseline run: https://github.com/OpenAEV-Platform/openaev/actions/runs/30557476448
Latest run: https://github.com/OpenAEV-Platform/openaev/actions/runs/30629176862

### 1. Unify Core CI and Nightly CI into one reusable pipeline
- [x] Before this branch, `core-ci.yml` (341 lines) and `nightly-ci.yml` (700 lines) each defined their own copy of most jobs — build, package, Docker, E2E, the final gate — and only a small subset (spotless, frontend-quality, api-tests, api-types-check, codecov) was shared via a `_quality-gates.yml` reusable workflow. Any fix to, say, the Docker build had to be made (or drifted) in two places.
- [x] The entire pipeline — migrations guard, build, packaging, Docker build/merge, Snyk, all test stages, coverage, and the required gate — is now one `workflow_call` reusable workflow, `.github/workflows/_ci-pipeline.yml` (539 lines).
- [x] `core-ci.yml` and `nightly-ci.yml` shrank to 52 and 70 lines respectively: each just supplies its own `api-matrix` / `e2e-matrix` / `snyk-fail-on-error` inputs and calls the same pipeline. Nightly-only behavior (larger matrices, Snyk enforcement) is expressed entirely through those inputs, not duplicated jobs.
- [x] `_quality-gates.yml` and the old duplicated jobs are deleted; `.github/ci-pipeline-schema.md` documents the resulting job graph.
- Files: `.github/workflows/_ci-pipeline.yml` (new), `.github/workflows/core-ci.yml`, `.github/workflows/nightly-ci.yml`, `.github/workflows/_quality-gates.yml` (deleted), `.github/ci-pipeline-schema.md` (new)

### 2. Stop paying for a Docker layer cache that never hits
- [x] `cache-to: type=gha,mode=max` was costing ~6.4 min per build job (one export step alone took 384s) while producing **zero** `CACHED` layers on PRs, because GHA caches are branch-scoped and unreadable across PRs.
- [x] `cache-to` now only runs on `refs/heads/main`; PRs still read the warm cache via `cache-from` for free.
- Files: `.github/actions/docker-build/action.yml`

### 3. Stop emulating arm64 Docker builds with QEMU
- [x] The previous `docker-build-push` action built **both** `linux/amd64,linux/arm64` in a single buildx call on one `ubuntu-latest` runner, using QEMU to emulate arm64 — slow, and serial with the amd64 build.
- [x] Split into `docker-build` (single-platform, matrixed across 4 build combinations: standard/ubi9 × amd64/arm64) and a separate `docker-merge-platforms` step. arm64 variants now build natively on `ubuntu-24.04-arm` runners, in parallel with amd64, instead of emulated and serialized.
- Files: `.github/actions/docker-build-push/action.yml` (deleted, 285 lines), `.github/actions/docker-build/action.yml` (new), `.github/actions/docker-merge-platforms/action.yml` (new), `.github/workflows/_ci-pipeline.yml`

### 4. Add container vulnerability scanning
- [x] No Snyk (or equivalent) container scan existed anywhere in CI before this branch.
- [x] `snyk-docker` now scans the built image after every `docker-build` job. It is advisory (`continue-on-error`) in Core CI and enforcing in Nightly CI (`snyk-fail-on-error` input), and is intentionally **excluded** from `pipeline-gate`'s required checks so vulnerability findings never block a PR merge, while still failing Nightly outright.
- Files: `.github/actions/snyk-docker/action.yml` (new), `.github/workflows/_ci-pipeline.yml`

### 5. Stop moving Docker images through gzip/artifact round-trips between jobs
- [x] Images were built → gzip'd → uploaded as artifacts → polled → downloaded → gunzip'd → `skopeo copy` → `docker load` in every E2E job, and merged into a multi-arch index via a temporary local `registry:2` + `crane`.
- [x] Docker Build now pushes each platform image straight to `ghcr.io/<repo>-ci` alongside the existing OCI-archive artifact (two buildx exporters, one build).
- [x] E2E pulls directly from GHCR (`imagetools inspect` polling + `docker pull`) instead of the artifact round-trip.
- [x] `docker-merge-platforms` uses `docker buildx imagetools create` to assemble the multi-arch index server-side instead of downloading ~1.9 GB per job.
- [x] **Constraint respected:** images must still be inspectable/downloadable after the run — every job still uploads/keeps the OCI-archive artifact; the registry only changes how *jobs* consume the image mid-run.
- [x] Fork PRs (read-only `GITHUB_TOKEN`, no `packages:write`) automatically fall back to the original artifact path via a shared `.github/actions/ci-registry` resolver.
- [x] A best-effort `registry-cleanup` job deletes each run's GHCR tags so the package doesn't grow unbounded.
- Files: `.github/actions/ci-registry/action.yml`, `.github/actions/docker-build/action.yml`, `.github/actions/e2e-tests/action.yml`, `.github/actions/docker-merge-platforms/action.yml`, `.github/workflows/_ci-pipeline.yml`

### 6. Rebalance API test shards so no single shard dominates
- [x] The old catch-all shard did 41% of all API test time (15.3 min) vs 6-8 min for the other three shards.
- [x] Shards are now numbered (`1`..`6`) and bin-packed by **measured** `mvn test` cost (fitted from real run data — file count was a poor predictor: one 21-file shard ran more tests, faster, than a 63-file shard), instead of grouped by feature/package.
- [x] `remaining` (catch-all) is kept intentionally small — just the loose top-level test classes — so newly added packages land there by default and the deliberate buckets stay balanced.
- [x] Include/exclude patterns moved out of hand-maintained YAML env blocks into versioned files (`.github/shards/api-*.txt`); the catch-all's exclude list is *derived* from those files instead of duplicated, removing a drift hazard where a pattern could be added to a shard but forgotten in the exclude list.
- [x] Added `.github/scripts/verify-test-shards.py` (simulates Surefire Ant-pattern matching) and `.github/scripts/pack-api-shards.py` / `balance-api-shards.py` (regenerate the split from measured costs) so the shard assignment is reproducible and checked, not hand-edited.
- Files: `.github/shards/`, `.github/actions/api-tests/action.yml`, `.github/scripts/*.py`, `.github/workflows/_ci-pipeline.yml`, `core-ci.yml`, `nightly-ci.yml`

### 7. Shard E2E tests instead of running one Playwright project serially per browser
- [x] `playwright.config.ts` keeps `fullyParallel: false` (tests share one backend), so parallelism has to come from splitting *which spec files* each job runs, not from Playwright workers.
- [x] E2E jobs now take `includes`/`excludes` path filters, with a `catchall` shard that runs whatever isn't explicitly claimed (mirrors the API sharding model).
- [x] Handled two correctness traps found while implementing this: (a) positional Playwright filters would silently exclude `auth.setup.ts` and break `storageState` for every dependent project if not always included; (b) an empty catch-all would make Playwright run *everything* instead of nothing — guarded with an explicit early exit.
- Files: `.github/actions/e2e-tests/action.yml`, `.github/workflows/core-ci.yml`

### 8. Stop running the full ubi9 × webkit matrix on every PR
- [x] Core CI dropped all `ubi9` variants and all `webkit` runs from its E2E matrix (26 → 8 jobs); every dropped combination was verified to already exist in Nightly before removal, so release-image coverage is unchanged, just deferred to Nightly for PR-time feedback.
- Files: `.github/workflows/core-ci.yml`

### 9. Decouple the required status check from best-effort reporting
- [x] `codecov` (coverage upload, already `continue-on-error`) removed from `pipeline-gate`'s `needs`, so a slow/flaky coverage upload can no longer delay the branch-protection check.
- [x] `sanity-check` rewrote its artifact verification from downloading **all** run artifacts (9.3 GB, to `stat` six of them) to reading name/size from the Actions artifacts API. Image internals (JAR present, Java runtime, layer count) are already validated where the image is built; platform completeness is validated in the merge step.
- Files: `.github/workflows/_ci-pipeline.yml`, `.github/actions/sanity-check/action.yml`

### 10. Remove redundant compression and caching-key over-invalidation
- [x] `compression-level: 0` on artifacts that are already compressed (2 Docker OCI archives, 2 backend JARs, the Maven repo archive) — re-zipping a gzip/jar for ~0% size gain was pure CPU cost on both upload and download.
- [x] Playwright's browser cache key changed from `hashFiles(yarn.lock)` to the pinned `@playwright/test` version, so an unrelated dependency bump no longer forces a full browser re-download.
- Files: `.github/actions/docker-build/action.yml`, `docker-merge-platforms/action.yml`, `backend-package/action.yml`, `backend-package-musl/action.yml`, `backend-compile/action.yml`, `e2e-tests/action.yml`

### 11. Make job names describe what they test, not their matrix plumbing
- [x] E2E job names changed from `E2E (std-amd64 chrome-std-amd64-arsenals)` (variant repeated 3×, driven by `artifact_suffix`) to `E2E (arsenals · chrome · std-amd64)` — leads with the tested area. Nightly rows (no explicit shard) render as `E2E (full · <project> · <variant>)`.
- Files: `.github/workflows/_ci-pipeline.yml`, `core-ci.yml`

guillaumejparis:
…ainability

### Proposed changes

*
*

### Testing Instructions

1. Step-by-step how to test

### Related issues

* Closes #ISSUE-NUMBER
*

### Checklist

### Further comments


.github/versions.env
```diff
@@ -0,0 +1,11 @@
+# Shared container image versions for CI and local dev.
+# Sourced by CI composite actions and by openaev-dev/docker-compose.yml.
+# Renovate tracks these via regexManagers (see renovate.json5).
+
+POSTGRES_IMAGE=postgres:17-alpine
```
RomuDeuxfois: nitpick: should be good to handle this on docker compose also

.github/actions/backend-compile/action.yml
```diff
@@ -34,6 +34,8 @@ runs:
       with:
         name: backend-compiled
         path: ~/.m2/repository/io/openaev/
+        # JARs are already zips — re-zipping costs CPU for ~0% gain
+        compression-level: 0
```
RomuDeuxfois, line 38: todo: good call, this should be put everywhere ? in api-tests, backend-compiled, backend-package-musl

Dockerfile_ubi9
```diff
@@ -4,45 +4,50 @@ USER root
 
 WORKDIR /opt/openaev-build/openaev-front
```
RomuDeuxfois, line 5: todo: by playing with github action I suppose we can have just one docker file

renovate.json5
```diff
@@ -104,6 +104,24 @@
             ],
             rebaseWhen: 'auto',
         },
+        // Group container image updates (docker-compose.yml + openaev-dev/versions.env)
+        // into a single PR per image so CI and local dev stay in sync.
+        {
+            groupName: 'container images',
```
RomuDeuxfois: todo: if you are using versions.env in docker-compose.yml you don't need this part

renovate.json5
```diff
@@ -113,4 +131,25 @@
...
+    // Renovate tracks docker-compose.yml natively (the built-in docker-compose
+    // manager resolves `${VAR:-image:tag}` defaults). The regex manager below
+    // tracks versions.env. The groupName ensures both are updated in a SINGLE PR.
+    customManagers: [
+        {
+            customType: 'regex',
+            fileMatch: [
+                '^openaev-dev/versions\\.env$',
+            ],
+            matchStrings: [
```
RomuDeuxfois, line 127: todo: could be replace because it's seems to be error prone

matchStrings: [
      '\\w+_IMAGE=(?<depName>[^:]+):(?<currentValue>[^\\s]+)',
  ],

openaev-dev/docker-compose.yml
```diff
@@ -8,7 +11,7 @@ services:
   # Development PostgreSQL database (persistent storage)
   openaev-dev-pgsql:
     container_name: openaev-dev-pgsql
-    image: postgres:17-alpine
+    image: ${POSTGRES_IMAGE:-postgres:17-alpine}
```
RomuDeuxfois: todo: remove the default to be sure evrything is in one place, what do you think ?

.github/ci-pipeline-schema.md
```diff
@@ -0,0 +1,191 @@
+# CI Pipeline Schema
```
RomuDeuxfois, line 1: todo: should be improve to be human-readable

.github/workflows/test-feature-branch.yml
```diff
@@ -66,7 +66,7 @@ jobs:
         uses: docker/build-push-action@v7
         with:
           context: .
-          file: ./Dockerfile_ga
+          file: ./Dockerfile_ga_deprecated
```
RomuDeuxfois, line 69: todo: why not using the current dockrfile ?

.github/shards/api-1.txt
```diff
@@ -0,0 +1,14 @@
+io/openaev/aop/**/*Test.java
```
RomuDeuxfois, line 1: prais: so nice !

.github/actions/e2e-tests/action.yml
```diff
@@ -41,33 +58,47 @@ runs:
     - name: Start all services (non-blocking)
       shell: bash
       run: |
+        source openaev-dev/versions.env
```
RomuDeuxfois: todo: you can create a  github/actions/start-dev-services/action.yml  I suppos.

openaev-dev/docker-compose.yml
guillaumejparis: We can but it will force devs to make the change :)

.github/workflows/copilot-request-review.yml
```diff
@@ -14,7 +14,12 @@ jobs:
     name: "🤖 On-demand agent review"
     if: >-
       github.event.issue.pull_request &&
-      startsWith(github.event.comment.body, '/review')
+      startsWith(github.event.comment.body, '/review') &&
```
RomuDeuxfois, line 18: todo; can we ask copilot to use our reviewer agent ?

scripts/download-binaries.sh
```diff
@@ -23,16 +23,42 @@ RESOURCES="openaev-api/src/main/resources"
 AGENT_LOCAL="${RESOURCES}/agents/openaev-agent"
 IMPLANT_LOCAL="${RESOURCES}/implants/openaev-implant"
 
-CURL_OPTS=(-L --fail --retry 3 --retry-delay 5 --silent --show-error)
+# Word-split on use, so keep every option space-free.
+export CURL_OPTS="-L --fail --retry 3 --retry-delay 5 --silent --show-error"
```
RomuDeuxfois: qs: array was stronger than string, what do you think ?

Dockerfile_ubi9
guillaumejparis: Maybe but Dockerfile are not only for use in our CI but can be used by anyone

.github/workflows/test-feature-branch.yml
guillaumejparis: Can't do that, I want to preserve the action as is

scripts/download-binaries.sh
guillaumejparis: as i need to export the env var can't keep the array no

RomuDeuxfois (approved):

PR #6464, opened 2026-06-29
feat(db): optimize findForIndexing sql queries (#6535)

### LINKED ISSUE #6535: fix(simulation): Tab analysis empty
## Description

The "Analysis" tab in simulations is completely empty. No data or content is displayed when navigating to this tab, even when the simulation has been executed and results should be available.

## Environment

1. OS: MacOs
2. Version: 2.260623.0

## Reproducible steps

1. Navigate to a simulation that has been executed
2. Click on the "Analysis" tab
3. The tab content is blank — no data, charts, or analysis results are displayed

## Expected output

The "Analysis" tab should display the simulation analysis data .

## Actual output

The tab is completely empty with no content rendered.

## Additional information

Image

guillaumejparis:
### Proposed changes

*
*

### Testing Instructions

1. Step-by-step how to test

### Related issues

* Closes #6535 
*

### Checklist

### Further comments

### DIFF
--- a/openaev-api/src/test/java/io/openaev/engine/IndexingRegressionIntegrationTest.java
+++ b/openaev-api/src/test/java/io/openaev/engine/IndexingRegressionIntegrationTest.java
@@ -1,5 +1,8 @@
 package io.openaev.engine;
 
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.await;
+
 import io.openaev.IntegrationTest;
 import io.openaev.database.model.IndexingStatus;
 import io.openaev.database.raw.RawGrant;
@@ -18,11 +21,6 @@ import io.openaev.utils.fixtures.composers.*;
 import io.openaev.utils.mockUser.WithMockUser;
 import io.openaev.utils.pagination.Pagination;
 import io.openaev.utilstest.RabbitMQTestListener;
-import org.junit.jupiter.api.*;
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.test.context.TestExecutionListeners;
-import org.springframework.transaction.annotation.Transactional;
-
 import java.io.IOException;
 import java.time.Duration;
 import java.time.Instant;
@@ -30,9 +28,10 @@ import java.time.temporal.ChronoUnit;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
-
-import static org.assertj.core.api.Assertions.assertThat;
-import static org.awaitility.Awaitility.await;
+import org.junit.jupiter.api.*;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.test.context.TestExecutionListeners;
+import org.springframework.transaction.annotation.Transactional;
 
 /**
  * End-to-end regression tests for the {@code findForIndexing} query in {@link
--- a/openaev-api/src/test/java/io/openaev/engine/IndexingRegressionIntegrationTest.java
+++ b/openaev-api/src/test/java/io/openaev/engine/IndexingRegressionIntegrationTest.java
@@ -1,0 +1,943 @@
+package io.openaev.engine;
+
+import io.openaev.IntegrationTest;
+import io.openaev.database.model.IndexingStatus;
+import io.openaev.database.raw.RawGrant;
+import io.openaev.database.raw.RawUserAuth;
+import io.openaev.engine.api.ListConfiguration;
+import io.openaev.engine.api.ListRuntime;
+import io.openaev.engine.model.endpoint.EsEndpoint;
+import io.openaev.engine.model.inject.EsInject;
+import io.openaev.engine.model.injectexpectation.EsInjectExpectation;
+import io.openaev.engine.model.scenario.EsScenario;
+import io.openaev.engine.model.vulnerableendpoint.EsVulnerableEndpoint;
+import io.openaev.engine.query.EsEntities;
+import io.openaev.utils.CustomDashboardTimeRange;
+import io.openaev.utils.fixtures.*;
+import io.openaev.utils.fixtures.composers.*;
+import io.openaev.utils.mockUser.WithMockUser;
+import io.openaev.utils.pagination.Pagination;
+import io.openaev.utilstest.RabbitMQTestListener;
+import org.junit.jupiter.api.*;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.test.context.TestExecutionListeners;
+import org.springframework.transaction.annotation.Transactional;
+
+import java.io.IOException;
+import java.time.Duration;
+import java.time.Instant;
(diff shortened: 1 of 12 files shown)


guillaumejparis:
The sync job was not marked as completed, causing it to reprocess ~1 month of injects on each run.

Root cause: In ElasticService.bulkProcessing, models.stream().parallel() runs all 13 models concurrently, but saveAll only triggers after the slowest model finishes. Since the Inject query was scanning all 75K changed rows on every cycle, it took ~40 minutes per batch — but only advanced 500 rows. Status updates for all 12 other (fast) models were blocked waiting for it.

Catch-up at 500 rows/40min = ~100 hours to fully sync
Each cycle looked like it was hanging — because it technically was
Fix: Optimized the Inject query to process batches in ~3.6s instead of 40min → catch-up drops to ~9 minutes (152 batches × 3.6s). Pure Java query change, no DB migration required (migration only adds indexes for further improvement).
Other requests that use the same kind of structure have been fixed too.

Added WARN logs on the Elasticsearch/OpenSearch service to surface which table causes issues in future runs.

laugiov:
Nice optimization. Before merging, could you attach EXPLAIN ANALYZE before/after on realistic data? 
You're also adding 4 indexes on tables that are already well indexed. 
I'd like to confirm the planner actually uses them, otherwise it's just write overhead.

laugiov:
The key change here is moving LIMIT into the ranked CTEs. 
That's also the one thing that can break the indexing cursor: it advances one batch at a time using the last row's updated_at, so if the ranking key differs from that field, rows can get skipped between ticks. 
The current tests only cover a single batch. 
Could you add a test with more entities than the batch size, looping fetch while advancing the cursor, and asserting nothing is skipped or duplicated?

openaev-model/src/main/java/io/openaev/database/repository/InjectExpectationRepository.java
```diff
@@ -370,105 +370,82 @@ void insertSignature(
...
+      ie.asset_id,
+      ie.asset_group_id,
+      i.tenant_id,
+      i.inject_title as inject_title,
+      MAX(ins.tracking_sent_date) AS tracking_sent_date,
+      array_agg(DISTINCT ap.attack_pattern_id) FILTER ( WHERE ap.attack_pattern_id IS NOT NULL ) AS attack_pattern_ids,
+      array_agg(DISTINCT ic_d.domain_id) FILTER (WHERE ic_d.domain_id IS NOT NULL ) AS domain_ids,
+      MAX(se.scenario_id) AS scenario_id,
+      array_agg(DISTINCT c.collector_security_platform) FILTER ( WHERE c.collector_security_platform IS NOT NULL ) ||
+      array_agg(DISTINCT a.asset_id) FILTER ( WHERE a.asset_id IS NOT NULL ) AS security_platform_ids
```
laugiov: ⚠️ Blocking. This query no longer rolls up the security platforms detected by agent-level (child) expectations into the agentless parent. 
That roll-up was added by #6497 ("inject issue undetected by security platform", merged Jul 3) via the agent_security_platforms CTE. 
This rewrite was authored Jun 30, before #6497, so it looks like the roll-up was dropped during a rebase onto main. 
Effect: a parent expectation's security_platforms_side is empty when detections live on agent children, the #6497 bug comes back. 
Could you re-apply the roll-up on top of the new query, or confirm + document if the removal is intentional?

openaev-api/src/test/java/io/openaev/engine/IndexingRegressionTest.java
```diff
@@ -310,23 +621,9 @@ void expectation_reindexed_when_linked_inject_updated() {
...
-      entityManager.persist(agent);
-      entityManager.flush();
-      Agent persistedAgent = entityManager.getReference(Agent.class, agent.getId());
-
-      DetectionInjectExpectation agentExpectation =
+      DetectionInjectExpectation agentlessExpectation =
           InjectExpectationFixture.createDefaultDetectionInjectExpectation();
-      agentExpectation.setAgent(persistedAgent);
-      agentExpectation.setResults(
+      agentlessExpectation.setResults(
```
laugiov: This test still claims "from agent-level expectation results", but it now puts the collector result directly on the agentless parent and removes the agent expectation, so it passes even though the agent→parent roll-up is gone (see my comment on InjectExpectationRepository). 
Please restore the original scenario: result on an agent child, asserted on the parent, so the test actually guards the behavior.

openaev-model/src/main/java/io/openaev/database/repository/InjectExpectationRepository.java
```diff
@@ -370,105 +370,82 @@ void insertSignature(
...
-    JOIN changed_expectations ce ON ie.inject_expectation_id = ce.inject_expectation_id
+    JOIN ranked_expectations re ON ie.inject_expectation_id = re.inject_expectation_id
+    LEFT JOIN exercises ex ON ex.exercise_id = ie.exercise_id
     LEFT JOIN injects i ON i.inject_id = ie.inject_id
     LEFT JOIN injects_statuses ins ON ins.status_inject = i.inject_id
     LEFT JOIN injectors_contracts ic ON ic.injector_contract_id = i.inject_injector_contract
     LEFT JOIN injectors_contracts_attack_patterns ic_ap ON ic_ap.injector_contract_id = ic.injector_contract_id
     LEFT JOIN attack_patterns ap ON ap.attack_pattern_id = ic_ap.attack_pattern_id
     LEFT JOIN injectors_contracts_domains ic_d ON ic_d.injector_contract_id = ic.injector_contract_id
+    LEFT JOIN users u ON u.user_id = ie.user_id
```
laugiov: These joins (users u, teams t, assets asset, asset_groups ag, exercises ex) aren't referenced in the SELECT/GROUP BY, dead joins. 
Safe to drop.

openaev-api/src/test/java/io/openaev/engine/IndexingRegressionIntegrationTest.java
```diff
@@ -0,0 +1,942 @@
...
+      // -- ASSERT --
+      awaitEndpointIndexedAssertion(
+          () -> {
+            EsEntities result = queryModel("inject");
+            assertThat(result.getTotal()).isGreaterThan(0);
+            assertThat(result.getEsDatas())
+                .anySatisfy(
+                    e ->
+                        assertThat(((EsInject) e).getInject_title())
+                            .as("inject_title doit etre present dans les documents indexes")
```
laugiov: Translate in english please ;)

laugiov (commented):
Thanks ! solid optimization.

I re-checked the query rewrites; they're semantically equivalent except one thing that needs to be resolved before merge.

Requesting changes, 2 blockers:

The agent→parent security-platform roll-up from #6497 is gone (looks lost in a rebase), and the test that guarded it no longer exercises it. Details inline.

laugiov (approved):

Issue #6471, opened 2026-06-29
fix(backend): ExecutorApi.getExecutor @AccessControl uses wrong resourceId (#collectorId instead of #executorId)


Seb-MIGUEL:
### Description

`ExecutorApi.getExecutor()` has a copy-paste error in its `@AccessControl` annotation: `resourceId = "#collectorId"` should be `resourceId = "#executorId"`.

The method parameter is `executorId`, but the SpEL expression references `#collectorId` which doesn't exist. This means the resource-level RBAC check is performed with a null resource ID, potentially bypassing resource-level authorization for this endpoint.

### Location

`openaev-api/src/main/java/io/openaev/rest/executor/ExecutorApi.java:101-104`

```java
@AccessControl(
    resourceId = "#collectorId",  // BUG: should be "#executorId"
    actionPerformed = Action.READ,
    resourceType = ResourceType.ASSET)
public Executor getExecutor(@PathVariable String executorId) {
```

### Fix

```java
@AccessControl(
    resourceId = "#executorId",
    actionPerformed = Action.READ,
    resourceType = ResourceType.ASSET)
```

### Severity

High — RBAC misconfiguration on a resource-level access check.

### Discovered during

Security review of PR for #6469.

PR #6472, opened 2026-06-29
fix(connectors): prevent JPA entity name mutation in connector read path (#6469)

### LINKED ISSUE #6469: fix(backend): integrations cards dynamically swap positions due to JPA entity name mutation in read path
### Description

Integration cards (collectors, injectors, executors) dynamically change positions on the integrations page after a few seconds, without any user interaction. For example, on `/admin/integrations/collectors`, the **Microsoft Defender** collector swaps position with **MITRE ATT&CK** after the page loads.

### Steps to Reproduce

1. Navigate to `/admin/integrations/collectors`
2. Wait 3-5 seconds
3. Observe collector cards swap positions (e.g., Microsoft Defender ↔ MITRE ATT&CK)

**Environment**: https://filigran.oaev.filigran.io (observed on collectors page, potentially also affects injectors/executors)

### Root Cause Analysis

The bug is caused by a **JPA entity name mutation side-effect in the GET endpoint**, combined with collector heartbeat registration that resets the name, creating a ping-pong cycle visible to the frontend via SSE.

#### The Cycle

**Step 1 — Page load (GET /api/collectors):**
- `AbstractConnectorService.toExistingConnectorOutput()` (line 74-76) calls `connector::setName()` on the **managed Hibernate entity** to override the name with the instance-config name (`COLLECTOR_NAME` key)
- Since the controller method is `@Transactional` (`CollectorApi.java` line 56), Hibernate dirty-checks the entity at flush → `UPDATE` SQL is executed
- `@PostUpdate` in `ModelBaseListener` fires → SSE event `DATA_UPDATE_SUCCESS` is dispatched with the **overridden name**
- Frontend sorts by `name` → cards rendered in position A

**Step 2 — Collector heartbeat (POST /api/collectors, a few seconds later):**
- `CollectorService.register()` (line 217) resets `collector.setName(input.getName())` to the collector's **self-reported registration name** (which may differ from the instance-config name)
- Transaction commits → Hibernate dirty-checks → name changed back → SSE event with **original name**
- Frontend receives SSE update → `useDataLoader.js` dispatches to Redux → `Referential.ts` merges the update → `normalizeSingle` maps `collector_name` → `name`
- Sort runs again with different names → **cards swap to position B**

**Step 3 — Next GET request or page reload repeats the cycle.**

#### Why Specific Cards Swap

Both "Microsoft Defender" and "MITRE ATT&CK" start with "Mi", making them adjacent in alphabetical sort. If the instance-config name differs even slightly from the registration name (e.g., "Microsoft Defender" vs "Microsoft Defender for Endpoint"), the relative sort order toggles. The sort has **no tiebreaker** (only `name`), making the swap deterministic.

### Affected Files

| File | Lines | Role |
|------|-------|------|
| `openaev-api/.../AbstractConnectorService.java` | 74-76 | **ROOT CAUSE**: mutates managed JPA entity name in read path |
| `openaev-api/.../AbstractConnectorService.java` | 105-112 | `getConfiguredConnectorName()` reads from instance config |
| `openaev-api/.../CollectorService.java` | 217 | `register()` resets name → completes the ping-pong |
| `openaev-api/.../CollectorApi.java` | 56 | `@Transactional` on GET enables dirty-check flush |
| `openaev-model/.../ModelBaseListener.java` | 157-160 | `@PostUpdate` → creates SSE event |
| `openaev-front/.../SortingFiltering.jsx` | 66-70 | Sort with single comparator (no tiebreaker) |

### Proposed Fix

**Primary fix (backend)** — Don't mutate managed entities in the read path. Set the display name on the **DTO only**:
```java
// AbstractConnectorService.java — instead of connector.setName(...)
String displayName = (instance instanceof ConnectorInstancePersisted pi)
    ? getConfiguredConnectorName(pi).orElse(connector.getName())
    : connector.getName();
// Pass displayName to the DTO/output mapper
```

**Secondary fix (frontend)** — Add a sort tiebreaker to prevent position instability even when names are temporarily equal:
```javascript
// SortingFiltering.jsx line 66-70
R.sortWith([R.ascend(R.propOr('', order.sortBy)), R.ascend(R.propOr('', 'id'))])
```

### Scope

- [x] Collectors page (confirmed)
- [ ] Injectors page (to verify — same `AbstractConnectorService` code path)
- [ ] Executors page (to verify — same `AbstractConnectorService` code path)

Seb-MIGUEL:
## Summary

Fixes #6469 — Integration cards (collectors, injectors, executors) dynamically swap positions on the integrations page after a few seconds.

## Root Cause

`AbstractConnectorService.toExistingConnectorOutput()` called `connector.setName()` on a **managed Hibernate entity** to override the display name from instance config. Since the GET endpoints used `@Transactional` (read-write), Hibernate dirty-checked the entity at flush, triggering:

1. `UPDATE` SQL on the connector row
2. `@PostUpdate` → SSE event with the overridden name
3. Collector heartbeat (`POST /api/collectors`) resets name to registration value
4. Another SSE event with the original name
5. Frontend re-sorts by name → cards swap positions

## Fix

**Root cause fix**: Resolve the display name as a local variable and pass it through `mapToOutput(... , String displayName)` instead of mutating the managed entity. The `displayName` parameter flows through the service subclasses to the mappers, which use it instead of `connector.getName()`.

**Defense-in-depth**: Added `@Transactional(readOnly = true)` on the GET list endpoints for collectors, injectors, and executors, following backend conventions for read methods.

## Scope

Affects all 3 connector types equally (shared `AbstractConnectorService` base class):
- ✅ Collectors
- ✅ Injectors
- ✅ Executors

## Files Changed (10 files, +32/-20)

| File | Change |
|------|--------|
| `AbstractConnectorService.java` | Resolve displayName without entity mutation, pass through `mapToOutput` |
| `CollectorService.java` | Accept `displayName` in `mapToOutput` override |
| `InjectorService.java` | Accept `displayName` in `mapToOutput` override |
| `ExecutorService.java` | Accept `displayName` in `mapToOutput` override |
| `CollectorMapper.java` | Use `displayName` instead of `collector.getName()` |
| `InjectorMapper.java` | Use `displayName` instead of `injector.getName()` |
| `ExecutorMapper.java` | Use `displayName` instead of `executor.getName()` |
| `CollectorApi.java` | `@Transactional(readOnly = true)` on GET list |
| `InjectorApi.java` | `@Transactional(readOnly = true)` on GET list |
| `ExecutorApi.java` | `@Transactional(readOnly = true)` on GET list |

## Review Notes

- Code review ✅ — approved, no blocking issues
- Security review ✅ — changes are security-safe and improve security posture
- Pre-existing RBAC bug discovered during review → tracked in #6471

### DIFF
--- a/openaev-api/src/main/java/io/openaev/service/connectors/AbstractConnectorService.java
+++ b/openaev-api/src/main/java/io/openaev/service/connectors/AbstractConnectorService.java
@@ -38,8 +38,7 @@ public abstract class AbstractConnectorService<T extends BaseConnectorEntity, Ou
       T connector,
       CatalogConnector catalogConnector,
       ConnectorInstance instance,
-      boolean existingConnector,
-      String displayName);
+      boolean existingConnector);
 
   protected abstract T createNewConnector();
 
@@ -72,18 +71,15 @@ public abstract class AbstractConnectorService<T extends BaseConnectorEntity, Ou
         isVerified && instance instanceof ConnectorInstancePersisted
             ? ((ConnectorInstancePersisted) instance).getCatalogConnector()
             : catalogConnectorService.findBySlug(connector.getType()).orElse(null);
-    // Resolve display name from instance config without mutating the managed entity
-    String displayName = connector.getName();
     if (instance instanceof ConnectorInstancePersisted persistedInstance) {
-      displayName = getConfiguredConnectorName(persistedInstance).orElse(displayName);
+      getConfiguredConnectorName(persistedInstance).ifPresent(connector::setName);
     }
-    return mapToOutput(connector, catalogConnector, instance, true, displayName);
+    return mapToOutput(connector, catalogConnector, instance, true);
   }
 
   private T createExternalConnector(String collectorId, ConnectorInstancePersisted instance) {
     T newConnector = createNewConnector();
     newConnector.setId(collectorId);
-    // Safe: newConnector is transient (not managed by Hibernate), no dirty-check risk
     newConnector.setName(resolveConnectorName(instance));
     newConnector.setExternal(true);
     newConnector.setType(instance.getCatalogConnector().getSlug());
@@ -160,8 +156,7 @@ public abstract class AbstractConnectorService<T extends BaseConnectorEntity, Ou
                         newConnector,
                         entry.getValue().getCatalogConnector(),
                         entry.getValue(),
-                        false,
-                        newConnector.getName()));
+                        false));
               });
     }
 
--- a/openaev-api/src/main/java/io/openaev/service/connectors/AbstractConnectorService.java
+++ b/openaev-api/src/main/java/io/openaev/service/connectors/AbstractConnectorService.java
@@ -38,7 +38,8 @@ public abstract class AbstractConnectorService<T extends BaseConnectorEntity, Ou
       T connector,
       CatalogConnector catalogConnector,
       ConnectorInstance instance,
-      boolean existingConnector);
+      boolean existingConnector,
+      String displayName);
 
   protected abstract T createNewConnector();
 
@@ -71,16 +72,18 @@ public abstract class AbstractConnectorService<T extends BaseConnectorEntity, Ou
         isVerified && instance instanceof ConnectorInstancePersisted
             ? ((ConnectorInstancePersisted) instance).getCatalogConnector()
             : catalogConnectorService.findBySlug(connector.getType()).orElse(null);
-    // Override connector name with instance-specific name (e.g. EXECUTOR_NAME) if present
+    // Resolve display name from instance config without mutating the managed entity
+    String displayName = connector.getName();
     if (instance instanceof ConnectorInstancePersisted persistedInstance) {
-      getConfiguredConnectorName(persistedInstance).ifPresent(connector::setName);
+      displayName = getConfiguredConnectorName(persistedInstance).orElse(displayName);
     }
-    return mapToOutput(connector, catalogConnector, instance, true);
+    return mapToOutput(connector, catalogConnector, instance, true, displayName);
   }
 
   private T createExternalConnector(String collectorId, ConnectorInstancePersisted instance) {
     T newConnector = createNewConnector();
     newConnector.setId(collectorId);
+    // Safe: newConnector is transient (not managed by Hibernate), no dirty-check risk
     newConnector.setName(resolveConnectorName(instance));
     newConnector.setExternal(true);
     newConnector.setType(instance.getCatalogConnector().getSlug());
@@ -157,7 +160,8 @@ public abstract class AbstractConnectorService<T extends BaseConnectorEntity, Ou
                         newConnector,
                         entry.getValue().getCatalogConnector(),
                         entry.getValue(),
-                        false));
+                        false,
+                        newConnector.getName()));
               });
     }
 
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -138,6 +138,26 @@ openaev-front/
 
 **Types**: `feat`, `fix`, `docs`, `style`, `refactor`, `test`, `chore`
 
+## PR Title Format
+
+PR titles follow the same Conventional Commits format as commit messages:
+
+```
+<type>(<scope>)?: <short description> (#<issue-number>)?
+```
+
+> ⚠️ The `openaev-pr-checks` GitHub App validates PR titles against this pattern. Titles with extra prefixes
+> (e.g. `[backend]`) will be rejected.
+
+## Pre-PR Checklist
+
+Before creating a pull request, validate locally:
+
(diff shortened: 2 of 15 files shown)


openaev-api/src/main/java/io/openaev/service/connectors/AbstractConnectorService.java
```diff
@@ -71,16 +72,18 @@ private Output toExistingConnectorOutput(
...
     }
-    return mapToOutput(connector, catalogConnector, instance, true);
+    return mapToOutput(connector, catalogConnector, instance, true, displayName);
   }
 
   private T createExternalConnector(String collectorId, ConnectorInstancePersisted instance) {
     T newConnector = createNewConnector();
     newConnector.setId(collectorId);
+    // Safe: newConnector is transient (not managed by Hibernate), no dirty-check risk
     newConnector.setName(resolveConnectorName(instance));
```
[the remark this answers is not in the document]
Seb-MIGUEL: Acknowledged — collectorId is pre-existing naming from when this class was collector-only. Renaming to connectorId is a good cleanup but out of scope for this bug fix PR (minimal changes principle). Will track as follow-up refactoring.

openaev-api/src/main/java/io/openaev/service/connectors/AbstractConnectorService.java
```diff
@@ -71,16 +72,18 @@ private Output toExistingConnectorOutput(
...
             : catalogConnectorService.findBySlug(connector.getType()).orElse(null);
-    // Override connector name with instance-specific name (e.g. EXECUTOR_NAME) if present
+    // Resolve display name from instance config without mutating the managed entity
+    String displayName = connector.getName();
     if (instance instanceof ConnectorInstancePersisted persistedInstance) {
-      getConfiguredConnectorName(persistedInstance).ifPresent(connector::setName);
+      displayName = getConfiguredConnectorName(persistedInstance).orElse(displayName);
     }
-    return mapToOutput(connector, catalogConnector, instance, true);
+    return mapToOutput(connector, catalogConnector, instance, true, displayName);
```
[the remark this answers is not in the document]
Seb-MIGUEL: The existing CollectorApiTest, InjectorApiTest, and ExecutorApiTest already exercise these GET list endpoints via MockMvc. The fix is purely mechanical (passing a parameter instead of mutating the entity) — the functional behavior is unchanged. Adding a specific test for 'entity name unchanged after GET' would require setting up connector instances with COLLECTOR_NAME config keys, which is disproportionate to the fix size. The @Transactional(readOnly = true) annotation provides a compile-time-visible guarantee against dirty flushes.

openaev-api/src/main/java/io/openaev/service/connectors/AbstractConnectorService.java
```diff
@@ -71,16 +72,18 @@ private Output toExistingConnectorOutput(
...
   }
 
   private T createExternalConnector(String collectorId, ConnectorInstancePersisted instance) {
     T newConnector = createNewConnector();
     newConnector.setId(collectorId);
+    // Safe: newConnector is transient (not managed by Hibernate), no dirty-check risk
```
RomuDeuxfois: **question**: what's the goal of this comment ?

openaev-api/src/main/java/io/openaev/service/connectors/AbstractConnectorService.java
```diff
@@ -71,16 +72,18 @@ private Output toExistingConnectorOutput(
         isVerified && instance instanceof ConnectorInstancePersisted
             ? ((ConnectorInstancePersisted) instance).getCatalogConnector()
             : catalogConnectorService.findBySlug(connector.getType()).orElse(null);
-    // Override connector name with instance-specific name (e.g. EXECUTOR_NAME) if present
+    // Resolve display name from instance config without mutating the managed entity
+    String displayName = connector.getName();
     if (instance instanceof ConnectorInstancePersisted persistedInstance) {
-      getConfiguredConnectorName(persistedInstance).ifPresent(connector::setName);
+      displayName = getConfiguredConnectorName(persistedInstance).orElse(displayName);
```
RomuDeuxfois: **todo**: Can we simplify things here ? We have some mapper with a lot of property and we are adding one more..

openaev-api/src/main/java/io/openaev/service/connectors/AbstractConnectorService.java
Seb-MIGUEL: Done — removed the comment. The transient nature is implicit from createNewConnector() returning a plain new instance; no annotation needed.

openaev-api/src/main/java/io/openaev/service/connectors/AbstractConnectorService.java
Seb-MIGUEL: Done — simplified: instead of threading displayName as an extra parameter through mapToOutput and all mappers, we now call connector.setName(displayName) directly before delegating to mapToOutput. Safe because all GET list endpoints are @Transactional(readOnly = true), so Hibernate will not flush the change. Mappers are back to using connector.getName() with no extra parameter.

RomuDeuxfois (approved):

PR #6475, opened 2026-06-29
fix(integrations): add fallback for missing catalog connector icons (#6474)

Seb-MIGUEL:
## Summary

Fixes missing connector logos on the catalog page. Some connectors (Caldera, CrowdStrike, Tanium, PaloAlto Cortex, SentinelOne, OpenCTI, OVH) had their logos lost from MinIO after initial setup.

## Root Cause

\IntegrationFactory.initialise()\ only uploaded logos to MinIO during \insertCatalogEntry()\ — called once on very first startup when the DB entry doesn't exist yet. If logos were later lost from MinIO (e.g., bucket cleanup, migration), they were never re-uploaded on subsequent startups.

## Fix

**Backend-only fix** (no frontend changes):

- Added \nsureCatalogLogo()\ hook in \IntegrationFactory\ base class, called in the \lse\ branch of \initialise()\ when the catalog entry already exists
- Each factory overrides \nsureCatalogLogo()\ to re-upload its logo to MinIO (idempotent \putObject\)
- \insertCatalogEntry()\ delegates to \nsureCatalogLogo()\ internally to avoid duplication
- Extracted \getLogoFilename()\ private method in each factory to centralize the logo filename string
- Made \nsureCatalogLogo()\ **best-effort** (try-catch with \log.warn\) so MinIO outages don't block application startup

## Files Changed

| File | Change |
|------|--------|
| \IntegrationFactory.java\ | Added \nsureCatalogLogo()\ hook + best-effort try-catch in \initialise()\ |
| 5 executor factories | Override \nsureCatalogLogo()\ + extract \getLogoFilename()\ |
| 2 injector factories | Override \nsureCatalogLogo()\ + extract \getLogoFilename()\ |
| 2 test factories | Override \nsureCatalogLogo()\ + extract \getLogoFilename()\ |

Closes #6474

PR #6491, opened 2026-06-30
fix(engine): fix ES indexing cursor skipping expectations with same timestamp (#6490)

### LINKED ISSUE #6490: fix(dashboard): inconsistency between dashboard detection % and inject detail page
## Problem

On the custom dashboard, some injects show **100% detected** in the widget. Clicking that percentage opens a drawer showing expectations with `inject_expectation_status = SUCCESS` from Elasticsearch. Clicking one of those expectations navigates to the inject detail page, which shows **100% not detected** in the header.

This inconsistency occurs because Elasticsearch becomes permanently out-of-sync with the database.

## Root Cause

The ES indexing job polls for updated expectations using:

```
WHERE inject_expectation_updated_at > :from
ORDER BY inject_expectation_updated_at ASC
LIMIT :limit  -- default 500
```

When many expectations are saved in the same batch (e.g. bulkComputeTechnicalExpectations during expiration), Hibernate's @UpdateTimestamp sets the **same millisecond timestamp** for all of them.

If the batch contains more than 500 expectations with the same timestamp:
1. The first 500 are indexed, lastIndexing = T
2. Next cycle: WHERE updated_at > T - all remaining items also have updated_at = T, so they are **permanently skipped**

The dashboard drawer (ES) still shows the old SUCCESS status, while the inject detail header reads from the database and shows the correct status.

## Fix

Implement compound keyset pagination on (updated_at, inject_expectation_id):
- New findForIndexingAfter(from, lastId, limit) query with compound cursor
- New column indexing_last_id in indexing_status table (V5_35 migration)
- IndexingStatus entity gets a lastId field
- Handler interface gets a new default method fetch(from, lastId, limit)
- InjectExpectationHandler overrides it to use the compound cursor
- Both OpenSearchService and ElasticService track lastId per cycle

When a full batch is returned, lastId is saved so the next cycle continues right after the last processed item. When a partial batch is returned, lastId is cleared.

Seb-MIGUEL:
## Problem

The custom dashboard shows **100% detected** for some injects. Clicking opens a drawer with `inject_expectation_status = SUCCESS` from Elasticsearch. Clicking an inject navigates to the inject detail page which shows **100% not detected** — inconsistent data between ES and DB.

Fixes #6490

## Root Cause

`findForIndexing` uses a strict `>` comparison on `updated_at`:
```sql
WHERE inject_expectation_updated_at > :from
ORDER BY inject_expectation_updated_at ASC
LIMIT 500
```

When `bulkComputeTechnicalExpectations` expires hundreds of expectations in one `saveAll()` call, Hibernate's `@UpdateTimestamp` assigns the **same millisecond** to all of them. If there are >500 with the same timestamp, the indexing cursor advances to `T`, then `updated_at > T` permanently skips the rest.

## Fix: Compound keyset cursor

Add a secondary `inject_expectation_id` cursor so the query can resume within a timestamp:

```sql
WHERE (inject_expectation_updated_at > :from)
   OR (inject_expectation_updated_at = :from AND inject_expectation_id > :lastId)
ORDER BY inject_expectation_updated_at ASC, inject_expectation_id ASC
```

### Changes

| File | Change |
|---|---|
| `V5_35__Add_indexing_last_id.java` | Migration: add `indexing_last_id TEXT NULL` to `indexing_status` |
| `IndexingStatus.java` | Add `lastId` field for compound cursor |
| `InjectExpectationRepository.java` | Add `findForIndexingAfter(from, lastId, limit)` query |
| `Handler.java` | Add default `fetch(from, lastId, limit)` method (backward-compatible) |
| `InjectExpectationHandler.java` | Override compound-cursor fetch + extract shared mapping helper |
| `OpenSearchService.java` | Track `lastId` in `bulkProcessing`, pass to handler |
| `ElasticService.java` | Same as OpenSearch |

### Cursor logic

- Full batch returned (size == 500): save `lastId = lastItem.id`, keep same `lastIndexing`
- Partial batch (size < 500): clear `lastId = null`, advance `lastIndexing` as before

### DIFF
--- a/.github/skills/review-code/SKILL.md
+++ b/.github/skills/review-code/SKILL.md
@@ -98,7 +98,17 @@ grep -rn "TenantBase\|tenant_id\|TenantContext" --include="*.java" $(git diff --
 git diff --name-only HEAD~1 | grep -E "\.tsx$|\.ts$" | head -10
 ```
 
-## Step 8 — Compile review
+## Step 8 — Check common anti-patterns (learned from reviews)
+
+Apply these checks based on past review feedback:
+
+- **Root cause vs workaround**: If a bug is backend-originated, don't add frontend workarounds (onError handlers, fallback states). Fix the root cause in the correct layer.
+- **String/value duplication**: When introducing new methods that share data with existing methods (e.g., logo filenames, config keys), extract the shared value to a private method or constant — never compute the same formatted string in multiple places.
+- **Non-critical operations**: Startup operations that interact with external services (MinIO, S3, etc.) for non-critical assets (logos, thumbnails) should be best-effort: wrap in try-catch with `log.warn` so failures don't block application startup.
+- **Test proportionality**: Don't require tests for small mechanical changes (1-line additions, parameter threading). Tests should be proportionate to the risk and complexity of the change.
+- **Pre-existing issues**: Don't fix unrelated pre-existing issues (e.g., alt text, parameter naming) in a bug fix PR. Track them as follow-up.
+
+## Step 9 — Compile review
 
 Generate the Code Review Summary following the output format
 defined in `code-reviewer.agent.md`.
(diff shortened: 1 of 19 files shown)


Seb-MIGUEL:
Closing: la branche incluait les commits de PR #6475 (InjectorFactory). Remplacé par une PR propre depuis main.

PR #6499, opened 2026-06-30
fix(chaining): add WorkflowUpdateEvent for legacy feature flag (#4824)

EvaE-Filigran:
### Proposed changes

* Add `@WorkflowUpdateEvent` annotation on the legacy REST API callback path (`InjectExecutionService.handleInjectExecutionCallback`) so that the inject chaining pipeline is notified when execution traces are processed via the legacy route

### Testing Instructions

1. Enable `LEGACY_INGESTION_EXECUTION_TRACE` feature flag (or use `main-ff` where all flags are active)
2. Create a scenario with chained injects (inject A → condition → inject B)
3. Execute inject A
4. Verify that step receives the structured output from inject A
5. Verify that the condition evaluates and inject B is triggered

### Checklist

### Further comments

On `main-ff`, all feature flags are enabled (`FEATURE_FLAG_ALL = "*"`), including `LEGACY_INGESTION_EXECUTION_TRACE`. This forces execution traces to always go through the legacy REST API callback instead of the queue path. The callback was missing `@WorkflowUpdateEvent`, so the chaining system was never notified of inject completion — steps never received outputs and conditions were never evaluated.


openaev-api/src/main/java/io/openaev/rest/inject/service/InjectExecutionService.java
```diff
@@ -40,6 +41,7 @@ public class InjectExecutionService {
   @Resource protected ObjectMapper mapper;
 
   @Transactional
+  @WorkflowUpdateEvent(injectId = "#injectId")
```
savacano28, line 44: We currently have this annotation on the controller side in InjectApi.injectExecutionCallback. Being in InjectApi, shouldn't it cover both cases of trace processing? @Dimfacion 

Maybe you could add some explanation about why we have this annotation in both services: batchingInjectStatusService and InjectExecutionService? 

Thanks!

savacano28 (approved):

openaev-api/src/main/java/io/openaev/rest/inject/service/InjectExecutionService.java
Dimfacion: Sorry, I don't really have the context but from what I can understand, if you put it in InjectApi, you will create an event about a trace that has not been processed since if the batching mode is activated, InjectApi will only publish an event in rabbitMQ. The processing of that event might happen several minutes later ...

impolitepanda (approved):

savacano28:
Tests for FF in chaining: OK
* All : openaev.enabled-dev-features=* => OK
image
image
image

* Only chaining: openaev.enabled-dev-features=INJECT_CHAINING => OK
image
image
image
image

* No chaining:
Atomic: OK
image
Simualtion time based:
image

impolitepanda (approved):
Re approved after fixes from tests

EvaE-Filigran (commented):
I propose not throwing an exception if the step is not found, but instead returning an Optional<Step> or Optional<String> for the step ID. This code will be called for every inject trace update regardless of whether chaining is involved. If a step has been deleted, no further processing can be performed anyway. findStepIdByInjectId only exists to find a step for a specific inject in order to update it.

PR #6506, opened 2026-06-30
feat(signature-refactor): refactor inject expectations signatures process (#5151)

### LINKED ISSUE #5151: [Expectation/Result] Manual result set on Asset propagates to Agents but does not trigger Asset-level score recomputation
## Description

When a user manually sets an expectation result on an **Asset** (e.g., adding a HarfangLab detection result), the result is correctly propagated **downward** to all Agents of that Asset. However, the **Asset-level score itself is never recomputed**, and consequently the **Asset-Group level score** is also never updated.

The downward propagation works, but the upward re-aggregation (Agent → Asset → Asset-Group) is not triggered, leaving the hierarchy in an inconsistent state.

## Environment

1. OS (where OpenAEV server runs): All
2. OpenAEV version: Current
3. OpenAEV client: Backend — Expectation result manual update + aggregation engine
4. Other environment details: Affects all inject types when results are set manually on an Asset. Related to aggregation rules (MAX at Agent level, worst-case at Asset/Group level).

## Reproducible Steps

1. Create a technical inject with a **Prevention** expectation
2. Target the inject at an **Asset-Group** (AG1) containing:
   - Asset A (with Agent 1 and Agent 2)
3. Run the inject
4. Verify initial state — all scores are `null`:
Asset-Group AG1: score = null
Asset A: score = null
Agent 1: score = null
Agent 2: score = null
5. Manually set a result **"Prevented" (score=100)** on **Asset A**
6. Check Agent-level scores:
Agent 1: score = 100 ✅ (propagated correctly)
Agent 2: score = 100 ✅ (propagated correctly)

7. Check Asset A score:
Asset A: score = null ❌ (NOT recomputed)

8. Check Asset-Group AG1 score:
Asset-Group AG1: score = null ❌ (NOT recomputed)

## Expected Output

After manual result propagation to Agents, the system should **re-aggregate upward** through the entire hierarchy:
Step 1: Propagate DOWN → Manual result → All Agents ✅
Step 2: Re-aggregate UP → Agent scores → Asset score = 100 ✅
Step 3: Re-aggregate UP → Asset scores → Asset-Group score = 100 ✅

Expected final state:
Agent 1: score = 100
Agent 2: score = 100
Asset A: score = 100 ← recomputed from Agent scores
Asset-Group AG1: score = 100 ← recomputed from Asset scores

## Actual Output

Only Step 1 (downward propagation) is executed. Steps 2 and 3 (upward re-aggregation) are **not triggered**.

Final state (inconsistent):
Agent 1: score = 100 ✅
Agent 2: score = 100 ✅
Asset A: score = null ❌
Asset-Group AG1: score = null ❌

**Consequences:**
- Security Coverage widget shows `null` instead of actual coverage for the affected Asset/Group
- CISO dashboard metrics are incomplete
- MITRE ATT&CK heatmap does not reflect manually validated results
- The user sees correct data at Agent level but incorrect/missing data at Asset and Group level
- No workaround exists — the user cannot force a recomputation

## Additional information

**Suspected root cause:** The manual result setter invokes the downward propagation logic (`propagateToChildren()` or equivalent) but does **not** call the upward re-aggregation logic (`reAggregateToParent()` or equivalent) after propagation completes. The re-aggregation logic exists and works correctly when triggered by Collectors — it is simply not invoked in the manual update path.

Source: [FigJam architecture analysis](https://www.figma.com/board/1hfcComk4HCiZe3NUVM17T/Inject-expectation-brainstorm?node-id=0-1&p=f&t=3n81HadfSguQBCvt-0)

NB  *"the result will be applied to all agents of that asset — however, no result will be computed at the asset level"*

gabriel-peze:
### Proposed changes

* Move signatures frm a column in JSONB format from injects_expectations table to a dedicated injects_expectations_signatures table

### Testing Instructions

1. Eveyrhing should run as usual, signature shoul be store as always

### Related issues

* Closes #5151


openaev-model/src/main/java/io/openaev/database/model/InjectExpectationSignature.java
```diff
@@ -1,72 +1,72 @@
...
+import jakarta.validation.constraints.NotNull;
+import java.io.Serializable;
+import java.time.Instant;
+import lombok.*;
+import org.hibernate.annotations.SQLInsert;
+import org.hibernate.jdbc.Expectation;
+
+@Getter
+@Setter
 @Builder
```
MarineLeM: I think this annotation is not used, we can may be remove it as @AllArgsConstructor ? Wdyt?

openaev-api/src/main/java/io/openaev/service/InjectExpectationService.java
```diff
@@ -1207,8 +1188,19 @@ private void addDateSignatureToInjectExpectationsByAgent(
       @NotBlank final String agentId,
       @NotBlank final Instant date,
       @NotBlank final String signatureType) {
-    // Insert the signature for all agent and inject in one query
-    injectExpectationRepository.insertSignature(signatureType, date.toString(), injectId, agentId);
+    // Load all expectations for the inject/agent, append the signature, then persist the changes.
+    List<TechnicalInjectExpectation> injectExpectations =
+        injectExpectationRepository.findAllByInjectAndAgent(injectId, agentId);
```
MarineLeM, line 1193: This function seems heavy: it loads and save the entire InjectExpectation just to manipulate expectationSignatures.
Perhaps you should consider adding an ExpectationSignatureRepository to manage these objects directly, wdyt ?

MarineLeM (approved):

PR #6523, opened 2026-07-01
fix(composite-key): use composite key on java side

RomuDeuxfois:
## What

Completes the composite-key migration for connector entities (Injector, Collector, Executor) and the injector-contract relationship, so the Hibernate model matches the composite primary key `(id, tenant_id)` that already ships in the schema.

## Why

Built-in connector ids are shared across tenants, so `id` alone cannot be the primary key. The database already uses a composite `(id, tenant_id)` key, but the JPA mapping was still single `@Id`, with native-query workarounds for the resulting multi-tenant bugs (a cross-tenant contract leak, and an UPDATE that matched every tenant). This PR moves the model to composite keys and removes those workarounds, fixing the bugs at the mapping level.

## Key changes

* `Injector`, `Collector`, `Executor` use a composite `@IdClass(ConnectorCompositeId)` `(id, tenant_id)`; `InjectorContract` uses `@EmbeddedId`.
* The injector/contract link is remodelled from `@ManyToMany` to a dedicated join entity `InjectorInjectorContract`. A `@ManyToMany` cannot express the shared `tenant_id` column in both join groups; the join entity can, and it makes cross-tenant links structurally impossible at the database level.
* `Inject.injector` uses `@NotFound(IGNORE)` so an uninstalled connector degrades to null instead of failing audit serialization on every inject update.
* Tenant provisioning copies the default contracts without their links; the links are recreated in a foreign-key-safe way when the built-in injectors register (they do not exist yet at copy time).
* The cross-tenant link invariant is enforced in the `InjectorInjectorContract` constructor, the single point where every link is created.

## Testing

* Full `openaev-api` suite: 3688 tests, no regressions attributable to these changes.
* New regression coverage: tenant provisioning correctly links its built-in connectors (real Postgres and tenant filter), plus a unit test for the cross-tenant link guard.

## Follow-ups (documented separately)

Three latent, design-level items were deliberately deferred rather than patched here, to avoid regression risk on working code:
* `equals` / `hashCode` alignment with the composite key,
* observability on silently-null `@NotFound` associations,
* the asymmetric `@NotFound` on the join entity.


laugiov:
## Known follow-ups (deliberately deferred)

These items were found while finishing this PR and intentionally NOT fixed here, to keep the PR regression-free. They are latent (no live path today) and better decided at the design level than patched under pressure.


---

## 1. `equals` / `hashCode` are not aligned with the new composite key (deferred on purpose)

### What
The connector primary key became composite `(id, tenant_id)`. But identity in Java code was left keyed on `id` only:
- `Injector.equals` / `Injector.hashCode` compare `id` only.
- `Collector` and `Executor` inherit Lombok `@Data` equality from `BaseConnectorEntity`, which covers `id/name/type/external` but NOT `tenant_id`.
- The in-memory dedup inside `InjectorContract.addInjector` (`current.contains(injector)`) is therefore also `id`-only.

So two connectors that share a business id across two tenants (for example the built-in `email` injector in tenant A and in tenant B) are considered equal by Java, even though they are two distinct rows.

### Why it does not break anything today
Every connector read is scoped to a single tenant by the Hibernate `tenantFilter`. Java collections of connectors are therefore single-tenant, where business ids are unique, so `id`-only equality is correct in practice. This is a latent trap, not a live bug. No current code path iterates connectors across tenants in one collection.

### Why it was deferred (not fixed in the PR)
1. It is pre-existing: the `id`-only equality was already there before this PR, it is not a regression introduced by the composite-key completion work.
2. Fixing it is genuinely risky. `equals` / `hashCode` are used everywhere (every `Set`, `Map`, `.contains`, Hibernate's own entity dedup, the `addInjector` dedup). Making them composite-aware can change subtle behavior in code that currently relies on `id`-only identity, so it carries real regression risk. That conflicts with the "no regression" bar this PR was held to.
3. It is a design-level call about the composite-key model as a whole.

### Recommendation
Fix it in a dedicated follow-up, not in this PR, and validate with the full multi-tenant suite. Concretely:
- `Injector.equals` / `hashCode`, and the `BaseConnectorEntity` equality for `Collector` / `Executor`, should compare `(id, tenant_id)`.
- The `addInjector` dedup should then also be composite-aware.
- Alternatively, if the team decides the `tenantFilter` guarantee is a sufficient guard forever, document that decision explicitly next to the entities so the `id`-only equality is a conscious choice rather than an oversight.

An independent code audit rated this MAJOR (it is the invariant a multi-tenancy model should hold), while also confirming it is latent with no confirmed live path today. Both statements are true: important to close, not urgent, and worth doing carefully on its own.

### Locations
- `openaev-model/.../database/model/Injector.java` (equals/hashCode)
- `openaev-model/.../database/model/BaseConnectorEntity.java` (`@Data` equality for Collector/Executor)
- `openaev-model/.../database/model/InjectorContract.java` (`addInjector` `contains` dedup)

---

## 2. Observability for silent-null connector links (deferred, needs a targeted approach)

Two audit passes flagged that a connector link which does not resolve is dropped silently:
- `Inject.injector` and `InjectorInjectorContract.injector` use `@NotFound(action = IGNORE)`, so a dangling reference (connector uninstalled) becomes null with no log.
- `Injector.getContracts()` and `InjectorContract.getInjectors()` filter nulls out of the derived views.

Why it was not patched in this PR: a naive WARN is either noisy or wrong.
- Logging in the derived getters spams for any persistently-dangling link, because those getters are on the hot serialization path.
- Logging at inject resolution (`InjectUtils.resolveInjector`) false-positives on contracts that legitimately have no per-tenant injector (for example OVH and OpenCTI contracts, copied to a tenant that has no matching built-in injector).
- A blanket "every default contract must have a link" check has the same false-positive.

Recommendation: instrument this as a targeted signal that knows which contracts are expected to have an injector (the built-in-registered set), ideally a metric with tenant and contract-id context rather than a log line, or a health check run after provisioning. The provisioning gap itself is already guarded in CI by the strengthened test `TenantServiceTest.should_provision_and_link_builtin_injector_contracts_for_new_tenant`.

## 3. Asymmetric `@NotFound` on the join entity (minor)

`InjectorInjectorContract.injector` has `@NotFound(action = IGNORE)` but `injectorContract` does not. If a link's contract were ever missing, `getContracts()` would throw `EntityNotFoundException` before its null filter runs. Largely shielded today by the `ON DELETE CASCADE` on the join FK and the same-tenant guarantee. Not patched here because adding `@NotFound` to that side forces it EAGER, which risks a load cycle with the already-EAGER `InjectorContract.injectorLinks`. Worth a careful look if the asymmetry ever bites.

## 4. Strategic note (design level, for the record)

The composite-key approach is the right fix for connectors (built-in ids repeat across tenants, so id alone cannot be the primary key), and it removes real native-query workarounds that main still carries (cross-tenant contract leak, all-tenant UPDATE). But it makes connectors the only entities using composite keys, diverging from the platform norm (surrogate key plus `tenantFilter`), with a recurring Hibernate 6 cost. The `equals`/`hashCode` gap above is one symptom of that cost. Worth deciding whether composite keys stay a scoped exception for connectors or become a pattern, and documenting that choice.

openaev-api/src/main/java/io/openaev/service/InjectorService.java
```diff
@@ -478,23 +495,16 @@ private void updateExistingBuiltinInjector(
...
-        name,
-        contractor.getType(),
-        category,
-        Boolean.TRUE.equals(isCustomizable),
-        Boolean.TRUE.equals(isPayloads));
+    injector.setName(name);
+    injector.setType(contractor.getType());
+    injector.setCategory(category);
+    injector.setExternal(false);
+    injector.setCustomContracts(Boolean.TRUE.equals(isCustomizable));
```
Dimfacion, line 502:
```suggestion
    injector.setCustomContracts(isCustomizable);
```
Not sure of why we're doing the equals here

openaev-api/src/main/java/io/openaev/service/InjectorService.java
```diff
@@ -478,23 +495,16 @@ private void updateExistingBuiltinInjector(
...
+    injector.setPayloads(Boolean.TRUE.equals(isPayloads));
```
Dimfacion, line 503: **nitpick :** same as above (https://github.com/OpenAEV-Platform/openaev/pull/6523/changes#r3542487325)

openaev-api/src/main/java/io/openaev/service/InjectorService.java
Dimfacion: Should have drank more coffee : isCustomizable can be technically null so it's better to use the equals. Closing the comment

openaev-api/src/main/java/io/openaev/service/InjectorService.java
Dimfacion: Same answer as the original comment

openaev-model/src/main/java/io/openaev/database/model/InjectorInjectorContract.java
```diff
@@ -0,0 +1,125 @@
...
+  @Id
+  @Column(name = "injector_contract_id")
+  private String injectorContractId;
+
+  @Id
+  @Column(name = "tenant_id")
+  private String tenantId;
+
+  @NotFound(action = NotFoundAction.IGNORE)
+  @ManyToOne(fetch = FetchType.EAGER)
```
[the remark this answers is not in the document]
laugiov: Fair EAGER smell, but here it is load-bearing. Two consumers read a real property off the injector, not just its id:

MailingService.sendEmail, reached from UserService.requestPasswordReset which is @Async and not @Transactional, reads injector.getType(). With @Async there is no open-in-view and the contract is already detached after findById, so LAZY would throw LazyInitializationException on the password-reset email path.
Serialization reads injector.getName() through the injector_contract_injector_names JSON property, so any contract output loads the injectors regardless.
LAZY would therefore break the async mail path and give no gain on the serialization path (injectors are fetched either way, batched via default_batch_fetch_size=50). Keeping EAGER. A real reduction would need a names projection plus a transactional mail path, tracked separately.

Dimfacion (approved):
Tested OK. No issue found on my end

PR #6528, opened 2026-07-02
fix(executor): override document argument default_value with inject content in executable payload (#6509)

### LINKED ISSUE #6509: fix(export): inject-attached documents missing from scenario export
## Problem

When exporting a scenario, documents directly attached to injects (\inject_documents\) are not included in the \scenario_documents\ list of the export ZIP. As a result, re-importing the scenario fails because the inject references documents that are absent from the archive.

## Root Cause

In \ScenarioService.exportScenario()\ (lines ~546-560), the code collects documents from:
- \scenario.getDocuments()\ ✅
- \inject.getPayload().getAttachedDocument()\ (Payload-attached files only) ✅

But it **skips**:
- \inject.getDocuments()\ → \InjectDocument.document\ (documents attached directly to the inject via \injects_documents\ table) ❌

## Expected Behaviour

All documents referenced by any inject (whether via \inject.getDocuments()\ or \inject.getPayload().getAttachedDocument()\) must appear in \scenario_documents\ and be bundled in the export ZIP.

## Steps to Reproduce

1. Create a scenario with an inject
2. Attach a document directly to the inject (not via a Payload)
3. Export the scenario
4. Inspect the exported JSON → \scenario_documents\ is empty / does not include the inject document
5. Re-import the scenario → error because the inject references a missing document

## Fix

Add \inject.getDocuments().stream().map(InjectDocument::getDocument)\ to the document-collection stream in \ScenarioService.exportScenario()\.

Seb-MIGUEL:
## Summary

Fixes the \[FAIL] zip_file arg did not stage on disk\ error when executing DEEP#DRIVE (and any other payload with document-type arguments).

## Root Cause

The implant calls \GET /api/tenants/{tenantId}/injects/{injectId}/{agentId}/executable-payload\ to retrieve the payload to execute. It iterates \payload_arguments\, and for each argument of type \document\, it downloads the file using \default_value\ as the document UUID.

Previously, \processCommandPayload()\ correctly encoded the command content (replacing \#{zip_file}\ with \#{location}/filename.zip\ using the inject content value), but it returned \payload_arguments\ **unchanged** — still carrying the payload template's \default_value\ UUID, not the one selected in the inject.

Result: the implant downloaded the wrong (or non-existent) document, causing the file staging to fail.

## Fix

Added \
esolveDocumentArgumentsFromInjectContent()\ in \ExecutableInjectService\: after encoding the command, this method iterates the payload arguments and, for any \Document\-type argument, creates a copy with \default_value\ set to the actual inject content value (falling back to the original default if not set).

Copies are created to avoid mutating shared Hibernate-managed \PayloadArgument\ instances.

## Test

Added \given_documentArgumentOverriddenInInjectContent_should_returnActualDocumentIdInPayloadArguments()\ in \InjectApiTest.RetrievingExecutablePayloadInject\: creates a Command payload with a document arg (random default UUID), overrides it in the inject content with \DOCUMENT1\, and asserts that \payload_arguments[0].default_value\ in the response equals \DOCUMENT1.getId()\.

Closes #6509

### DIFF
--- a/openaev-api/src/main/java/io/openaev/rest/inject/service/ExecutableInjectService.java
+++ b/openaev-api/src/main/java/io/openaev/rest/inject/service/ExecutableInjectService.java
@@ -288,16 +288,22 @@ public class ExecutableInjectService {
       Inject inject,
       List<ObjectNode> injectorContractFields,
       String obfuscator) {
-    switch (contract.getPayload().getTypeEnum()) {
-      case PayloadType.COMMAND:
-        return processCommandPayload(
-            payloadToExecute, contract, inject, injectorContractFields, obfuscator);
-      case PayloadType.DNS_RESOLUTION:
-        return processDnsResolutionPayload(payloadToExecute, inject);
-      default:
-        // All other payload types are intentionally passed through unchanged.
-        return payloadToExecute;
-    }
+    Payload processed =
+        switch (contract.getPayload().getTypeEnum()) {
+          case PayloadType.COMMAND ->
+              processCommandPayload(
+                  payloadToExecute, contract, inject, injectorContractFields, obfuscator);
+          case PayloadType.DNS_RESOLUTION -> processDnsResolutionPayload(payloadToExecute, inject);
+          default ->
+              // All other payload types are intentionally passed through unchanged.
+              payloadToExecute;
+        };
+    // Override the default_value of document-type arguments with the actual inject content value
+    // for all payload types. The implant uses payload_arguments[].default_value to download
+    // documents before execution; without this override it would download the payload's default
+    // document instead of the one configured on the inject.
+    resolveDocumentArgumentsFromInjectContent(processed, inject.getContent());
+    return processed;
   }
 
   private Payload processCommandPayload(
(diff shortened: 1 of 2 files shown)


Seb-MIGUEL:
Addressed in commit 1d8ec33: moved \
esolveDocumentArgumentsFromInjectContent()\ to take \Payload\ (base class) and call it in \processPayloadToExecute()\ after the type switch — now applies to all payload types (COMMAND, EXECUTABLE, FILE_DROP, NETWORK_TRAFFIC, AI_ATTACK, and future ones).

openaev-api/src/main/java/io/openaev/rest/inject/service/ExecutableInjectService.java
```diff
@@ -319,6 +325,37 @@ private Payload processCommandPayload(
     return payloadCommand;
   }
 
+  private void resolveDocumentArgumentsFromInjectContent(
```
RomuDeuxfois, line 328: **todo**: We already do something in ExecutableInjectService.replaceArgumentsByValue line 121L Why adding something new here ?

Seb-MIGUEL:
@RomuDeuxfois re: resolveDocumentArgumentsFromInjectContent - the two methods serve different roles. replaceArgumentsByValue (l.84) processes the command string: replaces #{zip_file} with #{location}/filename.zip - the path where the file will be once downloaded on disk. resolveDocumentArgumentsFromInjectContent (this PR) updates payload_arguments[].default_value - that is what the implant reads to know which document UUID to download from the API before running the command (implant iterates payload_arguments, finds type==document, calls api.download_file(default_value)). Without this fix, default_value always holds the payload template UUID (not the inject-configured one), so the implant downloads the wrong document even though the command path is correct.

RomuDeuxfois (approved):

PR #6537, opened 2026-07-02
fix(multi_tenancy): external reference for agent can't be the agent id (#6526)

### LINKED ISSUE #6526: fix(scope):  fix(multitenancy): 2 crowdskike installations on 2 different tenants clashes
## Description
When two or more tenants deploy the same CrowdStrike or SentinelOne executor, agents from the background sync job (ManagerIntegrationsSyncJob) collide on their primary key, causing:
```
ERROR: insert or update on table "agents" violates foreign key constraint "agent_executor_id_fk"
  Detail: Key (agent_executor, tenant_id)=(..., <wrong_tenant>) is not present in table "executors".
```
 **Root cause**
EndpointService.setNewAgentAttributes sets agent.setId(input.getExternalReference()) for CS/S1 agents — using the external device ID as the DB primary key. When a second tenant syncs the same platform, CrudRepository.save() calls merge() (since ID is non-null), finds the existing agent from tenant A, and issues an UPDATE with tenant B's executor — violating the composite FK (agent_executor, tenant_id).
 **Additional issue**
The unique_security_platform_name_type_ci_idx index is not tenant-scoped, preventing a second tenant from installing the same security platform (e.g. "Microsoft Defender" / EDR).
 **Fix**
• Decouple agent_id from externalReference — use UUID as PK, keep externalReference for external API communication
• Add tenant-scoped unique index on security_platforms
• Data migration to reassign UUIDs to existing CS/S1 agents

## Environment

1. OS: { e.g. macOS 14, Windows 11, Ubuntu 22.04, etc. }
2. Version: { e.g. 1.2.3 }
3. Other environment details:

## Reproducible steps

1. { e.g. Run ... }
2. { e.g. Click ... }
3. { e.g. Error ... }

damgouj:
### Proposed changes

*
*

### Testing Instructions

1. Step-by-step how to test

### Related issues

* Closes #6526

### Checklist


openaev-api/src/main/java/io/openaev/migration/V6_20260701180000000__Reassign_agent_ids_for_external_executors.java
```diff
@@ -0,0 +1,95 @@
...
+ *   <li>{@code asset_agent_jobs.asset_agent_agent}
+ *   <li>{@code agents.agent_parent} (self-reference)
+ * </ol>
+ */
+@Component
+public class V6_20260701180000000__Reassign_agent_ids_for_external_executors
+    extends BaseJavaMigration {
+
+  @Override
+  public void migrate(Context context) throws Exception {
```
corinnekrych, line 32: i like the idea of migrating old ids so we dont mix in the code looking uuid vs CS/S1 external id. Much cleaner. 🤞 that the migration is working well on main/main-ff

corinnekrych:
With this PR, you redesign the way we send request to CS/S1 not using batch processing anymore. 
- it simplifies the design
- but have impact of performance

## What are the performance consequences of `entities/sessions/v1` vs batch?

**slower — potentially significantly so at scale.** Here's why:

| | Old batch API | New entity API |
|---|---|---|
| **CS API calls to open session** | 1 call for N devices → 1 `batch_id` | 1 call **per device** → N `session_id`s |
| **CS API calls to run command** | 1 call with `batch_id` covers all devices | 1 call **per device** with `session_id` |
| **Total CS API calls for N devices** | 2 calls | 2×N calls |
| **Pagination limit** | 2500 agents per batch | 100 actions per schedule slot |

For 100 Windows agents, the old approach made **2 CS API calls**. The new approach makes **200 CS API calls** (100 session opens + 100 command executions), sent in chunks of 100 with 5-second delays between chunks.

The pagination limit being reduced from 2500 → 100 also means more scheduling slots and more wait time for large deployments, since each slot adds a 5-second delay:
- 1000 agents: old = 2 API calls; new = 2000 API calls spread over 10 schedule slots = **~45 seconds of forced delay** just from throttling, before accounting for actual API latency.

@damgouj was the initial choice to use batch processing a performance choice?
I think there are options to keep batch processing let's discuss it

corinnekrych:
@damgouj with 4e1f4b423 i've added an ADR to document our decision.
@laugiov I've moved the ADR template to code base. we can discus further next weekly how we want to move froward with ADR and numbering  (not sure this one should be 001)

corinnekrych (approved):

PR #6555, opened 2026-07-03
feat(chaining): add primitive and complex types in chaining (#6536)

### LINKED ISSUE #6536: feat(chaining):  refactor chaining engine to include complex and primitives types
## Use case

**Objective**
Review and refactor ContractOutputType, ArgumentType, ArgumentSubtype, ConditionKeyType and ConditionKeySubtype in backend and front end side.

**Actions**

1. Create or prepare a back end registry that centralizes:

- Primitive types
- Complex output types
- Mapping metadata between complex fields and primitive types
- Type usage rules for:
 - contract outputs
 - action/payload arguments
 - event conditions

2. Update UI calls: Check Payloads arguments, Events, Actions, WorkflowScopeVariables

3. Update chaining engine to limit depending by complex or primitives ⇒ User should only see primitive types

</body>
</html>

**Dod**
- [x]  A backend chaining type registry is created or prepared as the central source of truth.
- [x]  Primitive types are defined in the registry.
- [x]  Complex output types are defined in the registry.
- [x]  Complex-field-to-primitive mapping metadata is defined in the registry.
- [ ]  `ContractOutputType` is reviewed/refactored to align with the registry.
- [x]  `ArgumentType` and `ArgumentSubtype` are reviewed/refactored to align with primitive-compatible input types.
- [x]  `ConditionKeyType` and `ConditionKeySubtype` are reviewed/refactored to align with primitive-compatible condition types.
- [x] Ui: In actions we can only link with primitive types
- [x] Ui: In event, we can only link with primitives types
- [x] Ui: in Payload argument, we can only link with primitive types

savacano28:
### Proposed changes

- Introduce PrimitiveType / ComplexType plus a ChainingTypeRegistry and mapping model (ChainingMappedType, ChainingOutputType, ChainingTypeKind) to translate ContractOutputType into chaining semantics.
- Replace legacy enums (ConditionKeyType, ArgumentType, ArgumentSubType, ConditionKeySubtype) across backend and frontend, and update APIs/DTOs/tests accordingly; payload argument types are now served by `GET /api/threat_arsenals/argument-types/` (gated by the INJECT_CHAINING feature flag).
- Add scope-aware validation of primitive values before persisting them into workflow state (PrimitiveValueValidator + PrimitiveValidationContextBuilder): IPs, subnets, domains and asset/asset-group IDs are checked for format and against workflow scope allow/deny rules; Port and Number get format checks; other types are accepted as-is until dedicated rules exist.
- Add a migration that removes the deprecated `subtype` field from payload arguments and unwraps legacy wrapped JSON in inject status payload output.

Review follow-ups included in this PR:

- Restored the `condition_key_type` JSON property on `Condition` so scenario/simulation exports keep the key type and the V1 importer can read it back; dropped the removed `condition_key_subtype` from the export mixin.
- Renamed the migration after the latest main migrations so it is not skipped by Flyway on environments with out-of-order disabled.
- Renamed `isValidForPrimitiveType` to `isAcceptedForPrimitiveType`, added Port/Number format validation, added a warn log for unknown output keys, and guarded mapper conditions with a null key type.
- Added unit tests for PrimitiveValueValidator and PrimitiveValidationContextBuilder.

image
image
image

### NOT included

- Correlated data (next PR)
- Mapping from correlated data to primitive values (next PR)
- Safeguard for new output processors
- EE license gating for chaining: to be decided holistically when the INJECT_CHAINING feature flag is removed (see review discussion)

### Threat Arsenal

**Payload with primitive argument (e.g., IP)**
- Create payload, example: `Ping #{ip}`
- Add arguments
- Check primitive types in payload argument types
- Create/execute atomic with that payload

**Payload with document argument type**
- Create payload with an argument type: document
- Create/execute atomic with that payload

**Payload management**
- Update payload
- Export/Import
- Delete payload

### Scenario/Simulation Chaining

**Scope variables management**
- CRUD
- Select primitive type for scope variables

**Create scenario/simulation**
- Fill scope configuration with endpoints, IPs, domains
- Add an action with argument input, example: IP
- Check primitive types in the mapping options
- Add event
- When adding a condition, check primitive types - use a condition with IP, for example
- Execute
- Check that the action execution used the IP defined in the scope

### Related issues

* Closes #6536

### Checklist


openaev-api/src/main/java/io/openaev/migration/V6_20260720178453597__Remove_Argument_Subtype_And_Unwrap_Json.java
```diff
@@ -0,0 +1,119 @@
...
+            FROM json_array_elements(payload_arguments) argument
+          )
+          WHERE payload_arguments IS NOT NULL
+            AND json_typeof(payload_arguments) = 'array'
+            AND EXISTS (
+              SELECT 1
+              FROM json_array_elements(payload_arguments) argument
+              WHERE argument::jsonb ? 'subtype'
+            );
+          """);
```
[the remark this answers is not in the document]
savacano28: N/A. New feature, no legacy cases to consider.

openaev-model/src/main/java/io/openaev/database/model/PayloadArgument.java
```diff
@@ -11,11 +11,7 @@ public class PayloadArgument {
 
   @NotNull
   @JsonProperty("type")
-  private ArgumentType type;
-
-  @JsonProperty("subtype")
-  @Schema(nullable = true, description = "Optional sub-field key for structured output types")
-  private ArgumentSubType subtype;
+  private PrimitiveType type;
```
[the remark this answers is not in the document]
savacano28: N/A. New feature, no legacy cases to consider.

openaev-model/src/main/java/io/openaev/database/model/Condition.java
```diff
@@ -44,16 +43,8 @@ public class Condition implements Base {
...
   @Schema(description = "Key type")
-  private ConditionKeyType keyType;
-
-  @Column(name = "condition_key_subtype")
-  @Nullable
-  @Enumerated(EnumType.STRING)
-  @JdbcTypeCode(SqlTypes.NAMED_ENUM)
-  @Schema(description = "Key subtype")
-  private ConditionKeySubtype keySubtype;
+  private PrimitiveType keyType;
```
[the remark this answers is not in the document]
savacano28: N/A. New feature, no legacy cases to consider.

openaev-model/src/main/java/io/openaev/database/model/ChainingTypeKind.java
```diff
@@ -0,0 +1,7 @@
+package io.openaev.database.model;
+
+public enum ChainingTypeKind {
+  PRIMITIVE,
+  COMPLEX,
+  NON_CHAINABLE
```
EvaE-Filigran: should maybe be NO_CHAINABLE or NOT_CHAINABLE ou UNCHAINABLE ? "NON" is french :)

openaev-api/src/main/java/io/openaev/service/chaining/WorkflowStateService.java
```diff
@@ -87,18 +87,18 @@ private void propagateToLocalStates(
...
-  private Set<ConditionKeyType> resolveOutputKeyTypes(Set<String> typeNames) {
+  private Set<PrimitiveType> resolveOutputKeyTypes(Set<String> typeNames) {
     return typeNames.stream()
         .map(
             typeName -> {
               try {
-                return ConditionKeyType.valueOf(typeName);
+                return PrimitiveType.valueOf(typeName);
               } catch (IllegalArgumentException e) {
                 return null;
```
impolitepanda, line 106: Could you add a warn log here so that if this error occurs, we can more easily find out which key triggered the exception ?

openaev-api/src/main/java/io/openaev/service/chaining/PrimitiveValueValidator.java
```diff
@@ -0,0 +1,98 @@
...
+   * @param value candidate value
+   * @param context precomputed scope validation context
+   * @return true when the value is accepted for this primitive type
+   */
+  public static boolean isValidForPrimitiveType(
+      PrimitiveType primitiveType, String value, PrimitiveValidationContext context) {
+    if (value == null) {
+      return false;
+    }
+    return switch (primitiveType) {
```
impolitepanda, line 41: I have an issue with this function, but I don't know how to formalize it so I think it could be good to just talk about it, maybe ?

My issue is the default case which returns true, meaning that we don't do any validation for other cases at all. And it seems strange to me. What do you think?

impolitepanda: For example, if somebody calls this function with a value "toto" and a type "Port", it will return true, meaning it's a valid data, and it's definitely not.

openaev-api/src/main/java/io/openaev/api/threat_arsenal/ThreatArsenalApi.java
```diff
@@ -128,6 +146,13 @@ public List<Collector> collectorsFromAction(@PathVariable String actionId) {
     return threatArsenalService.getCollectorsForActionRemediation(actionId);
   }
 
+  private List<PrimitiveType> resolveAvailableTypes() {
+    if (!previewFeatureService.isFeatureEnabled(INJECT_CHAINING)) {
+      return List.of(PrimitiveType.Text, PrimitiveType.Document, PrimitiveType.TargetedAsset);
```
impolitepanda, line 163: How will this behave with the legacy engine (for time based simulation) when we remove the feature flag ?

openaev-api/src/main/java/io/openaev/service/chaining/PrimitiveValueValidator.java
savacano28: Yes, for the moment the validation is only based on the defined scope rules :)

impolitepanda: Maybe we should change the name of the function to avoid confusion, then ?

openaev-api/src/main/java/io/openaev/api/threat_arsenal/ThreatArsenalApi.java
savacano28: By removing the feature flag (ff), we will expose all the different types for defining payload arguments; however, I have a question regarding EE (?), so I should add a check here for that as well.

openaev-api/src/main/java/io/openaev/service/chaining/WorkflowStateService.java
SamuelHassine: Done in 17e59ec1b: resolveOutputKeyTypes now logs a warn with the offending key when it does not match any PrimitiveType, so the trigger is identifiable in the logs.

openaev-api/src/main/java/io/openaev/service/chaining/PrimitiveValueValidator.java
SamuelHassine: Addressed in 17e59ec1b, in two parts:

- Renamed the function to isAcceptedForPrimitiveType and documented the semantics in the class javadoc: scope-restrictable types (IPs, subnets, domains, asset/asset-group IDs) get format + scope-rule checks, and types without defined rules are accepted as-is. The "accepted" naming makes explicit that this is an ingestion gate, not a full semantic validation of every primitive type.
- Tightened the obvious gap you pointed out: Port now requires an integer in 0-65535 and Number must parse as a number, so "toto" with type "Port" is rejected. Other types (Text, Username, Hash, ...) intentionally remain pass-through until dedicated rules are defined, since they cannot be constrained by scope rules today.

Unit tests were added for all branches (PrimitiveValueValidatorTest).

openaev-api/src/main/java/io/openaev/api/threat_arsenal/ThreatArsenalApi.java
SamuelHassine: I looked at how EE gating is done elsewhere in the repo to answer this. Current state: the whole chaining stack is gated only by the INJECT_CHAINING preview flag (WorkflowApi.checkWorkflowFeatureEnabled, WorkflowService.isPreviewFeatureChainingEnable, and this resolver); there is no license check anywhere in chaining today.

Behavior when the flag is removed: this endpoint will return all primitive types for payload arguments to every engine, including the legacy time-based one. That is coherent, because argument types are payload metadata (what a payload declares it needs), not a chaining runtime capability; the chaining runtime itself stays protected by the workflow endpoints' own guards.

On the EE question: the established pattern is enterpriseEditionService.isLicenseActive(licenseCacheManager.getEnterpriseEditionInfo()) (see ScenarioService.throwIfScenarioNotLaunchable or ConnectorOrchestrationService.throwIfEnterpriseLicenseNotActive). If chaining is meant to be EE-only, that check should be introduced consistently across all chaining entry points at the moment the feature flag is removed, not only in this resolver - adding it here alone would be inconsistent (the rest of the chaining API would remain non-EE) and is premature while the flag still exists. I suggest tracking the EE gating decision in the follow-up issue that will remove the INJECT_CHAINING flag.

SamuelHassine:
## Review and fix summary

Full review and fix pass completed on this PR (head `17e59ec1b`).

**Merge with main**: the branch was conflicting with main; merged `origin/main` and resolved conflicts in `ExecutableInjectDTOMapper` (took main's equivalent endpoint-unproxy logic), `PayloadComponent.tsx` (re-applied the subtype removal and `formatPrimitiveTypeLabel` on top of main's `PlatformIconGroup` change), `api-types.d.ts` (kept main's new `AiTargetTarget`, dropped the removed `ArgumentTypeOutput`), and the 9 translation files (3-way key merge, re-sorted with `sort-translation-files.js`).

**Findings fixed during the independent review** (commit `17e59ec1b`):

- must-fix: `Condition.keyType` had lost its `@JsonProperty("condition_key_type")`; the export mixin (`Mixins.ConditionExport`) whitelists that name, so scenario/simulation exports would have silently dropped the condition key type while the V1 importer still reads it. Restored the annotation and removed the dead `condition_key_subtype` entry from the mixin.
- must-fix: after the merge, the migration `V6_20260717150000000__Remove_Argument_Subtype_And_Unwrap_Json` sorted before main's already-merged `V6_20260717220000000`/`221000000`; with Flyway `out-of-order` disabled (default) it would be skipped on environments tracking main. Renamed to `V6_20260718080000000`.
- should-fix: `InjectExecutionStep.applyMapping` could NPE on a MAPPER condition with a null key type; now guarded with a warn log.
- nits: javadocs said `NON_CHAINABLE` while the enum is `NOT_CHAINABLE`; removed a no-op `.map()` in `useArgumentTypes.ts`.

**Review comments addressed** (all threads replied and resolved):

- Warn log added in `WorkflowStateService.resolveOutputKeyTypes` with the offending key (@impolitepanda).
- `PrimitiveValueValidator.isValidForPrimitiveType` renamed to `isAcceptedForPrimitiveType` with documented scope-based semantics, plus new format validation for Port (integer 0-65535) and Number, so "toto" as Port is now rejected (@impolitepanda).
- EE gating for `resolveAvailableTypes`: investigated the repo's EE pattern (`EnterpriseEditionService.isLicenseActive`); no chaining code is EE-gated today, so adding a license check only in this resolver would be inconsistent. Replied with the recommendation to decide EE gating holistically when the INJECT_CHAINING flag is removed.

**Coverage**: added `PrimitiveValueValidatorTest` (all validator branches) and `PrimitiveValidationContextBuilderTest` (rule collection, normalization, null mappings). Codecov patch/project checks are green on the new head. Remaining gaps: `InjectExecutionStep` run/update paths are only partially covered (they need heavy execution mocking and are better served by integration tests), and `WorkflowStateService.newOutput` correlated-path logic still contains TODO placeholders.

**CI**: all GitHub Actions checks green (migrations guard, backend compile, API tests, E2E chrome/webkit, API types check, CodeQL, codecov, signed commits). PR is `MERGEABLE` / `CLEAN`. Description updated to reflect the final state; title already matches the repo convention.

Not approving as part of this pass - human review still required.

impolitepanda (approved):

PR #6595, opened 2026-07-07
feat(multitenancy): add an "activate-tenant-table" skill for v2 table activations (#6594)

### LINKED ISSUE #6594: feat(multitenancy): add an "activate-tenant-table" skill for v2 table activations
**Epic**: [feat(multitenancy): epic for refactor for tenant scoped API](https://github.com/OpenAEV-Platform/openaev/issues/6393)

---

## Context

The `import_mappers` pilot (#6255) defined the pattern to switch a table from v1 isolation (`@Filter` on the entity) to v2 (statement inspector + `can_access_tenant`). The activation issues (#6399 to #6403) all follow the same recipe: eligibility checks, code-path inventory, isolation tests, write attribution, one-commit go-live, full regression.

This recipe is precise and repetitive. It is a good fit for a skill under `.github/skills/`, next to the existing ones (`add-migration`, `add-test`, `review-multi-tenancy`, ...). A skill makes the process the same for every table and every developer, with or without an AI assistant.

## Scope

Add `.github/skills/activate-tenant-table/SKILL.md`. The skill must:

- follow the pilot: every step points to a real file from #6255 as the model, no invented patterns
- work test-first (TDD): the isolation test is written before any production code, must fail for the expected reason, and the raw red/green outputs are kept as evidence
- cover only HTTP-only strict tables (`tenant_id` NOT NULL). Dual-scope tables and tables written by background jobs are stop conditions, reported instead of worked around (background work waits for #6398)
- include the eligibility gates: entity implements `TenantBase`, tenant column not nullable, unique constraints on business keys include `tenant_id` (example: `mitigations` still has a global unique index on `mitigation_external_id`, which blocks its activation until a prep migration lands)
- include the full code-path inventory: the table does not belong to one API (the pilot had to wire `ScenarioImportApi` and `ExerciseImportApi` too)
- enforce the one-commit go-live: remove the v1 `@Filter` and add the table to `openaev.tenant.active-tables` in the same commit, plus the config guard test and the arch test entries
- state the rollback rule: revert the whole go-live commit, never the property alone
- end with a full regression pass (spotless, compile, targeted isolation tests, full API suite) and a short report (degradations accepted, child tables, client impact)

## Definition of Done

- [ ] `SKILL.md` added under `.github/skills/activate-tenant-table/`, same format as the existing skills
- [ ] every referenced model file exists in the repository (paths checked)
- [ ] the skill is validated on one real table from the activation batch (dry run of the eligibility and inventory phases at minimum)
- [ ] stop conditions are explicit: dual-scope table, background writer, global unique index on a business key
- [ ] TDD evidence rule is explicit: red and green outputs are part of the activation report

laugiov:
Closes #6594.

Adds `.github/skills/activate-tenant-table/SKILL.md`, the runbook to switch a table from v1 `@Filter` isolation to v2 (statement inspector + `can_access_tenant`), based on the `import_mappers` pilot (#6255).

The skill works test-first, keeps the red/green outputs as evidence, and encodes the guardrails: eligibility gates (strict table, HTTP-only, tenant-aware unique constraints), full code-path inventory, one-commit go-live, atomic rollback rule, full regression pass.

Validation done for the DoD:
- every model file referenced by the skill exists on `main` (paths checked)
- eligibility and inventory phases dry-run on a real candidate, `mitigations` (#6400), which surfaced a real blocker the gates now catch: a global unique index on `mitigation_external_id` not yet tenant-aware.

### DIFF
--- a/.github/skills/activate-tenant-table/SKILL.md
+++ b/.github/skills/activate-tenant-table/SKILL.md
@@ -404,9 +404,9 @@ Before marking the issue done, write down:
 - the endpoints wired and the arch-test entries added
 - background readers left degraded (from Phase 0/1), each with a one-line impact
 - child tables and how they are covered
-- client impact: the frontend does not send a tenant selector on resource
-  calls today, so multi-tenant users get 400 on create until it does. Say it
-  explicitly so the rollout plans the frontend work.
+- client impact: writes now require a single-tenant scope. Calls using the tenant path
+  (`/api/tenants/{tenantId}/...`) already satisfy this; callers using the header route or no selector
+  may get 400 on create/import until they switch to a single-tenant selector.
 
 ## Definition of Done
 
--- a/.github/skills/activate-tenant-table/SKILL.md
+++ b/.github/skills/activate-tenant-table/SKILL.md
@@ -320,7 +320,7 @@ For every other API or service found in Phase 1 that reads the table:
 - add a `TxCtx` parameter on its `@Transactional` entrypoint
 - add an isolation test proving the cross-tenant case through that path.
   Models: `openaev-api/src/test/java/io/openaev/rest/scenario/ScenarioImportApiTenantIsolationTest.java`
-  and `ExerciseImportApiTenantIsolationTest.java` (same folder pattern).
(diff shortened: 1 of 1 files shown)


corinnekrych (approved):
the best way to review is to test it, let's merge this one and we can refine it later as we progress on generating more v2 api

PR #6596, opened 2026-07-07
feat(multitenancy): v2 tenant-isolation test skill (#6390)

### LINKED ISSUE #6390: feat(Multi-tenancy): skill/template to generate v2 tenant-isolation tests
## Context
Every activable table needs an end-to-end v2 isolation test. Writing one by hand per table (7 now,
31 once background scope lands) does not scale and drifts in style.

## Goal
Improve the `TENANT_ISOLATION.md` skill so it produces a v2-correct isolation test for a given table,
modeled on `ImportMapperHttpIsolationTest`:
- one tenant path per test (the per-request scope is set once; the aspect forbids redefining it);
- seed straight into the table (no scope set during setup);
- raw-JDBC ground-truth reads (the inspector never rewrites them);
- v2 semantics in the assertions (a cross-tenant delete is a 2xx scoped no-op, not a 404);
- cover the `X-Tenant-Ids` header route and the fail-closed case.
- write-attribution cases on `create`/`import` once `TenantWriteScopeResolver` is wired on the endpoint: the created row gets the scope's tenant; a supplied out-of-scope tenant is refused (400); a multi-tenant scope without an explicit selector is refused (400). Reference: `TenantWriteScopeResolver` (wired on `MapperApi` today).
- for a dual-scope table (`roles`, `groups`, `parameters`): a platform-row write case, once the sanctioned platform-write path exists.

## Consolidation
Fold the overlapping `MapperApiTenantIsolationTest` (#5718 adapted to v2) into the generated
canonical form once available.

## Reference
Corinne's review on #6255 (thread 4), johanah29 #5718, `ImportMapperHttpIsolationTest`.

## Pairs with
The "tenant scope coverage" CI gate issue.

laugiov:
Closes #6390.

Rewrites `.github/skills/add-test/TENANT_ISOLATION.md` and its templates for multi-tenancy v2, modeled on the real pilot (`ImportMapperHttpIsolationTest`, PR #6255). The old v1 rules that are now wrong are called out explicitly (cross-tenant delete is a 2xx scoped no-op, not a 404; no scope switch inside a test). Cases are generated from the API shape (probe first), write-attribution and upsert cases are the red of the activation TDD, the dual-scope platform-write case stays a pending scaffold (Q7).

Also folds `MapperApiTenantIsolationTest` into the pilot: its 4 cases all have exact equivalents in `ImportMapperHttpIsolationTest` (mapping in the skill history / issue).

Validation done:
- pilot suite green on a pristine database (17/17) after the fold
- a test generated with the skill for `kill_chain_phases` (#6402) compiles and fails only for the predicted pre-wiring reasons (12 run, 6 red, 0 errors, 1 disabled header-route case). That file is deliberately not in this PR; it will open #6402.

Review note: the canonical form is a dedicated test file, not a `TenantIsolation` nested class. `@TestPropertySource` is class-level and forks the Spring context, so nesting it in `*ApiTest` would impose the activated context on every test of the class. This diverges from the previous convention on purpose; the skill documents the reason.


laugiov:
Second-pass comments addressed in a0be63b7e:
- Case matrix POST-create row now asserts `tenant_id` with a raw-JDBC read (not an entityManager query), matching the raw-JDBC rule stated elsewhere in the skill. That was a leftover from an earlier edit.
- Templates intro no longer claims every snippet is from the pilot. It now states the core read/write patterns come from the pilot, and each snippet with no pilot equivalent (rawTenant, upsert, options/batch lookup) is marked inline with its real provenance.

For the two first-pass comments about the `activate-tenant-table` skill: it lands in #6594, so that PR should merge first. Both references now cite the PR number so the pointer is valid regardless of merge order.

.github/skills/add-test/TENANT_ISOLATION.md
```diff
@@ -1,110 +1,165 @@
...
+reason; the reasons are written down so a hand-written test does not drift.
 
-- An API endpoint has a tenant-scoped path: `/api/tenants/{tenantId}/...`
-- The entity table has a `tenant_id` column (listed in `TenantScopedTables.java`)
-- The existing `*ApiTest.java` does NOT have a `TenantIsolation` nested class
+Full code templates: [`examples/tenant-isolation-templates.md`](examples/tenant-isolation-templates.md).
 
-## Prerequisites
+This skill pairs with:
+- the `activate-tenant-table` skill (PR #6594): its isolation-test phase
```
corinnekrych: let's not reference PR in the skills but rather file path

.github/skills/add-test/examples/tenant-isolation-templates.md
```diff
@@ -1,259 +1,351 @@
...
+    tenantB = tenantHelper.createTenantWithCurrentUser("http-iso-b").getId();
+    rowA = seedRow(tenantA, "row-a");
+    rowB = seedRow(tenantB, "row-b");
+  }
+
+  // ... cases from the matrix, see below ...
+
+  // -- helpers, all from the pilot --
+
+  private String seedRow(String tenantId, String name) {
```
corinnekrych, line 83: Since we're doing API level test why not use API call for the create (for ex for sceanrio api):
```
      String createResponse =
          mvc.perform(
                  post("/api/tenants/" + tenantX.getId() + "/scenarios")
                      .content(asJsonString(input))
                      .contentType(MediaType.APPLICATION_JSON)
                      .accept(MediaType.APPLICATION_JSON)
                      .with(csrf()))
              .andExpect(status().is2xxSuccessful())
              .andReturn()
              .getResponse()
              .getContentAsString();
```
why do we need to inser tin DB?

.github/skills/add-test/examples/tenant-isolation-templates.md
```diff
@@ -1,259 +1,351 @@
...
+        // one setParameter per seed column
+        .executeUpdate();
+    return id;
+  }
+
+  // Ground-truth reads, bypassing the scope: raw JDBC on the test's own
+  // connection sees the uncommitted seed and the rewriter does not touch a
+  // statement it never generated. A flush first forces any pending scoped
+  // UPDATE/DELETE to reach the database.
+  private String rawName(String id) {
```
corinnekrych, line 99: this lots of boilerplate code to add to each teant tes, using utils/*Fisture class to crete DTO for CREATE endpoint could be simplier i dont get why we need to insert in DB

.github/skills/add-test/examples/tenant-isolation-templates.md
```diff
@@ -1,259 +1,351 @@
...
+          .andReturn().getResponse().getContentAsString();
+  assertTrue(response.contains(rowA), "A's row must appear in A's search results");
+  assertFalse(response.contains(rowB), "B's row must not appear in A's search results");
+}
 
-  String entityId = JsonPath.read(createResponse, "{entity_id_json_path}");
+@Test
+@DisplayName("via the X-Tenant-Ids header (no path tenant): search returns A's row and not B's")
+void searchViaHeaderReturnsOnlyA() throws Exception {
+  // While the v1 @Filter is still on the entity, this case fails (v1 falls back
```
corinnekrych, line 171: In the DoD of each v2 api ticket we have "remove the table's v1 @Filter and add it to openaev.tenant.active-tables in the same change (avoid the window where v1 and v2 contradict)" 

no need to disabled it.

corinnekrych: ok got it it's temporay step

.github/skills/add-test/examples/tenant-isolation-templates.md
laugiov: The native seed is a v2 constraint, not a style choice: two API creates seed two tenants and set the scope twice in one transaction, the second create throws, verified directly on lessons_templates: IllegalStateException: Tenant scope … is already set for this transaction … tried to change it. Native inserts set no scope, so both rows land cleanly and a read test doesn't depend on the create endpoint. 
On boilerplate: agreed, the create-test input can use the entity's *Fixture, and the generic JDBC plumbing can move to a shared base; only the seed stays native.

.github/skills/add-test/examples/tenant-isolation-templates.md
laugiov: see my reply above on the native-seed constraint + Fixture point

corinnekrych (approved):

PR #6599, opened 2026-07-08
fix(injectors): manage base url on command and cleanup command at runtime (#6575)

### LINKED ISSUE #6575: fix(injectors): base url is not updated in implant commands when changed
## Description

Changing the base URL environment variable once OpenAEV has been started once will cause implants failure as they won't be able to download injects. They will still try to download from the originally set base url instead of the new one.

## Environment

1. OS: all
2. Version: all
3. Other environment details:

## Reproducible steps

1. Deploy an openaev instance with a base URL, any one. (I suggest using ngrok to get a temporary public URL) Launch the openaev backend.
2. Once successfull, shut it down (and shutdown ngrok if it was used) and change the base url to something else (like localhost) and relaunch OpenAEV's backend
3. Install an agent and try to launch an attack on this agent

## Expected output

The attack should go through.

## Actual output

The implant fails because it's trying to download injects/payloads from the previous URL which isn't correct any more.

## Additional information

The core issue stems from the "injectors" table, more specifically the value of columns "injector_executor_commands" and "injector_executor_clean_commands" for the OpenAEV agent row. These values are baked on the first startup of the backend and contain direct URL and tokens for implants to reach the backend. They are never updated after that s if the baseurl changes, the implants will still target the old base url.

Two main choices here: 
- compute the commands with the correct URL and tokens everytime the value is read (cleanest, but longest)
- recompute the commands with the correct URLs and token at each startup.
There may be other solutions I have not seen, of course.

gabriel-peze:
### Proposed changes

* Modify executor command management to add the base url define at runtime

### Testing Instructions

1. Run the platform as usual, injectors should not have the base url hard coded in injectors table anymore (new ones and existing ones)
2. Every injects should run as usual
3. Modify your base url, and injects should still works

Feature tests can be done in local with the OAEV agent, and no regression tests can be done on the feature-branch, executors are deployed there (please don't run recurrent scenarios or heavy scenario).

### Related issues

* Closes #6575


openaev-api/src/main/java/io/openaev/integration/impl/injectors/openaev/OpenaevImplantCommandBuilder.java
```diff
@@ -34,7 +34,7 @@ record CommandVars(
     CommandVars(OpenAEVConfig cfg) {
       this(
           "token=\"#{token}\"",
-          "server=\"" + cfg.getBaseUrlForAgent() + "\"",
+          "server=\"#{baseUrl}\"",
           "max_size=\"" + cfg.getLogsMaxSize() + "\"",
```
impolitepanda: I think we should extend the scope of this PR to all cfg set params in this command. Otherwise, we take the chance of the same issue occuring down the line with someone changing the logs max size, unsecured certificate and with proxy env var but them being already baked in the command and thus staying unchanged.

openaev-api/src/main/java/io/openaev/executors/ExecutorHelper.java
```diff
@@ -34,14 +34,22 @@ public static String replaceArgs(
...
+      String baseUrl,
+      String maxSize,
+      String unsecuredCertificate,
+      String withProxy) {
     if (platformType == null
         || command == null
         || injectId == null
         || agentId == null
-        || tenantId == null) {
+        || tenantId == null
```
impolitepanda, line 46: I think you're missing the token null check

impolitepanda (approved):

openaev-api/src/main/java/io/openaev/executors/ExecutorHelper.java
RomuDeuxfois: @copilot can you handle this comment ?

Issue #6401, opened 2026-06-26
feat(multitenancy): API isolation v2 lessons template


corinnekrych:
**Epic**: GH issue epic generated by script: [feat(multitenancy): epic for refactor for tenant scoped API](https://github.com/OpenAEV-Platform/openaev/issues/6393)

---

## Context

`LessonsTemplateApi` is mapped from codebase analysis to tenant table(s): `lessons_templates`.
Current state: activation = activable now, isolation mechanism = v1, test coverage = no isolation nested class.

## Definition of Done

- [ ] API behavior is aligned with v2 tenant-scoped isolation expectations
- [ ] Existing v1-only assumptions are removed where required for this endpoint
- [ ] Create tenant isolation integration tests for this API (`TenantIsolation` nested class or dedicated isolation test file)
- [ ] Add/validate tenant A vs tenant B isolation and out-of-rights access behavior
- [ ] Wire `TenantWriteScopeResolver` on the create/import endpoints (attribution: row gets the scope's tenant; out-of-scope supplied tenant → 400; multi-tenant scope without selector → 400)
- [ ] Coordinated go-live: remove the table's v1 `@Filter` and add it to `openaev.tenant.active-tables` in the same change (avoid the window where v1 and v2 contradict)

PR #6621, opened 2026-07-08
chore(multitenancy): v2 api isolation lesson template (#6401)

corinnekrych:
### Proposed changes

*
*

### Testing Instructions

1. Step-by-step how to test

### Related issues

* Closes #ISSUE-NUMBER
*

### Checklist

### Further comments


.github/skills/activate-tenant-table/SKILL.md
```diff
@@ -376,32 +377,94 @@ Rebase right before go-live and re-read the merged allowlist line; a bad
...
+
+**Classification and action:**
+
+| Finding | On `{table}`? | Action |
+|---------|---------------|--------|
+| `TenantContext.getCurrentTenant()` used to look up `{Entity}` | Yes | **Remove** — the inspector scopes the query |
+| `TenantContext.getCurrentTenant()` used to look up a *different* entity | No | **Report only** — out of scope for this activation |
+| `findByIdAndTenantId` on `{EntityRepository}` | Yes | **Replace with `findById`** — inspector handles scoping, explicit tenant filter blocks multi-tenant reads |
+| `findByIdAndTenantId` on a *different* repository | No | **Report only** — that table is still on v1 |
+| `TenantBaseListener` on `{Entity}` | Yes | **Remove from `@EntityListeners`** — write attribution is now explicit |
```
laugiov:
```suggestion
| `TenantBaseListener` on `{Entity}` | Yes | **Remove only if every write path attributes the tenant explicitly** (resolver / `setTenant`); otherwise **keep** — a listener-dependent path (e.g. an importer that `save()`s without `setTenant`) would write a NULL tenant |
```
I verified this on kill_chain_phases: its import (V1_DataImporter.createKillChainPhase) saves without setTenant, so removing the listener writes a NULL tenant. 
For lessons_templates it's safe (only createLessonsTemplate writes, and it attributes explicitly). 
Note the file currently contradicts itself, line 312 says "Leave TenantBaseListener" while the go-live step (line 349) and this row say "remove"; worth aligning all three to the conditional rule.

.github/skills/activate-tenant-table/SKILL.md
```diff
@@ -376,32 +377,94 @@ Rebase right before go-live and re-read the merged allowlist line; a bad
...
 catches, do not rely on it alone.
 
-### Phase 7 — Full regression pass
+### Phase 7 — v1 remnant audit
+
+Before running the regression suite, audit the full call stack of every public
+method on `{Api}` (and on the other APIs from Phase 5) for v1 isolation
+patterns that conflict with or duplicate the v2 mechanism.
+
+**What to search for:**
```
laugiov, line 410: Different concern from v1 remnants, but same activation impact and worth a companion check (here or in the inventory phase): adding a TxCtx parameter to an endpoint breaks tests a repository grep misses 

- a standaloneSetup / @WebMvcTest MockMvc test hitting the URL returns 500 (No primary or single unique constructor found for interface TxCtx), that setup has no TxCtxArgumentResolver;
- a test calling the controller method directly in Java breaks at compile time.

I hit both (OpenCTIApiTest, KillChainPhaseApiTest). Scan:

- grep -rln "standaloneSetup" openaev-api/src/test/java | xargs grep -l "{Api}\|/{entities}"
- grep -rn "{Api}" openaev-api/src/test/java   # direct calls

Fix: register the resolver on the standalone builder (.setCustomArgumentResolvers(new TxCtxTestArgumentResolver(...))) or move to the full-context IntegrationTest base; add the TxCtx arg to direct calls.

openaev-api/src/test/java/io/openaev/rest/lessons_template/LessonsTemplateHttpIsolationTest.java
```diff
@@ -0,0 +1,255 @@
...
+            .andExpect(status().isOk())
+            .andReturn()
+            .getResponse()
+            .getContentAsString();
+    assertTrue(response.contains(templateA), "A's template must appear in A's search results");
+    assertFalse(response.contains(templateB), "B's template must not appear in A's search results");
+  }
+
+  @Test
+  @Disabled(
```
laugiov: This case is disabled because the X-Tenant-Ids header selector returns an empty set. 
On the import_mappers pilot this exact case was re-enabled at go-live and passes, so it's worth confirming, before merge, whether the header selector is genuinely broken for lessons_templates or the test setup differs from the pilot's. 
Any idea about the root cause ?

laugiov: Update: I couldn't reproduce the empty result , it passes both in isolation and within the full LessonsTemplateHttpIsolationTest on a clean DB. 
So it doesn't look like the header selector is broken; more likely a stale local DB or cross-class test state when you saw it empty. 
Could you try re-running it on a fresh DB? 
And if it only fails inside the full API suite, that points to a test-isolation issue rather than a product bug.

corinnekrych: fixed a5b3ccaa5

I think AI miss the removal of @Disabled. we should remove thsoe step https://github.com/OpenAEV-Platform/openaev/blob/4fd373f9088f0599597675e66833b42195df5fe4/.github/skills/activate-tenant-table/SKILL.md#L248-L252 and keep it simple no temporary @disable

.github/skills/activate-tenant-table/SKILL.md
corinnekrych: fixed with 776684eb2
I've added your comment

we can refine skills further in next PR if needed

laugiov (approved):

PR #6624, opened 2026-07-08
feat(executor): add Microsoft Defender for Endpoint executor (#6545)

### LINKED ISSUE #6545: feat(executor): add Microsoft Defender for Endpoint executor
Add a new MDE (Microsoft Defender for Endpoint) executor following the same pattern as CrowdStrike, SentinelOne, Tanium and Palo Alto Cortex executors. Uses MDE Live Response API to run scripts on endpoints.

Seb-MIGUEL:
## Goal

Add a **Microsoft Defender for Endpoint (MDE)** executor so OpenAEV can run threat arsenal actions on endpoints through the MDE **Live Response** API, reusing the already-deployed Defender sensor (no extra agent to install).

Closes #6545.

## What it does

- New EE executor (`io.openaev.executors.mde` + `io.openaev.integration.impl.executors.mde`): OAuth2 client-credentials auth, device/device-group sync, and per-machine Live Response dispatch.
- On **Windows**, the implant is launched from a **detached SYSTEM scheduled task** that **self-deletes** after execution, so it survives the Live Response session teardown (same rationale as Palo Alto Cortex).
- Per-inject **self-cleanup** of stale implant/payload directories (piggybacks on the inject's own session) instead of a periodic all-agents garbage collector.
- Catalog-first configuration (Integrations > Executors UI): Azure tenant ID, client ID/secret, optional device group, and the Live Response script names.

## Agent identity (aligned with #6526)

Consistent with the CrowdStrike/SentinelOne rework in #6526:
- `--agent-id` embeds the **OpenAEV agent UUID** (`agent.getId()`), so the execution callback resolves the agent via `findById`.
- The **MDE Device ID** (`agent.getExternalReference()`) is used **only** to target the Live Response API — the agent primary key is never set to the external reference.

## Documentation

- `docs/.../ecosystem/executors.md`: full MDE deployment section (Azure app permissions `Machine.Read.All` + `Machine.LiveResponse.All`, Live Response setup incl. **unsigned script execution**, script upload, config fields, checks, and MDE-specific troubleshooting).
- Downloadable subprocessor scripts: `mde_subprocessor_windows.ps1`, `mde_subprocessor_unix.sh`.
- Cross-references added in `enterprise.md` and `assets.md`.
- Screenshots to be added.

## Testing

- Full reactor build passes (`mvn install -DskipTests`).
- `MdeExecutorServiceTest` (Mockito unit test) covers sync + per-OS Live Response dispatch.
- Validated end-to-end against a real MDE tenant: inject dispatch → implant download → detached scheduled task → callback → execution traces → `EXECUTED`.

## Notes for reviewers

- MDE dispatches **per agent**; the 5s pagination throttle applies within an action's agent list. Fine for typical inject sizes; very large fan-outs could approach MDE Live Response rate limits (batch size configurable).
- MDE is a new/unreleased executor, so no data migration is required for existing agents.


Seb-MIGUEL:
Thanks for the review 🙏 All points addressed in `773d5b276` (rebased on latest `main`, all commits signed):

**Blocking**
- `MdeExecutorConfig` `@NotBlank` on `Integer` fields → now `@NotNull` + `@Min(1)` (NotBlank only supports CharSequence and would fail Hibernate Validator at runtime).
- `InjectorService` long single-line call / `keysOf`/`valuesOf` ordering → **removed entirely**: main's #6523 (composite key on the Java side) supersedes the native-query workaround, so this now uses a plain `setExecutorCommands(...)` + `save(injector)` (mirrors `updateExistingExternalInjector`). No more native hstore update.

**Non-blocking**
- `MdeExecutorServiceTest` `Thread.sleep` → `verify(client, timeout(2000))` for deterministic async verification.
- `MdeDeviceFixture` lastSeen now formatted in UTC (`ZoneOffset.UTC`).

Also aligned the executor with #6526/#6523: `--agent-id` uses the OpenAEV agent UUID (callback resolves via `findById`), the MDE Device ID is used only to target the Live Response API, and tenant access uses `executor.getTenantId()`.

RomuDeuxfois (approved):

openaev-api/src/main/java/io/openaev/service/InjectorService.java
```diff
@@ -501,6 +505,11 @@ private void updateExistingBuiltinInjector(
     injector.setExternal(false);
     injector.setCustomContracts(Boolean.TRUE.equals(isCustomizable));
     injector.setPayloads(Boolean.TRUE.equals(isPayloads));
+    // Refresh config-derived executor commands (hstore) so existing deployments pick up new
+    // per-executor launch strategies (e.g. the MDE detached scheduled task) instead of keeping the
+    // values baked at first registration — mirrors updateExistingExternalInjector.
```
RomuDeuxfois: **todo**: remove this unused comment

openaev-api/src/main/java/io/openaev/migration/V6_20260703000000000__Fix_MDE_device_group_not_required.java
```diff
@@ -0,0 +1,26 @@
+package io.openaev.migration;
+
+import java.sql.Statement;
+import org.flywaydb.core.api.migration.BaseJavaMigration;
+import org.flywaydb.core.api.migration.Context;
+import org.springframework.stereotype.Component;
+
+@Component
+public class V6_20260703000000000__Fix_MDE_device_group_not_required extends BaseJavaMigration {
```
RomuDeuxfois: **todo**:Since MDE Executor is new, I am not sur this migration is needed

openaev-api/src/main/java/io/openaev/executors/mde/client/MdeExecutorClient.java
```diff
@@ -0,0 +1,368 @@
...
+      synchronized (this) {
+        if (Instant.now()
+            .isAfter(tokenExpiresAt.minusSeconds(AUTH_REFRESH_BEFORE_EXPIRY_SECONDS))) {
+          authenticate();
+        }
+      }
+    }
+  }
+
+  private void authenticate() throws IOException {
```
RomuDeuxfois, line 340: **todo**: Could be better to have a MdeAuthenticationClient for this part.

openaev-api/src/main/java/io/openaev/executors/mde/config/MdeExecutorConfig.java
```diff
@@ -0,0 +1,139 @@
...
+import jakarta.validation.constraints.NotBlank;
+import jakarta.validation.constraints.NotNull;
+import lombok.Getter;
+import lombok.Setter;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.stereotype.Component;
+
+@Setter
+@Component
+@ConfigurationProperties(prefix = "executor.mde")
```
RomuDeuxfois, line 20: **question**: Since everything go to the catalog featute, do we want to expose config properties and let user create executor through this ?

openaev-api/src/main/java/io/openaev/executors/mde/config/MdeExecutorConfig.java
```diff
@@ -0,0 +1,139 @@
...
+  public static final int DEFAULT_API_REGISTER_INTERVAL = 1200;
+  public static final int DEFAULT_STALE_PENDING_THRESHOLD_MINUTES = 30;
+
+  @IntegrationConfigKey(
+      key = "EXECUTOR_ID",
+      description = "ID of the builtin Microsoft Defender for Endpoint executor",
+      isRequired = true)
+  @Getter
+  @NotBlank
+  private String id = MDE_EXECUTOR_DEFAULT_ID;
```
RomuDeuxfois, line 35: **todo**: We should avoir static UUID if it's not necessary

openaev-api/src/main/java/io/openaev/executors/mde/model/MdeAuthentication.java
```diff
@@ -0,0 +1,16 @@
+package io.openaev.executors.mde.model;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import lombok.Getter;
+
+/**
+ * Bearer token response from Azure AD. Using {@code @Getter} only to prevent {@code toString()}
+ * from exposing the token.
+ */
```
RomuDeuxfois: **todo**: remove unused comment

openaev-api/src/main/java/io/openaev/executors/mde/service/MdeExecutorContextService.java
```diff
@@ -0,0 +1,229 @@
...
+
+    Injector injector = inject.getInjector();
+    if (injector == null) {
+      injector =
+          inject
+              .getInjectorContract()
+              .map(InjectorContract::getFirstInjector)
+              .orElseThrow(
+                  () -> new UnsupportedOperationException("Inject does not have a contract"));
+    }
```
RomuDeuxfois, line 74: **todo**: Rmove this part, it's deprecated

openaev-api/src/main/java/io/openaev/integration/impl/executors/mde/MdeExecutorIntegrationFactory.java
```diff
@@ -0,0 +1,146 @@
...
+  protected final String getClassName() {
+    return MdeExecutorIntegrationFactory.class.getCanonicalName();
+  }
+
+  @Override
+  protected void runMigrations() throws Exception {
+    // MDE is a catalog-first executor: it has no legacy application.properties configuration to
+    // migrate, so there is nothing to do here (same pattern as PaloAltoCortex). Running a
+    // ConfigurationMigration would spawn a spurious empty PROPERTIES_MIGRATION instance alongside
+    // the one deployed from the catalog UI.
```
RomuDeuxfois: **todo**: Remove this comment

openaev-api/src/main/java/io/openaev/integration/impl/executors/mde/MdeExecutorIntegration.java
```diff
@@ -0,0 +1,165 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ScheduledFuture;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskScheduler;
+
+@Slf4j
+public class MdeExecutorIntegration extends Integration {
+
+  public static final String MDE_EXECUTOR_DEFAULT_ID = "f9c4a0e1-3b7d-4f8a-9e2c-1d5b6a0f8e3c";
```
RomuDeuxfois, line 39: **todo**: please avoid static UUID if is not needed

openaev-api/src/main/java/io/openaev/executors/mde/service/MdeExecutorContextService.java
```diff
@@ -0,0 +1,229 @@
...
+              platform, command, inject.getId(), agent.getId(), inject.getTenant().getId(), token);
+      command =
+          command.replaceFirst(
+              "\\$?x=.+location=.+;\\[Environment]::CurrentDirectory",
+              Matcher.quoteReplacement(implantLocation));
+      // Self-clean stale implant/payload dirs (>24h) at the start of every inject. This piggybacks
+      // on the inject's own Live Response session, so cleanup happens only on agents that actually
+      // run injects — replacing the periodic all-agents garbage collector that flooded the MDE API
+      // and starved inject dispatch.
+      command = WINDOWS_CLEAN_PAYLOADS_COMMAND + ";" + command;
```
RomuDeuxfois, line 177: **question**: This should be handled by a garbage collector service no ?

docs/docs/deployment/assets/mde_subprocessor_unix.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
```
RomuDeuxfois, line 1: **todo**: add the same sanity check than ps1 script.

openaev-api/src/test/java/io/openaev/executors/mde/client/MdeExecutorClientTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+import io.openaev.executors.mde.model.MdeMachineAction;
+import io.openaev.utils.fixtures.MdeMachineActionFixture;
+import java.lang.reflect.Method;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+class MdeExecutorClientTest {
```
RomuDeuxfois, line 14: **question**: uefull to test a client class ?

openaev-api/src/main/java/io/openaev/executors/mde/config/MdeExecutorConfig.java
```diff
@@ -0,0 +1,119 @@
...
+  private String unixScriptName = "openaev-subprocessor.sh";
+
+  @IntegrationConfigKey(
+      key = "EXECUTOR_MDE_API_BATCH_EXECUTION_ACTION_PAGINATION",
+      jsonType = CONNECTOR_CONFIGURATION_TYPE.INTEGER,
+      description =
+          "Number of machines sent per 5 seconds to MDE Live Response API to execute a payload (MDE rate limit is stricter than CrowdStrike)")
+  @Getter
+  @NotBlank
+  private Integer apiBatchExecutionActionPagination = 10;
```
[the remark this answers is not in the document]
SamuelHassine: ﻿Addressed: these Integer config fields now use @NotNull + @Min(1) instead of @NotBlank (NotBlank only supports CharSequence and would fail Hibernate Validator at runtime). See MdeExecutorConfig.

openaev-api/src/main/java/io/openaev/executors/mde/config/MdeExecutorConfig.java
```diff
@@ -0,0 +1,119 @@
...
+
+  @IntegrationConfigKey(
+      key = "EXECUTOR_MDE_API_REGISTER_INTERVAL",
+      jsonType = CONNECTOR_CONFIGURATION_TYPE.INTEGER,
+      description =
+          "MDE API interval to register/update device groups and agents in OpenAEV (in seconds)")
+  @Getter
+  @NotBlank
+  private Integer apiRegisterInterval = 1200;
```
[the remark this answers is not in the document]
SamuelHassine: ﻿Addressed: this Integer field now uses @NotNull + @Min(1) instead of @NotBlank, for the same reason (NotBlank only supports CharSequence).

openaev-api/src/main/java/io/openaev/service/InjectorService.java
```diff
@@ -492,6 +496,11 @@ private void updateExistingBuiltinInjector(
         Boolean.TRUE.equals(isCustomizable),
         Boolean.TRUE.equals(isPayloads));
 
+    // Refresh config-derived executor commands (hstore) so existing deployments pick up command
+    // changes (e.g. new per-executor launch strategies) instead of keeping the values baked at
+    // first registration. Same tenant-scoped native UPDATE for the same composite-PK reason.
+    refreshBuiltinExecutorCommands(injector.getId(), tenantId, executorCommands, executorClearCommands);
```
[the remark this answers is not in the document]
SamuelHassine: ﻿Addressed: the native hstore update was removed entirely. updateExistingBuiltinInjector now uses setExecutorCommands(...) + save(injector), so the flagged long single-line call no longer exists.

openaev-api/src/main/java/io/openaev/service/InjectorService.java
```diff
@@ -596,6 +605,35 @@ private Injector createNewBuiltinInjector(
...
+    }
+    return map.entrySet().stream().map(Map.Entry::getKey).toArray(String[]::new);
+  }
+
+  private static String[] valuesOf(Map<String, String> map) {
+    if (map == null || map.isEmpty()) {
+      return new String[0];
+    }
+    return map.entrySet().stream().map(Map.Entry::getValue).toArray(String[]::new);
+  }
```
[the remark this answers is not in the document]
SamuelHassine: ﻿Addressed: keysOf()/valuesOf() were removed together with the native hstore workaround; the code now relies on setExecutorCommands(...) + save(injector), so the ordering concern no longer applies.

openaev-api/src/test/java/io/openaev/executors/mde/service/MdeExecutorServiceTest.java
```diff
@@ -0,0 +1,185 @@
...
+    when(executorService.manageWithoutPlatformAgents(agents, injectStatus)).thenReturn(agents);
+
+    // Act
+    mdeExecutorContextService.launchBatchExecutorSubprocess(
+        inject, new HashSet<>(agents), injectStatus, "token");
+    // Scheduled execution: wait for the scheduled task to run
+    Thread.sleep(1000);
+
+    // Assert
+    verify(client).executeAction(any(), eq("openaev-subprocessor.ps1"), any());
```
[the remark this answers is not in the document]
SamuelHassine: ﻿Addressed: replaced Thread.sleep(...) with verify(client, timeout(2000)) for deterministic async verification.

openaev-api/src/test/java/io/openaev/executors/mde/service/MdeExecutorServiceTest.java
```diff
@@ -0,0 +1,185 @@
...
+    InjectStatus injectStatus = InjectStatusFixture.createPendingInjectStatus();
+    when(executorService.manageWithoutPlatformAgents(agents, injectStatus)).thenReturn(agents);
+
+    // Act
+    mdeExecutorContextService.launchBatchExecutorSubprocess(
+        inject, new HashSet<>(agents), injectStatus, "token");
+    Thread.sleep(1000);
+
+    // Assert
+    verify(client).executeAction(any(), any(), any());
```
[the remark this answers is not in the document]
SamuelHassine: ﻿Addressed: this test now uses verify(client, timeout(2000)) instead of Thread.sleep(...).

openaev-api/src/test/java/io/openaev/utils/fixtures/MdeDeviceFixture.java
```diff
@@ -0,0 +1,35 @@
...
+    device.setComputerDnsName("mde-test-host.example.com");
+    device.setOsPlatform("Windows10");
+    device.setOsArchitecture("64-bit");
+    device.setLastIpAddress("192.168.1.10");
+    device.setLastExternalIpAddress("203.0.113.10");
+    device.setRbacGroupId("42");
+    device.setRbacGroupName("Test Device Group");
+    device.setHealthStatus("Active");
+    DateTimeFormatter formatter =
+        DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss'Z'").withZone(ZoneId.systemDefault());
```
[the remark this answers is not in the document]
SamuelHassine: ﻿Addressed: the fixture formatter now uses ZoneOffset.UTC explicitly (see MdeDeviceFixture.MDE_LAST_SEEN_FORMATTER), so the trailing Z is consistent with the actual value.

openaev-api/src/main/java/io/openaev/service/InjectorService.java
SamuelHassine: ﻿Done: trimmed the comment down to a concise two-line note about why executor commands are refreshed on existing deployments.

openaev-api/src/main/java/io/openaev/migration/V6_20260703000000000__Fix_MDE_device_group_not_required.java
SamuelHassine: ﻿Agreed - removed the migration. MDE is a new/unreleased executor and its catalog config already declares EXECUTOR_MDE_DEVICE_GROUP as optional (isRequired = false), so a fresh catalog insert is already correct and no data migration is needed.

openaev-api/src/main/java/io/openaev/executors/mde/client/MdeExecutorClient.java
SamuelHassine: ﻿Kept inline for now. authenticate() is a single private method co-located with the client that consumes the token, which matches the other executor clients (CrowdStrike, SentinelOne, PaloAltoCortex all authenticate inline). Extracting a dedicated MdeAuthenticationClient would add a class and indirection for one token endpoint without a functional benefit. Happy to revisit if we later centralize OAuth2 client-credentials handling across executors.

openaev-api/src/main/java/io/openaev/executors/mde/config/MdeExecutorConfig.java
SamuelHassine: ﻿This is the shared BaseIntegrationConfiguration pattern used by every catalog executor (CrowdStrike, Tanium, SentinelOne, PaloAltoCortex): @ConfigurationProperties(prefix) provides the application.properties binding the base class expects, while @IntegrationConfigKey drives the catalog UI. Configuration is created through the catalog, exactly like the other executors - there is no MDE-specific deviation here.

openaev-api/src/main/java/io/openaev/executors/mde/config/MdeExecutorConfig.java
SamuelHassine: ﻿This mirrors every existing executor: each defines a stable *_EXECUTOR_DEFAULT_ID used as the EXECUTOR_ID config default (see CrowdStrike, Tanium, SentinelOne, PaloAltoCortex, Caldera, OpenAEV). The fixed UUID gives the builtin executor a stable identity across catalog (re)deployments; generating a random id per boot would create duplicate executor rows. Kept for consistency with the established convention.

openaev-api/src/main/java/io/openaev/executors/mde/model/MdeAuthentication.java
SamuelHassine: ﻿Done: replaced the Javadoc block with a single-line comment explaining why this DTO uses @Getter only (so the bearer token is never exposed via toString()).

openaev-api/src/main/java/io/openaev/executors/mde/service/MdeExecutorContextService.java
SamuelHassine: ﻿This is the same legacy-inject fallback that every executor context service uses (CrowdStrike, Tanium, PaloAltoCortex, Caldera and OpenAEV all have the identical inject.getInjector() == null -> injectorContract.getFirstInjector() block), and it is covered by the given_legacyInjectWithoutInjector test. It is not MDE-specific dead code, so I kept it aligned with the others; if the fallback is being retired it should be a single cross-executor change rather than an MDE-only divergence.

openaev-api/src/main/java/io/openaev/integration/impl/executors/mde/MdeExecutorIntegrationFactory.java
SamuelHassine: ﻿Done: trimmed the runMigrations comment to a single line (same spirit as PaloAltoCortex's comment).

openaev-api/src/main/java/io/openaev/integration/impl/executors/mde/MdeExecutorIntegration.java
SamuelHassine: ﻿Same rationale as the config EXECUTOR_ID default: this MDE_EXECUTOR_DEFAULT_ID constant matches the convention of every other executor (CrowdStrike, Tanium, SentinelOne, PaloAltoCortex, Caldera, OpenAEV), giving the builtin executor a stable identity across catalog (re)deployments. Kept for consistency.

openaev-api/src/main/java/io/openaev/executors/mde/service/MdeExecutorContextService.java
SamuelHassine: ﻿Intentional. The earlier design used a periodic all-agents garbage collector (like CrowdStrike's), which flooded the MDE Live Response API - MDE throttles it far more aggressively than CrowdStrike - and starved actual inject dispatch. The per-inject self-clean piggybacks on the Live Response session the inject already opens, so cleanup only runs on machines that actually execute injects and never competes with dispatch for the rate-limit budget. This is documented in the code comment.

docs/docs/deployment/assets/mde_subprocessor_unix.sh
SamuelHassine: ﻿Done: added the same guard as the .ps1 script - the Unix subprocessor now exits 1 with an error message if no Base64 argument is provided, before attempting to decode.

openaev-api/src/test/java/io/openaev/executors/mde/client/MdeExecutorClientTest.java
SamuelHassine: ﻿Kept. MdeExecutorClientTest exercises the stale-pending detection boundary (isStalePendingAction): the age threshold plus null/unparseable-timestamp handling. That is real logic independent of the HTTP layer, it runs fast (no network or token), and it guards a subtle branch the service-level test does not cover.

SamuelHassine:
## Review and fix pass - summary

Ran a full independent senior-reviewer pass over the MDE executor (whole files, not just the diff), cross-checked it against the existing CrowdStrike / Tanium / SentinelOne / PaloAltoCortex executors, addressed every open review thread, resolved the merge conflict with `main`, and drove CI green.

### Merge conflict resolved
- `main` refactored `OpenaevImplantCommandBuilder` to build implant commands from placeholders (`#{baseUrl}`, `#{maxSize}`, ...) resolved in `ExecutorHelper.replaceArgs(...)`, dropping the `OpenAEVConfig` dependency from the builder.
- Reconciled by keeping main's placeholder approach while preserving the MDE detached scheduled-task launch, and wiring `OpenAEVConfig` through `MdeExecutorContextService` so it calls the new 10-arg `replaceArgs(...)` (same as CrowdStrike/PaloAltoCortex). Updated the integration + factory constructors and the unit test accordingly.

### Fixes applied
- Removed migration `V6_..._Fix_MDE_device_group_not_required`: MDE is new/unreleased and the catalog config already declares `EXECUTOR_MDE_DEVICE_GROUP` as optional, so a fresh catalog insert is correct without a data migration.
- Added the missing-argument sanity check to the Unix subprocessor script (matches the PowerShell script: exit 1 with an error if no Base64 arg).
- Trimmed the verbose comments flagged in review (InjectorService, MdeExecutorIntegrationFactory, MdeAuthentication) and removed a non-ASCII ellipsis from an MDE log message.

### Review verdict on the executor internals (no further changes needed)
- Auth: OAuth2 client-credentials, token cached with volatile double-checked refresh 5 min before expiry; client secret is only sent in the token form body and never logged; the token DTO uses `@Getter` only so `toString()` cannot leak it; error bodies are truncated to 500 chars before logging.
- Dispatch: per-agent Live Response is throttled in 5s batches (MDE rate limit is stricter than CrowdStrike), with null/non-positive pagination guarded to the default; stale Pending sessions are cancelled only past a configurable threshold so concurrent injects are never disturbed.
- Status: device active state prefers near real-time Advanced Hunting activity and falls back to sensor health / inventory `lastSeen` when Advanced Hunting is unavailable, with null-safe parsing throughout.

### Review threads
All 19 threads addressed and resolved: the 7 outdated Copilot findings (all already fixed on the branch: `@NotBlank` -> `@NotNull`+`@Min(1)` on Integer config, native hstore removed, `Thread.sleep` -> `verify(timeout)`, UTC fixture) plus the 12 from @RomuDeuxfois - implementing the concrete ones (migration removal, Unix sanity check, comment trims) and justifying the ones that intentionally follow the established cross-executor conventions (static builtin `EXECUTOR_ID`, catalog config binding, inline auth, legacy-inject fallback, per-inject self-clean vs a periodic GC, testing the client's stale-detection helper).

### CI
All required checks green: Backend Compile, API Tests (all shards), E2E (chrome + webkit), E2E Infra (x64 + arm64), Frontend Build/Quality, Spotless, API Types, Migrations Guard, CodeQL, signed-commits, PR title. Codecov (patch + project, all flags) is also green after rebasing coverage on the updated base.

### Notes
- Coverage: `MdeExecutorClient` (HTTP/token plumbing) and the integration/factory wiring remain largely uncovered by unit tests; the service-level behavior (sync, active-status resolution, batch dispatch, null-guards) and the stale-pending helper are covered. Codecov is green, so this is informational only.
- New signed merge commit authored by the reviewer; the original author's commits and attribution are preserved.
- The build was verified via CI: Maven is not available in the review environment, so `mvn clean install` / `spotless:apply` were not run locally (edited Java files were formatted with the pinned google-java-format 1.24.0 and Spotless Check is green in CI).

docs/docs/deployment/ecosystem/executors.md
```diff
@@ -430,6 +431,175 @@ Endpoint on the OpenAEV endpoint page.
 
     You are now ready to leverage your SentinelOne platform to run OpenAEV threat arsenal actions!
 
+---
+<a id="mde-agent"></a>
+
+## Microsoft Defender for Endpoint (MDE) Agent
```
RomuDeuxfois, line 437: **nitpick**: we are starting to have a big file here. Can be good to split it, what do you think ?

RomuDeuxfois (approved):

PR #6639, opened 2026-07-09
chore(multitenancy): v2 api isolation mitigation (#6400)

johanah29:
### Proposed changes

Api isolation v2 for mitigations


corinnekrych (approved):
This cas be tested at API level only. i've added an http vile to test it

Issue #6398, opened 2026-06-26
feat(multitenancy): carrying the tenant context onto background tasks


corinnekrych:
**Epic**: GH issue epic generated by script: [feat(multitenancy): epic for refactor for tenant scoped API](https://github.com/OpenAEV-Platform/openaev/issues/6393)

---

## Context

Tenant-scoped filtering is linked to multiple background paths (jobs, inject execution, Elasticsearch indexing, startup DataPack seeding, event listeners).  
Before broad activation, tenant context must be propagated explicitly onto these non-HTTP flows.

Use `TagRuleApi` / `tag_rules` as the concrete example of a path that requires propagation before safe activation.

## Definition of Done

- [ ] Background paths propagate tenant context explicitly (no implicit thread-local fallback)
- [ ] Unit test verifies propagation behavior on a background component
- [ ] Integration test verifies tenant-correct behavior for one background execution path
- [ ] TagRule-related background path is covered by at least one validation scenario
- [ ] Background transactions open through the tenant-aware primitive (no `@Transactional` on background paths, no thread-local fallback)

laugiov:
⚠️ Superseded on the transaction-mechanism point by the decision comment below: background transactions use a tenant-aware primitive on TransactionTemplate, not @Transactional. Kept for the decision trail.

**Design notes, not a final decision.**

**Propagation mechanism.** The tenant-scope aspect fires only on `@Transactional` methods (it reads the
`TxCtx` from the method args), never on a programmatic `TransactionTemplate`. So:
- Default: `@Transactional` + a `TxCtx` parameter → the scope is set automatically.
- For jobs that legitimately avoid a job-level `@Transactional` (programmatic `TransactionTemplate` /
  parallel workers, e.g. ES indexing, the perf pattern in #6187 that wraps ES/OpenSearch network bulk
  calls), set the scope manually inside the short transaction via a small helper.
- Not "@Transactional vs TransactionTemplate": `@Transactional` + `TxCtx` by default, manual scope-set
  for the perf/parallel cases.

**Global ("see all") scope.** The few genuinely-global jobs (ES indexing) use `TxCtx.allTenants()`.
How to implement it is still open : enumerate all tenant ids, or a typed, code-only, grep-audited
platform token honored by `can_access_tenant`.

**Platform-write, still open.** Writing genuinely-platform rows (`tenant_id` NULL: the platform
[...]

Issue #6393, opened 2026-06-26
🤖[Multi-Tenancy - v2] epic for refactor for tenant scoped API v2


corinnekrych:
Thus main epic drives the migration from v1 tenant isolation (`@Filter` + tenant thread-local behavior) to v2 isolation (`TxCtx` request scope + SQL rewriter + write-scope policy).  

The mechanism is already implemented and validated on the `import_mappers` pilot. 
Remaining work is endpoint-by-endpoint rollout with two constraints:
1. API/controller refactor to v2 scope propagation.
2. Table activation order based on background-path safety (7 activable now, 31 blocked until background context propagation is complete).

laugiov:
Decision update. Two points that shape the blocked half of this epic:

1. Background transactions will NOT use @Transactional. They open through a tenant-aware primitive built on TransactionTemplate, which sets the tenant scope at transaction open. Reason: @Transactional has a silent self-invocation trap (an intra-class call bypasses the Spring proxy: no transaction, no scope).@Transactional(REQUIRES_NEW) alternative was considered and set aside: it keeps the self-invocation trap, so it would need the same compile-time guard without removing the root cause. The current aspect only fires on @Transactional, so the primitive replaces it on the background side; HTTP keeps @Transactional + TxCtx. A compile-time detection of the self-invocation trap is required (see #6391).
2. allTenants() (global jobs, e.g. ES indexing) is an intention resolved, at scope-set time, into an explicit list of active tenant ids tied to the caller. No wildcard ever reaches the database. Enumeration is the starting implementation; switching to a special value later is non-structuring and reversible.
Detailed impact is tracked in #6398 (mechanism), #6391 (compile-time guard), #6462 (ES read side).

Issue #6391, opened 2026-06-26
feat(Multi-tenancy): @RequiresTxScope marker + compile-time rule for tenant-scoped entrypoints


laugiov:
## Context
v2 tenant isolation rides on a `TxCtx` parameter that looks unused in the handler. A refactor could
drop it and silently disable scoping on that endpoint. #6255 added an immediate floor: an ArchUnit
test (`TenantScopedEntrypointsTxCtxArchTest`) checking a hardcoded list of entrypoints declares
`TxCtx`. That list is fragile: renaming a method or adding a new scoped endpoint silently escapes the
guard.

This is the "long run" guardrail Corinne proposed in her #6255 review (thread 1):
> compile-time rule that any @transactional method on active-table paths must declare TxCtx, plus an
> explicit marker annotation (e.g. @RequiresTxScope) and tests that fail if scope is missing.

## Goal
Replace the hardcoded ArchUnit list with a self-maintaining, derive-from-code rule.

## Proposed approach
- A marker annotation `@RequiresTxScope` on the tenant-scoped entrypoints.
- A compile-time rule in `openaev-annotation-processor` (alongside `EndpointTransactionalRule`,
  `NoRepositoryRule`, ...) that fails the build if a `@RequiresTxScope` method does not declare a
  `TxCtx` parameter.
- Once in place, retire the hardcoded list in `TenantScopedEntrypointsTxCtxArchTest`.

## Reference
- Corinne's review on #6255 (thread 1).
- The interim floor: `TenantScopedEntrypointsTxCtxArchTest`, `TenantScopeFailClosedTest`.
- The processor framework: `AbstractTransactionalRule`, `EndpointTransactionalRule`.

## Note
The runtime safety net already holds (dropping `TxCtx` is fail-closed: 0 rows, proven by
`TenantScopeFailClosedTest`). This issue is about catching the mistake at build time rather than
discovering a broken feature at runtime.

## Scope extension (background transaction decision)
Background transactions will open through a tenant-aware primitive built on `TransactionTemplate` (not `@Transactional`; see #6398). This issue therefore also covers, beyond the `TxCtx`-parameter rule:
- the `TxCtx`-parameter rule stays an **annotation-processor** rule (it works on declarations, which is enough);
[...]

PR #6642, opened 2026-07-09
feat(multi-tenancy): background transaction primitive, allTenants resolution, frozen build-time guards and first background-ready table activation (#6398)

laugiov:
Part of #6398. References #6391 (build-time guard, widened scope) and the epic #6393.

## What this PR materialises

Three settled decisions of the background-scope work, on running code, plus one real table
activation that exercises the whole chain end to end:

**1. Background transactions open through a tenant-aware primitive, not `@Transactional`.**
`TenantScopedTransaction` (built on `TransactionTemplate`) opens the transaction and sets the tenant
scope in one explicit move. `@Transactional` has a silent self-invocation trap: an intra-class call
bypasses the Spring proxy, so neither the transaction nor the scope is set. The
`@Transactional(REQUIRES_NEW)` alternative was considered and set aside: it keeps the trap. HTTP
keeps `@Transactional` + `TxCtx` through the existing aspect, untouched.

Semantics, proven by integration tests on the active pilot table (`import_mappers`):
- the scope is transaction-local: visible inside, gone after;
- fail-closed: a raw `TransactionTemplate` with no scope reads 0 rows, the primitive reads them;
- `execute()` refuses to open inside an active transaction (a primitive call is an explicit scope
  boundary), and `executeNew()` refuses to run OUTSIDE one: nesting is deliberate, never an escape
  hatch around the guard (Runnable overloads covered too);
- a nested transaction never inherits the outer scope; `executeNew` re-sets it;
- a joined `@Transactional` service keeps the primitive's scope, the aspect stays inert without a
  `TxCtx` argument;
- an exception rolls back the writes and the next transaction carries no scope;
- `null` and `Missing` scopes are refused: no background transaction without a real scope.

**The per-tenant loop idiom, `forEachTenant`.** The poisoning finding (below) makes the naive
pattern (outer `allTenants()` + nested `executeNew` per tenant) a trap. The structural answer is
built and proven: `forEachTenant(Consumer<TxCtx>)` runs the work once per active tenant, each in its
OWN top-level transaction. A failing tenant is rolled back alone and cannot poison the others
(proven by a test where the failing tenant's write is verified rolled back while the others
committed); failures are collected and rethrown as one aggregate carrying the original causes,
never swallowed. The tenant list is a snapshot taken at loop start (a stable loop domain), unlike
`execute(allTenants(), …)` which re-resolves per transaction (freshness). Jobs cannot enumerate
tenants themselves: the resolver is package-private, the loop is the door. `forEachTenant` refuses
to run inside an active transaction, upfront, with the reason named. Known ceilings, accepted for
now: the loop is serial and single-threaded (no batching, parallelism or per-tenant timeout), and on
partial failure the job is marked failed even when most tenants succeeded (the per-tenant `log.warn`
and the aggregate's suppressed causes carry the detail; a summary metric is a follow-up).
`forEachTenant` has no production caller yet - its first is the migration-engine conversion,
tracked as the direct follow-up.

**2. `allTenants()` is an intention, resolved into an explicit list. Never a wildcard.**
`TxCtx.allTenants()` cannot serialize itself (`toGuc()` throws), so no wildcard can ever reach
`set_config` / `can_access_tenant`; misuse through the aspect fails loudly (tested). The resolver is
package-private, so only same-package code - in practice the primitive - resolves the intention,
inside the opening transaction, into the explicit list of active tenants (a convention-and-review
barrier, not a hard one). A long job therefore re-resolves at every short transaction and naturally
sees tenants created in between (tested). Write attribution refuses it: a row belongs to exactly one
tenant (tested).

**The poisoning rule, discovered and proven by this PR.** Inside a resolved `allTenants()` scope, a
joined `@Transactional` service carrying a narrower `TxCtx` trips the aspect's nesting guard - and
the trip POISONS the joined transaction (rollback-only). This is general: ANY runtime exception from
a joined service poisons the background transaction; catching it and carrying on still dies at
commit (`UnexpectedRollbackException`), proven by a dedicated test. Recovery belongs around an
`executeNew` boundary (REQUIRES_NEW isolates the failure) or around a `forEachTenant` iteration
(each tenant is its own transaction), never inside the same transaction. Both the trap and the
working idioms are tested.

Measured: a 1000-id scope round-trips in 24-49 ms across runs on the pilot table. The GUC mechanics
hold; the per-row cost of `can_access_tenant` on a realistically sized table is NOT measured here
and remains the perf question to settle (existing T2 measurements are the baseline to extend).

Showcase conversion: `UrlAccessTokenPurgeJob` now opens through the primitive with the `allTenants`
intention (a bulk delete by predicate: no per-row attribution involved). Its table carries no
tenant, so no behavior change on any real path (one theoretical edge: with zero active tenants in
the registry, the purge now fails loudly instead of running; the default tenant makes this
unreachable in practice).

**3. Build-time guards, frozen-baseline mode (widened #6391).**
Four ArchUnit rules, wrapped in `FreezingArchRule`:
- no intra-class call to a `@Transactional` method (the self-invocation trap), codebase-wide;
- no `@Transactional` on job classes (`io.openaev.scheduler.jobs..`);
- no raw transaction plumbing (`TransactionTemplate` / `PlatformTransactionManager`) in job classes;
- no dependency on the primitive from HTTP-side classes - matched by package (`io.openaev.rest..`,
  `io.openaev.api..`) AND by controller annotation, because real controllers live outside those
  packages (`SchemaApi`, `FullTextSearchApi`, `OpenCTIApi`; a test pins that all three are matched).

A fifth, source-level tripwire pins the scope channel itself: the inspector does not guard
`set_config` (a `SELECT set_config(...)` touches no tenant table), so any native query could
silently widen the scope; until the inspector learns to reject unsanctioned writes to
`app.current_tenants` (arbitration item, which must also cover the `SET LOCAL` idiom), a test walks
every `openaev-*` production module (roots derived, a new module cannot escape the scan) and asserts
that exactly two classes - the aspect and the primitive - contain that call.

The committed freeze store records the existing violations; only a NEW violation breaks the build,
so this lands green. The store is locked by configuration (`allowStoreCreation=false`,
`allowStoreUpdate=false`, pinned by a test; a manual before/after checksum during development
confirmed a run leaves the store byte-identical). A fixture per rule proves each guard fires,
evaluated through `hasViolation()` so a pass can only come from a real violation.

**Deliberate ratchet consequence to be aware of:** with the store locked, FIXING a frozen violation
makes the arch test fail with a store-update refusal (`StoreUpdateFailedException`, verified against
ArchUnit 1.4.2 internals: solved violations trigger a store write) until the store is deliberately
regenerated (one-off `-Darchunit.freeze...` overrides, recipe in `archunit.properties`). In other
words, every conversion PR must re-freeze the store in the same PR. That is intentional (a shrink is
an audited, reviewed event, and a store diff is where a new violation could hide) but it is a
workflow decision: see the arbitration list.

**4. CAPSTONE - `cwes` activated on v2, through the revised activation skill.**
The `activate-tenant-table` skill (in-repo) was revised for the background era: background writers
are no longer an automatic stop; they convert through the primitive (a new Phase 5b covers the scope
choice - `forEachTenant` / `forTenant` / `allTenants` -, write attribution, the poisoning idiom, the
native-vs-raw-SQL distinction, and the frozen-store refresh procedure). Then the skill was exercised
for real: `cwes` (reached only through a vulnerability's `@ManyToMany`) is now fully on v2:
- `TxCtx` wired on the whole Vulnerability/CVE surface (12 handlers, registered in the entrypoints
  arch test); the two mappers that materialize the association are behind them (verified inventory);
- reads scoped by the inspector: cross-tenant reads see empty CWEs (fail-closed, proven end to end
  through the real endpoints);
- writes attributed EXPLICITLY: a payload carrying CWEs resolves its single write tenant first
  (`TenantWriteScopeResolver`) and looks the CWE up by the per-tenant unique key
  (`cwe_external_id, tenant_id`). An ambiguous multi-tenant scope is refused with 400 - without
  this, the preset CWEs seeded per tenant make a bare external-id lookup crash (or silently link
  another tenant's row) as soon as a caller is authorized on two tenants; proven by a dedicated
  red-first test;
- the v1 `@Filter` is removed from the entity; `cwes` joins `openaev.tenant.active-tables`
  (config-guarded so dropping it fails the build); `TenantBaseListener` stays as a null-fill
  fallback (the resolver-set value wins);
- the tenant-provisioning datapack (the one background writer of `cwes`) ships UNCONVERTED under a
  documented waiver: it only INSERTs fresh rows (VALUES inserts pass the inspector) with correct
  `TenantContext` attribution and never reads the table - the exact shape is pinned by a
  provisioning-style test, and the conversion is the tracked follow-up where `forEachTenant` gets
  its first production caller.

**Behavior changes shipped by the capstone (to arbitrate alongside the primitive):**
- a vulnerability/CVE payload carrying CWEs now REQUIRES a resolvable single-tenant write scope:
  the plain and bulk endpoints return 400 for a multi-tenant caller (v2 semantics: a row belongs to
  exactly one tenant). Collectors using the tenant-prefixed path are unaffected;
- any context without a tenant scope reading a vulnerability's CWEs sees an empty list (fail-closed)
  - today the only CWE readers are the wired endpoints, but a FUTURE unwired reader would go dark
  silently (see limits).

## The frozen baseline is a deliverable: the conversion work list

- **41 self-invocation call sites across 21 classes** (each one is a latent "no transaction, no
  scope, silently" bug), including `InjectApi`, `ExerciseService`, `TenantService`,
  `PlatformRoleService`, `PlatformGroupService`, `WorkflowService`, `UserService`. This count is a
  floor, not an exact census (see limits).
- **`ScenarioExecutionJob`, `SecurityCoverageJob`**: `@Transactional` on job classes.
- **`ComchecksExecutionJob`**: raw `TransactionTemplate` (field + constructor + call).
- HTTP-path rule: zero violations today (the rule exists to keep it that way).

The burn-down has its own in-repo runbook (`.github/skills/reduce-tx-baseline/SKILL.md`): triage of
the violation type, sanctioned fix pattern per flavor, red-first tests when behavior changes, and
the deliberate store re-freeze. The store directory also carries a README stating what the
baseline is and the rules for changing it.

## Assumed limits (each one named, with its follow-up)

- **A job calling a repository directly, with no primitive, passes all rules**: the repository opens
  its own transaction, no scope is set, and fail-closed makes the job read 0 rows silently. The
  failure direction is right (correctness, not a leak) but "the job runs and sees nothing" will be
  the first support cost of the conversions. Converting those paths IS the G1-G4 work.
- **The "active table → scoped access" invariant has an interim static guard, not a runtime one.**
  `TenantActiveTableAccessArchTest` pins the REVIEWED access surface of every active table: only
  allowlisted classes may touch its repository or call an association accessor that lazy-loads it
  (for `cwes`: `Vulnerability#getCwes`, which bypasses the repository entirely), and activating a
  table without extending the guard fails the build. Floor semantics, stated plainly: it verifies
  WHO accesses, not that the access carries a scope at runtime; each allowlist entry documents its
  scope mechanism and adding one is a review event. The structural fixes are on the arbitration
  list: the compile-time scope rule (#6391) and/or a fail-loud `can_access_tenant`.
- **No guard is transitive.** The jobs rules see only the job classes themselves; the HTTP-path rule
  sees only DIRECT dependencies, so a service reachable from a non-transactional controller could
  still carry the primitive (mitigated by the `execute()` guard on transactional HTTP flows; the
  structural answer - e.g. refusing the primitive when a request context is present - is on the
  arbitration list). Rule boundaries exclude the `@Scheduled` surfaces (`SmtpService`,
  `ImapService`), `ExpectationsExpirationManagerJob` (under `io.openaev.collectors..`), listeners,
  seeding and the detached thread hand-offs (`StreamApi`/`streamExecutor`, Tanium/CrowdStrike
  `scheduledExecutorService`, the notification `taskScheduler`): #6398 perimeter decisions.
- **Under `allTenants()`, the resolved list is not accessible to the work**: a joined
  `@Transactional(TxCtx)` service can only be called with `null` (aspect inert) - passing
  `allTenants()` throws, passing a narrower ctx trips the guard. Subtle; the structural answer (an
  overload handing the resolved ctx to the work) is deliberately deferred. For per-tenant work,
  `forEachTenant` hands each iteration its own single-tenant ctx, which closes the common case.
- The self-invocation detection reads bytecode: method references (`this::txMethod`), `super`
  calls, class-level-only `@Transactional`, inherited methods and meta-annotations are not seen;
  reflection is invisible. The 41-item baseline is a floor. The standard self-injection workaround
  is flagged too, deliberately.
- The scope-channel tripwire is textual: string concatenation or the `SET LOCAL` idiom would evade
  it (hence the inspector arbitration item). The caller-bound resolver flavour (`CallerScopeSource`)
  is posed, not wired.
- Two invariants are argued by construction rather than by test: cross-thread scope isolation (the
  GUC is per connection) and the "gone after" assertion running on a possibly different pooled
  connection. The store-lock test pins the configuration file, not a runtime `-D` override.
- The tenant proofs for the primitive run on a test harness against `import_mappers`
  (`@TestPropertySource`); the capstone (`cwes`) is the real-code proof of the activation chain, on
  the HTTP side. The one background writer of an active table (the provisioning datapack on `cwes`)
  is waived-and-pinned, not converted (follow-up above).

## For arbitration - one blocking question, the rest has proposed defaults

**Blocking:**

- [ ] **The `AllTenants` type.** The PR makes it a `TxCtx` variant that cannot serialize itself
      (misuse fails loudly at runtime, proven through the aspect). The alternative is a separate
      intention type accepted only by the primitive: misuse impossible at compile time, at the cost
      of a second type in background signatures. The PR's choice is defended, not assumed.

**Shipped shapes to ratify (or amend - they are in the diff, tested):**

- [ ] The per-tenant loop `forEachTenant` (flat loop, snapshot list, aggregate failure, serial) and
      the `execute`/`executeNew` strictness (refuse inside / refuse outside) vs a silent
      `REQUIRES_NEW`.
- [ ] The capstone behavior change: CWE-bearing payloads require a single-tenant write scope (400
      otherwise). Right v2 semantics, but it lands on collector-facing endpoints.
- [ ] Freeze-store governance - proposed: locked everywhere, deliberate re-freeze in each
      conversion PR, burn-down of the 41 items owned and tied to the G1-G4 milestones.
- [ ] Guard boundaries - proposed: keep `scheduler.jobs..` + controller-annotation matching now;
      extend to `@Scheduled` surfaces, listeners and seeding as each G-phase converts them.
- [ ] Scope-channel hardening - proposed: yes, the inspector should reject unsanctioned writes to
      `app.current_tenants` (both `set_config` and `SET LOCAL`), folded into the ScopeChannel
      follow-up; the committed tripwire is the interim guard.
- [ ] Fail-loud option for missing scopes - today an active table queried with NO scope returns 0
      rows (fail-closed, silent). The structural alternative: `can_access_tenant` RAISES when the
      scope GUC is absent, turning every forgotten path into a loud error instead of silent
      emptiness. Runtime-complete (covers new callers, transitive paths, native SQL) where every
      static guard is a floor - but it changes the platform's failure philosophy (a missed path
      500s instead of rendering empty) and needs a carve-out for legitimately scope-less contexts.
      Proposed: debate at the first conversion wave, with the interim access guard as the bridge.

**Asynchronous (no meeting time needed):**

- [ ] The primitive's name (`TenantScopedTransaction`; alternatives `TenantTx`, `BackgroundTx`).
- [ ] Semantics of global maintenance on tenant-less tables (`allTenants()` today; a dedicated
      platform/maintenance scope may be more truthful - parked with Q7).
- [ ] The `TxCtx` javadoc wording ("no wildcard ever reaches the scope channel"; the invariant from
      the merged v2 design #6255 is strictly preserved).

## Plan after validation

The guard commit will be extracted into its own PR attached to #6391 when this leaves draft; the
primitive, resolution and capstone land through #6398. Direct follow-ups, in order: the
migration-engine conversion (per-tenant datapacks through `forEachTenant` - its first production
caller - and the cwes waiver retired), the active-table scope guard (closing the "new unscoped
reader goes dark" limit), and folding the duplicated `set_config` call (primitive + aspect) into one
shared scope-channel helper. `executeNew` stays reserved for punctual narrowing, never for loops.


.github/skills/activate-tenant-table/SKILL.md
```diff
@@ -22,6 +24,15 @@ PR #6255 description once before starting, plus the javadoc of
 `TenantStatementInspector`, `TenantWriteScopeResolver` and
 `TenantScopeTransactionAspect`.
 
+If the table has a background writer (a scheduler job, queue consumer or
+startup task that writes it), read the background transaction primitive too:
+`TenantScopedTransaction` (`openaev-model/src/main/java/io/openaev/context/TenantScopedTransaction.java`).
+The HTTP path carries its scope through `@Transactional` + `TxCtx` + the aspect;
+the background path must NOT use `@Transactional` (its self-invocation trap
+silently skips both the transaction and the scope) and opens transactions through
+the primitive instead. Phase 5b converts those writers; it is the prerequisite for
```
corinnekrych: what is phase 5b?

.github/skills/activate-tenant-table/SKILL.md
```diff
@@ -365,6 +458,183 @@ run it:
...
+  job that writes an already-active table without going through the primitive
+  would read and write zero rows with no failing test. The existing rules forbid
+  the wrong SHAPE (`@Transactional`, raw plumbing, raw JDBC) but do not assert
+  that every writer of an active table carries a real scope. Until that guard
+  exists, converting a table's writers is a point-in-time fact, not an invariant
+  — say so in the report.
+- The per-tenant loop is serial and single-threaded, one transaction per tenant.
+  For a job over thousands of tenants, watch total runtime against the job's
+  window (`@DisallowConcurrentExecution` means an overrun skips the next fire).
+  No batching or parallelism is provided; if runtime is a concern, raise it
```
corinnekrych: For ManagerIntegrationsSyncJob which deliberately uses CompletableFuture + a dedicated executor for parallel per-tenant sync, forEachTenant() would be a regression.
How do we plan to work with this case?

.github/skills/activate-tenant-table/SKILL.md
laugiov: Phase 5b refers to this runbook's own phase numbering: it is defined further down in this same file, between Phase 5 and the go-live. This is the step that converts a table's background writers to the primitive.

.github/skills/activate-tenant-table/SKILL.md
laugiov: There is no plan to convert that job to forEachTenant. The loop is a helper for the simple case, when tenants are processed one after another. Nothing more.

A job that already runs its tenants in parallel keeps its executor. The only change at conversion time: each parallel task opens its own transaction with tenantTx.execute(TxCtx.forTenant(id), work). The scope is set inside that transaction only, and a test already proves a scope never crosses threads.

Compared to forEachTenant, such a job only gives up the built-in error aggregation, which it already handles itself. I will make the parallel pattern explicit in the Phase 5b table.

laugiov: Done. The parallel pattern is now a row in the Phase 5b scope table, and there is a dedicated test for your exact case: concurrentScopedTransactionsStayIsolated holds two scoped transactions open at the same time and proves each one sees and writes only its own tenant's row.

Two cautions were also added to the skill for whoever converts that job: size the executor against the connection pool (each task holds one connection for its whole transaction), and await all tasks before the job returns, otherwise @DisallowConcurrentExecution no longer protects against overlapping runs.

corinnekrych:

corinnekrych (approved):
I've refined SKILL.md in my next PR for #6399 collectors table. All good so far!

corinnekrych (approved):

PR #6641, opened 2026-07-09
feat(engine-sync): quartz parallelism for engine sync (#6640)

antoinemzs:
### Proposed changes

* Refactor code to use quartz for parallelism and job concurrency gatekeeping

### Testing Instructions

1. Watch engine sync and database resource consumption, expecially whether there are multiple occurences of the same indexing query in PG

### Related issues

* Closes #6640 

### Checklist

- [ ] I consider the submitted work as finished
- [ ] I tested the code for its functionality
- [ ] I wrote test cases for the relevant uses case
- [ ] I added/update the relevant documentation (either on github or on notion)
- [ ] Where necessary I refactored code to improve the overall quality
- [ ] For bug fix -> I implemented a test that covers the bug

### Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #6644, opened 2026-07-10
fix(tests): restore profile exception for tests (#0000)

antoinemzs:
### Proposed changes

* Don't load the engine sync job in test profile


antoinemzs:
@laugiov fix for your comment in #6641

laugiov (approved):

Issue #6647, opened 2026-07-10
feat(chaining): attack path execution step POC


laugiov:
## Context

During a chaining simulation, OpenAEV runs many actions that pivot across endpoints and discover findings (endpoints, credentials, files, users, CVEs, open ports). Product wants an Attack Path view: a graph of injector to endpoint to finding type to finding that a user can load, read, and expand.

Today the chaining engine keeps a step's execution data in a single JSONB column (`steps.step_data`, a serialized `Inject`). That blob is a weak base for the attack path:

- Write cost: every update rewrites the whole document (write amplification, MVCC bloat).
- Read cost: getting one value means parsing the whole blob.
- Aggregation: counting or filtering over buried JSON fields has no clean index.

## Goal

Prove, with a bounded proof of concept, that a normalized PostgreSQL model can be the source of truth for the attack path, and that the interactive graph rebuilds from flat, indexed reads (no blob, no giant recursive query) and stays fast at large volume.

## What the POC proves

- Execution data lives in a small set of normalized tables, not a JSONB blob.
- The full graph of a simulation rebuilds from two flat reads plus one in-memory pass into `{ nodes, edges }`, with deterministic IDs. No recursive walk.
- Each interaction (load the graph, expand an endpoint, get an endpoint's relations) is a single indexed query.
- It scales: at a large volume, per-simulation graph load stays within target, with the database optimizations shown as evidence.
- A minimal front renders the graph from the API and does a couple of manipulations.

## Scope

In scope:

- Three additive, isolated tables (`attackpath_execution`, `attackpath_finding`, `attackpath_execution_finding`), seeded directly.
- A rebuild service producing `{ nodes, edges }` aligned to the render contract.
- POC-only API endpoints behind a feature flag.
- A parameterized generator producing large, skewed volumes with deliberate single-simulation outliers.
- A reproducible benchmark with `EXPLAIN ANALYZE` evidence.
- A minimal front tab.

Non-goals:

- No production UI.
- No path computation (reachability, shortest path, path discovery).
- No change to how the engine stores `step_data`. Additive only.
- No Elasticsearch.
- No live streaming.
- No ingestion from the real engine. This is the number one follow-up, and it is not de-risked here.

## Key facts confirmed against the code

- "Simulation" is the `exercises` table. There is no `simulation` table.
- Endpoint is an `assets` discriminator. There is no `endpoint` table.
[...]

Issue #5048, opened 2026-03-02
feat(chaining): attack path view


mergenhan:
As a SOC user, 
I want to watch the simulation live 
so that I can see executions in real time.

Issue #204, opened 2022-10-11
Problem to upload file higher than 13MB


Bapth-R:
Hello, 
Thanks for your great work :)

## Description
I want to upload a zip file (14,7 MB) to the Document part but the file does not want to be upload to the OpenEx platform. 
I tested to uploading zip files smaller than 10MB and it worked. 
 
## Environment
1. OS (where OpenEx server runs): Debian 11
2. OpenEx version: OpenEx 3.0.2

Thanks,
BR

SamuelHassine:
Hello,

I've just tested on the latest version (https://demo.openex.io) a 30MB file and it works well. Can you try to upgrade your platform?

Kind regards,
Samuel

Bapth-R:
Thanks for your answer. I updated the platform to the 3.2.0 version but it's the same problem. 
Where can I find the log file ?

SamuelHassine:
Hello @Bapth-R,

Can you check the response of upload query in the network console of your browser?

Do you have any output in the platform Docker container when you upload the file?

Kind regards,
Samuel

Bapth-R:
Hello @SamuelHassine,

I have the 413 error on my browser : 
![image](https://user-images.githubusercontent.com/76159854/196144545-a2a10a2a-9f33-43e7-8767-70d4291eec73.png)

In the Docker container, I have any output related to the document uploading in the /logs/ part. Maybe I'm not looking on the right folder. 

Regards, 
BR

Issue #202, opened 2022-10-10
Be able to schedule injects based on the results / expectations of previous injects


SamuelHassine:
## Problem to Solve

Be able to schedule injects based on the results / expectations of previous injects

Chunk of https://github.com/OpenBAS-Platform/openbas/issues/311

EllynBsc:
Duplicate of [#202](https://github.com/OpenBAS-Platform/openbas/issues/202)

PR #6652, opened 2026-07-10
feat(chaining): attack path view #6647 #5048 #6855

### LINKED ISSUE #6855: feat(chaining): attack path - part 1
## Description

This issue covers **Part 1** of the Attack Path feature in the chaining engine. It introduces the full backend and frontend infrastructure to visualize, explore, and analyze a simulation's attack path as an interactive graph.

#### 🗄️ Backend — Model & persistence

- New JPA entities: `AttackPathExecution`, `AttackPathExecutionFinding`, `AttackPathFinding`
- New SQL projections (rows): `AttackPathExecutionRow`, `AttackPathFindingRow`, `AttackPathEndpointFindingRow`, `AttackPathEdgeGroupRow`, `AttackPathEndpointGroupRow`, `AttackPathInjectorMetaRow`, `AttackPathInjectorPatternRow`, `AttackPathInjectorPatternRow`, `AttackPathSimSummaryRow`, etc.
- New repositories: `AttackPathExecutionRepository`, `AttackPathFindingRepository`
- Flyway migration `V6_20260719200000000__Add_attack_path_tables`: 3 tables (`attackpath_execution`, `attackpath_finding`, `attackpath_execution_finding`), indexes, FK to `tenants` with `ON DELETE CASCADE`, `NOT NULL` constraint on `tenant_id`
- Additional migration `V6_20260720000000000__Add_attackpath_execution_vulnerability_status`

#### ⚙️ Backend — Services & API

- `AttackPathGraphService`: graph rebuild (full mode and collapsed mode for large simulations)
- `AttackPathExecutionIngestionService` + `AttackPathSourceTargetResolver`: ingestion of execution rows with asset-group member resolution (both direct assets and group members are expanded and recorded as individual endpoint rows)
- `AttackPathSeedService`: realistic data generator for benchmarks and tests — uses raw JDBC with `@AllowRawJdbc` (allowlisted exception, see ADR-003); **must be removed before production** (`TODO(#6647)`)
- `AttackPathIds`: deterministic ID helper
- `AttackPathApi`: REST endpoints gated behind the `ATTACK_PATH` preview feature flag — every endpoint calls `requireAttackPathFeature()` and returns **404** if the flag is disabled
  - main graph, expand, relations, paginated findings, execution detail (Result & Terminal)
  - seed endpoint (dev/bench) — `@AccessControl(skipRBAC = true)` while fed by synthetic seed (intentional concession, documented)
- Tenant isolation on all native reads (`WHERE tenant_id`); `ON DELETE CASCADE` on FK ensures attack path rows are cleaned up when a simulation is deleted
- Attack path rows deletion when `ExerciseService` cleans a simulation
- Expectation update (`InjectExpectationService`) during ingestion with endpoint context enrichment (`agent_id`, `asset_id`, `asset_group_id`)
- ATT&CK techniques and injector type exposed on injector nodes
- Frozen attributes of a source-only endpoint exposed on the graph
- `InjectExecutionStep` and `WorkflowService` updated
- Import detection remediation at execution time

#### 🧪 Backend Tests

- `AttackPathSchemaMigrationTest`: verifies tables, indexes, FK, `NOT NULL`, idempotency
- `AttackPathRepositoryTest`, `AttackPathGraphServiceTest`, `AttackPathSeedServiceTest`
- `AttackPathApiTest`, `AttackPathApiDisabledTest` (verifies 404 when flag is off), `AttackPathHttpIsolationTest`
- `AttackPathCollapsedTest`, `AttackPathModeConsistencyTest`
- `AttackPathFindingsListTest`, `AttackPathExecutionDetailTest`
- `AttackPathIngestionTenantAttributionTest`: verifies `tenant_id` attribution on execution rows
- `AttackPathBenchmark` / `AttackPathCreateBenchmark` / `AttackPathPartitioningDemoTest`: volume and partitioning benchmarks (raw JDBC ~17× faster than JPA for seed inserts)
- `ExecutionTraceRepositoryTest`, `InjectStatusRepositoryTest`
- `WorkflowServiceTest`, `ExerciseServiceTest`, `ExerciseServiceIntegrationTest`
- `TenantActiveTableAccessArchTest` architecture test updated to allowlist the raw-JDBC attack path seed

#### 🖥️ Frontend

Built on **`@xyflow/react` 12.10.2** (already in `package.json`) with a **manual column layout** (no external layout lib).

- New `attack_path/` module under `simulations/simulation/`:
  - `SimulationAttackPath.tsx`: main tab with simulation picker dropdown — rendered only when `isFeatureEnabled('ATTACK_PATH')` is true
  - `AttackPathFlow.tsx`: ReactFlow graph with minimap, legend, expand/collapse mode, auto-fit on collapse, view preserved on expand
  - `AttackPathLegend.tsx`, `AttackPathTableView.tsx`: sortable table view with **CSV export** (`attack-path-chokepoints.csv`)
  - Nodes: `AssetNode`, `FindingNode`, `FindingTypeNode`, `InjectorNode`, `EndpointClusterNode`, `FindingClusterNode`
  - Edges: `GroupedEdge`, `CausalEdge` (kill-chain — behind mock, `TODO(#6647)`)
  - Side panels: `FindingDetailPanel`, `EndpointDetailPanel`, `ExecutionResultTerminalPanel`
  - `ImageWithFallback.tsx`: security platform icon fallback helper
  - `attack-path-colors.ts`: verdict, chokepoint (violet accent), selection (blue reserved), severity palette
  - `attack-path-flow-helpers.ts`: ReactFlow node and edge builder
  - `attack-path-killchain-mock.ts`: temporary causal kill-chain edges (`TODO(#6647)`: drop when backend exposes the real data)
  - `node-sizes.ts`
- `attack-path-actions.ts`: Redux/API actions (graph, expand, relations, findings, execution detail, seed)
- `api-types.d.ts` updated (types generated from the API)
- i18n: keys added to all locales (de, en, es, fr, it, ja, ko, ru, zh)
- Dark theme: readable canvas and node selection
- Search box to focus an endpoint, injector or finding type
- Captured secrets protected — never displayed in clear text across all attack-path views
- **Chokepoint strip**: top N most-exposed endpoints (ranked by finding count); badge + popover card, focusable from both graph and table views, stable CSV rank independent of current table sort
- `vite.config.ts` updated
- Unit tests: `attack-path-actions.test.ts`, `attack-path-flow-helpers.test.ts`, `SimulationAttackPath.test.tsx`

#### 📖 Documentation

- `adr/ADR-002-attack-path-execution-store-on-postgresql.md`: storage strategy, partitioning decision
- `docs/docs/development/attack-path.md`: full developer guide (updated `mkdocs.yml`)
- `ATTACK_PATH_ANALYSIS.md`, `ATTACK_PATH_BACKEND_REQUIREMENTS.md`, `ATTACK_PATH_UX_IMPROVEMENTS.md`
- `specs/001-attack-path-ux-hardening/spec.md`
- `.github/skills/activate-tenant-table/SKILL.md`
- `docs(multi-tenancy)`: documented the insert-only raw-JDBC exception and nested `executeNew` connection cost

---

## Environment

1. OS: N/A (backend + frontend feature)
2. Version: in development (branch `feat/chaining-attack-path-part1`)
3. Required dependencies: PostgreSQL, RabbitMQ, MinIO, Elasticsearch — standard project Docker services

## Reproducible steps

This issue is a **feature**, not a bug. To validate:

1. Start Docker services: `cd openaev-dev && docker-compose up -d`
2. Build the backend: `mvn clean install -DskipTests -Pdev`
3. Add `ATTACK_PATH` to `openaev.enabled-dev-features` in `application.properties` (e.g. `openaev.enabled-dev-features=INJECT_CHAINING,ATTACK_PATH`)
4. Navigate to a simulation → **Attack Path** tab
5. Select a simulation from the dropdown
6. Explore the graph (expand/collapse, click a finding, side panel, table view, CSV export, chokepoint strip)

## Expected output

- The attack graph displays endpoints, injectors and findings grouped into clusters
- Clicking a finding opens the detail panel with result and terminal output
- The table view lists endpoints with sorting and CSV export (`attack-path-chokepoints.csv`)
- Chokepoints (most-exposed endpoints by finding count) are highlighted with a violet accent
- Data is strictly isolated per tenant
- Captured secrets are never displayed in clear text
- Endpoints are **404** if the `ATTACK_PATH` flag is not enabled

## Actual output

Feature implemented and tested (see tests above).

## Additional information

- Related issues: #5048, #6647, #6642, #6729, #6732, #6803
- **Collapsed mode** added for large simulations (performance): aggregated view groups endpoints into clusters
- **Causal kill-chain edges** are behind a temporary mock (`attack-path-killchain-mock.ts`) pending the real backend (`TODO(#6647)`)
- **`skipRBAC = true`** on the seed endpoint is intentional: synthetic seeded simulations are not real exercises, so resource-level RBAC cannot apply (documented)
- **Raw JDBC** used in `AttackPathSeedService` for performance (~17× faster than JPA for bulk inserts); allowlisted with `@AllowRawJdbc`, sets `tenant_id` explicitly on every row; **must be removed before production** (`TODO(#6647)`)
- Volume and partitioning benchmarks included to validate the PostgreSQL storage strategy (ADR-002)
- Multi-tenant isolation verified by dedicated tests (`AttackPathHttpIsolationTest`, `AttackPathIngestionTenantAttributionTest`)
- Several `TODO(#6647)` remain for Part 2: real per-finding expectation verdicts, real injector type/slug from backend, kill-chain causal edges, seed removal

laugiov:
## Update (2026-07-23) — reconciled with main, reviewed, merge-ready as an off-by-default POC

- Reconciled with `main`; re-verified: full build green, attack-path suite 117/117, front type-check clean.
- Multi-agent review pass. Two flag-independent bugs fixed here (with a regression test):
  the injector `target_selector` compared to `"asset"`/`"asset_group"` while the stored value is `"assets"`
  (agentless asset scans recorded nothing); and `attackpath_execution_id` was `varchar(255)` while the id
  embeds an unbounded target (long targets overflowed and silently dropped the inject's edges → now `text`).
- Still behind `PreviewFeature.ATTACK_PATH` (off by default): write path, all read endpoints, and the front
  tab/route are gated.
- Before enabling the flag on real data (tracked, not blockers for an off-by-default merge): status write-back
  (#204, else nodes render RED), placeholder front telemetry, resource-level RBAC on the reads, rollback-orphan rows.

Closes #6855

## Summary

Storing a chaining simulation's attack-path data in a small set of normalized PostgreSQL tables, so the interactive graph rebuilds from flat, indexed reads instead of parsing the `step_data` JSONB blob.

A proof of concept, feature-flagged and off by default. It landed as small, test-first commits (below).

Refs #6647. Also carries the first interactive drawers for feature #5048 (see the next section).

## Update — interactive attack-path drawers (feature #5048)

Two interactive increments landed on top of the store POC, on this same branch, to unblock the product feature #5048 (attack path creation and live tracking) so the team can build on them. Both read the frozen snapshot only (no live inject resolution), stay behind the `ATTACK_PATH` preview feature, and are tenant-isolated like the rest of the POC.

- **Finding widgets and drawer.** The tab header shows per-category finding counters (Credentials, Users, CVEs, Files). Clicking a counter opens a right drawer listing that category's findings, paginated, with credential secrets masked server-side (the username is kept, only the secret is hidden). Clicking a finding highlights it on the map and focuses its producing execution in the feed.
  New read: `GET /api/attack-path/simulations/{id}/findings?category=&page=&size=`.
- **Execution result and terminal panel.** Clicking an execution in the feed opens an in-flow panel with Result and Terminal tabs. The Terminal tab reuses the platform's shared terminal renderer, fed by the snapshot's masked command and output; the Result tab shows the target, the prevention and detection status, and the produced findings. A "Focus on map" action re-centers the execution's endpoint.
  New read: `GET /api/attack-path/simulations/{id}/executions/{executionId}`.

Caveat (see "Not in scope"): the drawers render from the snapshot columns and have so far been exercised against seeded data. Real-engine ingestion has now started (Phase A below), so the execution rows are written from a real run, but the command, terminal, status and findings columns the drawers rely on are not engine-populated yet. Completing that ingestion, and honoring its value-shape contract (finding type = the `ContractOutputType` label, credential value `username:password`, a command and terminal synthesized from the real execution traces), is the number-one follow-up (#204 / #202).

## Update: real-engine ingestion, Phase A (feature #5048, issues #203 / #204)

Two increments started wiring the store to the real chaining engine, so the drawers above can eventually read live data instead of seeded rows. Both stay behind the `ATTACK_PATH` preview feature and are additive and guarded (an ingestion failure never fails the inject execution).

- **Execution rows from a real run (#203, #6729).** At a real agent-triggered run, one `attackpath_execution` row is created per (target, agent): source = the agent's asset (`source_kind` = `AGENT_ASSET`, `source_asset_id`, plus the **frozen** source hostname/ip/platform), the target resolved per payload (DNS hostname, command endpoint or file), the target's frozen hostname/ip/platform, `agent_name` / `agent_privilege`, `payload_name`, `source_injector` and `executed_at`. A queryable `inject_id` column is written at create as the lookup key #204 uses to find a run's rows by `(inject_id, agent_id)`. Asset/agent/source/target columns only; command, terminal, status and findings are the following increments.
- **Command and trace extraction (#204, #6732).** Helper methods to build the executed `command` (from the payload command blocks) and the per-endpoint execution traces (the terminal), both null-guarded. Trace ordering is made deterministic at the model level (`@OrderBy` on the status traces, plus `ORDER BY` on the trace-lookup queries), so the rebuilt terminal is stable. These are the building blocks for #204 writing command / terminal / `executed_at` back onto the rows.

Follow-up: the Phase A create runs in a background/queue context, so making its idempotent re-access production-safe under the tenant inspector needs the background tenant-scope primitive (#6642); today it is proven by tests and stays behind the flag.

## Why

`step_data` (a serialized `Inject` in one JSONB column) is a weak base for the attack path: every update rewrites the whole blob (write amplification, bloat), reading one value parses the whole document, and there is no clean index for aggregation. This POC shows a normalized relational source of truth that rebuilds the graph cheaply and stays fast at scale.

## Approach

- Three additive, isolated tables: `attackpath_execution` (one row = one edge, carrying the run snapshot), `attackpath_finding`, `attackpath_execution_finding`.
- Rebuild = two flat JPQL reads (Read A: executions; Read B: findings joined to their producing execution) plus one in-memory pass into `{ nodes, edges, counters }` with deterministic IDs. No recursion.
- Tenant isolation via the v2 statement inspector: the two tenant tables are added to `openaev.tenant.active-tables` (no v1 `@Filter`), and each read endpoint declares a `TxCtx` so the transaction aspect writes the request's tenant scope (from the `{tenantId}` path) for the inspector to filter against. Without that parameter the reads would fail closed.
- Isolated and droppable: prefixed tables, no hard FK to real product tables (except `tenant`), reads are pure JPQL (no native SQL).
- Feature-flagged with the platform's standard mechanism, not an ad-hoc property: a new `ATTACK_PATH` value in the `PreviewFeature` enum, checked at runtime in the controller (which reads `openaev.enabled-dev-features`). Off by default, so the routes return 404 unless the feature is enabled.

## Design docs

Two committed docs, both with mermaid diagrams: an ADR (`adr/ADR-002-attack-path-execution-store-on-postgresql.md`) for the decision and the options weighed, and a developer guide (`docs/docs/development/attack-path.md`) that teaches the model, the rebuild, the two modes, tenant isolation, the measured results, and how to continue.

## Progress

- [x] T1 Flyway migration: three tables plus indexes (additive, idempotent)
- [x] T2 entities (`TenantBase`) plus JPQL projection repositories plus tenant activation
- [x] T3 deterministic, collision-safe ID helper (`AttackPathIds`)
- [x] T4 rebuild service: two reads plus one in-memory pass; the read path is a constant two reads independent of graph size
- [x] T5 graph endpoint, gated by the `ATTACK_PATH` preview feature (off by default)
- [x] T6 expand and relations endpoints, one indexed query each (asset and discovered/raw endpoints)
- [x] T7 tenant isolation proven end-to-end through the real HTTP endpoints: under its own tenant path a simulation's graph/expand/relations are visible, under another tenant's path they are empty, and a scope-less read is fail-closed
- [x] T8 seed generator: batched raw-JDBC bulk insert deliberately bypassing the inspector (ADR-002; ~17x faster than the through-Hibernate path and safe because `tenant_id` is explicit), skewed sizes with single-simulation outliers, spray/stuffing fan-out, shared findings and discovered endpoints; admin-only, flag-gated `POST /seed`
- [x] T9 benchmark harness (`AttackPathBenchmark`, env-gated with `ATTACKPATH_BENCHMARK=true`, skipped in CI): seeds prod-fidelity data (rows carry realistic TOAST-able `command`/`terminal_output`, power-law fan-out, endpoints scaling with size), times `buildGraph` per size bucket (typical + 100k/300k/500k outliers) with p50/p95, reads/assembly split and peak heap; ORM A/B and a heavy-column Read-A-vs-`SELECT *` A/B; `EXPLAIN`. Numbers and the honest verdict in the workspace `results.md`
- [x] Collapsed graph mode: a DB-aggregated response for large simulations (endpoint groups + grouped edges + per-type finding counts, no per-execution/per-finding rows), auto-selected above a threshold or forced via `?mode=`, reusing `AttackPathDTO` with a `mode` field
- [x] T10 to T12 minimal front (React Flow): a POC "Attack path" simulation tab behind the `ATTACK_PATH` feature, rendering full and collapsed graphs (injector / endpoint / finding-type / finding nodes, grouped `+N` edges), a collapsed/full/auto mode toggle, and endpoint click → lazy-load its findings (expand) and executions (relations side panel); typed api actions with a Vitest action test and a DTO-to-flow mapping test
- [x] Severity colours combine prevention **and** detection (GREEN prevented, ORANGE detected-not-prevented, RED neither), worst-case per endpoint, applied to nodes and edges
- [x] Front polish: a simulation picker (dropdown listing the tenant's simulations with their endpoint counts, backed by a `GET /simulations` endpoint), a `nav` seed preset (~100 endpoints) and a tenant-targetable seed (`tenantId` + presets), a circular status-coloured node design with a grid layout and viewport culling for large graphs
- [x] T13 ADR-002 (`adr/ADR-002-attack-path-execution-store-on-postgresql.md`): context, options A (JSONB+index) / B (normalized PostgreSQL) / C (Elasticsearch), decision B with the measured evidence, consequences, scope boundary; erDiagram of the model and a "feared recursive query vs actual two-reads-and-a-pass" flowchart
- [x] T14 developer guide (`docs/docs/development/attack-path.md`, in the mkdocs nav): the big picture, the model, the rebuild, the two modes, severity, tenant isolation, the API, run/seed/explore, the measured results, and how to continue — with the required mermaid diagrams
- [x] T15 results: the method, numbers, and what they mean are written up (developer guide section 9 and the workspace `results.md`); this PR description carries the headline

### Feature #5048 drawers (on top of the POC)

- [x] Finding widgets + right drawer per category (Credentials / Users / CVEs / Files), paginated, credential secrets masked server-side; finding click cross-focuses the map and the feed. New read `GET /simulations/{id}/findings?category=&page=&size=`
- [x] Execution result & terminal panel: clicking a feed execution opens an in-flow Result / Terminal panel (shared terminal renderer, masked command and output), with a "Focus on map" action. New read `GET /simulations/{id}/executions/{executionId}`
- [x] Tests: backend read + masking + endpoint (admin, flag-off 404, cross-tenant not found) + HTTP isolation; front action encoding and a component test (feed click fetches and opens the panel, both tabs render, terminal shows the masked command/output, focus-on-map sets the map focus)

## Testing

Real Postgres, from `openaev-api`:

```
mvn -pl openaev-api test -Dtest="AttackPath*"
```

- Migration test: the three tables and their indexes exist, additive, idempotent (re-run is a no-op).
- Repository test: the JPQL projection reads return the expected flat rows filtered by simulation. Read A omits the heavy columns (`command`, `terminal_output`).
- ID helper unit test: deterministic and collision-safe (delimiter-in-value and null cases).
- Rebuild service test: nodes/edges/counters, endpoint colours, `+N` grouping with `executionIds`, shared-finding dedup, execution→finding cross-reference, empty simulation, expand and relations, discovered (raw) endpoint, and the constant-two-queries proof.
- Endpoint tests: the graph, expand and relations routes (base and tenant-prefixed) return the DTOs; with the feature off the routes return 404.
- Tenant isolation test (HTTP, `active-tables` on): modeled on the `ImportMapperHttpIsolationTest` pilot. Through the tenant-path endpoints the owner tenant sees its graph/expand/relations, another tenant sees empty (no cross-tenant leak on any of the three reads), and a scope-less repository read is fail-closed with raw-JDBC ground truth proving the rows exist.
- Seed generator test (`active-tables` on, proving the raw-JDBC seed writes correctly even with the tables activated): a smoke run asserts the shape (spray fan-out, findings shared across endpoints, discovered raw endpoints, a status mix, populated run-snapshot columns, a skewed distribution with an outlier far above the median) and determinism (same seed → same counts), with raw-JDBC ground truth. Endpoint tests: `POST /seed` returns the counts for an admin and is 403 for a non-admin.
- Feature #5048 drawers: the findings-list read (pagination, deterministic ordering, credential masking) and the execution-detail read (result + terminal, credential secrets masked in the command, the output and the finding values); the two endpoints (admin, feature-off 404, another tenant not found) and their HTTP isolation; on the front, the action URL/id encoding and a component test of the execution panel.

## Key results (benchmark at 5.78M executions, prod-fidelity data)

Full numbers and methodology in the workspace `results.md`. Headline:

- The relational model and the `simulation_id` index work: the read plan is a bounded index scan, no recursion, a constant two reads regardless of graph size (an auto-served request adds one O(1) `COUNT` to pick the mode).
- **A large single simulation is the ceiling**: rebuilding its whole graph costs ~1.4 s (100k executions) to ~6.3 s and ~3.2 GB heap (500k). Even ~30k typical simulations are ~0.75–1.4 s at 5M total, over the 300 ms target.
- **Per-simulation cost is not volume-independent at scale**: once the wide-row table exceeds cache the heap fetch is IO-bound, so the same simulation is ~1.6x slower at 5M than at 0.5M. (The skinny first-pass data hid this; realistic rows exposed it.)
- **The short-column Read A is worth ~2x** (measured: 500k read 1.3 s vs 2.5 s for `SELECT *`), and **ORM overhead is negligible** (no native read fallback needed).
- **The collapsed backend lever is built and measured**: at 5.78M total, forcing the aggregated mode on the 500k outlier is **~1.2 s / 16 MB heap vs ~6.3 s / 3.2 GB full (~5x, ~200x heap)** — one indexed `GROUP BY` returning ~2,050 endpoint nodes, with per-endpoint findings/executions lazy-loaded on click in single-digit-millisecond reads (an endpoint expand touches 0.18% of the rows). A simulation above the collapse threshold is served collapsed, so it stays renderable; the front renders both modes. It is not a free 300 ms (the aggregation scan is still IO-bound at 5M), but it removes the heap and render walls.

## Try it (reviewer)

Full steps in the developer guide: **`docs/docs/development/attack-path.md`** (enable, seed, explore).
In short:

1. Enable `ATTACK_PATH` (in `openaev.enabled-dev-features`, or `*`); start the backend and the front
   (`cd openaev-front && yarn start`). Log in as admin.
2. The tables are tenant-active, so seeded data is only visible under **your** tenant — the seed takes a
   `tenantId` (the single-tenant default is `2cffad3a-0001-4078-b0e2-ef74274022c3`). One `medium` seed
   gives the whole spectrum (20 simulations, ~100→450 endpoints, one 100k outlier):
   ```
   curl -X POST http://localhost:8080/api/attack-path/seed \
     -H "Authorization: Bearer <admin-token>" -H "Content-Type: application/json" \
     -d '{"preset":"medium","seed":7,"tenantId":"2cffad3a-0001-4078-b0e2-ef74274022c3"}'
   ```
   Presets: `smoke` (tiny), `nav` (~100 endpoints, smooth), `medium` (~0.5M), `large` (~2M), `full`
   (~5.8M — use the `AttackPathBenchmark` harness, not the endpoint).
3. Open any simulation → the **Attack path** tab. The **Simulation id** field is a dropdown of your
   seeded simulations with their **endpoint count**, largest first. Pick one, then:
   - Toggle **Collapsed / Full / Auto** on `ap-seed-7-sim-0` (the 100k outlier) to feel the lever:
     collapsed ≈ ~450 nodes vs full ≈ ~2,800 nodes + a 100k-row feed (backend rebuild ≈ 0.3 s
     collapsed vs ≈ 1.4 s full; the front latency itself is reasoned from node counts, not
     browser-measured — see results.md).
   - Click an endpoint to lazy-load its findings (collapsed) and its executions (side panel).
   - Click a finding widget (Credentials / Users / CVEs) to open its drawer, then a finding to
     cross-focus the map and the feed; click a feed execution to open its Result / Terminal panel.

## Not in scope

No production UI, no path computation, no change to `step_data` storage, no Elasticsearch, and no live streaming. Real-engine ingestion has **started** on this branch (Phase A, see the ingestion update below): the execution rows are now written from a real agent-triggered run with their asset/agent/source/target details. The remaining ingestion (command and terminal for #204, status and findings with the value-shape contract for #202) is the number-one follow-up.

## Screenshots

**1. The full attack-path graph** (small `demo-attack-path`). Left to right: injectors (hexagons) → endpoints (circles coloured by severity: green = prevented, orange = detected but not prevented, red = neither) → finding-type icon nodes → finding leaves. Edges are coloured by the endpoint's severity and labelled by finding type; a `+N` badge groups repeated executions on the same hop.

**2. Selecting an endpoint** highlights its path and lists its executions in the side panel (CORP-DC-01 and its two executions, with agent and privilege).

**3. The simulation picker** — a dropdown of the tenant's simulations, each showing its endpoint count (the number of graph nodes) and execution count, largest first.

**4. Collapsed mode on a ~100-endpoint simulation** — one node per endpoint in a grid layout, so hundreds of endpoints stay navigable on one screen.

**5. Full mode on the same simulation** — every finding node materialized into a dense mesh. This is what collapsed mode avoids, and why large simulations are served collapsed.

**6. Selecting an endpoint on a large graph** — the side panel lists all of its executions (here 22, across several payloads and agents).


openaev-api/src/main/java/io/openaev/service/attackpath/AttackPathGraphService.java
```diff
@@ -0,0 +1,518 @@
...
+ *
+ * <p>The execution is carried on the source-to-target edge (its {@code executionIds}), not as a
+ * standalone map node (design O2), while the left feed still lists every execution.
+ */
+@Service
+@RequiredArgsConstructor
+public class AttackPathGraphService {
+
+  private static final String SOURCE_INJECTOR = "INJECTOR";
+  private static final String PREVENTED = "Prevented";
```
RomuDeuxfois: ExpectationType.PREVENTION.successLabel
Check others

openaev-front/src/admin/components/simulations/simulation/attack_path/edges/GroupedEdge.tsx
```diff
@@ -0,0 +1,75 @@
...
+            className="nodrag nopan"
+            style={{
+              position: 'absolute',
+              transform: `translate(-50%, -50%) translate(${labelX}px,${labelY}px)`,
+              background: theme.palette.background.paper,
+              border: `1px solid ${alpha(color, 0.6)}`,
+              borderRadius: 10,
+              padding: '0 6px',
+              fontSize: 11,
+              color: theme.palette.text.secondary,
```
RomuDeuxfois, line 70: Hardcoded CSS

openaev-model/src/main/java/io/openaev/database/repository/attackpath/AttackPathExecutionFindingRepository.java
```diff
@@ -0,0 +1,8 @@
+package io.openaev.database.repository.attackpath;
+
+import io.openaev.database.model.attackpath.AttackPathExecutionFinding;
+import io.openaev.database.model.attackpath.AttackPathExecutionFinding.AttackPathExecutionFindingId;
+import org.springframework.data.repository.CrudRepository;
+
+public interface AttackPathExecutionFindingRepository
```
RomuDeuxfois: seems not used

adr/ADR-003-attack-path-execution-store-on-postgresql.md
```diff
@@ -0,0 +1,166 @@
+# ADR-002: Attack path execution store on PostgreSQL
+
+|  |  |
+| --- | --- |
+| Status | Accepted (POC-backed) |
+| Related | https://github.com/OpenAEV-Platform/openaev/issues/6647 |
+
+> **In one paragraph.** The attack-path view rebuilds a graph of what a simulation did (which injector
+> hit which endpoint, what it found). Today that data lives in a JSONB blob (`step_data`) that is slow to
```
RomuDeuxfois, line 9: step_data is a property introduce for the chaining.
Can we direclty add some structure here ? No migration required since it's under FF for now.

openaev-api/src/main/java/io/openaev/api/attackpath/AttackPathPocApi.java
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  /**
+   * Expand an endpoint into its finding types and findings, from a single indexed read. {@code ref}
+   * is the endpoint key (asset id or the raw value of a discovered endpoint), URL-encoded.
+   */
+  @GetMapping("/simulations/{simulationId}/endpoint/findings")
+  @Transactional(readOnly = true)
+  @AccessControl(skipRBAC = true)
+  public AttackPathExpandDTO expand(
```
RomuDeuxfois: Should be suffix by finding to be clearer no ?

openaev-api/src/main/java/io/openaev/api/attackpath/AttackPathApi.java
```diff
@@ -0,0 +1,117 @@
...
+  private final AttackPathSeedService seedService;
+
+  /**
+   * Full graph of a simulation, built from two flat reads plus one in-memory pass. {@code
+   * skipRBAC}: synthetic seeded simulations are not real {@code exercises}, so resource-level RBAC
+   * cannot resolve them; tenant isolation is still enforced by the statement inspector.
+   */
+  @GetMapping("/simulations/{simulationId}/graph")
+  @Transactional(readOnly = true)
+  @AccessControl(skipRBAC = true)
```
RomuDeuxfois, line 78: Should not be skipRBAC I think

openaev-api/src/main/java/io/openaev/api/attackpath/AttackPathPocCondition.java
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Registers the attack-path POC endpoints only when the {@code ATTACK_PATH_POC} preview feature is
+ * enabled (in {@code openaev.enabled-dev-features}, or {@code *}). Mirrors {@code
+ * InjectChainingCondition}, the platform's standard way to gate a bean behind a preview feature.
+ */
+public class AttackPathPocCondition implements Condition {
+
+  @Override
+  public boolean matches(ConditionContext context, @NotNull AnnotatedTypeMetadata metadata) {
```
RomuDeuxfois: Use PreviewFeatureService to homogenize code

openaev-model/src/main/java/io/openaev/database/model/attackpath/AttackPathExecution.java
```diff
@@ -0,0 +1,114 @@
...
+@Getter
+@Setter
+@Entity
+@Table(name = "attackpath_execution")
+@EntityListeners(TenantBaseListener.class)
+public class AttackPathExecution implements TenantBase {
+
+  @Id
+  @ControlledUuidGeneration
+  @Column(name = "id")
```
RomuDeuxfois: prefix by table/column

adr/ADR-003-attack-path-execution-store-on-postgresql.md
laugiov: Good point, and you're right: it needs no migration (the tables already exist) and it's safe under the flag. Doing it means the actual ingestion: hooking the chaining step write (StepService / StepEventService) to write an attackpath_execution row (plus its findings and links) per executed step, extracting source/target/agent/status out of step_data and resolving source→target (with/without agent). 
That extraction + resolution is the genuinely tricky part, the product spec even flags per-endpoint status as an open point, so I'd rather ship it as a focused follow-up right after this PR than bolt it onto the naming/cleanup we're finalizing here.

openaev-api/src/main/java/io/openaev/api/attackpath/AttackPathApi.java
laugiov: skipRBAC is needed while the tables are fed by the synthetic seed: seeded simulations aren't real
`exercises`, so resource-level RBAC can't resolve their ids, removing it would 404 every read on
seeded data. Tenant isolation still applies (the statement inspector). I added a TODO that real
resource RBAC must apply once these tables are fed from real simulations (the ingestion follow-up).
OK to keep it with that TODO?

PR #6666, opened 2026-07-13
fix(multi-tenancy): changed the way executors are registered to always use the register service method (#6547)

### LINKED ISSUE #6547: fix(multi-tenancy): agent list on endpoint not filtered by tenant
## Description

When a simulation agent is installed on the same physical endpoint (same machine) from two different tenants (tenant A then tenant B), the agent configuration files on the machine are overwritten (only one configuration persists on the filesystem since they share the same install path). However, in the database, both agent records remain referenced. As a result, when a user connected to tenant A views the list of agents for an endpoint, they see **all installed agents without tenant distinction**, including those belonging to tenant B.

Note: there is no credential leakage since each agent has its own token scoped to its tenant.

## Environment

1. OS: All 
2. Version: 2.260623.0
3. Other environment details: Multi-tenancy enabled, at least 2 tenants configured, same physical machine targeted from both tenants

## Reproducible steps

1. Connect to tenant A and install an agent on a physical endpoint (machine X)
2. Verify the agent is registered and visible in tenant A
3. Connect to tenant B and install an agent on the **same physical machine X** (same MAC addresses / hostname)
4. The agent configuration on the machine is overwritten by tenant B's configuration
5. Reconnect to tenant A and display the list of agents for that endpoint

## Expected output

- A user connected to tenant A should only see **agents belonging to tenant A** for that endpoint
- Agents from tenant B should not appear in the agent list when viewed from tenant A's context
- The platform should handle the configuration conflict on the physical machine (warning, error, or support for co-existing configurations per tenant)

## Actual output

- The user connected to tenant A sees **all agents installed on that endpoint**, including those from tenant B — no tenant filtering is applied on the agent list
- On the physical machine, the configuration is overwritten: only the last configured tenant is operational (the agent only heartbeats to one tenant), making the first tenant's agent appear stale/inactive

## Additional information

- **Dual issue**:
  1. **DB / API layer**: the endpoint's agent collection is not filtered by tenant when loaded, so agents from all tenants are returned indiscriminately
  2. **Filesystem**: the agent configuration on the physical machine is a singleton, installing from a second tenant overwrites the first tenant's config, rendering the first tenant's agent inactive without notification
- **No credential leakage**: each agent has its own token scoped to its tenant, so there is no risk of cross-tenant authentication
- **Suggested fix direction**:
  - Ensure the tenant filter is properly applied when loading the agents collection within an endpoint

Image

Image

Image

impolitepanda:
### Proposed changes

* Changed the register api level function to use the register service method, which will prevent bypassing the multi tenancy safety guards

### Testing Instructions

1. Step-by-step how to test

### Related issues

* Closes #6547 

### Checklist

### Further comments

### DIFF
--- a/openaev-api/src/main/java/io/openaev/rest/executor/ExecutorApi.java
+++ b/openaev-api/src/main/java/io/openaev/rest/executor/ExecutorApi.java
@@ -7,7 +7,9 @@ import static io.openaev.utils.SecurityUtils.validateJFrogUri;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.openaev.aop.AccessControl;
 import io.openaev.context.TenantContext;
-import io.openaev.database.model.*;
+import io.openaev.database.model.Action;
+import io.openaev.database.model.Executor;
+import io.openaev.database.model.ResourceType;
 import io.openaev.database.repository.ExecutorRepository;
 import io.openaev.executors.ExecutorService;
 import io.openaev.rest.catalog_connector.dto.ConnectorIds;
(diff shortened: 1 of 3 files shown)


openaev-api/src/main/java/io/openaev/rest/executor/ExecutorApi.java
```diff
@@ -191,17 +182,26 @@ public Executor registerExecutor(
...
+      InputStream iconStream =
+          icon.isPresent() && "image/png".equals(icon.get().getContentType())
+              ? icon.get().getInputStream()
+              : null;
+      InputStream bannerStream =
+          banner.isPresent() && "image/png".equals(banner.get().getContentType())
+              ? banner.get().getInputStream()
+              : null;
+
+      return executorService.register(
```
corinnekrych, line 196: @impolitepanda i was wondering what the main change with this refactor (which make sense reuse register instead of duplicating the logic)
The main differenrent is with a new create executor in registor we do a setTenantId which was missing in the old duplicated code.

corinnekrych (approved):

openaev-api/src/main/java/io/openaev/rest/executor/ExecutorApi.java
impolitepanda: That's exactly that, yes :) It _should_ fix the issue in the bug, but I didn't test it properly, hence why it's only a draft PR. But if you validate it, then I can merge (+ fix tests + rebase), no problem :)

impolitepanda:
Closed as source issue also closed

PR #6709, opened 2026-07-15
feat(chaining): attack path finding widgets and findings drawer

laugiov:
## What this adds

The Attack Path tab's top-bar finding counters become **clickable widgets**. Clicking one opens a right-side
**findings drawer** listing that category's findings, and clicking a finding **cross-focuses** the graph: the
map centers on the endpoint where it was found, and the execution feed highlights the executions that
discovered it.

This is the "navigate from findings" direction product asked for (US5 of #5048). It reuses the POC's
normalized store from #6647, so the reads stay cheap (the `attackpath_finding` table is far smaller than the
executions).

## Backend

- **Paginated category read** (`AttackPathGraphService.listFindings`): lists a widget category's findings for
  a simulation. A widget maps to a set of finding types (`credentials`, `users = {username, admin_username}`,
  `cves`, `files`); an unknown category returns an empty page.
- **Two reads, not a join**: page the findings (`findPageByTypes`, `EXISTS` + explicit `countQuery`), then
  read that page's producing-execution links (`findExecutionLinks`). Paginating a `finding JOIN
  execution_finding` would split a finding's executions across pages and miscount the total.
- **Honours the graph invariant (CF1)**: the read only lists findings a producing execution links to, so the
  drawer never shows something the graph omits. A stable `ORDER BY (endpointKey, value, id)` keeps scroll
  pagination from repeating or skipping rows.
- **HTTP endpoint**: `GET /simulations/{id}/findings?category=&page=&size=` on `AttackPathApi`, flag-gated by
  the `ATTACK_PATH` preview feature (404 when off), base and tenant-prefixed, size-capped at 200. Tenant
  isolation is enforced by the statement inspector, like every other attack-path read.
- **Credential masking** is server-side: for a `username:password` pair the username is kept and only the
  secret is masked with a fixed-length mask (`admin:••••`), so rows stay distinguishable without leaking the
  secret or its length. The clear secret never leaves the server.
- Execution feed nodes now carry their raw execution id in `ref`, so a drawer item can match and highlight
  its producing executions in the feed.
- The synthetic seed now produces realistic `username:secret` credentials so the masking is meaningful.

## Frontend

- The counter chips for Files, Credentials, Users and CVEs become buttons that open a right-side MUI drawer
  (Ports is dropped, it is not a product widget; Files reads 0 until ingestion adds a `file` finding type).
- The drawer lists each finding's value and endpoint (credentials masked), with the total in the header and a
  `+N more` when the page is shorter than the total. Drawer rows are keyboard-accessible.
- Clicking a finding centers the endpoint node on the map, selects it (which highlights the edges into it),
  loads that endpoint's execution feed, and highlights and scrolls to the producing executions.
- One new i18n key (`No findings`) added to all 9 locales.

## Testing

- Backend: `AttackPath*` suite (62) plus `TenantNonOrmAccessArchTest` green. Covers the paginated read
  (masking, orphan exclusion, shared-value-per-endpoint, stable pagination), the HTTP endpoint (admin
  happy-path, flag-off 404, two-tenant isolation), and the feed cross-focus id.
- Frontend: `check-ts`, `i18n-checker`, eslint `--max-warnings 0`, and the attack-path vitest (12) green,
  including a component test that drives the drawer and asserts the cross-focus wiring.

## Open questions for product (none blocking)

- **Widget count vs drawer total**: the widget counts distinct values, the drawer total counts (value,
  endpoint) rows, so a shared credential reads lower on the widget (6) than in its drawer (9). Reconcile to
  one metric, or label them distinctly?
- **Masking format**: confirm `user:••••` as the final format.
- **Files finding type**: no `file` type exists today (widget reads 0). Real source is a model/ingestion
  decision.
- **Finding uniqueness per (type, value, endpoint)**: the drawer assumes at most one finding row per
  (type, value, endpoint). Worth making a store invariant on the ingestion side.
- **Status granularity**: the endpoint status is a two-state simplification of the product's four-state model.
- **Pagination**: page size and the scroll threshold for the drawer.

## Notes

- Stacked on #6647 (the POC). Base this PR on `lgi/6647-attack-path-execution-step-poc`, not `main`.
- Part of #5048 (Attack Path Creation / Live Tracking), user story US5.

### DIFF
--- a/openaev-api/src/main/java/io/openaev/api/attackpath/AttackPathApi.java
+++ b/openaev-api/src/main/java/io/openaev/api/attackpath/AttackPathApi.java
@@ -14,10 +14,12 @@ import io.openaev.service.attackpath.AttackPathSeedService;
 import io.openaev.service.attackpath.dto.AttackPathDTO;
 import io.openaev.service.attackpath.dto.AttackPathEndpointRelationsDTO;
 import io.openaev.service.attackpath.dto.AttackPathExpandDTO;
+import io.openaev.service.attackpath.dto.AttackPathFindingPageDTO;
 import io.openaev.service.attackpath.dto.AttackPathSeedInput;
 import io.openaev.service.attackpath.dto.AttackPathSeedResultDTO;
 import java.util.List;
 import lombok.RequiredArgsConstructor;
+import org.springframework.data.domain.PageRequest;
 import org.springframework.http.HttpStatus;
 import org.springframework.transaction.annotation.Transactional;
 import org.springframework.web.bind.annotation.GetMapping;
@@ -47,6 +49,9 @@ import org.springframework.web.server.ResponseStatusException;
 public class AttackPathApi extends RestBehavior {
 
   public static final String ATTACK_PATH_URI = "/api/attack-path";
+
+  /** Upper bound on the findings page size, so a client cannot request an unbounded read. */
+  private static final int MAX_FINDINGS_PAGE_SIZE = 200;
 
   private final AttackPathGraphService graphService;
   private final AttackPathSeedService seedService;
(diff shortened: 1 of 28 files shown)


RomuDeuxfois:
@copilot run security reviewer

RomuDeuxfois:
@copilot run security reviewer

PR #6716, opened 2026-07-15
fix(ci): e2e-infra-tests artifact name collision between x64/arm64 legs (#6715)

### LINKED ISSUE #6715: fix(ci): e2e-infra-tests artifact name collision between x64/arm64 legs
## Bug

CI fails intermittently with:

```
Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run
```

## Root cause

In `.github/workflows/_quality-gates.yml`, job `e2e-infra-tests` matrixes over `arch` (`x64` / `arm64`) but both legs call `./.github/actions/e2e-tests` with a hardcoded `browser: chromium` input (not derived from `matrix.arch`):

```yaml
e2e-infra-tests:
  strategy:
    matrix:
      include:
        - arch: x64
          runner: ubuntu-latest
        - arch: arm64
          runner: ubuntu-24.04-arm
  steps:
    - uses: ./.github/actions/e2e-tests
      with:
        playwright-config: playwright.infra.chromium_config.ts
        browser: chromium
```

`.github/actions/e2e-tests/action.yml` names the uploaded Playwright report artifact only from `inputs.browser`:

```yaml
- name: Upload Playwright report (${{ inputs.browser }})
  uses: actions/upload-artifact@v7
  with:
    name: playwright-report-${{ inputs.browser }}
```

Since both matrix legs run inside the **same workflow run** (this job is invoked from `core-ci.yml` via the reusable `_quality-gates.yml`), both the `x64` and `arm64` legs try to upload an artifact named `playwright-report-chromium` — `actions/upload-artifact@v7` (v4 API) treats artifact names as unique per run, causing the 409 on whichever leg uploads second.

All other matrixed jobs in the repo (`e2e-tests`, `api-tests`, nightly `e2e-tests`) already derive the artifact name from a matrix dimension that's actually unique per leg — this job is the outlier.

## Fix

Add an optional `artifact-suffix` input to `.github/actions/e2e-tests/action.yml` and pass `matrix.arch` as the suffix from `e2e-infra-tests`, so the uploaded artifacts become `playwright-report-chromium-x64` / `playwright-report-chromium-arm64`. No other caller is affected (suffix defaults to empty, preserving existing artifact names).

RomuDeuxfois:
## Problem
`actions/upload-artifact@v7` treats artifact-name collisions **within the same workflow run** as a hard `409 Conflict` (unlike v3's silent merge). The `e2e-infra-tests` job in `_quality-gates.yml` runs a 1-dimensional matrix (`arch: x64` / `arch: arm64`), and both legs call the shared `./.github/actions/e2e-tests` composite action with `browser: chromium` hardcoded — so both legs uploaded `playwright-report-chromium` / `playwright-coverage-chromium`, and the second upload failed with the 409.

## Fix
Suffix the artifact names in `.github/actions/e2e-tests/action.yml` with `${{ github.job }}-${{ strategy.job-index }}` instead of the browser name alone:

```
playwright-report-${{ inputs.browser }}-${{ github.job }}-${{ strategy.job-index }}
playwright-coverage-${{ inputs.browser }}-${{ github.job }}-${{ strategy.job-index }}
```

Why this and not a simpler suffix:
- A static per-caller suffix (e.g. `-e2e-infra`) would only disambiguate *between* jobs, not *within* the `e2e-infra-tests` matrix — it wouldn't fix the actual bug.
- `runner.arch` fixes today's collision, but only because the matrix currently has a single dimension (`arch`). If a second dimension is ever added to that matrix (or to any other caller), `runner.arch` alone would collide again — silently reintroducing the 409.
- `strategy.job-index` is GitHub Actions' dedicated, always-unique index per matrix leg within a job, regardless of how many matrix dimensions exist now or later. Combined with `github.job`, it also disambiguates between call sites even if they ever share the same browser input. No maintenance burden if the matrix shape changes.

## Scope / impact
- Single file changed: `.github/actions/e2e-tests/action.yml` (composite action used by `e2e-tests` and `e2e-infra-tests` jobs in `_quality-gates.yml`).
- No change needed to `_quality-gates.yml` itself — this fix is fully contained in the reusable action, so no `workflows` scope permission required.
- Downstream consumer (`codecov` job) downloads via glob pattern `playwright-coverage-*` with `merge-multiple: true` — insensitive to the exact suffix, verified no other hardcoded references to these artifact names exist in the repo.

Fixes #6715.

### DIFF
--- a/.github/actions/e2e-tests/action.yml
+++ b/.github/actions/e2e-tests/action.yml
@@ -17,6 +17,18 @@ inputs:
   browser:
     description: "Browser name (chrome, chromium, webkit, or firefox) — used for artifact naming and browser install."
     required: true
+  artifact-suffix:
+    description: >
+      Optional suffix appended to uploaded artifact names (e.g. a matrix
+      dimension like arch). Needed when a caller job runs this action
+      multiple times with the SAME browser value inside a single matrix
+      (e.g. e2e-infra-tests: x64/arm64 both use browser=chromium) — without
+      a suffix, the parallel legs would upload identically-named artifacts
+      in the same workflow run and actions/upload-artifact would fail with
+      a 409 Conflict. Leave empty when browser alone is already unique per
+      run (default, backward compatible).
+    required: false
+    default: ""
 
 runs:
   using: composite
@@ -243,11 +255,14 @@ runs:
       if: ${{ !cancelled() }}
       uses: actions/upload-artifact@v7
       with:
-        name: playwright-report-${{ inputs.browser }}
+        # artifact-suffix keeps names unique when the same browser value
+        # is used by multiple parallel matrix legs in the same run
+        # (e.g. e2e-infra-tests x64/arm64) — see input description above.
+        name: playwright-report-${{ inputs.browser }}${{ inputs.artifact-suffix != '' && format('-{0}', inputs.artifact-suffix) || '' }}
         path: openaev-front/test-results/
     - name: Upload E2E coverage report (${{ inputs.browser }})
       if: ${{ !cancelled() && inputs.browser == 'chrome' }}
       uses: actions/upload-artifact@v7
       with:
-        name: playwright-coverage-${{ inputs.browser }}
+        name: playwright-coverage-${{ inputs.browser }}${{ inputs.artifact-suffix != '' && format('-{0}', inputs.artifact-suffix) || '' }}
         path: openaev-front/test-results/coverage/
--- a/.github/actions/e2e-tests/action.yml
+++ b/.github/actions/e2e-tests/action.yml
@@ -239,13 +239,6 @@ runs:
       env:
         APP_URL: http://localhost:8080
         E2E_COVERAGE: "true"
-    # ── Artifact naming ─────────────────────────────────────────────
-    # `github.job` + `strategy.job-index` guarantee uniqueness within
-    # the workflow run regardless of how many matrix dimensions the
-    # calling job has (now or in the future). A suffix like
-    # `runner.arch` would only cover today's single-dimension matrix
-    # in e2e-infra-tests and would silently collide again if a second
-    # dimension were ever added.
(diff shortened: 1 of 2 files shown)


RomuDeuxfois:
## ⚠️ Residual risk — flagging explicitly, not hiding it in a code comment

The current fix disambiguates artifact names via `${{ inputs.browser }}-${{ runner.arch }}`. This is **correct for the current matrix** in `_quality-gates.yml`:

- `e2e-tests`: `browser=chrome` vs `browser=webkit` (both `ubuntu-latest`/X64) → already unique on `browser` alone.
- `e2e-infra-tests`: `browser=chromium` on both legs, but `arch=x64` (`ubuntu-latest`) vs `arch=arm64` (`ubuntu-24.04-arm`) → disambiguated by `runner.arch`.

**When this breaks again:** if a future matrix leg shares *both* the same `browser` value *and* the same `runner.arch` as an existing leg (e.g. sharding `e2e-infra-tests` into two `ubuntu-latest`/x64 workers both running `browser: chromium`), the 409 returns — `runner.arch` can't discriminate a matrix dimension it has no visibility into.

**Why not the fully robust fix now:** the structurally correct fix is an explicit `artifact-suffix` input, filled by the *caller* with its actual matrix key — because composite actions do **not** inherit the calling job's `matrix`/`strategy` context (only `github`, `inputs`, `runner`, `env`, `steps` propagate). That requires editing `_quality-gates.yml` (a workflow file), which this PR's token/scope cannot do.

**Follow-up:** once someone with `workflows` scope can edit `_quality-gates.yml`, migrate to an explicit `artifact-suffix` input per matrix leg, making uniqueness the caller's responsibility instead of an inferred side-effect of `runner.arch`. Filing this as a separate hardening issue rather than blocking this minimal fix on it.

guillaumejparis (approved):

PR #6719, opened 2026-07-15
feat(chaining): be able to execute injectors with chaining engine (#6600)

### LINKED ISSUE #6600: feat(chaining): be able to execute injectors with chaining engine
## Use case

When configuring an injector (e.g., Nuclei / Nmap) in Chaining, the user will not have to select exactly one target category: target assets, target asset groups, or manual targets.  

## Proposed solution

In one action, the user could have multiple asset and/or asset group and/or manual items for the injectors target asset.

camrrx:
### Proposed changes

* Manage the injector contrat to execute injectors in chaining 

### Testing Instructions

2. Add asset in scope definition
3. Add an injector (nmap/nuclei/netexec) as an action
4. Launch the simulation

### Related issues

* Closes #6600 

### Checklist

### Further comments


openaev-api/src/main/java/io/openaev/service/chaining/ScopeService.java
```diff
@@ -18,6 +18,31 @@ public class ScopeService {
   private final AssetService assetService;
   private final AssetGroupService assetGroupService;
 
+  /**
+   * Returns the asset groups (ASSET_GROUP_ID allowlist rules) that are not explicitly denied as a
+   * whole group. Used at inject execution time to populate inject.assetGroups.
+   */
+  public List<AssetGroup> getValidAssetGroupsFromScope(String workflowId) {
```
savacano28, line 25: I have created a PrimitiveValueValidator. Can you check if you can reuse those classes?

openaev-api/src/main/java/io/openaev/service/chaining/ScopeService.java
```diff
@@ -116,6 +141,51 @@ private Set<String> getDeniedAssetIds(
...
+   * their CIDR notation as-is, since tools like nmap accept them directly.
+   *
+   * <p>Denylist filtering: an individual IP is removed if it exactly matches a denied IP rule or
+   * falls inside a denied IP_SUBNET rule. A subnet value is removed only on exact match against a
+   * denied IP_SUBNET rule (partial overlap is not computed).
+   */
+  public List<String> getValidIpsFromScope(String workflowId) {
+    List<WorkflowScopeRule> allRules = workflowScopeRuleRepository.findAllByWorkflowId(workflowId);
+
+    Set<String> deniedIps =
```
savacano28: Same here: Check PrimitiveValueValidator, Context :)

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
```diff
@@ -186,31 +198,58 @@ public Optional<Step> run(Step readyStep) throws ChainingException {
...
+
+        // Execute on manual IPs from scope rules
+        if (manualInject != null) {
+          Inject manualInjectBuilt = injectService.createInject(manualInject);
+          if (lastCreatedInjectId == null) {
+            lastCreatedInjectId = manualInjectBuilt.getId();
+          }
+
+          // Only track the manual inject in step data if no asset inject was created
+          if (!hasAssetTargets) {
```
savacano28: We need still this condition? we have already check assetTargets in l223?

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
```diff
@@ -186,31 +198,58 @@ public Optional<Step> run(Step readyStep) throws ChainingException {
...
-              inject.getAssetGroups(),
-              List.of(),
-              true);
-
-      // TODO Check add documents? Executable Payloads
-      // executableInject.addDirectAttachment(inject.getDocuments());
-
-      executor.directExecute(executableInject);
+      // PAYLOAD
+      if (!isExternalInjector) {
```
savacano28: What happens with payload with arguments like targeted-asset?

Image

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
```diff
@@ -170,12 +173,21 @@ public Optional<Step> ready(Step stepTemplate, String input, Workflow workflowRu
...
+    ScopeTargets scope =
+        new ScopeTargets(
+            scopeService.getValidAssets(workflowId),
+            scopeService.getValidAssetGroupsFromScope(workflowId),
+            scopeService.getValidIpsFromScope(workflowId));
+
     // CALL BY QUEUE READY
-    Inject inject = getInjectFromDataStep(readyStep);
-    // CREATE & SAVE INJECT
+    Inject inject = getInjectFromDataStep(readyStep, scope);
```
savacano28: Questions:
* In this getInjectFromDataStep, we add targeted assets for injectors, but later in buildManualTarget we remove these values and keep only IPs from scope rules for injectors, is it normal?
* What happens when we have a payload with a payloadArgument like target-asset?

Todo:
* Add more details about different cases we have: Injectors, payloads, payloads with asset-targets

savacano28:
## Tests

### UI: Forms

**Nuclei**
- We should remove "Link to output" from fields that don't need it.
- Expectations: Should be only "Not vulnerable".

image

**Nmap**
- Form correct.
- Expectations: Should be only "Not vulnerable".

image

** Payload with target-assets**
- Form not correct: we should remove "linked to output" 
- Expectations: OK based on what was defined in the payload

image

openaev-api/src/main/java/io/openaev/service/chaining/ScopeService.java
camrrx: Yes i will use it thanks

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
camrrx: Yes, it's still needed : the two checks serve different purposes. Line 223 guards the asset inject execution. Line 238 guards setInjectId for the manual inject: if an asset inject was already created (line 226), we don't want to overwrite step.data with the manual inject ID. Removing it would cause the step to track the wrong inject ID when both asset and IP targets are present.

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
camrrx: I've added an autolink method to manage the autolink between targeted asset from the injector contract and the primitive type targeted asset. For a payload we have : 
- source asset which is coming from the scope definition
and we could also have : 
- targeted asset which is autolink to "targeted asset" type

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
camrrx: 1) Yes, it's intentional :  

- getInjectFromDataStep populates the inject with assets/asset groups from scope rules : it's a first execution with the assets. 
- buildManualTargetInject creates a separate execution for IP execution, explicitly clearing assets since those IPs don't map to known endpoints.

2) I've add an auto-link to map targeted asset to the primitive type "targeted asset", but I think we should do something in inject service, but it's out of the scope of my PR

EvaE-Filigran (approved):

Issue #205, opened 2022-10-11
Editing challenge leads to errors


SamuelHassine:
## Description

Editing challenge leads to errors:

```
react_devtools_backend.js:4026 TypeError: Cannot read properties of null (reading 'take')
    at openex-4RDBVOFV.js:6914:3958
    at t (openex-4RDBVOFV.js:6914:1377)
    at openex-4RDBVOFV.js:7238:105764
    at Array.map (<anonymous>)
    at $w0 (openex-4RDBVOFV.js:7238:105145)
    at f91 (openex-4RDBVOFV.js:5:53525)
    at va1 (openex-4RDBVOFV.js:5:63102)
    at pN1 (openex-4RDBVOFV.js:9:9809)
    at lN1 (openex-4RDBVOFV.js:9:968)
    at W2t (openex-4RDBVOFV.js:9:891)

```

PR #6729, opened 2026-07-16
feat(chaining): attack-path execution rows with asset and agent details (#5048)

laugiov:
## What this PR does (#203 — asset & agent details)

At each agent-based run, it creates one `attackpath_execution` row per (target, agent). Each row holds the **asset and agent details**: the source (the agent's endpoint), the target (resolved from the payload — a DNS host, a command target, or file assets), and the frozen `source_kind`, `source_asset_id`, `target_kind/key/hostname/ip/platform`, `agent_id/agent_name/agent_privilege`, `payload_name`.

Two parts:
1. **`AttackPathSourceTargetResolver`** — a pure function. It turns an inject into source→target edges.
2. **`AttackPathExecutionIngestionService`** — at RUN (`InjectExecutionStep.run()`), it resolves the edges and creates the rows. The hook is **flag-gated** by `ATTACK_PATH` and **guarded** (a failure never breaks the inject execution).

The row id is **deterministic**: `AttackPathIds.executionNode(injectExecId, targetKey, agentId)`. So a retry writes the same row, never a duplicate.

> The source/target resolution is nominally **#205** ("endpoint & injector definition"). It is covered here, so #205 can be absorbed into #203 or closed.

Tests: `AttackPathSourceTargetResolverTest` (pure) and `AttackPathExecutionIngestionServiceTest` (real Postgres).

---

## For #204 (command, executed_at, traces) and #202 (status, outputs)

This PR writes **only** the asset/agent columns. Your columns (`command`, `terminal_output`, `executed_at`, `prevention_status`, `detection_status`, findings) are **not** written here. Here is what you need to build Phase B on top of it:

- **Same row.** Your Phase B updates the same rows I create. Join on the deterministic id above — so your data lands on the same row, no duplicate.

- **Finding the rows (open point — let's confirm).** In the callback you have `injectId + agentId`. The row id also embeds `targetKey`, which you do not know from the callback, so you cannot rebuild the id. You need a **queryable `inject_id` column** (`simulation_id + agent_id` is too wide — one agent runs many injects). This is the *"il faut mettre à jour l'id de l'inject"* from our chat: it is a **create-time value**, so #203 can set it when it creates the row. I prototyped the column and removed it during the scope split — putting it back on the #203 create is easy. **Let's confirm who adds it.**

- **Where to hook.** Both callback paths — the single one (`InjectExecutionService.handleInjectExecutionCallback`) and the batch one (`BatchingInjectStatusService`) — go through **`InjectExecutionService.processInjectExecutionWithAgent`**. That is the right place to hook. Do **not** use the AOP aspect (`@WorkflowUpdateEvent`): it only has the inject id, not the agent or the traces.

- **Per-agent terminal.** `ExecutionTrace` has an `agent`, a `message`, and a `time`. Filter the inject's traces by agent, order by time, join the messages. **Overwrite** (do not append) so a re-processing stays idempotent.

- **⚠️ Tenant scoping — please read this.** The ingestion runs in a **background/queue context** with **no `TxCtx`**. On the active tables (`attackpath_execution` is in `openaev.tenant.active-tables`), the statement inspector **fails closed** without a scope: your `find` will return **empty in production**. You must set the tenant scope with the background-tx primitive (**PR #6642**, `TenantScopedTransaction.executeNew(TxCtx.forTenant(injectTenant), ...)`). Two warnings:
  - The **test harness does not enable the inspector** (no `active-tables` in the test config). So your tests can be **green even if the scope is wrong**. Activate the inspector in a test to check the real tenant behaviour.
  - My Phase A create has the same issue (its idempotent merge-SELECT fails closed too). I will convert it to the primitive when #6642 lands. We stay on the current base until then.

- **Open question — `executed_at`.** The board lists it under #204 (yours). But the column is **NOT NULL**, so the create cannot leave it empty. I set it as a creation stamp (the RUN time) for now. We should decide together: keep it here, or make it **nullable** and you own it fully.

- **Open point — source endpoint info (a #203 gap).** You noted #203 should also freeze the **source** endpoint's `hostname/ip/platform` (the agent's endpoint, for agent-based runs). Right now the create sets `source_asset_id` + `agent_name` (= the endpoint hostname) but **not** the source `ip/platform`. So we need `source_hostname/ip/platform` columns filled at create — a small #203 addition. Confirm and I add them.

---

## Not in this PR (follow-ups)

- Asset-group expansion (`computeDynamicAssets`) — Phase A covers the inject's **direct** assets only.
- Injector-triggered path — TBD (injectors do not run through chaining yet).
- Tenant scoping of the create — waits for **#6642**, then converted at rebase.
- Performance gate.

> Not production-ready until the tenant scoping (#6642) is applied. It is safe on the base branch because the whole path is behind the `ATTACK_PATH` flag (off by default).

### DIFF
--- a/openaev-api/src/main/java/io/openaev/service/attackpath/ingestion/AttackPathExecutionIngestionService.java
+++ b/openaev-api/src/main/java/io/openaev/service/attackpath/ingestion/AttackPathExecutionIngestionService.java
@@ -66,6 +66,9 @@ public class AttackPathExecutionIngestionService {
     row.setSourceKind(edge.sourceKind());
     row.setSourceInjector(edge.sourceInjector());
     row.setSourceAssetId(edge.sourceAssetId());
+    row.setSourceHostname(edge.sourceHostname());
+    row.setSourceIp(edge.sourceIp());
+    row.setSourcePlatform(edge.sourcePlatform());
     row.setAgentId(edge.agentId());
     row.setAgentName(edge.agentName());
     row.setAgentPrivilege(edge.agentPrivilege());
--- a/openaev-api/src/main/java/io/openaev/service/attackpath/ingestion/AttackPathExecutionIngestionService.java
+++ b/openaev-api/src/main/java/io/openaev/service/attackpath/ingestion/AttackPathExecutionIngestionService.java
@@ -24,7 +24,7 @@ import org.springframework.transaction.annotation.Transactional;
  * Attack-path ingestion — Phase A (issue 5048, #203). At RUN, create one EXECUTION row per resolved
  * edge from the run's source/target resolution, on the store columns the read already consumes. The
  * tenant is set by {@code TenantBaseListener} from the current tenant context. #204/#202 update
- * these rows later (Phase B), found by the same deterministic id.
+ * these rows later (Phase B), found by the queryable {@code (inject_id, agent_id)} written here.
  */
 @Service
 @RequiredArgsConstructor
@@ -56,6 +56,9 @@ public class AttackPathExecutionIngestionService {
     // Deterministic natural key so #204/#202 (Phase B) and retries converge on this same row.
     row.setId(AttackPathIds.executionNode(ctx.injectExecId(), edge.targetKey(), edge.agentId()));
     row.setSimulationId(ctx.simulationId());
+    // Queryable lookup key so #204's per-output update finds the run's rows by (inject id, agent
+    // id).
+    row.setInjectId(ctx.injectExecId());
     row.setStepId(ctx.stepId());
     row.setStepTemplateId(ctx.stepTemplateId());
     row.setExecutedAt(ctx.executedAt());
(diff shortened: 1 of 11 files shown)


openaev-api/src/main/java/io/openaev/service/attackpath/ingestion/AttackPathExecutionIngestionService.java
```diff
@@ -0,0 +1,174 @@
...
+    createRows(
+        context(step, inject, contract), resolver.resolve(resolutionInput(inject, contract)));
+  }
+
+  private ExecutionContext context(Step step, Inject inject, InjectorContract contract) {
+    Payload payload = contract.getPayload();
+    return new ExecutionContext(
+        inject.getExercise() != null ? inject.getExercise().getId() : null,
+        step.getId(),
+        null, // stepTemplateId — wired from the step template later
```
EvaE-Filigran: Step step hold step template id

laugiov: good catch

PR #6732, opened 2026-07-16
feat(chaining): add methods to extract command and execution trace

EvaE-Filigran:
### Proposed changes

- Add support to extract command information and execution traces from inject runs.
- Group execution results by endpoint source to make outputs easier to read and follow.
- Clarify command availability: runs executed with an injector and a non-Command payload may have an empty `command` field, because command content is not available for those payload types.
- Executed at is handle by AttackPathExecution. 

### Testing Instructions

1. Create and run an Chaining simulation with a Command action, then verify command and execution trace information are extract.
2. Create and run an Chaining simulation with a non-Command action, then verify execution traces are present and `command` can be empty.

### Related issues
- #6647 

### Checklist

### Further comments

- For some executions, the `command` field can be empty by design when the payload or injector does not expose command.

### DIFF
--- a/openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
+++ b/openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
@@ -528,6 +528,9 @@ public class InjectExecutionStep implements ActionStep {
 
     InjectStatus status = inject.getStatus().get();
     StatusPayload statusPayload = status.getPayloadOutput();
+    if (statusPayload == null || statusPayload.getPayloadCommandBlocks() == null) {
+      return "";
+    }
     StringBuilder command = new StringBuilder();
     statusPayload
         .getPayloadCommandBlocks()
--- a/openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
+++ b/openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
@@ -549,7 +549,9 @@ public class InjectExecutionStep implements ActionStep {
     if (inject.getInjector() == null) {
       executionTraces.forEach(
           executionTrace -> {
-            // TODO BUILD INDEX
+            if (executionTrace.getAgent() == null || executionTrace.getAgent().getAsset() == null) {
+              return;
+            }
             String agentId =
                 executionTrace.getAgent().getId() + executionTrace.getAgent().getAsset().getId();
             tracesByEndpointSource
--- a/openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
+++ b/openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
@@ -523,6 +523,56 @@ public class InjectExecutionStep implements ActionStep {
     return setField(dataStep, "inject_id", injectId);
   }
 
+  private String getCommand(Inject inject){
+      if(inject.getStatus().isEmpty()) return "";
+
+      InjectStatus status = inject.getStatus().get();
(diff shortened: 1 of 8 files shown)


openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
```diff
@@ -523,6 +523,58 @@ private String setInjectId(String injectId, String dataStep) {
...
+    if (inject.getStatus().isEmpty()) return tracesByEndpointSource;
+
+    InjectStatus status = inject.getStatus().get();
+    List<ExecutionTrace> executionTraces = status.getTraces();
+
+    if (inject.getInjector() == null) {
+      executionTraces.forEach(
+          executionTrace -> {
+            // TODO BUILD INDEX
+            String agentId =
```
laugiov, line 558: Heads up: error/info/warning traces have a null agent (see InjectStatus.addErrorTrace/addInfoTrace/addWarningTrace), so getAgent().getId() will NPE here. Could you skip agent-less traces and add a test with a trace that has no agent?

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
```diff
@@ -523,6 +523,58 @@ private String setInjectId(String injectId, String dataStep) {
     return setField(dataStep, "inject_id", injectId);
   }
 
+  private String getCommand(Inject inject) {
+    if (inject.getStatus().isEmpty()) return "";
+
+    InjectStatus status = inject.getStatus().get();
+    StatusPayload statusPayload = status.getPayloadOutput();
```
laugiov, line 530: payloadOutput is not initialized, so it can be null and getPayloadCommandBlocks() would NPE. Can it be null here? A null check returning "" would be safer.

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
```diff
@@ -523,6 +523,58 @@ private String setInjectId(String injectId, String dataStep) {
...
+            });
+    return command.toString();
+  }
+
+  private Map<String, StringBuilder> getExecutionTracesByEndpointIndex(Inject inject) {
+    Map<String, StringBuilder> tracesByEndpointSource = new HashMap<>();
+    if (inject.getStatus().isEmpty()) return tracesByEndpointSource;
+
+    InjectStatus status = inject.getStatus().get();
+    List<ExecutionTrace> executionTraces = status.getTraces();
```
laugiov, line 550: Small question: are these traces time-ordered? 
There's no @OrderBy, so the terminal lines could come out unordered when loaded from the DB.

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
EvaE-Filigran: done

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
EvaE-Filigran: yes it can be null, fix done

openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
EvaE-Filigran: You’re right, without explicit ordering traces could come back unordered from DB.
I fixed this by adding deterministic ordering both at mapping and query levels. I also added integration tests to verify ordering on all updated queries.

laugiov (approved):

PR #6751, opened 2026-07-17
chore(multitenancy): api v2 collectors (#6399)

### LINKED ISSUE #6399: feat(multitenancy): API isolation v2 collector GCatalog
**Epic**: GH issue epic generated by script: [feat(multitenancy): epic for refactor for tenant scoped API](https://github.com/OpenAEV-Platform/openaev/issues/6393)

---

## Context

`CollectorApi` is mapped from codebase analysis to tenant table(s): `collectors, collector_types`.
Current state: activation = activable now, isolation mechanism = v1, test coverage = TenantIsolation nested exists.

## Definition of Done

- [ ] API behavior is aligned with v2 tenant-scoped isolation expectations
- [ ] Existing v1-only assumptions are removed where required for this endpoint
- [ ] Update tenant isolation integration tests for this API (`TenantIsolation` nested class or dedicated isolation test file)
- [ ] Add/validate tenant A vs tenant B isolation and out-of-rights access behavior
- [ ] Wire `TenantWriteScopeResolver` on the create/import endpoints (attribution: row gets the scope's tenant; out-of-scope supplied tenant → 400; multi-tenant scope without selector → 400)
- [ ] Coordinated go-live: remove the table's v1 `@Filter` and add it to `openaev.tenant.active-tables` in the same change (avoid the window where v1 and v2 contradict)

corinnekrych:
### Proposed changes

API v2 fro `collectors` table

#### Integrations > Collectors page: CRUD catalog
1.List collectors — Navigate to Integrations > Collectors; verify the list loads correctly with all expected collectors
2.View collector detail — Click a collector; verify its detail page (status, type, last run) displays
3.Collector image — Verify collector icons/images render (tests the getCollectorImageById endpoint)
4.Delete a collector — Delete a test collector; confirm it disappears from the list
Logged a bug for this flow https://github.com/OpenAEV-Platform/openaev/issues/6825  (outside scope of this PR)
5.Update a collector — Edit collector config (if UI allows); save and verify changes persist
6. tenant scope: create 2 tenant, verify we can install Crowstrike collector isnatnce on both tenants

Step tested

- create tenant A
- create tenant B
- add collector crowdstrike to tenant B => check isolation
- add collector crowdstrike to tenant A => check multiple instances can be associated to different tenants
- delete collector in tenant A

##### Runtime: Open AEV agent + test ExpectationsExpirationManagerJob
in `application-dev.properties`
```
collector.expectations-expiration-manager.expiration-time=120
openaev.expectation.technical.expiration-time=120
# 1h(3600  seconds)  ? asset-level expectations
collector.expectations-expiration-manager.expiration-time-for-asset=60
# job runs every 60s
collector.expectations-expiration-manager.interval=60
# disable to stop the job
collector.expectations-expiration-manager.enable=true
# Logging for ExpectationsExpirationManager
logging.level.io.openaev.collectors.expectations_expiration_manager=debug
```
- create an atomic testing
- create a threat arsenal not detected - echo
- update detection/prevention with expiration time of 1min
- launch atomic testing
See debug trace once the job is picking up:
```
[OpenAEV API] [TaskScheduler12] e.s.ExpectationsExpirationManagerService : Found 4 pending expired expectations for tenant 90c449fe-3480-4158-bd61-88f86a9c3b5a (expirationTime=120s, assetExpirationTime=60s)
[OpenAEV API] [TaskScheduler12] e.s.ExpectationsExpirationManagerService : Expired 4 expectations for tenant 90c449fe-3480-4158-bd61-88f86a9c3b5a
[OpenAEV API] [TaskScheduler12] i.o.c.e.ExpectationsExpirationManagerJob : Finished expectations expiration for tenant 90c449fe-3480-4158-bd61-88f86a9c3b5a
[OpenAEV API] [TaskScheduler12] i.o.c.e.ExpectationsExpirationManagerJob : ExpectationsExpirationManagerJob completed
```
Screenshot 2026-07-21 at 14 51 06

##### Runtime Crowdstrike exec + Crowdstrike detection
Pre-requisites: 
* have a local VM running with crowstrike installed
Test:
- create tenant B
- add collector crowdstrike to tenant B 
- create payload with:
```
bash -c # Mimics a known malware dropper pattern mkdir -p /tmp/.hidden && curl -o /tmp/.hidden/update http://crowdstrike.com/test && chmod +x /tmp/.hidden/update
```
- create an atomic testing with this payload, select the VM
- launch atomic testing
Check attactkis detected

Screenshot 2026-07-22 at 11 55 09

### Related issues

* Closes https://github.com/OpenAEV-Platform/openaev/issues/6399
*

### Checklist

### Further comments


corinnekrych:

corinnekrych:

corinnekrych:

openaev-model/src/main/java/io/openaev/database/repository/CollectorRepository.java
```diff
@@ -18,6 +18,10 @@ public interface CollectorRepository
 
   Optional<Collector> findByIdAndTenantId(@NotNull String id, @NotNull String tenantId);
 
+  /** Finds a collector by its ID only. Tenant scoping is handled by the v2 SQL inspector. */
+  @Query("SELECT c FROM Collector c WHERE c.id = :id")
+  Optional<Collector> findByCollectorId(@Param("id") @NotNull String id);
```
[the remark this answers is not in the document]
corinnekrych: fixed in 4ed580257

openaev-api/src/main/java/io/openaev/rest/collector/CollectorApi.java
```diff
@@ -96,7 +101,8 @@ private Collector updateCollector(
       resourceId = "#collectorId",
       actionPerformed = Action.READ,
       resourceType = ResourceType.COLLECTOR)
-  public Collector getCollector(@PathVariable String collectorId) {
+  // TxCtx scopes this read to the caller's tenants via the inspector.
+  public Collector getCollector(TxCtx ctx, @PathVariable String collectorId) {
     return collectorService.collector(collectorId);
```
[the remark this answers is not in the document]
corinnekrych: fixed in 4ed580257

openaev-api/src/main/java/io/openaev/rest/collector/CollectorApi.java
```diff
@@ -194,8 +202,10 @@ public Collector updateCollector(
           "Removes a registered collector. Intended for stopped collectors that no longer ping;"
               + " an active collector re-registers on its next heartbeat.")
   @Transactional(rollbackFor = Exception.class)
-  public void deleteCollector(@PathVariable String collectorId) {
-    collectorRepository.deleteByIdAndTenantId(collectorId, TenantContext.getCurrentTenant());
+  public void deleteCollector(TxCtx ctx, @PathVariable String collectorId) {
+    // TxCtx scopes this delete to the caller's tenants; the inspector rewrites the DELETE the
+    // same way it rewrites SELECTs, so a delete outside the scope matches no row and is a no-op.
+    collectorRepository.deleteByCollectorId(collectorId);
```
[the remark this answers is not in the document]
corinnekrych: 4ed580257

openaev-api/src/main/java/io/openaev/config/TenantStatementInspector.java
```diff
@@ -258,7 +260,16 @@ private void filterTables(PlainSelect select) {
    * rejected.
    */
   private FromItem filterFromItem(FromItem item) {
-    if (item instanceof ParenthesedSelect) {
+    // Sub-selects and lateral sub-selects are never real tables; they do not need tenant
+    // filtering. A LATERAL table function (e.g. "LEFT JOIN LATERAL jsonb_array_elements(...)")
+    // unnests a column of the row already being joined, never a whole table, so it is safe too.
+    // A non-lateral table function (e.g. "CROSS JOIN generate_series(1, 10)") is NOT unnesting an
+    // existing row and is not a shape this rewriter has reviewed; it stays rejected.
```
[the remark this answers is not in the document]
corinnekrych: 28e1f1781

openaev-model/src/main/java/io/openaev/database/repository/CollectorRepository.java
```diff
@@ -49,6 +53,14 @@ WHERE c.collectorType IN (
       value = "DELETE FROM collectors WHERE collector_id = :id AND tenant_id = :tenantId")
   void deleteByIdAndTenantId(@Param("id") String id, @Param("tenantId") String tenantId);
 
+  /**
+   * Deletes a collector by its ID only. Tenant scoping is handled by the v2 SQL inspector, which
+   * rewrites this DELETE the same way it rewrites SELECTs on active tables.
+   */
+  @Modifying
+  @Query("DELETE FROM Collector c WHERE c.id = :id")
+  void deleteByCollectorId(@Param("id") String id);
```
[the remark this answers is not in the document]
corinnekrych: fixed in https://github.com/OpenAEV-Platform/openaev/commit/4ed5802574ee1c3ab7838b95aa4d69c2b3283dc6

openaev-model/src/main/java/io/openaev/database/repository/CollectorRepository.java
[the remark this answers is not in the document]
RomuDeuxfois: Could this still happen, though ?

openaev-model/src/main/java/io/openaev/database/repository/CollectorRepository.java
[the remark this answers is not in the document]
RomuDeuxfois: Same here

openaev-api/src/main/java/io/openaev/rest/collector/service/CollectorService.java
```diff
@@ -77,9 +76,7 @@ protected List<Collector> getAllConnectors() {
 
   @Override
   protected Collector getConnectorById(String collectorId) {
-    return collectorRepository
-        .findByIdAndTenantId(collectorId, TenantContext.getCurrentTenant())
-        .orElse(null);
+    return collectorRepository.findByCollectorId(collectorId).orElse(null);
```
RomuDeuxfois, line 81: **TODO**: Something seems off to me here. Since Collector has a composite key, what happens if the TenantContext gets populated with two tenants? Wouldn't that give us two potential/ambiguous collectors?

corinnekrych: The composite PK concern is valid — here's how we've addressed it:
1.@RequireTenantSelector on the TxCtx parameter ensures a tenant ID is always present in the URL path. Without an explicit selector, the request is rejected with 400 before reaching the service layer. (it was already covered in the existing code)
2.To cover future multi-tenant selection (multiple tenants in the X-Tenant-Ids header, coming in a future version), all single-entity endpoints explicitly call writeScopeResolver.tenantForWrite(ctx, null) which returns 400 if multiple tenants are in scope — then use findById(ConnectorCompositeId.of(id, tenantId)) for an unambiguous full-PK lookup. (added with [442bb89](https://github.com/OpenAEV-Platform/openaev/pull/6751/commits/442bb89ee98941a9c02fe21555adbb1128a609e8) )

openaev-api/src/test/java/io/openaev/architecture/TenantActiveTableAccessArchTest.java
```diff
@@ -157,4 +164,27 @@ static void every_active_table_is_guarded(JavaClasses classes) throws Exception
...
+
+  @ArchTest
+  static final ArchRule collectors_repository_access_is_reviewed =
+      noClasses()
+          .that()
+          .doNotBelongToAnyOf(
+              // TxCtx-carrying entrypoints, pinned by TenantScopedEntrypointsTxCtxArchTest:
+              CollectorApi.class,
+              InjectExpectationTraceApi.class,
+              // Services behind those handlers; every caller is a wired handler:
```
laugiov: This comment is the root cause of a production break: it assumes every caller of CollectorService is a wired handler, but the rule only checks direct CollectorRepository dependencies, so service callers escape it. With collectors now fail-closed, several unwired readers break:

- PayloadApi#upsertPayload (no TxCtx) reaches PayloadUpsertService line 65: every payload pushed by a collector fails with "Collector not found". This is the hot path of the CrowdStrike and Sentinel collectors.
- PayloadApi#collectorsFromPayload and AtomicTestingApi#collectorsFromAtomicTesting: silently empty lists.
- InjectService.runChecks (securityPlatformCollectors) from scenario/exercise/inject handlers: health checks silently degrade.

CI stays green because the test classpath keeps the allowlist empty, so none of these paths run with collectors active. Suggested fix: wire TxCtx on these handlers, pin them in the entrypoints arch test, and guard CollectorService itself here so the next unwired caller fails the build.

openaev-api/src/main/java/io/openaev/scheduler/jobs/InjectsExecutionJob.java
```diff
@@ -452,7 +455,27 @@ public void execute(JobExecutionContext jobExecutionContext) throws JobExecution
                     .forEach(
                         executableInject -> {
                           try {
-                            this.executeInject(executableInject);
+                            String tenantId =
+                                executableInject.getInjection().getInject().getTenant().getId();
+                            // Scope the transaction so the v2 inspector can resolve
+                            // can_access_tenant for activated tables (e.g. collectors)
+                            tenantTx.execute(
```
laugiov, line 452: The transaction wraps the whole executeInject, which includes the executor-context call and the RabbitMQ publish, and it runs inside nested parallelStream. So each in-flight inject holds one DB connection (pool max 20) through network I/O: an inject burst starves the pool and takes the HTTP path down with it.

Same boundary, second effect: the EXECUTING status and the expectations now commit AFTER the publish. A fast injector can call back before they are visible.

Suggested fix: only computeAndSaveExpectations needs the collectors scope. Scope that DB section, commit, then publish outside the transaction. One change, both problems gone.

corinnekrych: Indeed opening a transaction at that level, for the case of external connector (internal path has no i/o rabbit) does not help the DB connection issue. 
computeAndSaveExpectations is not the only place where we need DB scope (collector, executor - next PR). We will need a bigger refactor to fix it.
Proposed solution: split in 3 phases:
```
// In Executor (inject TenantScopedTransaction)
public void directExecute(ExecutableInject executableInject) throws Exception {
    String tenantId = executableInject.getInjection().getInject().getTenant().getId();
    TxCtx ctx = TxCtx.forTenant(tenantId);

    // Phase 1: DB writes (status + validation)
    InjectData data = tenantTx.execute(ctx, () -> prepareExecution(executableInject));

    // Phase 2: network I/O — no connection held
    rabbitmqService.publish(data.injectorId(), data.jsonInject());

    // Phase 3: DB writes (expectations + final status)
    tenantTx.execute(ctx, () -> finalizeExecution(data));
}
```
However this is a much bigger refactor, phase1 is still heavy... More detailles analysis
```
InjectsExecutionJob.executeInject() (the current outer wrapper):
checkErrorMessagesPreExecution()  // DB read — Phase 1
healthCheckUtils.runContentChecks() // pure/no DB — pre-check, no tx needed
executor.execute(executableInject)  // → delegates to Executor.execute()
Executor.execute() line by line:
inject.getInjectorContract()        // already loaded (in-memory) — no DB
inject.getInjector()                // already loaded (in-memory) — no DB
injectorRepository.findByTypeAndTenantId() // DB read — Phase 1
connectorInstanceService.hasStartedConnectorInstance() // DB read — Phase 1
injectStatusService.initializeInjectStatus() // DB write — Phase 1
executionExecutorService.launchExecutorContext() // heavy DB: agents, endpoints,
                                                  // traces, executor context — Phase 1
Executor.executeExternal():
executableInjectDTOMapper.toExecutableInjectDTO() // serialization — no DB
injectStatusRepository.findByInjectId()           // DB read — Phase 1
injectService.resolveAllAssetsToExecute()         // DB read — Phase 1
injectExpectationService.computeAndSaveExpectations() // DB write — Phase 1 (last)
                                                      // ← COMMIT HERE
rabbitmqService.publish()                         // 🚫 network — Phase 2 (no tx)
injectStatus.addInfoTrace()                       // in-memory mutation
injectStatusRepository.save(injectStatus)         // DB write — Phase 3
```

@laugiov tbh not sure which path to go to keep this PR is scope. let's discuss it further.

openaev-api/src/test/java/io/openaev/architecture/TenantActiveTableAccessArchTest.java
corinnekrych: fixed in 43e7c092e

corinnekrych:

corinnekrych:

openaev-api/src/main/java/io/openaev/scheduler/jobs/InjectsExecutionJob.java
corinnekrych: I've opened https://github.com/OpenAEV-Platform/openaev/issues/6968 to track it and attached the ticket to the issue for v2 api migratino

openaev-api/src/main/java/io/openaev/collectors/expectations_expiration_manager/ExpectationsExpirationManagerJob.java
```diff
@@ -49,20 +49,26 @@ public void registerForTenant(String tenantId) throws Exception {
...
-        TenantContext.clearCurrentTenant();
-      }
-    }
+    log.debug("ExpectationsExpirationManagerJob starting (interval={}s)", config.getInterval());
+    tenantTx.forEachTenant(
+        scope -> {
+          // Bridge: set TenantContext so that the v1 Hibernate @Filter (enabled by
+          // HibernateFilterTransactionAspect) keeps working for tables not yet on v2.
+          // Once all tables touched by this job are activated on v2, remove this line.
+          String tenantId = ((TxCtx.Restricted) scope).tenantIds().getFirst();
```
damgouj, line 58: question: @laugiov could you explain the tenantIds please ? I don't understand why we do a getfirst here, thanks !

openaev-api/src/main/java/io/openaev/collectors/expectations_expiration_manager/ExpectationsExpirationManagerJob.java
```diff
@@ -49,20 +49,26 @@ public void registerForTenant(String tenantId) throws Exception {
...
+          TenantContext.setCurrentTenant(tenantId);
+          try {
+            log.debug("Processing expectations expiration for tenant {}", tenantId);
+            // Detection & Prevention expectation expiration
+            this.fakeDetectorService.computeExpectations(tenantId);
+            // Heartbeat: surface the run as the collector's last execution so the UI
+            // shows a truthful liveliness signal for this built-in collector.
+            this.collectorService.updateLastExecution(config.getId(), tenantId);
+            log.debug("Finished expectations expiration for tenant {}", tenantId);
+          } finally {
```
damgouj, line 68: suggestion: you removed the catch exception here

openaev-api/src/main/java/io/openaev/rest/collector/CollectorApi.java
```diff
@@ -97,8 +103,12 @@ private Collector updateCollector(
       resourceId = "#collectorId",
       actionPerformed = Action.READ,
       resourceType = ResourceType.COLLECTOR)
-  public Collector getCollector(@PathVariable String collectorId) {
-    return collectorService.collector(collectorId);
+  // Collector uses a composite PK (collector_id, tenant_id); the same ID exists per tenant.
+  // Require a selector so findByCollectorId never returns multiple rows under a multi-tenant scope.
+  public Collector getCollector(
+      @RequireTenantSelector TxCtx ctx, @PathVariable String collectorId) {
+    String tenantId = writeScopeResolver.tenantForWrite(ctx, null);
```
damgouj, line 110: question: could you explain again please ? Not sure to understand sorry.

damgouj (changes requested):

openaev-api/src/main/java/io/openaev/collectors/expectations_expiration_manager/ExpectationsExpirationManagerJob.java
laugiov: @damgouj 
forEachTenant gives the callback a scope with exactly one tenant (the loop builds it with TxCtx.forTenant(id)), so the cast is safe and getFirst() just means "the tenant of this iteration".

We only need the raw id to feed TenantContext on the next line: that is the v1 bridge for the tables this job touches that are not migrated yet. Once they are on v2, both lines disappear (see the comment above).

Fair point that the cast is noisy: the loop could hand the id directly to the callback. Noted as a small improvement on the primitive, outside this PR.

corinnekrych: @laugiov i've included a fix for it in my next PR

openaev-api/src/main/java/io/openaev/rest/collector/CollectorApi.java
corinnekrych: writeScopeResolver.tenantForWrite(ctx, null); is usually used when we do POST at creation time we need to ensure we have only one tenant in scope because we want each entiy to be attached to only one tenant. Here we reused the the reolver to ensure we have only one tenant in scope because we want this APO to return only one Collector (composite key - collectedId is a shared among several tenat, but with teantID this is uniquely scoped).
tl;dr: use `writeScopeResolver.tenantForWrite` to ensure only one tenant is in scope, the method throw relevant exception if this is not the case.

openaev-api/src/main/java/io/openaev/collectors/expectations_expiration_manager/ExpectationsExpirationManagerJob.java
corinnekrych: it's not lost, it moved — into forEachTenant itself.
Before, each caller had to hand-roll its own try { ... } catch (Exception e) { log.error(...) } around one tenant's work to keep other tenants going. Now TenantScopedTransaction.forEachTenant does exactly that centrally (TenantScopedTransaction.java:143-151):
for (String tenantId : tenantIds) {
  try {
    execute(scope, () -> work.accept(tenantId));
  } catch (RuntimeException failure) {
    failures.put(tenantId, failure);
    log.warn("forEachTenant: tenant {} failed, continuing with the others", tenantId, failure);
  }
}
So per-tenant isolation (one tenant's failure doesn't block the others) is preserved — it's just enforced once, in the primitive, for every caller, instead of being duplicated (and potentially inconsistently implemented) in each job.
If we'd kept the job's own catch (Exception e) inside the lambda, it would swallow the exception before forEachTenant ever sees it — defeating its aggregation, since forEachTenant re-throws one IllegalStateException at the end with all per-tenant failures attached as suppressed causes (that's the "1 of 1 tenant(s) failed" log you saw earlier). That single aggregate is intentional: it surfaces failures to the scheduler exactly once per run instead of leaving them silently logged per tenant with no visibility that the overall run had a problem.

damgouj (approved):
Tests OK

PR #6815, opened 2026-07-21
fix(chaining): add expectation update

EvaE-Filigran:
### Proposed changes

* Chaining step output now includes inject status and expectation updates (detection, prevention, vulnerability) so workflow progression can react to fresh expectation data.
* Attack path execution now supports vulnerability status end-to-end (model + seed + DB migration), and workflow progress evaluation is more robust when workflow runs are detached or incomplete.

### Testing Instructions

1. Run backend tests for workflow/chaining behavior and ensure they pass.
2. Apply migration and verify the new vulnerability status column exists.
3. Trigger an inject execution and confirm step output includes expectation updates and inject status fields.

### Related issues

* #6647 
* Related to chaining expectation propagation and attack path status tracking

### Checklist

### Further comments

This implementation focuses on functional consistency between expectation updates and chaining step outputs, while keeping workflow progress evaluation safer in asynchronous execution contexts.


laugiov (approved):
The expectation status in the step output and the detached-workflow reload are solid wins.

One heads-up for later (not blocking this PR): when you wire the attack-path write-back (the commented syncAttackPathExecutionStatus), the UPDATEs run on attackpath_execution, which is now tenant-active on our branch. The statement inspector adds its own can_access_tenant check on top of your e.tenant.id = :tenantId filter, and from the run flow no tenant scope is set, so the update would silently match zero rows. The create path solves this by writing through the primitive: tenantTx.executeNew(TxCtx.forTenant(injectTenantId), ...).

Heads-up 2: this branch is behind ours now (the MT v2 conversion plus a consolidated migration), so you'll need a rebase before merging.

Issue #6830, opened 2026-07-21
fix(executors): improve CrowdStrike error readability


GaetanSantucci:
### Context

During CrowdStrike execution, we observed errors that were not very actionable, including:

- `Unexpected character '<' (code 60)` (non-JSON response, often HTML),
- or valid JSON responses containing `errors` (e.g., `401 Unauthorized`) without entering the `catch` block.

Example payload returned by CrowdStrike:
```json
{
  "meta": {
    "trace_id": "88338d79-4db9-4cb4-909b-dcbc74709b36"
  },
  "errors": [
    {
      "code": 401,
      "message": "Unauthorized: Please provide trace-id='88338d79-4db9-4cb4-909b-dcbc74709b36' to support"
    }
  ]
}

PR #6832, opened 2026-07-21
fix(crowdstrike): propagate CrowdStrike API errors as exceptions (#6830)

GaetanSantucci:
CrowdStrike integration errors were difficult to troubleshoot:
- non-JSON responses could lead to generic parsing errors (`Unexpected character '<' (code 60)`),
- valid JSON error payloads (e.g. `401 Unauthorized`) were parsed successfully but not treated as functional failures,
- `trace_id` returned by CrowdStrike support payloads was not surfaced in exception messages.

#### `openaev-api/src/main/java/io/openaev/executors/crowdstrike/model/CrowdstrikeMeta.java`
- Added mapping for `trace_id` (`traceId`) to expose CrowdStrike support correlation id.

#### `openaev-api/src/main/java/io/openaev/executors/crowdstrike/model/ResourcesGroups.java`
- Added `meta` field so host-group responses can carry `trace_id`.

#### `openaev-api/src/main/java/io/openaev/executors/crowdstrike/client/CrowdStrikeExecutorClient.java`
- Added functional error guard `throwIfCrowdStrikeErrors(...)`.
- Updated `hostGroup(...)` to fail fast when CrowdStrike returns `errors[]`, including `trace_id` in the thrown message when available.
- Hardened `authenticate()`:
  - throw `ExecutorException` when authentication returns errors,
  - throw `ExecutorException` when access token is missing/blank,
  - removed silent `return` paths that allowed execution to continue with invalid auth state.

### Expected behavior after this change

- If CrowdStrike returns:
  ```json
  {
    "meta": {"trace_id":"..."},
    "errors":[{"code":401,"message":"Unauthorized..."}]
  }

### DIFF
--- a/openaev-api/src/main/java/io/openaev/executors/crowdstrike/client/CrowdStrikeExecutorClient.java
+++ b/openaev-api/src/main/java/io/openaev/executors/crowdstrike/client/CrowdStrikeExecutorClient.java
@@ -138,10 +138,16 @@ public class CrowdStrikeExecutorClient {
   }
 
   public ResourcesGroups hostGroup(String hostGroup) {
-    String jsonResponse;
     try {
-      jsonResponse = this.get(HOST_GROUPS_URI + "?ids=" + hostGroup);
-      return this.objectMapper.readValue(jsonResponse, new TypeReference<>() {});
+      String jsonResponse = this.get(HOST_GROUPS_URI + "?ids=" + hostGroup);
+      ResourcesGroups resourcesGroups =
+          this.objectMapper.readValue(jsonResponse, new TypeReference<>() {});
+      throwIfCrowdStrikeErrors(
+          "hostGroup",
+          "hostGroupId=" + hostGroup,
+          resourcesGroups.getErrors(),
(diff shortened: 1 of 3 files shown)


openaev-api/src/main/java/io/openaev/executors/crowdstrike/client/CrowdStrikeExecutorClient.java
```diff
@@ -255,19 +261,44 @@ private void authenticate() throws IOException {
...
+            CROWDSTRIKE_EXECUTOR_NAME);
       }
       this.token = auth.getAccess_token();
       this.lastAuthentication = Instant.now();
     } catch (IOException e) {
       throw new ClientProtocolException("Unexpected response", e);
     }
   }
+
+  private void throwIfCrowdStrikeErrors(
```
damgouj, line 268: suggestions:
- Maybe you can add it for other API calls ?
- You can refacto the for part and the string builder with the logErrors method I think

damgouj (changes requested):
Thanks for the improvement, just a quick comment ;)

damgouj (approved):

PR #6874, opened 2026-07-23
feat(ci): add docs deployment workflow (#6873)

### LINKED ISSUE #6873: ci(docs): automate documentation deployment on release
## Context

Documentation is currently deployed manually via `mike deploy --push`. This should be automated as part of the release process and also available as a manual trigger.

## Expected behavior

- Documentation is automatically published to GitHub Pages when a new release is performed via the release orchestrator
- Documentation can also be published manually via `workflow_dispatch` from the Actions tab

RomuDeuxfois:
### Related issues

* Closes https://github.com/OpenAEV-Platform/openaev/issues/6873

### DIFF
--- a/.github/workflows/docs-deploy.yml
+++ b/.github/workflows/docs-deploy.yml
@@ -9,11 +9,6 @@ on:
         description: 'Documentation version label (e.g. "2.260717.0")'
         required: true
         type: string
-      set_latest:
-        description: 'Update the "latest" alias to point to this version'
-        required: false
-        type: boolean
-        default: true
     secrets:
       RELEASE_TOKEN:
         required: false
@@ -23,11 +18,6 @@ on:
         description: 'Documentation version label (e.g. "2.260717.0")'
         required: true
         type: string
-      set_latest:
-        description: 'Update the "latest" alias to point to this version'
-        required: false
-        type: boolean
-        default: true
 
 concurrency:
   group: docs-deploy
@@ -45,6 +35,7 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v7
         with:
+          ref: ${{ inputs.version }}
           fetch-depth: 0 # full history required by mike and git-based plugins
           token: ${{ secrets.RELEASE_TOKEN || secrets.GITHUB_TOKEN }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/docs-deploy.yml
```diff
@@ -0,0 +1,77 @@
...
+      - name: Configure git identity
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+
+      - name: Deploy with mike
+        working-directory: docs
+        env:
+          MKDOCS_GIT_COMMITTERS_APIKEY: ${{ secrets.RELEASE_TOKEN || secrets.GITHUB_TOKEN }}
+          VERSION: ${{ inputs.version }}
```
antoinemzs, line 83: Sorry for the added complexity but it would be great to check the version is among the existing tags

.github/workflows/docs-deploy.yml
```diff
@@ -0,0 +1,77 @@
...
+          SET_LATEST: ${{ inputs.set_latest }}
+        run: |
+          if [ "$SET_LATEST" = "true" ]; then
+            mike deploy --push --update-aliases "$VERSION" latest
```
antoinemzs, line 87: perhaps an opportunity to remove the `SET_LATEST` arg, and figure this out based on the list of exoisting tags.

There is an action for "find latest tag" which could then assert whether the input version is indeed the latest tag

.github/workflows/docs-deploy.yml
```diff
@@ -0,0 +1,77 @@
...
+      - name: Install documentation dependencies
+        working-directory: docs
+        run: pip install -r requirements.txt
+
+      - name: Configure git identity
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+
+      - name: Deploy with mike
```
antoinemzs, line 79: It looks like it would be a good idea to checkout the git repo at the same tag as was supplied in the input. Thoughts?

.github/workflows/docs-deploy.yml
RomuDeuxfois: To be sure we are not deploying something wrong ?

.github/workflows/docs-deploy.yml
RomuDeuxfois: This can simplify a lot, I agree. Good point :)

.github/workflows/docs-deploy.yml
RomuDeuxfois: We can have some delay between the release process and the doc process, so.. Yes good point.

.github/workflows/docs-deploy.yml
antoinemzs: yes, to prevent a typo and ensure the tag is an actual released version

.github/workflows/docs-deploy.yml
antoinemzs: do you mean that we change the docs and publish after the actual release? If so (maybe that's not your point) then that could be a good process requirement to have all the relevant docs changes tagged at the same time as the code

RomuDeuxfois:
@antoinemzs I tried to address your comments.
PR ready for a second round.

.github/workflows/docs-deploy.yml
```diff
@@ -0,0 +1,82 @@
...
+      - name: Resolve latest tag
+        id: latest-tag
+        env:
+          VERSION: ${{ inputs.version }}
+        run: |
+          if ! git tag --list "$VERSION" | grep -q .; then
+            echo "::error::Tag '${VERSION}' does not exist. The version must match an existing git tag."
+            exit 1
+          fi
+          LATEST_TAG=$(git tag --sort=-v:refname | head -n1)
```
antoinemzs: why not reuse the find-latest-tag github action?

RomuDeuxfois: Like that ? https://github.com/OpenAEV-Platform/openaev/pull/6874/changes/09ce81857b601cd67770ae6591f624e0ccf0e8d6

.github/workflows/docs-deploy.yml
```diff
@@ -0,0 +1,91 @@
...
+          if ! git tag --list "$VERSION" | grep -q .; then
+            echo "::error::Tag '${VERSION}' does not exist. The version must match an existing git tag."
+            exit 1
+          fi
+
+      - name: Find latest tag
+        uses: ./.github/actions/find-latest-tag
+        id: find-latest-tag
+
+      - name: Resolve latest flag
```
antoinemzs:
```suggestion
      - name: Resolve latest tag
```

antoinemzs (approved):

PR #6883, opened 2026-07-23
fix(implant): add a timeout for the implant command (#6878)

### LINKED ISSUE #6878: fix(implant): add a timeout for the implant command
## Description

If an "infinite" command is launched, the implant is never killed. We need to add the same timeout we have on the OpenAEV UI.

## Reproducible steps

1. Launch a command that takes more than 10 minutes ("sleep XXX" or "Remote Process Injection in LSASS via mimikatz" from ART for example),
2. Look in the VM, the implant is not killed.

## Expected output

The implant is not killed and stays in background.

## Actual output

No implant lives after 10 minutes of timeout.

damgouj:
### Proposed changes

* Add a timeout when the implant is executed. If the timeout is reached, the implant is killed.

### Testing Instructions

1. Launch a command that takes more than 10 minutes ("sleep XXX" or "Remote Process Injection in LSASS via mimikatz" from ART for example),
2. Look in the VM, the implant is not killed.

### Related issues

* Closes #6878

### Checklist

### DIFF
--- a/openaev-api/src/main/java/io/openaev/integration/impl/injectors/openaev/OpenaevInjectorIntegrationFactory.java
+++ b/openaev-api/src/main/java/io/openaev/integration/impl/injectors/openaev/OpenaevInjectorIntegrationFactory.java
@@ -18,7 +18,7 @@ import io.openaev.service.connector_instances.ConnectorInstanceService;
 import java.lang.reflect.InvocationTargetException;
 import java.util.List;
 import java.util.Map;
-import org.springframework.core.env.Environment;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Service;
 
 @Service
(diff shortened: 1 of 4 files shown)


openaev-api/src/main/java/io/openaev/integration/impl/injectors/openaev/OpenaevInjectorIntegrationFactory.java
```diff
@@ -94,7 +99,13 @@ public Integration spawn(ConnectorInstance instance)
 
   @Override
   public void registerConnectorForTenant(String tenantId) throws Exception {
-    Map<String, String> executorCommands = OpenaevImplantCommandBuilder.buildExecutorCommands();
+    String threshold = env.getProperty("inject.execution.threshold.minutes");
```
RomuDeuxfois: **todo**: please use a safer way to do this

  @Value("${inject.execution.threshold.minutes:10}")                                                                                                                                                                                
  private int injectExecutionThreshold;

RomuDeuxfois (approved):

PR #6904, opened 2026-07-23
fix(chaining): scope step consumers to the run tenant (#6357)

### LINKED ISSUE #6357: feat(chaining): make chaining tenant compatible
Dod:
- all the chaining apis are multi  tenants compatible

laugiov:
### What
The chaining queue consumers run on worker threads that carry no tenant scope, so a step run resolved its
data under the default tenant. On a non-default tenant this means the run's tenant-active reads find
nothing and the inject does not execute. This scopes each event's run to its tenant.

- `StepEventService` (ready + external-update) opens each event under the run's tenant on both MT
  mechanisms: the v2 primitive (`tenantTx.execute`, GUC for tenant-active writes) and v1
  `TenantContext` (the `@Filter` reads the run still needs, not yet migrated to v2), set-then-finally-clear
  on the pooled worker (mirrors `ScenarioExecutionJob`).
- The tenant travels on the event (`StepEvent`/`ExternalUpdateEvent` carry `tenant_id`), stamped
  server-side by `QueueChainingService` from the workflow/step simulation tenant via projection queries
  (no lazy navigation: the producers run on scheduler/queue threads with no open session).
- Null `tenant_id` (in-flight/legacy messages) falls back to the default tenant: no rollout break.

### Scope / not in this PR
This covers **internal injectors** (e.g. the attack-path implant, `Executor.executeInternal`). It does
**not** fix **external injectors**: `Executor.executeExternal` publishes to the static base connector
exchange (`RabbitmqService.publish`, tenant-blind), which 404s on a non-default tenant. That is a separate
change (make the external publish tenant-aware) and must not regress classic-external / the default tenant.
Do not auto-close #6357.

### Tests
`AttackPathRunWiringIT` W4 (real consumer, no ambient scope → row under the run tenant), W5 (wrong tenant →
no row), W6 (tenant projections on a real graph); unit tests for both consumers; full chaining/attackpath
suite green.

### DIFF
--- a/openaev-api/src/main/java/io/openaev/service/chaining/ExternalUpdateEvent.java
+++ b/openaev-api/src/main/java/io/openaev/service/chaining/ExternalUpdateEvent.java
@@ -26,6 +26,14 @@ public class ExternalUpdateEvent implements Queueable {
   @JsonProperty("step_id")
   private String stepId;
 
+  /**
+   * The tenant this event's work belongs to. Carried so the consumer can restore the tenant context
+   * the producer thread had. Null on legacy/in-flight messages (pre-#6357); the consumer falls
+   * back.
+   */
+  @JsonProperty("tenant_id")
+  private String tenantId;
+
   /** The timestamp when this event was emitted, in milliseconds since epoch. */
   @JsonProperty("event_emission_date")
   private long emissionDate;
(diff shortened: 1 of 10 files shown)


laugiov:
Addressed the copilot review:

- findTenantIdByStepId / findTenantIdByWorkflowId are now NATIVE queries. `exercises`
  is a v1 `@Filter(tenantFilter)` entity, so a JPQL path through it would be re-filtered
  by the ambient TenantContext and could stamp null on a producer thread carrying the
  default/wrong tenant. Native SQL bypasses the Hibernate filter, so the stamp is
  context-free. (Good catch.)
- W6 now clears TenantContext before calling the projections, so it actually proves the
  stamping does not depend on ambient v1 scope.
- Tenant-propagation unit tests stub findById -> Optional.empty() explicitly so they take
  the intended not-found branch rather than relying on Mockito's default.

PR #6905, opened 2026-07-23
feat(attack-path): enforce resource RBAC on reads and scenario-scope the picker (#6647)

laugiov:
## What

Hardens the attack-path API on two fronts (tracked internally as B1 + B0):

- **B1 — resource-level RBAC on the reads.** The 5 per-simulation reads (`graph`,
  `endpoint/findings`, `endpoint/relations`, `findings`, `execution`) now enforce
  `SIMULATION READ` per resource, instead of being open to any authenticated user in
  the tenant.
- **B0 — scenario-scoped, grant-filtered picker.** `GET /simulations` accepts an optional
  `scenarioId` and, when given, returns only that scenario's simulations (server-side);
  the rows are then grant-filtered to the ones the caller can read.

## Why

All 7 attack-path endpoints were `@AccessControl(skipRBAC = true)`: any authenticated user
in the tenant could read any simulation's attack path, and the picker returned every
simulation with attack-path data in the tenant, regardless of scenario or the caller's
grants (e.g. one scenario's runs showing up under another). That was a POC concession while
the store was fed by synthetic seed data (seed ids are not real exercises, so resource RBAC
would 403 them). Real data now flows, so enforce real RBAC while keeping seed ids reachable.

## How

- New `AttackPathAccessControl` guard: `SIMULATION READ` via the platform `PermissionService`
  (capability-or-grant, same as every other simulation resource), **seed-tolerant** — a seed
  id (`ap-seed-…`) is allowed without a grant. The `@AccessControl(skipRBAC = true)` annotation
  stays because it cannot express the seed exception; the guard is the real gate, and one helper
  backs both the reads and the picker filter.
- `GET /simulations` gains an optional `scenarioId` -> `findSimulationSummariesByScenario`
  (JPQL subquery `simulation_id IN (SELECT ex.id FROM Exercise ex WHERE ex.scenario.id = :scenarioId)`),
  plus a per-row grant filter (`canReadSimulation`, seed rows kept).

## Scope / not in this PR

- The 5 per-simulation reads enforce READ (403 without access); `GET /simulations` grant-filters its
  rows (never 403); only `POST /seed` is left `skipRBAC` (a dev-only data generator). So 6 of the 7
  endpoints are now access-controlled.
- **Accepted pre-GA risk:** seed ids bypass RBAC (the data is synthetic). To revisit at GA / when the
  seed generator is removed.
- No payload/DTO change on the reads; the picker only gains an optional query param. Admins and
  capability users are unchanged; the front keeps working (the client-side scenario filter is now
  backed by the server). No migration.

## Tests

`AttackPathApiRbacIT` — real stack (MockMvc + real PG + real grants, no mocked RBAC), 8/8:
- no grant -> 403 on each of the 5 reads;
- a resource grant (no capability) -> not forbidden on all 5, 200 on graph (proves grant-level, not capability);
- seed id -> reachable without a grant (read and picker);
- picker scoped to a scenario returns only that scenario's sims;
- picker grant-filters (hides a simulation the caller cannot read);
- admin unchanged (reads not forbidden, picker not grant-filtered).

### DIFF
--- a/openaev-api/src/main/java/io/openaev/api/attackpath/AttackPathApi.java
+++ b/openaev-api/src/main/java/io/openaev/api/attackpath/AttackPathApi.java
@@ -9,6 +9,7 @@ import io.openaev.database.model.attackpath.projection.AttackPathSimSummaryRow;
 import io.openaev.rest.helper.RestBehavior;
 import io.openaev.rest.settings.PreviewFeature;
 import io.openaev.service.PreviewFeatureService;
+import io.openaev.service.attackpath.AttackPathAccessControl;
 import io.openaev.service.attackpath.AttackPathGraphService;
 import io.openaev.service.attackpath.AttackPathSeedService;
 import io.openaev.service.attackpath.dto.AttackPathDTO;
@@ -57,6 +58,7 @@ public class AttackPathApi extends RestBehavior {
   private final AttackPathGraphService graphService;
   private final AttackPathSeedService seedService;
   private final PreviewFeatureService previewFeatureService;
+  private final AttackPathAccessControl attackPathAccessControl;
 
   /** Runtime feature gate: return 404 unless the {@code ATTACK_PATH} preview feature is enabled. */
   private void requireAttackPathFeature() {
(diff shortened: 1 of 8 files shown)


laugiov:
Both addressed:
- Picker no longer runs per-row currentUser()/hasPermission: AttackPathAccessControl.retainReadable
  resolves the current user once and filters the rows with it. The capability/admin path is in-memory,
  so only a pure-grant user pays a lightweight indexed EXISTS per row (row-level authz, scenario-scoped
  in the common path).
- AttackPathApiRbacIT now stores enabledDevFeatures in setUp and restores it in tearDown (+ evicts the
  global cache), like WorkflowApiTest, so it never wipes another test's dev features.

PR #6911, opened 2026-07-24
fix(attack-path): restrict the graph to chained scenarios and fix the inject link (#6647)

Seb-MIGUEL:
## Summary

Two fixes on the Attack Path view (behind the `ATTACK_PATH` flag):

- **Chained-only exposure**: the attack-path tab and route were shown on time-based scenarios. They are now gated to chained scenarios (`INJECT_CHAINING` + `scenario_workflow_id`), mirroring the existing simulation gating. Time-based scenarios no longer show the tab, and the route is not registered.
- **"Action details" link**: from the scenario attack-path view the link used a relative `../injects/<id>`, resolving to the scenario (which has no such inject → 404). It now targets the selected simulation: `/admin/simulations/<simulationId>/injects/<injectId>`, so it resolves in both scenario and simulation contexts.

`check-ts` and eslint clean; verified live that a chained scenario still shows the tab.

Related issue: #6647

### DIFF
--- a/openaev-front/src/admin/components/scenarios/scenario/Index.tsx
+++ b/openaev-front/src/admin/components/scenarios/scenario/Index.tsx
@@ -48,7 +48,11 @@ const IndexScenarioComponent: FunctionComponent<{ scenario: ScenarioOutput }> = 
   const { t } = useFormatter();
   const location = useLocation();
   const isChainingFeatureEnabled = isFeatureEnabled('INJECT_CHAINING');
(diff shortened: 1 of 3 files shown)


SamuelHassine:
Closing: these changes have been incorporated into the improvements wave branch (PR #6910) as commit 3f691ee6f (co-authored). The chained-scenario attack-path tab gating was already fixed there; the master-detail execution drawer and the simulation-scoped inject link were cherry-picked on top. Thanks @Seb-MIGUEL!

PR #6912, opened 2026-07-24
feat(chaining): connect data for terminal view and expectations (#6647)

EvaE-Filigran:
### Proposed changes

* Refactored attack-path execution updates to use execution-id based writes instead of multiple step/agent/target specific repository methods:
  * Added `updateTerminalViewByExecutionIndex(...)`
  * Added `updateExpectationStatusByExecutionId(...)`
  * Removed legacy per-field/per-scope update methods (`updatePreventionStatusBy...`, `updateDetectionStatusBy...`, `updateVulnerabilityStatusBy...`)
* Added attack-path ingestion logic to:
  * Aggregate terminal traces per execution index and persist them (`updateTerminalView(...)`)
  * Aggregate expectation results per execution index by type (prevention/detection/vulnerability)
  * Resolve the highest-priority label per expectation type (`success > partial > pending > failure`) and persist statuses
* Added guards to skip trace aggregation/update when execution index resolution returns `null` (prevents null-key accumulation / invalid updates)
* Reduced transaction overhead by batching updates inside a single tenant-scoped transaction per operation (instead of one transaction per index)
* Updated chaining step integration:
  * `InjectExecutionStep.update(...)` now triggers attack-path terminal output sync
  * `formatExpectationToOutput(...)` now triggers attack-path expectation status sync
* Updated command rendering in `InjectExecutionStep.getCommand(...)`:
  * Replaces `#{argKey}` placeholders using payload argument `defaultValue`
* Added integration test coverage for expectation priority resolution:
  * `given_mixedExpectationResults_should_updateExecutionWithHighestPriorityLabels`

### Testing Instructions

1. Start a **chaining simulation** that includes at least one **Command payload** with variable placeholders (e.g. `#{var}`).
2. Execute the workflow and wait until inject execution and expectation updates are completed.
3. Verify that Command payload placeholders are correctly resolved:
   - Example: command template `echo #{var}` should be executed/stored as `echo <expected-value>`.
4. Verify that **expectation statuses** written to attack-path execution are consistent with inject expectations:
   - prevention status matches prevention expectation result
   - detection status matches detection expectation result
   - vulnerability status matches vulnerability expectation result
5. Verify that **execution traces** are correctly displayed in attack-path terminal output:
   - traces are grouped by execution index
   - stdout/stderr (or fallback message) is rendered as expected
   - no missing trace due to null/unresolved execution index
6. (Optional regression check) Re-run the same simulation path and confirm updates are still coherent and no duplicate/incorrect status mapping appears.


openaev-api/src/main/java/io/openaev/service/attackpath/ingestion/AttackPathExecutionIngestionService.java
```diff
@@ -286,14 +370,148 @@ public String getExecutionIndex(Inject inject, String target) {
...
+        String[] targets = inject.getContent().get("targets").asText().split(",");
+        if (targets.length < 1) return null;
+
+        target = inject.getContent().get("targets").asText().split(",")[0];
         return AttackPathIds.executionNode(inject.getId(), target, inject.getInjector().getId());
+
+      } else if (targetSelector.equals("asset") || targetSelector.equals("asset_group")) { // ASSET
+        if (inject.getAssets().isEmpty()) return null;
+        target = inject.getAssets().getFirst().getId();
+        AttackPathIds.executionNode(inject.getId(), target, inject.getInjector().getId());
```
laugiov: This line builds the id but does not return it, so the method falls through to return null below. For an injector that targets assets, the index is always null and nothing gets written back. Adding return in front fixes it.

openaev-api/src/main/java/io/openaev/service/attackpath/ingestion/AttackPathExecutionIngestionService.java
```diff
@@ -286,14 +370,148 @@ public String getExecutionIndex(Inject inject, String target) {
...
-        // injector is the source, mirroring the write side executionNode(inject, target, injector).
+
+      if (targetSelector.equals("manual")) { // DISCOVERY
+        String[] targets = inject.getContent().get("targets").asText().split(",");
+        if (targets.length < 1) return null;
+
+        target = inject.getContent().get("targets").asText().split(",")[0];
         return AttackPathIds.executionNode(inject.getId(), target, inject.getInjector().getId());
+
+      } else if (targetSelector.equals("asset") || targetSelector.equals("asset_group")) { // ASSET
```
laugiov: Two things here. First, this branch checks "asset", but the write side (getAttackPathExecution) acts on "assets" (plural), and that is the value that creates the rows. So the read looks for a different key than the write and finds nothing to update. We need the same string on both sides.

Second, on the next line you take getAssets().getFirst(). But the write creates one row per target, including assets that come from asset groups. Those are not in getAssets(), they are resolved from getAssetGroups(). So when there are several targets, only the first is matched and asset-group targets are missed. The executor path already passes the specific agent.getId() for each trace. Could the injector path do the same and pass the specific target, instead of deriving a single one? This is the point we discussed this morning.

EvaE-Filigran: "assets" fixed

openaev-api/src/main/java/io/openaev/service/attackpath/ingestion/AttackPathExecutionIngestionService.java
```diff
@@ -286,14 +370,148 @@ public String getExecutionIndex(Inject inject, String target) {
...
+        if (inject.getAssets().isEmpty()) return null;
+        target = inject.getAssets().getFirst().getId();
+        AttackPathIds.executionNode(inject.getId(), target, inject.getInjector().getId());
       }
```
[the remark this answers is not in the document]
EvaE-Filigran: fix

laugiov (approved):

Issue #5536, opened 2026-04-21
[CRED] Create and manage credential assets (CRUD)


DeniseMonreale:
As a Soc Manager Admin

I want to create or import credentials into OpenAEV,

so that they are encrypted, centrally managed

PR #6914, opened 2026-07-24
feat(secret-providers): add secret system integration (#5536)

MarineLeM:
### Description

This pull request introduces **Secrets Providers**, a new class of OpenAEV integration responsible for storing, retrieving, and deleting secrets used by the platform.

This PR ships the first implementation: the **Local** provider, backed by OpenAEV's own Postgres store (via `SecretService` / `SecretReferenceService`).

### Proposed changes

- Added `SecretsProviderApi`, a new REST controller exposing endpoints to list, retrieve, and query related IDs for secrets providers.
- Introduced the `SecretProvider` abstraction: a built-in integration responsible for storing, retrieving, and deleting secrets.
- Implemented the **Local** provider:
  - `LocalSecretsProviderIntegrationFactory` — builds the integration bean.
  - `LocalSecretsProviderIntegration` — wires the provider into the shared `Manager`.
  - `LocalSecretsProvider` — the provider implementation, delegating to:
    - `LocalSecretsResolver` — retrieves/looks up secrets.
    - `LocalSecretsWriter` — stores/deletes secrets.

### Related issues

* Chunk 1 : #5536


openaev-model/src/main/java/io/openaev/database/model/HashSecret.java
```diff
@@ -0,0 +1,30 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.openaev.database.audit.ModelBaseListener;
+import jakarta.persistence.Column;
+import jakarta.persistence.DiscriminatorValue;
+import jakarta.persistence.Entity;
+import jakarta.persistence.EntityListeners;
+import jakarta.validation.constraints.NotBlank;
+import lombok.Data;
+
+@Data
```
gabriel-peze: THe annotation Data will generate a toString method, with all the attributes, even the ones with @JsonIgnore, (used only by jackson).
This could lead to a leak of the secret_hash value in the applications logs, is it supposed to be displayed on the logs ?

For example, a "log.error(hashSecret)" will display the secret hash in the logs.

In addition, annotation Data will generate a new Equals and HashCode method, who will overide the one from Secret, that will ignore the id from Secret and could lead to an exception if two differents secrets from database had the same hash and hashAlgorithm.

Like in CredentialSecretReference.java, you could add the "@EqualsAndHashCode(callSuper = true)" annotation to avoid this.

openaev-model/src/main/java/io/openaev/database/model/UsernamePasswordSecret.java
```diff
@@ -0,0 +1,30 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.openaev.database.audit.ModelBaseListener;
+import jakarta.persistence.Column;
+import jakarta.persistence.DiscriminatorValue;
+import jakarta.persistence.Entity;
+import jakarta.persistence.EntityListeners;
+import jakarta.validation.constraints.NotBlank;
+import lombok.Data;
+
+@Data
```
gabriel-peze: THe annotation Data will generate a toString method, with all the attributes, even the ones with @JsonIgnore, (used only by jackson).
This could lead to a leak of the secret_username and secret_password values in the applications logs, is it supposed to be displayed on the logs ?

For example, a "log.error(usernamePasswordSecret)" will display the username and password in the logs.

In addition, annotation Data will generate a new Equals and HashCode method, who will overide the one from Secret, that will ignore the id from Secret and could lead to an exception if two differents secrets from database had the same username and password.

Like in CredentialSecretReference.java, you could add the "@EqualsAndHashCode(callSuper = true)" annotation to avoid this.

openaev-model/src/main/java/io/openaev/database/model/SecretReference.java
```diff
@@ -0,0 +1,105 @@
...
+
+  @Column(name = "secret_reference_status")
+  @JsonProperty("secret_reference_status")
+  @NotBlank
+  private String status = "ACTIVE";
+
+  @ManyToOne
+  @JoinColumn(name = "secret_reference_created_by")
+  @JsonProperty("secret_reference_created_by")
+  private User createdBy;
```
gabriel-peze, line 75: Is it usefull to return the full User object and not only his id ?

If it should return only the id you can use the @JsonSerialize(using = MonoIdSerializer.class)

openaev-front/src/actions/secrets_providers/secrets-providers-action.ts
```diff
@@ -0,0 +1,24 @@
...
+  const uri = `${SECRETS_PROVIDERS_URI}/${secrets_providerId}`;
+  return getReferential(schema.secretsProvider, uri)(dispatch);
+};
+
+export const fetchSecretsProviderRelatedIds = (secrets_providerId: string) => {
+  return simpleCall(`${SECRETS_PROVIDERS_URI}/${secrets_providerId}/related-ids`);
+};
+
+export const deleteSecretsProvider = (secrets_providerId: string) => (dispatch: Dispatch) => {
+  return delReferential(`${SECRETS_PROVIDERS_URI}/${secrets_providerId}`, 'secretsproviders', secrets_providerId)(dispatch);
```
gabriel-peze: The delete route doesn't exist in the API, is it normal to have a call to a delete route here ?

If not, is it possible to delete this code, or if it's for future developments, is it possible to at least have a comment ?

openaev-api/src/main/java/io/openaev/integration/IntegrationFactory.java
```diff
@@ -95,10 +93,5 @@ public List<ConnectorInstance> findRelatedInstances(String tenantId) {
             tenantId, this.getClassName()));
   }
 
-  public abstract Integration spawn(ConnectorInstance instance)
-      throws JsonProcessingException,
-          InvocationTargetException,
-          NoSuchMethodException,
-          InstantiationException,
-          IllegalAccessException;
+  public abstract Integration spawn(ConnectorInstance instance) throws Exception;
```
gabriel-peze: Is it normal to throw Exception here ? I'm not a big fan of throwing everything, i think it's better to keep the throws like before and control what we throws

openaev-front/src/admin/components/integrations/catalog_connectors/ConnectorMarketplace.tsx
```diff
@@ -78,6 +78,10 @@ const ConnectorMarketplace = ({ items, renderFooterAction, searchPlaceholder }:
         label: t('Executors'),
         icon: TerminalOutlined,
       },
+      SECRETS_PROVIDER: {
+        label: t('Secrets Provider'),
+        icon: TerminalOutlined,
```
gabriel-peze: The icon is the same as the Executors, is it normal ? is it the good icon ?

openaev-api/src/main/java/io/openaev/secrets/provider/impl/LocalSecretsProvider.java
```diff
@@ -0,0 +1,141 @@
...
+import io.openaev.database.model.UsernamePasswordSecret;
+import io.openaev.secrets.provider.SecretStoreRequest;
+import io.openaev.secrets.provider.SecretsProvider;
+import io.openaev.secrets.provider.SecretsProviderType;
+import io.openaev.secrets.service.SecretReferenceService;
+import io.openaev.secrets.service.SecretService;
+import io.openaev.service.connector_instances.NativeEncryptionService;
+import java.util.Objects;
+
+public class LocalSecretsProvider extends SecretsProvider {
```
gabriel-peze, line 19: I saw that there was no tests for this class, is it possible to add some ?

openaev-api/src/main/java/io/openaev/api/secrets_providers/SecretsProviderService.java
```diff
@@ -0,0 +1,97 @@
...
+   */
+  public ConnectorIds getSecretsProviderRelationsId(String secretProviderId) {
+    return getConnectorRelationsId(secretProviderId);
+  }
+
+  @Override
+  protected List<SecretsProvider> getAllConnectors() {
+    return managerFactory
+        .getManager(TenantContext.getCurrentTenant())
+        .requestManyAllStates(
```
gabriel-peze: requestManyAllStates could lead to a NoSuchElementException from Manager.java, i think this code should be included in a try/catch to return an empty list if a NoSuchElementException is throw.

gabriel-peze:
Is it normal that i have no icon for the integration ?
image

Edit, fixed :
image

openaev-api/src/main/java/io/openaev/secrets/provider/impl/LocalSecretsProvider.java
```diff
@@ -0,0 +1,141 @@
...
+      Secret existingSecret, SecretStoreRequest payload) {
+    UsernamePasswordSecret passwordSecret =
+        existingSecret == null
+            ? new UsernamePasswordSecret()
+            : expectUsernamePasswordSecret(existingSecret);
+    passwordSecret.setUsername(
+        Objects.requireNonNull(payload.username(), "request.username must not be null"));
+    passwordSecret.setPassword(
+        encryptRequired(payload.password(), "request.password must not be null"));
+    return passwordSecret;
```
antoinemzs, line 97: Looks like this should be more generic. Fine for the first two models, though expanding on this should require an encryption gateway component with declarative encrypted fields (set on the model classes).

pass for now

openaev-api/src/main/java/io/openaev/secrets/provider/SecretProvider.java
```diff
@@ -0,0 +1,35 @@
+package io.openaev.secrets.provider;
+
+import io.openaev.database.model.SecretReference;
+
+/**
+ * Contract for secret storage backends. Each implementation handles a specific storage mechanism
+ * (local DB, HashiCorp Vault, CyberArk, etc.).
+ */
+public interface SecretProvider {
```
antoinemzs: nitpick: a bit confusing to have an interface with a singular "Secret-" name and the abstract class implementing it a purlarl "Secrets-" name.

IMO we should choose between the two: either have an interface for the implementers to figure out everything or an abstract class that provides a functional shell and require implementing specifics.

From the current implementation, it looks like we lean on to the abstract class being more suited: we set some metadata on it, and also provide a placeholder subclass.

openaev-api/src/main/java/io/openaev/secrets/provider/SecretStoreRequest.java
```diff
@@ -0,0 +1,12 @@
+package io.openaev.secrets.provider;
+
+/**
+ * Payload for provider-side secret storage.
```
antoinemzs, line 4: noted, probably also consider subsuming this into a generic interface based on model in later iterations

openaev-front/src/admin/components/integrations/catalog_connectors/ConnectorMarketplace.tsx
antoinemzs: good catch, should be `KeyOutlined` (suggested in PoC, but debatable)

openaev-api/src/main/java/io/openaev/secrets/provider/impl/LocalSecretsProvider.java
MarineLeM: I prefer to add a test for this part in chunk-2, since I updated it

openaev-api/src/main/java/io/openaev/secrets/provider/impl/LocalSecretsProvider.java
MarineLeM: I will try to find a solution, and if I find one, I'll implement it in chunk-2 since I updated this part.

openaev-front/src/actions/secrets_providers/secrets-provider-helper.d.ts
```diff
@@ -0,0 +1,6 @@
+import { type SecretProvider, type SecretsProviderOutput } from '../../utils/api-types';
+
+export interface SecretsProviderHelper {
+  getSecretsProvider: (secretsProviderId: string) => SecretsProviderOutput;
+  getSecretsProvidersIncludingPending: () => SecretProvider[];
```
gabriel-peze: I think it should be "SecretsProvider[]" instead of "SecretProvider[]".

The SecretProvider type doesn't exist in the api-types, but SecretsProvider yes and is unused.

openaev-api/src/main/java/io/openaev/api/secrets_providers/form/SecretsProviderOutput.java
```diff
@@ -0,0 +1,28 @@
...
+  @JsonProperty("secrets_provider_name")
+  @NotBlank
+  private String name;
+
+  @JsonProperty("secrets_provider_type")
+  @NotBlank
+  private String type;
+
+  @JsonProperty("existing_secret_provider")
+  @NotBlank
```
gabriel-peze: Carefull, @NotBlank can be applied only on String types.

openaev-api/src/main/java/io/openaev/api/secrets_providers/SecretsProviderApi.java
```diff
@@ -0,0 +1,105 @@
...
+  public ConnectorIds getSecretsProviderRelatedIds(
+      TxCtx ctx, @PathVariable String secretsProviderId) {
+    return secretsProviderService.getSecretsProviderRelationsId(ctx, secretsProviderId);
+  }
+
+  @GetMapping(value = "/{secretProviderType}/image", produces = MediaType.IMAGE_PNG_VALUE)
+  @AccessControl(skipRBAC = true)
+  @Operation(summary = "Get injector image by type")
+  @Transactional
+  public ResponseEntity<InputStreamResource> getInjectorImage(
```
gabriel-peze: Is it a copy/past mistake ?
Should be "getProviderImage" from my point of view, the Operation annotation should change too.

gabriel-peze (approved):

PR #6920, opened 2026-07-24
feat(attack-path): chained-only exposure, unified drawer, live refresh, endpoint dedup, injector traces (#6647)

Seb-MIGUEL:
## Summary

Frontend improvements to the Attack Path view (all behind the `ATTACK_PATH` flag). Supersedes #6911 (which held the first two items — that PR can be closed).

- **Chained-only exposure**: the tab/route are gated to chained scenarios (`INJECT_CHAINING` + `scenario_workflow_id`), mirroring the simulation gating. Time-based scenarios no longer show it.
- **Unified master-detail drawer**: clicking an execution replaces the endpoint/finding panel in a single, wider drawer with a back arrow (no more double drawer); close dismisses the whole thing.
- **"Action details" target**: opens the inject on its **simulation** (not the scenario → 404); a payload-backed inject lands on Overview, a network injector on Execution details.
- **Live refresh**: the graph polls every 10s while mounted and updates in place (no spinner / selection reset / churn) so a running run populates on its own.
- **Endpoint dedup in chain mode**: several independent injectors at the same depth hitting the same asset now converge on one shared endpoint node (cross-depth causal chains are preserved).
- **Injector terminal view**: a network injector (no payload) shows its global execution traces (the "<tool> succeeded: …" output) instead of an empty per-target terminal; the tab is labelled "Execution details" vs "Terminal view" accordingly. Dropped the noisy `payload › <uuid>` crumb.

Tests: extended `attack-path-flow-helpers.test.ts` (endpoint dedup + updated depth-keyed ids) and kept the component test green.

Related issue: #6647

### DIFF
--- a/openaev-front/src/admin/components/simulations/simulation/attack_path/SimulationAttackPath.tsx
+++ b/openaev-front/src/admin/components/simulations/simulation/attack_path/SimulationAttackPath.tsx
@@ -384,11 +384,12 @@ const SimulationAttackPath = ({ scenarioExerciseIds, scenarioId }: SimulationAtt
     };
   }, [simulationId]);
 
-  // Causal overlay data source (issue 6647). The rendered graph is collapsed, which omits the
-  // per-execution kill-chain fields, so we fetch the full graph once — only to derive the per-injector
-  // meta — and gate it on the simulation's execution count (mirrors the backend collapse-threshold) so a
-  // large run never pulls a full payload just for the overlay. When the size is unknown or above the
-  // ceiling, the overlay is simply absent (additive: the graph is unchanged).
+  // Causal overlay data source (issue 6647) + live refresh. The collapsed graph omits the per-execution
+  // kill-chain fields, so we fetch the full graph (to derive the per-injector meta AND to drive the
+  // causal-chain layout), gated on the execution count (mirrors the backend collapse-threshold) so a
+  // large run never pulls a full payload. Within the gate we also POLL it, so a running chained run
+  // (which renders from fullDto, not dto) live-updates like the collapsed graph does. The gate uses the
+  // initial summary-row count; a run that grows past the ceiling mid-view keeps its overlay until reselect.
   useEffect(() => {
     if (!simulationId) {
       setKillChainMeta(new Map());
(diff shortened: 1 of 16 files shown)


openaev-front/src/admin/components/simulations/simulation/attack_path/SimulationAttackPath.tsx
```diff
@@ -339,6 +353,37 @@ const SimulationAttackPath = ({ scenarioExerciseIds, scenarioId }: SimulationAtt
     load();
   }, [load]);
 
+  // Live refresh: an on-going run keeps discovering endpoints/findings, so poll the graph on an interval
```
laugiov, line 392: SimulationAttackPath.tsx : the live refresh does not reach the causal (chained) view

The interval only calls setDto (L375, the collapsed graph), but a chained scenario renders from fullDto, not dto: chainMode = !!fullDto && … (L928) then buildCausalChainFlow(fullDto, t) (L955). fullDto is set only by the size-gated 'full' effect, whose deps are [simulationId, simulations] (L428), and simulations is fetched once (L442, no interval). Nothing refreshes fullDto on the interval.

So for a small chained run (chainMode), the visible causal graph does not live-update: new executions only appear after a manual reload. The interval still helps large runs (fullDto null → aggregated view from dto), but the causal view, which is the main case of this PR, stays frozen while the run is in progress. The interval also does not refresh the summary row, so executionCount (which gates the full fetch) stays stale.

Rather than duplicate the size gate in the collapsed poll, the cleanest fix is to make the existing 'full' effect poll: it already holds the gate, the simulations dep, the setFullDto / setKillChainMeta setters and the cancelled pattern. Suggested replacement for that effect:


  ```
// Causal overlay data source (issue 6647) + live refresh. The collapsed graph omits the per-execution
  // kill-chain fields, so we fetch the full graph (to derive the per-injector meta AND to drive the
  // causal-chain layout), gated on the execution count (mirrors the backend collapse-threshold) so a
  // large run never pulls a full payload. Within the gate we also POLL it, so a running chained run
  // (which renders from fullDto, not dto) live-updates like the collapsed graph does. The gate uses the
  // initial summary-row count; a run that grows past the ceiling mid-view keeps its overlay until reselect.
  useEffect(() => {
    if (!simulationId) {
      setKillChainMeta(new Map());
      setFullDto(null);
      return undefined;
    }
    const row = simulations.find(s => s.simulationId === simulationId);
    // No summary row yet, or above the ceiling: no overlay. Clear any state from a previous simulation.
    if (!row || (row.executionCount ?? 0) > CAUSAL_META_MAX_EXECUTIONS) {
      setKillChainMeta(new Map());
      setFullDto(null);
      return undefined;
    }
    let cancelled = false;
    let lastJson = '';
    const REFRESH_MS = 10000;
    const applyFull = () =>
      fetchAttackPathGraph(simulationId, 'full')
        .then((r) => {
          if (cancelled) {
            return;
          }
          // Skip the swap when nothing changed, so an open drawer / pan-zoom and a stable graph never
          // churn (same discipline as the collapsed poll).
          const json = JSON.stringify(r.data);
          if (json === lastJson) {
            return;
          }
          lastJson = json;
          setFullDto(r.data);
          setKillChainMeta(buildKillChainMeta(r.data));
        })
        .catch(() => {
          // Initial fetch failed: show no overlay (never leak the previous simulation's). After a first
          // success, keep the last good overlay on a transient failure; the next tick retries.
          if (!cancelled && lastJson === '') {
            setFullDto(null);
            setKillChainMeta(new Map());
          }
        });
    applyFull();
    const timer = setInterval(applyFull, REFRESH_MS);
    return () => {
      cancelled = true;
      clearInterval(timer);
    };
  }, [simulationId, simulations]);
```
The collapsed poll stays as it is (it still serves the aggregated view and the peripheral dto data). The lastJson check avoids any state swap when the payload is unchanged, so a stable graph never churns. On a transient poll error the last good overlay is kept; only a failed first fetch clears it, so a previous simulation's overlay never leaks. One assumed limit (in the comment): the gate uses the initial executionCount, so a run that grows past CAUSAL_META_MAX_EXECUTIONS mid-view keeps its overlay until the simulation is reselected. In practice a chained run rarely gets that large, and this matches the gate's current behaviour.

laugiov (changes requested):

laugiov (approved):

PR #6928, opened 2026-07-24
fix(chaining): use manual scope as target (#6924)

camrrx:
### Proposed changes

 - Added TargetType.MANUAL, buildManualTargetMap() (extracts target from inject_content), and manualMap passed through mapResultsToInjects 
 - Added "MANUAL" to TargetSimple.target_type in ItemTargets.tsx to display manual targets in the Target column

### Testing Instructions

1. Create a workflow scope with non-asset targeting criteria.
2. Launch a chaining simulation using scope-based targeting.
3. Compare the expected targets from the scope with the targets actually executed.

### Related issues

* Closes #6924


EvaE-Filigran (approved):
Code and tests ok

PR #6964, opened 2026-07-27
feat(attack-path): causal chain view, finding-panel correctness and drawer polish (#6647)

Seb-MIGUEL:
## Summary

Front follow-ups on the Attack Path view (behind the `ATTACK_PATH` flag), extending #6920. Groups the execution-drawer polish with the causal-chain view work, finding-panel correctness fixes, and the front-side alignment with the backend PRs merged in the meantime (#6912 expectations/terminal data, #6972 SMB shares as the native `file` type).

### Execution drawer & findings list
- **Findings tab** in the execution drawer (between "Execution details" and "Remediation"): renders the shared `FindingList` scoped to the inject (`searchDistinctFindingsForInjects`), so it works for both endpoint and injector-contract executions. Shown only when the execution carries an inject id.
- **`FindingList` gains `compact` + `hiddenFields` props** (defaults unchanged, existing usages untouched): `compact` hides the search/filters/pagination top bar; `hiddenFields` hides columns. In the narrow drawer it drops the asset-groups **and** first-seen / last-seen columns, leaving Type / Value / Assets.
- **Compact page size raised + "showing X of N" note** so scan findings are never silently hidden when the list is truncated.
- **Wider default drawer** (460 → 560px) so all four tabs (incl. the previously-clipped Remediation) and the findings table fit; still drag-resizable.
- **Empty state** is now truly centered in the graph area (`position: absolute; inset: 0`, 88px icon, h6 message) instead of a small block stuck at the top.

### Causal chain view (legibility)
- **Collapse duplicate causal edges on hub endpoints**: a key like `share_name IS_NOT_NULL` matches every produced finding of the type, which stacked N identical "Triggered …" labels over the consumer. Draw a single causal edge per (consumer, label), anchored to the nearest matching finding, so a one-asset/many-findings hub reads as cleanly as a linear chain.
- **Draw the causal edge for `IS_NOT_NULL` events** (any produced finding of the reconciled type satisfies the event).
- **Finding value labels**: show the full value flowing rightward instead of truncating, and give expanded findings enough vertical room so labels don't overlap the node above.
- Correct injector-highlight attribution and stabilize endpoint order on live refresh.

### Finding panel correctness (chain & focused views)
- **"Discovered on" shows the real endpoint** (host / IP) instead of the literal word "Endpoint": the finding carries its origin endpoint node id, resolved against the full graph (chain mode has no `focusedEndpoint` to fall back on).
- **Clicking a finding always updates the panel**: the resolution no longer bails when the collapsed-graph lookup misses, so clicking a share no longer left the previous portscan's panel on screen.
- **"Discovered by" resolves the real producer even for id-escaped values**: the backend escapes `\` and `|` in node ids, so a share value like `\\host\NETLOGON` never matched a rebuilt `NODE_FINDING|type|value`. Match producers by the node's actual / canonical id instead — fixed in both the chain and focused views.
- **A finding's highlight is scoped to its real production path**: skip forward causal edges in the upstream walk and restrict to the producing injector(s), so a plain portscan click no longer lit the whole downstream kill-chain and every injector on the shared endpoint.
- **Injector panel findings come from the full graph** (per-execution `findingsNodeIds`): shares now show for an injector that only produced shares, which the drawer category endpoint missed (it maps `files`→`file`).
- **Producing-action list falls back to the full graph** while the endpoint feed loads, removing the brief "no producing action" flash.

### Alignment with merged backend PRs
- **Real per-finding expectation verdicts** (#6912): the panel reads the finding node's `verdicts` (prevention / detection / vulnerability, aggregated worst-of across producers) instead of the previous placeholder. The DTO already serialises the exact `success` / `failed` / `unknown` labels.
- **SMB shares as the native `file` type** (#6972): the causal share reconciliation and the tests use `file`.

### i18n
- Restored the action/arsenal locale keys that an earlier rebase resolution had dropped, so the i18n checker passes.

check-ts and eslint clean; attack-path unit tests green. Verified live against a locally rebuilt backend carrying #6912 + #6972.

Related issue: #6647


laugiov (commented):
One thing worth confirming before merge: in compact mode `disablePagination` hides the pager, but the fetch still uses the default page size of 20. An attack-path inject from a scan (nmap, nuclei, netexec) often has more than 20 distinct findings, so the drawer would show 20 and silently hide the rest, with no "20 of N" and no way to reach the others. `totalElements` is already tracked in PaginationComponentV2, so a small "showing 20 of N" with a link to the inject's full Findings list (or a larger compact page size) would keep it from reading as "this inject has 20 findings". Is capping at 20 intended here?

Seb-MIGUEL:
Good catch — fixed in c756b8155. In compact mode I now raise the page size to 100 (covers the vast majority of scans without a pager), and when a run still produces more, a "Showing X of N findings — open the inject to see them all." note appears (total tracked from `totalElements`). So the drawer never silently reads as "this inject has N findings", and the full list stays reachable via the inject (Action details → its Findings tab).

laugiov (approved):

PR #6982, opened 2026-07-27
fix(attack-path): expand the focused endpoint's finding clusters and reclaim hidden columns' width (#6647)

Seb-MIGUEL:
## Description

Two UX fixes on the attack path view, reported from the chokepoint drill-down flow.

**1. Focusing an endpoint now opens its findings straight away.** Clicking a chokepoint (from the "Top chokepoints" popover, the table, or the search) rebuilt the graph as the focused endpoint path but left every finding-type cluster collapsed — the analyst had to click each one (portscan, files, credentials…) to see anything. The focused view now pre-expands all of the endpoint's finding-type clusters, fed by a single `fetchEndpointFindings` call split per type (no extra round-trips versus expanding them by hand).

**2. The findings drawer no longer truncates the Assets column.** `FindingList` column widths are percentages of the full six-column table. In compact mode the drawer hides three of them, so Type/Value/Assets only used 57% of the row: the asset chips ellipsized while the right half sat empty. The hidden columns' share is now redistributed over the visible ones, keeping their relative proportions. Full-page usage (no hidden fields) is unchanged.

Follow-up to #6964.

## Checks
- `yarn eslint`, `yarn tsc --noEmit` clean.
- `yarn vitest run` on the attack-path and findings suites: 27 tests pass.

### DIFF
--- a/openaev-front/src/admin/components/findings/FindingList.tsx
+++ b/openaev-front/src/admin/components/findings/FindingList.tsx
@@ -175,6 +175,24 @@ const FindingList = ({ searchDistinctFindings, filterLocalStorageKey, contextId,
 
   const visibleHeaders = headers.filter(h => !hiddenFields.includes(h.field));
 
+  // The column widths above are percentages of the full six-column table: hiding columns would
+  // otherwise leave the remainder crammed on the left with dead space on the right (the assets chips
+  // truncating while the row is half empty). Re-spread the hidden columns' share over the visible
+  // ones, keeping their relative proportions.
+  const visibleStyles: Record<string, CSSProperties> = (() => {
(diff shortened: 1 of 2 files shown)


Seb-MIGUEL:
Closing: nothing unique left in this PR.

- The attack-path change (auto-expand a focused endpoint's finding clusters) is superseded by #7003, which reimplements the same behaviour (seeded from the already-loaded endpoint findings, plus the cluster-dimming fix).
- The FindingList column rescale is already on main via #6995 (`visibleWidthTotal` / `visibleStyles` in FindingList.tsx) — same effect, different implementation.

No remaining delta to merge.

PR #6999, opened 2026-07-28
fix(attack-path): keep SMB findings under the share type instead of relabelling them as file (#6647)

Seb-MIGUEL:
## Why

#6972 introduced an interim stand-in presenting SMB `share` findings as `file` so the "Captured Files" card had real data in the MVP. There is no native `file` finding type — neither in `ContractOutputType` (openaev-model) nor in the shared `client-python` enum — and the NetExec injector only ever emits `share` (`contracts/contract_outputs.py`: `ContractOutputElement(type=ContractOutputType.Share, field="shares")`).

The relabel is not just cosmetic. A share is a **complex** finding (host + share name + permissions); a file would be a primitive (a name). Keeping the bridge would break two things the moment a native `file` type lands:

- **Counters**: the top bar dedupes on `type|value`, so one share plus one file would report "2 files".
- **Kill-chain correctness**: `AttackPathKeyMatcher` mapped the consumed key `share_name` to the finding type `file`, so a step consuming a share name would match a genuine file finding and draw a false causal edge.

## What changes

Backend — the relabel is gone, findings keep their stored type end to end:
- Removed `presentType()` and its six read-boundary mappers from `AttackPathGraphService`; repository rows now flow straight through.
- `categoryTypes`: `files` → `shares` (mapping to the `share` type); both counter switches count `share`.
- `AttackPathCounters.files` → `shares`.
- `AttackPathKeyMatcher`: `share_name` reconciles to `share`.

Frontend — the card becomes **Discovered Shares** (folder-network icon), with `COVERED_FINDING_TYPES`, `CATEGORY_OF_TYPE`, `FILTER_TO_FINDING_TYPES` and the front-side key matcher realigned on `share`. Translations updated in the nine locales.

Tests — `AttackPathFileTypeApiTest` becomes `AttackPathShareTypeApiTest`, asserting `share` survives every read (endpoint expand, graph node, drawer, execution detail, counter), plus a new case pinning the reason for this PR: **a share and a file on the same endpoint are counted separately**.

## New skill

`.github/skills/add-contract-output-type/SKILL.md` documents the end-to-end procedure for adding a finding type — platform enum, `client-python` enum, output processor, chaining registry (primitive vs complex), frontend, injector-side declaration, attack-path wiring and tests — with the pitfalls this PR ran into. Next step is a NetExec contract producing a real `file` finding, which will follow that procedure.

## Checks
- Backend: Spotless applied, `mvn test-compile` on `openaev-api` clean.
- Frontend: `yarn eslint`, `yarn tsc --noEmit` clean; `yarn vitest run` on the attack-path suites, 27 tests pass.
- The backend integration suite (`AttackPathShareTypeApiTest`) has not been run locally — it needs the DB-backed test stack; CI will cover it.


Seb-MIGUEL:
Superseded by #7016, which is stacked on this branch and now targets main directly — it contains this commit (the share/file stand-in removal) plus the native complex `file` finding type. Closing to avoid a redundant merge. (#6647)

PR #7059, opened 2026-07-29
feat(multitenancy): activate api v2 isolation for injectors and connector_instances (#6410)

### LINKED ISSUE #6408: feat(multitenancy): API isolation v2 connector instance GCatalog
**Epic**: GH issue epic generated by script: [feat(multitenancy): epic for refactor for tenant scoped API](https://github.com/OpenAEV-Platform/openaev/issues/6393)

---

## Context

`ConnectorInstanceApi` is mapped from codebase analysis to tenant table(s): `connector_instances`.
Current state: activation = linked (execution engine), isolation mechanism = v1, test coverage = TenantIsolation nested exists.

## Definition of Done

- [ ] API behavior is aligned with v2 tenant-scoped isolation expectations
- [ ] Existing v1-only assumptions are removed where required for this endpoint
- [ ] Update tenant isolation integration tests for this API (`TenantIsolation` nested class or dedicated isolation test file)
- [ ] Add/validate tenant A vs tenant B isolation and out-of-rights access behavior

### LINKED ISSUE #6410: feat(multitenancy): API isolation v2 injector GCatalog
**Epic**: GH issue epic generated by script: [feat(multitenancy): epic for refactor for tenant scoped API](https://github.com/OpenAEV-Platform/openaev/issues/6393)

---

## Context

`InjectorApi` is mapped from codebase analysis to tenant table(s): `injectors`.
Current state: activation = linked (execution engine), isolation mechanism = v1, test coverage = TenantIsolation nested exists.

## Definition of Done

- [ ] API behavior is aligned with v2 tenant-scoped isolation expectations
- [ ] Existing v1-only assumptions are removed where required for this endpoint
- [ ] Update tenant isolation integration tests for this API (`TenantIsolation` nested class or dedicated isolation test file)
- [ ] Add/validate tenant A vs tenant B isolation and out-of-rights access behavior

corinnekrych:
### Proposed changes

Multi-tenancy API v2 activation for the `injectors` and `connector_instances` tables: remove the v1 `@Filter`/listener isolation from both entities, add them to `openaev.tenant.active-tables` (with production config guards), wire `TxCtx` through every HTTP entrypoint that reads them (with `@RequireTenantSelector` on ID-based reads, options endpoints and writes on the composite-PK `injectors` table), make write attribution explicit via `TenantWriteScopeResolver`/`setTenant(...)`, and move datapack/runtime-migration transactions under `MigrationProcessor`'s tenant-scoped transactions (new `TenantScopedTransaction.setScopeOnCurrentTransaction` primitive for joining the tenant-onboarding ambient transaction). XTM Composer paths are explicitly cross-tenant (`TxCtx.allTenants()`), and a cross-tenant read hole in `ConnectorInstanceLogService` is closed. Extensive HTTP/non-admin isolation tests and architecture guards pin the behavior.

* Closes #6410
* Closes #6408

### Testing Instructions: Injectors v2 Activation - Manual QA Checklist

Run every scenario with **at least two tenants** present (Tenant A / Tenant B) unless noted otherwise, to catch cross-tenant leakage introduced by the injectors table's move to multi-tenancy v2.

#### 1. CRUD on injectors in the catalog (Settings -> Injectors / Integrations)

- [x] Log in as admin of Tenant A, open the injector catalog
- [x] Confirm only Tenant A's own injectors + platform built-ins are visible (email, manual, challenge, media pressure, OAEV implant)
- [x] Register/create Tenant B, log in as its admin
- [x] Confirm Tenant B's catalog does **not** show any custom injector created only under Tenant A
- [x] As Tenant A, edit an injector's config
- [x] Confirm the save succeeds and the change persists on reload
- [x] As Tenant B, try to fetch Tenant A's injector by ID directly (URL/API call with the known ID)
- [x] Confirm the response is 404 / not found, not leaked
- [x] Delete an injector under Tenant A
- [x] Confirm the row is removed from Tenant A's list only, and any linked inject/contract under Tenant A shows the injector as unavailable
- [x] Confirm Tenant B is unaffected by the deletion
- [x] As platform super-admin (all-tenants view, if it exists), confirm both tenants' injectors are listed distinctly and tagged correctly

#### 2. New tenant -> starter pack provisioning

- [x] Create a brand-new tenant via the UI (Settings -> Tenants -> Create)
- [x] Immediately open that tenant's injector catalog
- [x] Confirm built-in injectors are present
- [x] Create and launch an atomic test / inject using a built-in immediately after tenant creation, with no manual setup

#### 3. Fresh DB -> default tenant starter pack

- [x] Spin up a fresh environment (empty Postgres volume / `docker-compose down -v` + up)
- [x] Boot the API and let it fully start (startup path, not onboarding path)
- [x] Log in with the default tenant
- [x] Confirm the default tenant has all built-in injectors + executors + default contracts registered
- [x] Restart the API a second time on the same (now non-empty) DB
- [x] Confirm no duplicate rows are created (idempotency check)
- [x] Confirm startup does not error or hang

#### 4. Inject execution - OpenAEV agent (atomic testing)

- [x] Deploy/register an OpenAEV agent on an endpoint, confirm it appears under Assets for the tenant
- [x] Create an atomic test using the `openaev_implant` injector type, targeting that agent
- [x] Launch it
- [x] Confirm the inject executes, the agent picks it up, and status/result flows back correctly
- [x] Repeat under a second tenant with its own agent
- [x] Confirm Tenant A's inject never targets/executes on Tenant B's agent

#### 5. Inject execution nmap

- [x] Install OAV agent
- [x] Install nmap nuclei inject
- [x] Create an atomic test with contract "Nmap - TCP Connect Scan"
- [x] Launch it
- [x] Confirm execution status is reported back correctly
- [x] Confirm Tenant B cannot see Tenant A's atomic testing


corinnekrych:

corinnekrych:

corinnekrych:

corinnekrych:

corinnekrych:

corinnekrych:

corinnekrych:

corinnekrych:

corinnekrych:

openaev-api/src/main/java/io/openaev/api/xtm_composer/XtmComposerApi.java
```diff
@@ -36,6 +36,11 @@ public class XtmComposerApi extends RestBehavior {
   private final XtmComposerService xtmComposerService;
   private final ConnectorOrchestrationService orchestrationService;
 
+  // Tenant URI needed for RBAC only, NOT because the underlying data is tenant-scoped.
+  // Without the tenant path here, a user whose only CATALOG-capable group lives on a
+  // non-default tenant would resolve zero capabilities and get a 403, even though registration
+  // itself is not scoped to any tenant (issue #6485).
+  // TODO v2: #7248
```
corinnekrych, line 43: XTMComposerApi for now need Tenant in URI because of RBAC check, but in reality this is cross-tenant. I created #7248 to fix on its once on permission is v2

.github/skills/activate-tenant-table/SKILL.md
```diff
@@ -210,52 +210,135 @@ Write down every hit and classify it:
...
-method), not just one hop. Stopping at the first caller is the #7188-class
-regression: the executors activation (#6409) found `ExerciseService
-.throwIfExerciseNotLaunchable`'s callers `updateExerciseStart` and
-`deprecatedUpdateExerciseStart` and wired `TxCtx` on both, but
-`ExerciseApi#changeExerciseStatus` — a third, equally direct caller of the
-exact same gate — was never enumerated and shipped with the tenant-scope gap
-undetected. Every intermediate method found this way (a shared gate/helper
-like `throwIfXNotLaunchable`, a mapper, a projection builder, ...) becomes its
-own grep target:
+**Walk the FULL transitive closure of callers, not just one hop.** A single
```
corinnekrych, line 215: This is to ensure we check all APIs, we used to have lots of API TxCtx arg missed when on a complete separe aPi with embedded joins

.github/skills/activate-tenant-table/SKILL.md
```diff
@@ -449,6 +532,107 @@ Model: `openaev-api/src/main/java/io/openaev/rest/mapper/MapperApi.java`.
 Re-run the test class after each endpoint. Read tests go green one by one.
 Do not move to writes until all reads are green.
 
+### Phase 3b — Every direct or indirect link to the table, everywhere
+
+Phase 1 finds code that reads `{table}` through `{EntityRepository}` or a
```
corinnekrych, line 539: additional instructions to not miss join table

.github/skills/activate-tenant-table/SKILL.md
```diff
@@ -831,6 +1015,22 @@ patterns that conflict with or duplicate the v2 mechanism.
    attribution is now explicit via `TenantWriteScopeResolver`, the listener is
    dead code for this entity and should be removed to avoid a hidden fallback
    to `TenantContext`.
+4. **`// TODO v2:` markers anywhere in the codebase naming `{table}`** — a
```
corinnekrych, line 1020: I add this commodity of adding a placeholder when i knwo future tickets will sort out or have an impact on the current implementation

openaev-api/src/main/java/io/openaev/integration/impl/executors/crowdstrike/CrowdStrikeExecutorIntegrationFactory.java
```diff
@@ -90,8 +90,8 @@ protected final String getClassName() {
   }
 
   @Override
-  protected void runMigrations() throws Exception {
-    crowdStrikeExecutorConfigurationMigration.migrate();
+  protected void runMigrations(String tenantId) throws Exception {
+    crowdStrikeExecutorConfigurationMigration.migrate(tenantId);
```
corinnekrych, line 94: since connector_instance are fully v2, background job thta run connector migration needed an explicity tenantId rather that ambiant TenatContext.getCurrentTenant. This imply a large signature refactor.

openaev-model/src/main/java/io/openaev/context/TenantScopedTransaction.java
```diff
@@ -102,6 +102,33 @@ public void executeNew(TxCtx ctx, Runnable work) {
...
+   * FK-constrained insert against it fails immediately.
+   *
+   * <p>Unlike {@link #executeNew}, a failure in the caller's subsequent work marks the WHOLE
+   * ambient transaction rollback-only, same as any other code running in it. This is deliberate: a
+   * failed provisioning step must not leave a half-provisioned, committed tenant behind — the
+   * caller (the create-tenant HTTP transaction) rolls back everything, tenant row included.
+   *
+   * <p>Refuses to run outside an active transaction: there is no ambient scope to join.
+   */
+  public void setScopeOnCurrentTransaction(TxCtx ctx) {
```
corinnekrych, line 122: **TD;DR man-crafted 🤗** : New method created because executeNew does not cover all cases...
1/ for REST API, the creation of a new tenant that then trigger the datapack in a same txt need to update the tenant id to the newly created one.
2/ for background job, setting the scope at service layer cover multiple path

**Verbose 🤖 IA:** Why setScopeOnCurrentTransaction exists, and why ManagerCreator needed it
1. Origin (cdb8244339, "fix(datapack): create tenant flow need to set scope in ambient tx"). MigrationProcessor.createDependencyForTenant runs INSIDE TenantService.create()'s still-open transaction (the new Tenant row is uncommitted). It must write FK-constrained child rows (tags, datapacks) referencing that same tenant. executeNew() (REQUIRES_NEW) was the wrong tool: it suspends the ambient transaction and opens a new connection with its own snapshot, which can't see the uncommitted Tenant row — every FK insert fails immediately. So a third primitive was added: setScopeOnCurrentTransaction — set the GUC scope on the transaction already open, join it, open nothing new.
2. Reused for ManagerCreator (9773dec5aa, connector_instances activation). Once connector_instances went v2-active, ManagerCreator.createManager(tenantId) — which reads/writes that table — had to be scoped too. But this method is reached from wildly different callers with wildly different scoping states:
•an HTTP request already carrying its own TxCtx (correctly scoped already),
•ManagerIntegrationsSyncJob, which opens its transaction via tenantTx.execute(TxCtx.forTenant(id), …) (also already scoped),
•tenant-creation's ambient provisioning transaction (uncommitted Tenant row — same trap as #1, executeNew would break it),
•and a caller with no scope set at all (ComchecksExecutionJob's deliberately transaction-less lookup, which still joins createManager's own @Transactional).
Rather than auditing and fixing each caller's scoping individually (and re-auditing every future caller), ManagerCreator became the single choke point: it unconditionally (re-)asserts TxCtx.forTenant(tenantId) on whatever transaction it's joining. setScopeOnCurrentTransaction is safe for all four cases because it never opens a transaction — it can't break the uncommitted-tenant-row case, and it's a harmless no-op re-assertion when the caller was already correctly scoped.

openaev-api/src/main/java/io/openaev/processor/datapack/DataPack.java
```diff
@@ -17,20 +20,42 @@
...
   protected abstract boolean doProcess();
 
+  /**
+   * Enables the v1 Hibernate {@code tenantFilter} for the current tenant. Call this explicitly,
+   * when doing a datapack that touches v1 tables. Once the tables are migrated to v2, this call can
+   * be removed.
+   */
+  // TODO v2: once tags, tags_rules get v2 activated, remove this method and all calls to it (and
+  // the v1 filter itself)
+  protected void enableV1TenantFilter() {
```
corinnekrych, line 49: I try to be minimal in the changes on DataPack as there is a dedicated issue Antoine will look at.
but for some v1 table tga, tag_rule we still need v1 activation for nowq and with the start-up job we need to make sure this is set otherwise it will fail with no FK not found for tags

Image

openaev-api/src/main/java/io/openaev/utils/InjectUtils.java
```diff
@@ -79,8 +78,7 @@ public Injector resolveInjector(
     }
     if (injectorContract != null) {
       return injectorRepository
-          .findFirstByContractsCompositeIdIdAndTenantId(
-              injectorContract.getId(), TenantContext.getCurrentTenant())
+          .findFirstByContractsCompositeIdId(injectorContract.getId())
           .orElse(null);
     }
```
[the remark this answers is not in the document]
SamuelHassine: Fixed in the branch (and preserved in the squashed commit): the contract-based fallback now goes through the tenant-scoped findFirstByContractsCompositeIdIdAndTenantId (the non-tenant variant was dropped from InjectorRepository), using TenantContext as the tenant source because the inject entity itself is still on v1 - documented inline. The explicit injectorId branch stays on findByInjectorId, scoped by the v2 inspector through the endpoint's TxCtx; the ID-based endpoints now also force a single-tenant selector so the lookup cannot run under a multi-tenant scope.

openaev-api/src/main/java/io/openaev/rest/injector/InjectorApi.java
```diff
@@ -105,11 +109,9 @@ public Iterable<InjectorOutput> injectors(
       resourceId = "#injectorId",
       actionPerformed = Action.READ,
       resourceType = ResourceType.INJECTOR)
-  public Collection<JsonNode> injectorInjectTypes(@PathVariable String injectorId) {
+  public Collection<JsonNode> injectorInjectTypes(TxCtx ctx, @PathVariable String injectorId) {
     Injector injector =
-        injectorRepository
-            .findByIdAndTenantId(injectorId, TenantContext.getCurrentTenant())
-            .orElseThrow(ElementNotFoundException::new);
+        injectorRepository.findByInjectorId(injectorId).orElseThrow(ElementNotFoundException::new);
```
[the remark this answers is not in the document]
SamuelHassine: Fixed: injectorInjectTypes (GET /{injectorId}/injector_contracts) and the injector detail endpoint (GET /{injectorId}) now take @RequireTenantSelector TxCtx, matching the CollectorApi precedent for single-id lookups on composite-PK tables. A multi-tenant caller without a selector falls back to its single tenant / the default tenant instead of running an ambiguous findByInjectorId across tenants.

openaev-api/src/main/java/io/openaev/rest/injector/InjectorApi.java
```diff
@@ -278,6 +291,7 @@ public InjectorRegistration registerInjector(
   @Transactional
   @AccessControl(actionPerformed = Action.SEARCH, resourceType = ResourceType.INJECTOR)
   public List<FilterUtilsJpa.Option> optionsByName(
+      TxCtx ctx,
       @RequestParam(required = false) final String searchText,
       @RequestParam(required = false) final String sourceId) {
     return fromIterable(
```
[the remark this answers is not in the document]
SamuelHassine: Fixed: both options endpoints (GET and POST /options) now take @RequireTenantSelector TxCtx, so they always resolve a single tenant and cannot return duplicated built-in injector ids across tenants.

openaev-front/src/admin/components/integrations/common/ConnectorLayout.tsx
```diff
@@ -88,7 +90,15 @@ const ConnectorLayout = () => {
...
+    }).catch((error) => {
+      setLoading(false);
+      // The related-ids lookup returns 404 both when the connector genuinely
+      // does not exist and when it belongs to another tenant
+      if (error?.status === 404) {
+        MESSAGING$.notifyError(t('This item does not exist or you are not allowed to view it.'));
+        navigate(routes.list);
+      }
+    });
   }, [connectorId, apiRequest, dispatch]);
```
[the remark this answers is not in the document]
SamuelHassine: Intentionally NOT added, with an explanatory comment now in the code. This repo's eslint config does not enable react-hooks/exhaustive-deps, and the deps here are functionally load-bearing: useDataLoader registers the callback in a module-level SSE listener map keyed by identity and re-runs the effect whenever it changes. t from useFormatter is a new closure on every render and navigate changes identity on navigation, so adding them would re-register and refetch on every render/route change (potential refetch loop). MESSAGING$ is a module-level import (stable). All four values behave correctly when captured once for the lifetime of the mounted layout: the translation key is constant, routes.list is constant per connector type, and a captured navigate stays functional in react-router v6.

openaev-api/src/main/java/io/openaev/api/xtm_composer/XtmComposerApi.java
SamuelHassine: Thanks for the context - makes sense, resolving.

.github/skills/activate-tenant-table/SKILL.md
SamuelHassine: Thanks for the context - makes sense, resolving.

.github/skills/activate-tenant-table/SKILL.md
SamuelHassine: Thanks for the context - makes sense, resolving.

.github/skills/activate-tenant-table/SKILL.md
SamuelHassine: Thanks for the context - makes sense, resolving.

openaev-api/src/main/java/io/openaev/integration/impl/executors/crowdstrike/CrowdStrikeExecutorIntegrationFactory.java
SamuelHassine: Thanks for the context - makes sense, resolving.

openaev-model/src/main/java/io/openaev/context/TenantScopedTransaction.java
SamuelHassine: Thanks for the context - makes sense, resolving.

openaev-api/src/main/java/io/openaev/processor/datapack/DataPack.java
SamuelHassine: Thanks for the context - makes sense, resolving.

openaev-model/src/main/java/io/openaev/database/repository/InjectorRepository.java
```diff
@@ -16,38 +16,43 @@
 public interface InjectorRepository
     extends JpaRepository<Injector, ConnectorCompositeId>, JpaSpecificationExecutor<Injector> {
 
-  Optional<Injector> findByIdAndTenantId(@NotNull String id, @NotNull String tenantId);
+  @Query("SELECT i FROM Injector i WHERE i.id = :id")
+  Optional<Injector> findByInjectorId(@Param("id") @NotNull String id);
```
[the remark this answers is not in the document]
SamuelHassine: Partially fixed, remainder is tracked interim state. Every ID-based read on the API surface now carries @RequireTenantSelector (detail, /{id}/injector_contracts, both options endpoints - update/delete/register/related-ids already had it), so these lookups always run under a single-tenant scope. findByType's only production caller (InjectorService#injectorByType) currently has no callers. The remaining single-valued call sites (InjectUtils#resolveInjector's explicit-id branch) run under the tenant path or per-tenant background scopes today, and their inject-side callers migrate to explicit tenant handling with the injects v2 activation; a multi-tenant collision there fails loudly (non-unique) rather than silently crossing tenants.

openaev-api/src/main/java/io/openaev/processor/core/RuntimeMigration.java
```diff
@@ -29,7 +35,6 @@ protected RuntimeMigration(DataPackService dataPackService) {
   @Getter private final String migrationId = getProcessableId();
 
   @Override
```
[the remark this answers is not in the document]
SamuelHassine: Good catch - fixed. RuntimeMigration now exposes the same enableV1TenantFilter() helper DataPack got, and V20260725_Fix_starter_pack_payload_contracts calls it at the top of doMigrate(): its findById(String) on the composite-PK injectors_contracts table indeed matched every tenant's copy of a built-in contract once the @Transactional aspect (which used to enable the filter implicitly) no longer fires under MigrationProcessor's programmatic transactions. The other runtime migrations were audited: V20260420 reads only the v2-scoped injectors table (scoped by the GUC), and the remaining datapacks either enable the filter explicitly (starter pack, tags) or use tenant-parameterized queries.

openaev-api/src/main/java/io/openaev/rest/injector_contract/InjectorContractApi.java
```diff
@@ -167,7 +176,8 @@ public InjectorContract injectorContract(@PathVariable String injectorContractId
   @Transactional
   @AccessControl(actionPerformed = Action.CREATE, resourceType = ResourceType.INJECTOR_CONTRACT)
   public InjectorContract createInjectorContract(
-      @Valid @RequestBody InjectorContractAddInput input) {
+      @RequireTenantSelector TxCtx ctx, @Valid @RequestBody InjectorContractAddInput input) {
```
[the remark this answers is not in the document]
SamuelHassine: Fixed. InjectorContractService#createNewInjectorContract now stamps the new contract with its linked injector's tenant explicitly (a contract lives in its injector's tenant by construction), instead of relying on TenantBaseListener's TenantContext fallback. A header-selected tenant can no longer produce a contract stamped on the default tenant while linked to another tenant's injector.

openaev-api/src/main/java/io/openaev/api/threat_arsenal/ThreatArsenalApi.java
```diff
@@ -184,7 +188,8 @@ private List<PrimitiveType> resolveAvailableTypes() {
   @Transactional
   @AccessControl(actionPerformed = Action.CREATE, resourceType = ResourceType.THREAT_ARSENAL)
   public ThreatArsenalAction createAction(
-      @Valid @RequestBody ThreatArsenalActionCreateInput input) {
+      @RequireTenantSelector TxCtx ctx, @Valid @RequestBody ThreatArsenalActionCreateInput input) {
+    writeScopeResolver.tenantForWrite(ctx, null);
     return threatArsenalService.create(input);
   }
```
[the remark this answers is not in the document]
SamuelHassine: Acknowledged, deliberately out of scope for this activation. The entity written here is still on multi-tenancy v1, and its write attribution (TenantBaseListener + TenantContext) is unchanged by this PR - propagating an explicit tenant through the v1 service layers is exactly the work of that table's own activation issue, as this PR does for injectors and connector_instances. On the tenant path (the route the frontend and the documented API use) TenantContext and TxCtx are aligned by TenantInterceptor, so there is no mixed attribution there; the X-Tenant-Ids header mismatch on the global route is a pre-existing property of every v1-entity endpoint, not introduced by this change. The tenantForWrite guard added here is a fail-fast check that rejects ambiguous multi-tenant scopes before any write happens.

openaev-api/src/main/java/io/openaev/api/stix_process/StixApi.java
```diff
@@ -62,7 +63,7 @@ public class StixApi extends RestBehavior {
     @ApiResponse(responseCode = "500", description = "Unexpected server error")
   })
   @AccessControl(actionPerformed = Action.PROCESS, resourceType = ResourceType.STIX_BUNDLE)
-  public ResponseEntity<?> processBundle(@RequestBody @Validated CTIEvent ctiEvent)
+  public ResponseEntity<?> processBundle(@RequestBody @Validated CTIEvent ctiEvent, TxCtx ctx)
       throws ParsingException, ConnectorError, IOException {
     String tenantId = TenantContext.getCurrentTenant();
```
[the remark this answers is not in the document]
SamuelHassine: Acknowledged, deliberately out of scope for this activation. The entity written here is still on multi-tenancy v1, and its write attribution (TenantBaseListener + TenantContext) is unchanged by this PR - propagating an explicit tenant through the v1 service layers is exactly the work of that table's own activation issue, as this PR does for injectors and connector_instances. On the tenant path (the route the frontend and the documented API use) TenantContext and TxCtx are aligned by TenantInterceptor, so there is no mixed attribution there; the X-Tenant-Ids header mismatch on the global route is a pre-existing property of every v1-entity endpoint, not introduced by this change. The tenantForWrite guard added here is a fail-fast check that rejects ambiguous multi-tenant scopes before any write happens.

openaev-api/src/main/java/io/openaev/rest/atomic_testing/AtomicTestingApi.java
```diff
@@ -91,7 +91,10 @@ public StatusPayloadOutput findAtomicTestingPayload(@PathVariable String injectI
   @AccessControl(actionPerformed = Action.CREATE, resourceType = ResourceType.ATOMIC_TESTING)
   @Transactional(rollbackFor = Exception.class)
   public InjectResultOverviewOutput createAtomicTesting(
-      @Valid @RequestBody AtomicTestingInput input) {
+      // The TxCtx parameter is not used directly; it signals the transaction aspect to set the
+      // tenant scope for this write (createOrUpdate resolves the Injector via
+      // InjectUtils#resolveInjector, which reads the v2-scoped injectors table).
+      TxCtx ctx, @Valid @RequestBody AtomicTestingInput input) {
     return this.atomicTestingService.createOrUpdate(input, null);
```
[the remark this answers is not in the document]
SamuelHassine: Acknowledged, deliberately scoped. The Inject and its associations are still on v1; the bare TxCtx added here only makes the injector read (InjectUtils#resolveInjector) resolve under the request's scope, which the previously ambient-TenantContext lookup needed once injectors went v2-active. Requiring a single write tenant and threading it through createOrUpdate is the injects activation's own work - forcing @RequireTenantSelector now would change the endpoint contract ahead of that migration. On the tenant path TenantContext and TxCtx are aligned; the header-selector mismatch on the global route is a pre-existing property of every v1-entity endpoint. A duplicate built-in id under a genuinely multi-tenant scope fails loudly (non-unique) rather than silently resolving cross-tenant.

openaev-api/src/main/java/io/openaev/rest/payload/PayloadApi.java
```diff
@@ -71,7 +71,11 @@ public PayloadOutput payload(@PathVariable String payloadId) {
   @PostMapping({PAYLOAD_URI, TENANT_PAYLOAD_URI})
   @AccessControl(actionPerformed = Action.CREATE, resourceType = ResourceType.PAYLOAD)
   @Transactional(rollbackFor = Exception.class)
-  public PayloadOutput createPayload(@Valid @RequestBody PayloadCreateInput input) {
+  public PayloadOutput createPayload(
+      // unused directly: signals the transaction aspect so the read of the v2-scoped
+      // `injectors` table inside PayloadService#synchroniseInjectorContractBasedOnPayload
+      // (via InjectorRepository#findAllByPayloads) resolves the caller's tenant scope.
+      TxCtx ctx, @Valid @RequestBody PayloadCreateInput input) {
```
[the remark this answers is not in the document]
SamuelHassine: Acknowledged, deliberately out of scope for this activation. The entity written here is still on multi-tenancy v1, and its write attribution (TenantBaseListener + TenantContext) is unchanged by this PR - propagating an explicit tenant through the v1 service layers is exactly the work of that table's own activation issue, as this PR does for injectors and connector_instances. On the tenant path (the route the frontend and the documented API use) TenantContext and TxCtx are aligned by TenantInterceptor, so there is no mixed attribution there; the X-Tenant-Ids header mismatch on the global route is a pre-existing property of every v1-entity endpoint, not introduced by this change. The tenantForWrite guard added here is a fail-fast check that rejects ambiguous multi-tenant scopes before any write happens.

SamuelHassine:
## Review-and-fix pass summary

Full rebase onto latest `main`, independent senior review, Copilot thread handling, and a CI-fix loop. Not approving: the four infra E2E jobs are red (details below), and they were already red on the pre-rebase head - this needs the author's eyes on the agent execution path.

### Rebase onto main
- Rebased/merged the branch onto latest `main` (through `c195c4d33`) and squashed the 153-commit history into signed Conventional Commits (all `%G? == G`).
- Reconciled overlap with main PR #7234 ("stop a simulation without erasing what it produced"): a merge on this branch had silently reverted #7234 (ExerciseService cancel logic, ExerciseServiceIntegrationTest, ExerciseHeader.tsx wording, the 9 lang files). Restored main's version of all of those while preserving this PR's own new i18n key ("This item does not exist or you are not allowed to view it.").

### Independent review findings (fixed)
- `AbstractConnectorService.getConnectorRelationsId(id, tenantId)` returned 200 with `registered=false` for a connector not visible in the caller's tenant, while the single-arg overload throws 404. Restored the 404 guard so the API does not leak id existence across tenants (this is what `InjectorHttpIsolationTest.relatedIdsForCrossTenantInjectorIsNotFound` asserts).
- `InjectorApi`: added `@RequireTenantSelector` to the ID-based reads (`/{id}`, `/{id}/injector_contracts`) and both options endpoints - built-in injector ids repeat across tenants, so a multi-tenant scope made `findByInjectorId`/`findAll` ambiguous.
- `RuntimeMigration`: dropping `@Transactional` (migrations now run under `MigrationProcessor`'s programmatic tenant transaction) also stopped `HibernateFilterTransactionAspect` from enabling the v1 `tenantFilter`. `V20260725_Fix_starter_pack_payload_contracts.findById(String)` on the composite-PK `injectors_contracts` table then matched every tenant's copy of a built-in contract. Added an explicit `enableV1TenantFilter()` opt-in (mirrors `DataPack`).
- `InjectorContractService.createNewInjectorContract`: stamp the new contract with its injector's tenant explicitly instead of relying on `TenantBaseListener` + `TenantContext`, so a header-selected tenant cannot save a contract on the default tenant while linked to another tenant's injector.

### Copilot review threads (all resolved)
- Applied the three blocking `@RequireTenantSelector` suggestions and confirmed the `InjectUtils` fallback is tenant-scoped.
- `ConnectorLayout.tsx` `useCallback` deps: intentionally left narrow (this repo's eslint does not enable `react-hooks/exhaustive-deps`; the callback is an SSE reload listener keyed by identity, so adding `t`/`navigate` would refetch on every render). Documented inline.
- Balanced-review findings about v1 write attribution on `Payload`/`ThreatArsenalAction`/atomic-testing `Inject`: those entities are still on v1 and their attribution is out of scope for this activation (their own activation issues own that). The added `tenantForWrite` guards fail-fast on ambiguous scopes.

### CI-fix loop
- First run: 3 API-test shards red. Root cause was pre-existing test/prod drift on the branch, now fixed: EntityManager injection into manually-constructed datapacks (`StarterPackTest`, `TagRulesPackTest`), default-tenant membership grants for the mock user (`ThreatArsenalApiTest`, `ThreatArsenalApiImporterTest`, `InjectApiTest`, mirroring `ScenarioInjectTestApiTest`), and two arch-test allowlist entries (`SecurityPlatformApi`, `CapabilityResolverService` - both landed on main after the injectors allowlist was authored).
- After the fixes: all 8 API-test shards, all arch tests, frontend build/lint/tsc/i18n, spotless, migrations guard, docker builds, and the non-infra E2E shards (arsenals, multitenant, remaining) are green.

### Remaining blocker (not introduced here)
- `E2E (infra-agent ...)` and `E2E (infra-multitenant ...)` x2 arch each = 4 jobs red, plus the Pipeline Gate that depends on them. The failing assertion is `agent-implant.spec.ts` -> "Implant spawn by the agent" trace never appears: the atomic test with an `openaev_implant` payload launches but the agent never reports execution traces back (the inject reaches EXECUTED with no target traces).
- These same 4 infra jobs were already failing on the pre-rebase head, so the rebase and the fixes above did not introduce them. Reproduced the CI image + backing services locally, but the failure needs a live installed agent reporting back, which the local harness cannot provide. The background execution path (`InjectsExecutionJob.executeInTenant`) does set both `TxCtx.forTenant` and `TenantContext`, so the injector/executor reads during execution are scoped - the root cause looks deeper in the implant execution/trace path and needs the author's domain knowledge.

Everything else is green, all review threads are resolved, the title/description follow the convention and link the issues (Closes #6410, Closes #6408). Holding the approval until the infra E2E is resolved.

openaev-api/src/main/java/io/openaev/service/connectors/AbstractConnectorService.java
```diff
@@ -293,6 +293,54 @@ protected ConnectorIds getConnectorRelationsId(T connector, String connectorId)
...
+   *     scope, or from the specific connector when iterating a broader scope
+   * @return connector instance ID and catalog connector ID if available, null values if not found
+   * @throws ElementNotFoundException if the connector is not visible in the requesting tenant's
+   *     scope (wrong tenant, or the id does not exist at all)
+   */
+  public ConnectorIds getConnectorRelationsId(String connectorId, String tenantId) {
+    ConnectorInstanceConfigurationRepository.ConnectorIdsFromDatabase relatedIds =
+        connectorInstanceConfigurationRepository.findInstanceAndCatalogIdsByKeyValueAndTenantId(
+            this.connectorType.getIdKeyName(), connectorId, tenantId);
+    T connector = getConnectorById(connectorId);
```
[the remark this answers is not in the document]
SamuelHassine: Good point on the second half of the method. In practice this is safe for the two call sites: the request-scoped callers (ExecutorApi/InjectorApi related-ids) run under a single-tenant `@RequireTenantSelector` scope, so `getConnectorById` is already narrowed to one tenant. The broader-scope caller (`CalderaSettingsService`) passes each executor's OWN tenant id and reads the config lookup with the explicit `(connectorType, connectorId, tenantId)` predicate; the subsequent `getConnectorById` resolves the executor by its composite id through the inspector, and executors are single-instance per tenant there, so a non-unique result does not arise in that path. I have kept the explicit `tenantId` on the config lookup (the half that a native query would otherwise leave unscoped). Threading `tenantId` into an abstract tenant-aware `getConnectorById` to make both halves symmetric is a reasonable hardening, but it is a cross-connector-type refactor (collector/injector/executor all override `getConnectorById`) that is orthogonal to this activation and better done in its own change; I have not folded it in here to keep the activation focused.

SamuelHassine:
## Review-and-fix pass 2 - rebase + infra E2E root-cause

Rebased onto the latest main (through 3eab14040, the phishing capability #7149) and drove the whole non-infra rollup green. All 6 branch commits are signed (`%G? == G`). Remaining red: the 4 infra E2E jobs + the Pipeline Gate that depends on them.

### Rebase adaptation and CI fixes (all non-infra now green)
- `PhishingInjectorIntegrationFactory` (landed on main after this branch): adapted its `runMigrations()` override to the new `runMigrations(String tenantId)` contract this PR introduces, so the build compiles after the rebase.
- `TenantActiveTableAccessArchTest`: allowlisted `PhishingLandingPageService` (reads `InjectorRepository` via the tenant-explicit `findByTypeAndTenantId(PhishingContract.TYPE, tenantId)`, mirroring `PayloadService`).
- `InjectorActivationConfigTest` / `ConnectorInstanceActivationConfigTest`: parse `openaev.tenant.active-tables` into exact tokens instead of substring matching (balanced Copilot finding), so `injectors_archive` cannot satisfy the guard.
- Import ordering fix flagged by spotless.

All resolved review threads addressed; the last open thread (getConnectorById determinism) answered and resolved.

### Infra E2E root-cause (agent-implant.spec.ts)
The 4 infra jobs pass on main's tip (run 31313793238, attempt 1) and fail deterministically on this branch across 3 fresh runs/reruns, so the trigger is in this diff. I stood up the exact CI image (openaev:ci-31311027941 == this branch build) with real Postgres/ES/RabbitMQ/MinIO and reproduced the FULL agent flow, including running the real openaev-agent binary against it:

- Agent registers, polls jobs, picks up the implant job, and the backend records `Implant spawn by the agent` (verified in DB and via SSE in the failing CI traces).
- Every UI-facing endpoint returns that trace correctly under both the admin token and the agent's service-account token: `POST /endpoints/jobs`, `DELETE /endpoints/jobs/{id}`, `GET /injects/execution-traces` (AGENT and ASSETS target types), `GET /injects/status-all-traces`, and the target-search endpoints. The browser session itself receives the full trace payload.
- The atomic-testing launch transitions DRAFT -> QUEUING synchronously on the tenant path.

So the backend agent/implant/trace pipeline is functionally correct on this branch. In the failing CI traces the browser never calls `GET /injects/execution-traces` (0 calls) - the redesigned "Results by target" panel never mounts `ExecutionStatusDetail` for the selected target, so the (correctly produced and served) trace is never surfaced and the assertion times out. That panel is frontend code untouched by this PR (the only frontend diff here is `ConnectorLayout.tsx` + i18n keys), so it behaves identically to main.

I could not reproduce the red locally (the panel renders from correct data; my sandbox implant cannot report a real stdout so the inject only reaches spawn+timeout, which is not a faithful pass/fail of the UI panel). No backend read of `injectors` / `connector_instances` on the agent-callback, queue-consumer, scheduled-job or trace-ingestion paths was found running without a tenant scope; the `can_access_tenant` grant resolves for the agent service account.

Net: non-infra CI is fully green; the infra E2E remains red but I could not attribute it to a backend defect in this diff despite an end-to-end real-agent reproduction. Flagging for maintainer eyes on the redesigned atomic-testing "Results by target" selection/timing under the infra runners.

Not approving while the rollup is red.

openaev-api/src/main/java/io/openaev/rest/payload/service/PayloadService.java
```diff
@@ -88,8 +88,7 @@ public class PayloadService {
 
   public InjectorContract synchroniseInjectorContractBasedOnPayload(
       Payload payload, List<AttackPattern> attackPatterns, Set<Domain> domains, Set<Tag> tags) {
-    List<Injector> injectors =
-        this.injectorRepository.findAllByPayloadsAndTenantId(true, payload.getTenant().getId());
+    List<Injector> injectors = this.injectorRepository.findAllByPayloads(true);
```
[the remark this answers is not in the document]
SamuelHassine: Good catch. Fixed: `synchroniseInjectorContractBasedOnPayload` now narrows the `findAllByPayloads(true)` result to the payload's own tenant (`payload.getTenant().getId()`) before picking the reference injector and building the `injectors_injector_contracts` links, so the contract - stamped on the payload's tenant - can only reference that tenant's injector, and a multi-tenant ambient scope can no longer mix tenants or trip the join-table FK.

SamuelHassine (approved):
Approving: full CI rollup is green including the four infra E2E jobs (infra-agent and infra-multitenant on amd64/arm64) and the Pipeline Gate, all review threads are resolved, and all commits are signed.

The infra E2E regression is fixed at the root: the GET atomic-testing detail read had no TxCtx, so with injectors on v2 the inject's injector resolved with no tenant scope and inject_type came back null, and the frontend gates the "Results by target" panel on a non-null inject_type - so the panel never mounted, the execution-trace fetch never fired, and the agent's "Implant spawn by the agent" trace never rendered. Adding TxCtx to the detail read restores the type. Also narrowed the payload injector-contract sync to the payload's tenant.

SamuelHassine:
## Review-and-fix: infra E2E root-caused and fixed - full rollup green

The four infra E2E jobs (infra-agent + infra-multitenant, amd64/arm64) and the Pipeline Gate are now green, along with the rest of the rollup. All commits are signed and all review threads are resolved.

### Root cause of the infra E2E failure
I ran a main-vs-branch side-by-side of the identical agent-implant flow (both CI images, real Postgres/ES/RabbitMQ/MinIO, and the real openaev-agent binary) and diffed every API response the UI consumes. The seeded injectors/contracts and the target-search responses were identical; the only meaningful difference was in the atomic-testing detail: `inject_type` was `openaev_implant` on main but `null` on the branch.

`Inject#getType()` resolves the inject's injector through the contract's (eager) injector link on the `injectors` table. With injectors activated in v2, the tenant scope is a per-transaction GUC set by the transaction aspect only when the endpoint carries a `TxCtx`. Every write endpoint on `AtomicTestingApi` had one, but the GET detail read (`findAtomicTesting`) did not - so the injector load ran with no tenant scope, failed closed, and `inject_type` came back null. The frontend gates the "Results by target" panel on a non-null `inject_type` (`AtomicTesting.tsx`), so the panel never mounted, `GET /injects/execution-traces` was never called (matching the failing CI traces, which showed zero such calls), and the agent's `Implant spawn by the agent` trace never rendered - exactly the assertion that timed out.

Verified on the running branch image: for the same inject, the TxCtx-less detail returned a null `inject_type` while a TxCtx-bearing endpoint (duplicate) returned `openaev_implant`.

### Fixes
- `AtomicTestingApi#findAtomicTesting` and `#findAtomicTestingPayload`: add `TxCtx` so the injector resolves inside the tenant-scoped transaction and `inject_type` is populated. Regression test added in `AtomicTestingApiTenantIsolationTest`.
- `PayloadService#synchroniseInjectorContractBasedOnPayload`: narrow the payload-supporting injector lookup to the payload's own tenant before selecting/linking, so a multi-tenant ambient scope cannot mix tenants or trip the injector-contract join-table FK (addresses the Copilot blocking comment).

### Rebase and hygiene
Rebased onto the latest main (through the phishing capability); adapted `PhishingInjectorIntegrationFactory` to the new `runMigrations(String tenantId)` signature, allowlisted `PhishingLandingPageService` in the injectors arch guard, and tightened the activation-config guards to exact-token matching. All commits signed.

openaev-api/src/test/java/io/openaev/architecture/TenantActiveTableAccessArchTest.java
laugiov: This waiver comment is now stale: the datapack is converted in this very PR (DataPack is no longer @Transactional, and V20260330 writes cwes through the primitive scope set by MigrationProcessor). 
It is not an INSERT-only VALUES/TenantContext waiver anymore. 
Could you update the comment to reflect the real mechanism? 
Same for the cwes waiver text in the activate-tenant-table skill, which still cites this as the example.

openaev-model/src/main/java/io/openaev/context/TenantScopedTransaction.java
laugiov, line 122: Thanks for the detailed rationale above, it's convincing and the ManagerCreator choke-point makes sense.

One thing missing: the method has no test of its own. The allTenants() happy path is covered indirectly by XtmComposerConnectorInstancesCrossTenantTest, but the two behaviors that make this door distinct are untested: it refuses to run outside an active transaction, and a later failure marks the WHOLE ambient transaction rollback-only (the difference from executeNew). Could we pin those two in TenantScopedTransactionIntegrationTest before merge, to keep the primitive's test-backed standard?

laugiov:
I added 4 tests pinning setScopeOnCurrentTransaction, plus fixed the stale cwes waiver comment on the access guard.

laugiov:
Fixed the two allowlisted jobs flagged here. Once injectors is v2-active, the injector read fail-closed to null on these background transactions. Both now stamp the tenant of their own aggregate (comcheck's exercise, scenario) before resolving the injector, so the read runs under a real scope.

Also fixed the CI failure: V1_DataImporterTest built an injector with no tenant_id (now part of the composite PK), and re-froze the background-tx baseline (the Comcheck constructor grew two params).

damgouj (approved):

Issue #6894, opened 2026-07-23
feat(audit-logging): Phase 2 - US1 - Chunk 3 - Scheduled execution + target resolution


Dimfacion:
## Use case

Logging scheduled execution

1. Inject `Optional<AuditLogger>` into `InjectsExecutionJob`
2. Log after auto-start transition `SCHEDULED → RUNNING` in `handleAutoStartExercises()` (shared)
3. Enrich context with scenario info when the simulation originates from a scenario recurrence
4. Log resolved target list in `executeInject()` at inject execution time, once targets are known — including inactive agents and agentless assets in `contextData` (US.1a)
5. **Tests**:
    - Verify `handleAutoStartExercises()` emits `SCHEDULED_LAUNCH` audit event with `simulation_id`, `initiator = "scheduler"`
    - Verify scenario recurrence enriches context with `scenario_id`, `scenario_name`
    - Verify from-scratch simulation does NOT include scenario fields in context
    - Verify `executeInject()` emits `TARGET_RESOLUTION` audit event with structured `endpoints` list containing per-endpoint agent status (`AGENT_ACTIVE`, `AGENT_INACTIVE`, or `ASSET_AGENTLESS`)

Issue #5191, opened 2026-03-20
[bug][tabletop] Challenge inject expectation not initialized on simulation start — causes UI validation failure


nino-filigran:
## 🐛 Bug Report — Challenge Expectation Not Initialized at Simulation Start

---

## 📌 Context

When a simulation starts, all inject expectations are initialized with a default result:
- `inject-expectation.results[0].result = null` (for both user and agent)

This initialization is the trigger that allows the UI to display the expectation and let the facilitator/player interact with it.

**Challenge-type expectations are excluded from this initialization**, creating a critical inconsistency.

---

## 🔴 Problem Today

Challenge inject expectations are **not initialized** when the simulation starts.

**Root cause (identified):** Initializing a Challenge expectation with `score = null` and `result = null` breaks the UI — when these fields are null, the user can no longer validate the flag/answer. As a workaround, initialization was skipped entirely.

**Consequence:** The simulation can get stuck in `ONGOING` status indefinitely because:
- The Challenge expectation has no result object
- The completion check (`hasValidResults`) finds no results → never marks the inject as `COMPLETED`
- The simulation never transitions to a terminal state

This is directly related to issue #4831 (simulation stays ongoing) — the fix for injects without expectations did not cover this case.

---

## 🔁 Reproducible Steps

1. Create a scenario with a tabletop inject of type **Challenge**
2. Add a Challenge expectation (e.g., "Submit the correct flag")
3. Start the simulation
4. Observe: the Challenge expectation has no initialized result object
5. Attempt to validate the challenge as a player → UI is broken or unresponsive
6. Let all other expectations complete → simulation remains `ONGOING` indefinitely

---

## ✅ Expected Behaviour (Functional Target)

- **All** expectation types (Manual, Challenge, Article, Detection, Prevention) must be initialized with a result object when the simulation starts
- The Challenge expectation initialization must set `result = null` **without breaking the UI validation flow**
- The fix is: **repair the UI** to handle `result = null` gracefully (show the input form, not a broken state), then enable initialization for Challenge expectations like all other types
- Once initialized, the Challenge expectation participates in the simulation completion check normally

---

## ❌ Actual Behaviour

- Challenge expectations are skipped during initialization
- Players cannot validate Challenge injects
- Simulations with Challenge expectations never complete
- The `ONGOING` status persists indefinitely

---

## 💥 Impact

- **Severity:** Critical — blocks tabletop simulations with Challenge injects from completing
[...]

PR #7081, opened 2026-07-30
refactor(inject-expectation): wire behavior-based expectation creation into remaining executors (#5384)

### LINKED ISSUE #5384: refactor(inject-expectation): migrate initializeResults()
- `setupDefaultExpectationResults()` → calls `behavior.initializeResults()`
- Remove the old per-type duplicated initialization code
- Tests: verify that the initialized results are identical

⚠️In this task, you should include the standardization of initialization across each expectation type.
Today, the challenge is not initialized correctly, but fixing it will break the frontend. For more information, refer to this issue: https://github.com/OpenAEV-Platform/openaev/issues/5191

As a reminder, we initialize results like this:
Table top ⇒ one result per player
Technical ⇒ one result per security platform and per agent ( except for the vulnerability)

GaetanSantucci:
### Proposed changes

* Wire the behavior-based expectation creation into the two remaining executors (`Executor#executeExternal` and the OpenAEV implant executor) that were still calling the legacy `computeAndSaveExpectations`, and remove the legacy implementation, as part of the ongoing Inject Expectation refactoring (Vertical 2).
* Add `InjectExpectationService#computeAndSaveExpectationsFromInjectContent`, a single null-safe entry point shared by both executors: it converts the inject's stored JSON content, guards against a null content / null expectations field, and routes the raw expectations through `computeAndSaveExpectationsUsingBehaviors` (review feedback: NPE risk + duplicated block).
* Resolve asset targets at most once per creation call and share the resolved `AssetToExecute` list across all technical behaviors via a new `preResolvedAssets` parameter; the scheduled execution job's already-resolved targets (audit logging, #6894) are reused instead of re-resolving once per expectation type (review feedback: repeated `resolveAllAssetsToExecute` cost). `getValueTargetedAssetMap` is also hoisted out of the per-expectation loop.
* Add an `expectationConverter(ExecutableInject, io.openaev.model.inject.form.Expectation, ExpectationPropertiesConfig)` overload in `InjectExpectationUtils` converting a raw, untargeted content expectation into an untargeted `BaseInjectExpectation` template (common fields only); target resolution is deferred to the matching `ExpectationBehavior`.
* Move `expectedSecurityPlatforms` storage from `BaseInjectExpectation` to `TechnicalInjectExpectation` (same column, no migration needed) with deep-copy cloning, make `ExpectationBehavior` generic, and relocate collector-filtering / expiration-floor helpers into `InjectExpectationUtils`.
* Tests: replace the obsolete reflection-based tests of the removed legacy private methods with unit tests covering the null-content path and the "targets resolved exactly once" guarantee; update the security-platform helper tests to their new location.

### Testing Instructions

1. Run/trigger an inject (atomic testing or scheduled exercise) targeting an asset and/or asset group with at least one expectation of each type (PREVENTION, DETECTION, VULNERABILITY, MANUAL) declared in its content.
2. Verify that `inject_expectation` rows are created with the expected target (agent/asset/assetGroup) and score, matching what the legacy `computeAndSaveExpectations` produced for the same scenario.
3. Verify no expectations are created when the inject content has an empty `expectations` list (no-op / early return) and that an inject without stored content does not fail.
4. Check logs/traces for exceptions during JSON content deserialization (`contentConvert`) and behavior resolution (`resolveFor`).

### Related issues

* Relates #5384


openaev-api/src/main/java/io/openaev/executors/Executor.java
```diff
@@ -59,9 +59,13 @@ private InjectStatus executeExternal(ExecutableInject executableInject, Injector
     InjectStatus injectStatus =
         this.injectStatusRepository.findByInjectId(inject.getId()).orElseThrow();
 
-    List<AssetToExecute> assetToExecutes = this.injectService.resolveAllAssetsToExecute(inject);
-    injectExpectationService.computeAndSaveExpectations(
-        executableInject, inject, injector.getType(), assetToExecutes);
+    BaseInjectContent content =
+        injectExpectationService.contentConvert(executableInject, BaseInjectContent.class);
+    List<io.openaev.model.inject.form.Expectation> expectationsFromInjectContent =
+        content.getExpectations();
```
gabriel-peze: potential NPE, it could be great to make this line nullsafe

openaev-api/src/main/java/io/openaev/executors/Executor.java
```diff
@@ -59,9 +59,13 @@ private InjectStatus executeExternal(ExecutableInject executableInject, Injector
...
+
+    injectExpectationService.computeAndSaveExpectationsUsingBehaviors(
+        executableInject, expectationsFromInjectContent, injector.getType());
```
gabriel-peze: The 6 previous lines are duplicated on OpenAEVImplantExecutor.java, maybe you could mutualise it into the injectExpectationService

openaev-api/src/main/java/io/openaev/injectors/openaev/OpenAEVImplantExecutor.java
```diff
@@ -46,9 +47,13 @@ public ExecutionProcess process(Execution execution, ExecutableInject injection)
...
-    // Compute expectations
-    injectExpectationService.computeAndSaveExpectations(
-        injection, inject, OAEV_IMPLANT, assetToExecutes);
+    BaseInjectContent content =
+        injectExpectationService.contentConvert(injection, BaseInjectContent.class);
+    List<io.openaev.model.inject.form.Expectation> expectationsFromInjectContent =
+        content.getExpectations();
+
+    injectExpectationService.computeAndSaveExpectationsUsingBehaviors(
+        injection, expectationsFromInjectContent, OAEV_IMPLANT);
```
gabriel-peze: This code is a duplication of the one in Executor.java

openaev-api/src/main/java/io/openaev/service/expectation/AbstractTechnicalBehavior.java
```diff
@@ -28,7 +28,8 @@
  * InjectExpectation} refactoring (Vertical 2).
```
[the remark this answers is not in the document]
SamuelHassine: Fixed in the squashed commit: the stale "Dead code - not wired into any service yet" paragraph on AbstractTechnicalBehavior is removed now that the behaviors are wired into the executors (the other behavior classes had already been cleaned up in this PR).

openaev-api/src/main/java/io/openaev/service/expectation/AbstractTechnicalBehavior.java
```diff
@@ -45,12 +46,12 @@ public abstract class AbstractTechnicalBehavior implements ExpectationBehavior {
   @Override
   public void initializeAndSaveInjectExpectationsFromExecutableInject(
       ExecutableInject executableInject,
-      BaseInjectExpectation expectationTemplate,
+      TechnicalInjectExpectation expectationTemplate,
       String implantType) {
     Inject inject = executableInject.getInjection().getInject();
     List<AssetToExecute> assetToExecutes = this.injectService.resolveAllAssetsToExecute(inject);
```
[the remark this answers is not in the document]
SamuelHassine: Fixed: asset targets are now resolved at most once per creation call instead of once per expectation type. computeAndSaveExpectationsUsingBehaviors resolves the AssetToExecute list a single time (only when at least one technical template exists) and passes it to every behavior through a new preResolvedAssets parameter. Callers that already resolved targets (the scheduled execution job resolves them for audit logging and threads them through Executor#executeExternal) are reused instead of re-resolving. getValueTargetedAssetMap was also hoisted out of the per-expectation loop. A unit test now asserts resolveAllAssetsToExecute is called exactly once for PREVENTION+DETECTION+VULNERABILITY together.

openaev-api/src/main/java/io/openaev/executors/Executor.java
SamuelHassine: Fixed: both executors now call a single null-safe service entry point, InjectExpectationService#computeAndSaveExpectationsFromInjectContent. It guards against a null converted content (an inject without stored content makes contentConvert return null) and against a null expectations field, so the previous content.getExpectations() NPE can no longer happen. Covered by a new unit test (given_injectWithoutContent_should_notFailNorCreateExpectations).

openaev-api/src/main/java/io/openaev/executors/Executor.java
SamuelHassine: Done: the duplicated block (contentConvert + getExpectations + computeAndSaveExpectationsUsingBehaviors) is mutualised into InjectExpectationService#computeAndSaveExpectationsFromInjectContent, called by both Executor#executeExternal and OpenAEVImplantExecutor#process. This also made the line null-safe as requested in the other thread.

openaev-api/src/main/java/io/openaev/injectors/openaev/OpenAEVImplantExecutor.java
SamuelHassine: Done: this block is now a single call to the shared InjectExpectationService#computeAndSaveExpectationsFromInjectContent (same method used by Executor#executeExternal), which also reuses the assetToExecutes list already resolved a few lines above instead of resolving targets again.

openaev-api/src/main/java/io/openaev/executors/Executor.java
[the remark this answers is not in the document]
SamuelHassine: Fixed by encapsulating the content extraction inside the service API, as suggested: both executors now call InjectExpectationService#computeAndSaveExpectationsFromInjectContent, so ExecutorTest's mocked InjectExpectationService no longer needs to stub contentConvert (the mocked method is a no-op void call). The ExecutorTest external-execution cases pass again.

openaev-api/src/main/java/io/openaev/service/expectation/AbstractTechnicalBehavior.java
```diff
@@ -90,34 +101,36 @@ public void initializeAndSaveInjectExpectationsFromExecutableInject(
...
+                e ->
+                    isAgentExpectation(e)
+                        || isAgentlessAssetExpectationNecessary(e.getAsset(), inject))
+            .toList();
+    if (!leafExpectations.isEmpty()) {
+      // Fetched once per invocation (not per leaf), and only when a leaf actually needs it.
+      Map<String, Endpoint> valueTargetedAssetsMap = injectService.getValueTargetedAssetMap(inject);
+      leafExpectations.forEach(
+          e -> {
+            initializeResults(e);
```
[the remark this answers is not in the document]
SamuelHassine: Fixed in a05be5dcd: tenant security-platform collectors are now loaded once per behavior invocation (not once per leaf), the default pending results are restricted to the collectors matching the expectation's expectedSecurityPlatforms (empty/null keeps the legacy "every connected platform" behavior), and the expiration ordering floor is applied - the same rules as the legacy setupDefaultExpectationResults path. VulnerabilityBehavior keeps its vulnerability-specific default. Added the requested regression test with two connected platform types (EDR + XDR, expectation expecting EDR only) asserting only the expected type is seeded and collectors are fetched exactly once.

openaev-api/src/main/java/io/openaev/service/InjectExpectationService.java
```diff
@@ -101,62 +97,193 @@ public class InjectExpectationService {
...
+
+    ObjectNode storedContent = inject.getContent();
+    ObjectNode enrichedContent =
+        injectorContractContentUtils.setExpectations(
+            inject.getInjectorContract().get(),
+            storedContent != null ? storedContent.deepCopy() : null);
+    if (enrichedContent == null) {
+      return expectations;
+    }
+    return this.mapper.treeToValue(enrichedContent, BaseInjectContent.class).getExpectations();
```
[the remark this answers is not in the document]
SamuelHassine: Fixed in cbd676ed2: the contract-fallback return is now normalized (a null deserialized list becomes an empty list), so a stored explicit "expectations": null combined with a contract that declares no predefined expectations can no longer NPE on expectations.isEmpty(). Covered by a new unit test (given_contractFallbackWithoutPredefinedExpectations_should_notFail).

SamuelHassine (approved):
Full review and fix pass completed. The behavior-based wiring is correct, the legacy path removal is safe (remaining callers use buildAndSaveInjectExpectations), and all review feedback has been addressed: null-safe shared entry point for both executors, single asset-target resolution per creation call reusing the scheduled job's pre-resolved targets, security platform filtering and expiration floor restored in the technical behavior defaults, and the contract fallback normalized against null expectation lists. Rebased on main (audit-logging target resolution reconciled), squashed into signed commits, unit and architecture tests green, full CI green. Approving.

SamuelHassine:
## Review and fix summary (automated pass)

**What was done:**

- Independent full-file review of all 19 changed files plus callers (InjectsExecutionJob, chaining engine callers of buildAndSaveInjectExpectations, expectedSecurityPlatforms consumers).
- Reconciled the merge conflict with main: the audit-logging change (#6894) pre-resolves asset targets in the scheduled job; those targets are now threaded into the behavior-based expectation creation instead of being dropped.
- Mutualised the duplicated contentConvert block from Executor and OpenAEVImplantExecutor into a single null-safe service entry point, InjectExpectationService#computeAndSaveExpectationsFromInjectContent (fixes the NPE flagged in review and the ExecutorTest mock interaction).
- Asset targets are resolved at most once per creation call and shared across technical behaviors (was once per expectation type); getValueTargetedAssetMap and tenant collectors are fetched once per batch instead of per leaf.
- Restored the legacy seeding rules in the technical behavior defaults: expectedSecurityPlatforms collector filtering and the expiration ordering floor.
- Normalized the contract-fallback expectations list against explicit null content.
- Avoided @Transactional self-invocation (TenantBackgroundTransactionArchTest) via a private non-transactional worker.
- Replaced obsolete reflection-based tests of the removed legacy private methods with unit tests covering the null-content path, the single-resolution guarantee, the platform-filter regression (EDR vs XDR), and the null fallback list.

**Commits (all signed):** a2c6dbf22 (squash of the original 6 commits plus review fixes, Co-authored-by preserved), aed94ea43, a05be5dcd, cbd676ed2.

**Status:** all 8 review threads (Copilot lite + balanced, gabriel-peze) replied and resolved; CI fully green; mergeable CLEAN; PR title matches the Conventional Commits convention and the body links Closes #5384.

Not addressed here (pre-existing design decision from earlier verticals, worth a maintainer sanity check): MANUAL expectations are table-top only in the behavior model, while the removed legacy path could also produce asset-scoped manual expectations via getManualExpectationsByAsset.

openaev-api/src/main/java/io/openaev/service/InjectExpectationService.java
```diff
@@ -101,62 +97,197 @@ public class InjectExpectationService {
...
+  private void doComputeAndSaveExpectationsUsingBehaviors(
+      ExecutableInject executableInject,
+      List<io.openaev.model.inject.form.Expectation> expectationsFromInjectContent,
+      @Nullable String implantType,
+      @Nullable List<AssetToExecute> preResolvedAssets)
+      throws JsonProcessingException {
+
+    Inject inject = executableInject.getInjection().getInject();
+    List<io.openaev.model.inject.form.Expectation> expectations =
+        resolveExpectationsWithContractFallback(inject, expectationsFromInjectContent);
```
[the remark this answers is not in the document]
SamuelHassine: Fixed in 275527d2e: the behavior-based worker now applies the same guard as doBuildAndSaveInjectExpectations before resolving contract expectations or targets - a direct execution that is neither atomic testing nor a chaining run returns early and creates no expectations, matching the legacy computeAndSaveExpectations behavior. Covered by a unit test (given_directNonAtomicNonChainingExecution_should_notCreateBehaviorExpectations) asserting nothing is resolved or persisted.

openaev-api/src/main/java/io/openaev/service/InjectExpectationService.java
```diff
@@ -101,62 +97,197 @@ public class InjectExpectationService {
...
+    }
+    final List<AssetToExecute> resolvedAssets = assetToExecutes;
+
     injectExpectationsToApply.forEach(
         expectationTemplate -> {
-          ExpectationBehavior behavior = resolveFor(expectationTemplate);
+          ExpectationBehavior<BaseInjectExpectation> behavior = resolveFor(expectationTemplate);
           behavior.initializeAndSaveInjectExpectationsFromExecutableInject(
-              executableInject, expectationTemplate, implantType);
+              executableInject, expectationTemplate, implantType, resolvedAssets);
```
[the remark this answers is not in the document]
SamuelHassine: Fixed in 275527d2e: the group-level parent expectation is now created exactly once per asset group (tracked via a seen-group set inside the loop), instead of once per eligible asset of the group, so a group with N assets no longer produces N duplicate parent rows. The parent row still only appears when the group has at least one eligible asset, matching the legacy computeExpectationsForAssetGroup semantics. Covered by a unit test with one group containing two agentless assets asserting exactly one group parent row and one asset row per asset.

openaev-api/src/main/java/io/openaev/service/InjectExpectationUtils.java
```diff
@@ -400,4 +450,87 @@ private static boolean anyExpectationsMatch(
...
+      Double score,
+      boolean isGroup,
+      String name,
+      Long expirationTime,
+      BaseInjectExpectation.EXPECTATION_TYPE type,
+      ExpectationPropertiesConfig config) {
+
+    baseInjectExpectation.setExercise(executableInject.getInjection().getExercise());
+    baseInjectExpectation.setInject(executableInject.getInjection().getInject());
+    baseInjectExpectation.setExpectedScore(score);
```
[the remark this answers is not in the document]
SamuelHassine: Fixed in bc5fe160b: the content-based converter now normalizes a null score to the same 100.0 default the removed legacy factories applied (Objects.requireNonNullElse), so old or partially populated content can no longer produce an expectation whose null expected score gets unboxed on the first collector update. Covered by a unit test (given_nullScoreInContent_should_defaultExpectedScore).

SamuelHassine:
## Re-verification pass (automated)

The Copilot balanced re-review surfaced three additional legacy-parity regressions after the first pass; all confirmed valid and fixed in signed commits 275527d2e and bc5fe160b:

- Restored the direct-execution guard in the behavior-based worker: direct executions that are neither atomic testing nor chaining runs create no expectations, matching the removed legacy path.
- Group-level parent expectations are now created exactly once per asset group instead of once per eligible asset (no more N duplicate parent rows for a group with N assets).
- A null score in stored inject content defaults to 100.0 like the legacy factories, preventing a null unboxing on the first collector update.

Each fix ships with a unit test. All review threads are resolved, CI is fully green on bc5fe160b, and the PR remains approved and mergeable (CLEAN).

openaev-api/src/main/java/io/openaev/service/expectation/AbstractTechnicalBehavior.java
```diff
@@ -85,39 +103,45 @@ public void initializeAndSaveInjectExpectationsFromExecutableInject(
...
-      BaseInjectExpectation template,
+      TechnicalInjectExpectation template,
       @Nullable AssetGroup assetGroup,
       Asset asset,
       @Nullable Agent agent) {
-    TechnicalInjectExpectation expectation = (TechnicalInjectExpectation) template.clone();
+    TechnicalInjectExpectation expectation = template.clone();
     expectation.setAssetGroup(assetGroup);
     expectation.setAsset(asset);
     expectation.setAgent(agent);
```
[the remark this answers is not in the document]
SamuelHassine: Fixed in 36f9aa3e3: buildExpectationForTarget now restores the legacy factory semantics per hierarchy level - agent rows are always evaluated individually (flag false), asset rows are grouped exactly when they belong to an asset group (assetGroup != null), and only the group parent row keeps the content-configured flag from the template. Verified against the legacy factories (preventionExpectationForAsset/detectionExpectationForAsset/vulnerabilityExpectationForAsset all used setExpectationGroup(assetGroup != null); the ForAgent variants never set it). Covered by a unit test mixing a direct asset and a grouped asset with the content flag set to true.

SamuelHassine:
Re-verification round 3 (automated): the latest Copilot re-review flagged one more valid legacy-parity regression - the target clones propagated the content's expectationGroup flag to every hierarchy level, changing the grouped-asset scoring rules. Fixed in signed commit 36f9aa3e3 (level-specific flags restored: agent rows individual, grouped asset rows grouped, parent keeps the content flag) with a unit test. CI green, all threads resolved, PR remains approved and mergeable (CLEAN).

MarineLeM (changes requested):
Need to be tested

openaev-api/src/main/java/io/openaev/service/expectation/ChallengeBehavior.java
```diff
@@ -1,34 +1,91 @@
...
+    if (!cached.isEmpty()) {
+      return cached;
+    }
+    try {
+      ChallengeContent content =
+          mapper.treeToValue(
+              executableInject.getInjection().getInject().getContent(), ChallengeContent.class);
+      return fromIterable(challengeRepository.findAllById(content.getChallenges()));
+    } catch (Exception e) {
+      return List.of();
```
gabriel-peze, line 93: The error is silently lost here, is it possible to have something else than Exception, or if it's mandatory to catch Exception, at least log in warn ?

openaev-api/src/main/java/io/openaev/service/expectation/ArticleBehavior.java
```diff
@@ -27,7 +35,53 @@ public boolean supports(BaseInjectExpectation expectation) {
...
+    if (!cached.isEmpty()) {
+      return cached;
+    }
+    try {
+      ChannelContent content =
+          mapper.treeToValue(
+              executableInject.getInjection().getInject().getContent(), ChannelContent.class);
+      return fromIterable(articleRepository.findAllById(content.getArticles()));
+    } catch (Exception e) {
+      return List.of();
```
gabriel-peze, line 91: The error is silently lost here, is it possible to have something else than Exception, or if it's mandatory to catch Exception, at least log in warn ?

openaev-api/src/main/java/io/openaev/service/expectation/AbstractTechnicalBehavior.java
gabriel-peze, line 142: The same value is calculated here for all the expectations, could we calculate it once before and use the result instead ?

openaev-api/src/main/java/io/openaev/service/InjectExpectationUtils.java
```diff
@@ -22,15 +21,31 @@
...
+import lombok.extern.slf4j.Slf4j;
 
+@Slf4j
 public class InjectExpectationUtils {
 
   private InjectExpectationUtils() {}
 
   public static final double FAILED_SCORE_VALUE = 0.0;
 
+  //  public static buildChallengeExpectation
```
gabriel-peze: dead code commented ? if yes should be removed

openaev-api/src/main/java/io/openaev/service/expectation/ExpectationBehavior.java
```diff
@@ -106,4 +107,45 @@ default String resolveResultLabel(
...
+  /**
+   * Returns {@code true} if this behavior handles the given content-form expectation for the given
+   * inject. Defaults to a pure type match; inject-aware behaviors (e.g. phishing) override this to
+   * discriminate on the inject type for a shared expectation type.
+   *
+   * @param formExpectation the content-form expectation
+   * @param inject the inject the expectation is attached to
+   */
+  default boolean supportsFormExpectation(
+      io.openaev.model.inject.form.Expectation formExpectation, Inject inject) {
```
gabriel-peze: nitpick : The other Expectation have been removed from the class, you can directly use Expectation here.

openaev-api/src/main/java/io/openaev/service/expectation/ExpectationBehavior.java
```diff
@@ -106,4 +107,45 @@ default String resolveResultLabel(
...
+   * deferred to {@link #initializeAndSaveInjectExpectationsFromExecutableInject}, so each behavior
+   * expands the template into as many concrete expectations as its type requires.
+   *
+   * @param formExpectation the raw expectation declared in the inject content
+   * @param exercise the exercise the inject belongs to
+   * @param inject the inject the expectation is attached to
+   * @return an untargeted expectation template
+   */
+  default T convertFormExpectationToBaseInjectExpectation(
+      io.openaev.model.inject.form.Expectation formExpectation, Exercise exercise, Inject inject) {
```
gabriel-peze: nitpick : The other Expectation have been removed from the class, you can directly use Expectation here.

openaev-api/src/main/java/io/openaev/service/expectation/ExpectationBehaviorResolver.java
```diff
@@ -0,0 +1,61 @@
...
+   * expectation. Resolution is inject-aware so specialized behaviors (e.g. phishing) can win over
+   * their generic parent for the same expectation type. The unchecked cast is safe: the same
+   * behavior both builds the template and consumes it back.
+   *
+   * @param formExpectation the content-form expectation
+   * @param inject the inject the expectation is attached to
+   */
+  @SuppressWarnings("unchecked")
+  public ExpectationBehavior<BaseInjectExpectation> resolveForForm(
+      io.openaev.model.inject.form.Expectation formExpectation, Inject inject) {
```
gabriel-peze: nitpick : The other Expectation have been removed from the class, you can directly use Expectation here.

openaev-api/src/main/java/io/openaev/service/expectation/PhishingBehavior.java
```diff
@@ -0,0 +1,49 @@
...
+    if (!(expectation instanceof ManualInjectExpectation)) {
+      return false;
+    }
+    return expectation.getInject() != null
+        && PhishingContract.TYPE.equals(expectation.getInject().getType());
+  }
+
+  @Override
+  public boolean supportsFormExpectation(
+      io.openaev.model.inject.form.Expectation formExpectation,
```
gabriel-peze: nitpick : The other Expectation have been removed from the class, you can directly use Expectation here.

openaev-api/src/main/java/io/openaev/service/expectation/PhishingBehavior.java
```diff
@@ -0,0 +1,49 @@
...
+      io.openaev.database.model.Inject inject) {
```
gabriel-peze: nitpick : There is no other use of an Inject here, you can import him instead of using his FQN ?

gabriel-peze:
Made a few tests, specially on the signatures, and it looks good to me.

DeniseMonreale:
let's split this...

Please remove the challenge initialization path from this PR and drop "Closes #5384"  the executor wiring is good and can go in today...

Reason: enabling ChallengeBehavior creates a result with null score and ResultChip still renders any non-empty results array as "Validated" ....is exactly regression #5191. The frontend fix doesn't exist yet so idt that we should  gate this PR

#5384 stays open with two remaining items:
- the actual initialization standardization (the per-type branching is still in setupDefaultExpectationResults)
- a new frontend issue for ResultChip null-result handling

@GaetanSantucci @gabriel-peze @MarineLeM please telle me what is your POV

gabriel-peze:
Hello @DeniseMonreale 

I don't know if we should remove something from this PR, but we can protect some behavior by using a feature flag for the challenge i think.

In that case we will have a protection for the deployments and an environment (main-ff) to test and fix the behavior were the issue occur.

The change is done on the description to remove the "Closes" text.

Do not hesitate to correct me if i'm wrong @GaetanSantucci & @MarineLeM.

openaev-front/src/admin/components/common/challenges/ChallengesPreview.tsx
```diff
@@ -70,12 +70,14 @@ const ChallengesPreview: FunctionComponent<Props> = ({
...
     const sortedChallenges = challenges?.reduce<Record<string, ChallengeInformation[]>>((acc, challenge) => {
-      const category = challenge.challenge_detail?.challenge_category || '-';
+      const category = challenge.challenge_detail?.challenge_category || 'null';
       acc[category] = acc[category] || [];
       acc[category].push(challenge);
       return acc;
     }, {});
 
+    console.log('sortedChallenges', sortedChallenges);
+
```
antoinemzs: todo: remove debug trace

MarineLeM:
The fact that the initialized challenge produced a bug is fixed by this PR, @DeniseMonreale 

It's what we already planned to do!

MarineLeM (approved):

gabriel-peze (approved):

PR #7083, opened 2026-07-30
feat(crowdstrike): add trace on failed api post call (#3029)

### LINKED ISSUE #3029: fix: manage errors from CrowdStrike API execute action
## Use case

When I launch a payload from a CrowdStrike agent, if there is an error with the CS API which executes the attack, the inject stays in pending.

## Proposed Solution

See if it is possible to get the errors for each agent and log an error trace for the inject and agent concerned.

Yes

gabriel-peze:
### Proposed changes

* Add log to add more detail on failure of Crowdstrike API calls

### Testing Instructions

1. Make the crowdstrike API call fail
2. you should see in the log errors explaining what is the failure

### Related issues

* Closes #3029

### DIFF
--- a/openaev-api/src/main/java/io/openaev/executors/crowdstrike/client/CrowdStrikeExecutorClient.java
+++ b/openaev-api/src/main/java/io/openaev/executors/crowdstrike/client/CrowdStrikeExecutorClient.java
@@ -177,8 +177,6 @@ public class CrowdStrikeExecutorClient {
     } catch (IOException e) {
       logErrorMessage(e.getMessage(), e);
       throw new ExecutorException(e, e.getMessage(), CROWDSTRIKE_EXECUTOR_NAME);
-    } catch (Exception e) {
-      logErrorMessage(e.getMessage(), e);
     }
   }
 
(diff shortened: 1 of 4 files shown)


openaev-api/src/main/java/io/openaev/executors/crowdstrike/client/CrowdStrikeExecutorClient.java
```diff
@@ -207,21 +208,36 @@ private String post(@NotBlank final String uri, @NotNull final Map<String, Objec
       // Body
       StringEntity entity = new StringEntity(this.objectMapper.writeValueAsString(body));
       httpPost.setEntity(entity);
-      return httpClient.execute(httpPost, response -> EntityUtils.toString(response.getEntity()));
+      String responseEntity =
+          httpClient.execute(httpPost, response -> EntityUtils.toString(response.getEntity()));
+      ResourcesGroups resourcesGroups =
```
damgouj, line 213: question: ResourcesGroups work for the 2 post API ? The first one to get the session id and the second one after the attack ?

damgouj (approved):

openaev-api/src/main/java/io/openaev/executors/crowdstrike/client/CrowdStrikeExecutorClient.java
gabriel-peze: Yes, we receive in the response into responseEntity an object with the same pattern as the previous one :)

PR #7084, opened 2026-07-30
fix(chaining): payload steps without a condition mapper only ran once for a single asset

mergenhan:
## Summary

Refs OpenAEV-Platform/filigran-private#237

A chaining step template with a payload and **no condition mapper** (e.g. a root step, or a step gated only by a non-mapper condition such as `DEPEND_ON`) was permanently short-circuited by `StepService#createReadySteps` as soon as it had produced one `READY` step for a single scope asset. On every subsequent scheduling cycle, the following guard returned an empty list *before* `expandTargetBatches` ever ran, silently skipping every other asset in scope:

```java
if (!conditionService.hasConditionMapper(persistedTemplate)
    && isStepAlreadyExecutedOnce(persistedTemplate.getId(), workflowRun.getId())
    && injectExecutionStep.hasPayload(persistedTemplate)) {
  return List.of();
}
```

### Reported symptom

In a scenario with 3+ assets in scope and a mix of a payload-based step and an injector-contract-based step (rate limit 1 execution/min):
- the injector-contract step correctly fired once per asset across successive minutes,
- but the payload step only ever fired **once, for a single asset**, and was never retried for the remaining assets.

### Root cause

This guard predates per-target expansion for payload steps (introduced by #7060, "one execution per payload per asset"). It is now redundant and actively harmful:

- `expandTargetBatches` already fans out **every** `INJECT_EXECUTION` step — payload or not — into one batch per scope target.
- The per-target hash dedup (`committedTargetHashes` / `commitHashes`) a few lines further down in `createReadySteps` already prevents an already-executed `(combo, target)` pair from being re-created.

Because `hasPayload()` is `false` for pure injector-contract steps, the stale guard never triggered for them — only for payload steps — which is exactly the asymmetry that was reported.

## Fix

- Removed the obsolete template-level guard in `StepService#createReadySteps`.
- Removed the now-unused `StepService#isStepAlreadyExecutedOnce` helper. The repository method it wrapped (`existsByStepTemplateIdAndWorkflowId`) is still used elsewhere by `ConditionService` for `DEPEND_ON` conditions, so it was left untouched.

## Testing

- Added `StepServiceTest$Ready$PayloadStepMultiAssetExpansion`, which reproduces the exact reported scenario (payload step, no mapper, already executed once) and asserts the remaining scoped assets still get a `READY` step each.
  - Verified this new test **fails** against the previous (reverted) guard and **passes** with the fix.
- Full targeted suite: `StepServiceTest` + `ConditionServiceTest` → 160/160 passing, `BUILD SUCCESS`.

### DIFF
--- a/openaev-api/src/main/java/io/openaev/service/chaining/StepService.java
+++ b/openaev-api/src/main/java/io/openaev/service/chaining/StepService.java
@@ -113,16 +113,6 @@ public class StepService {
     Step persistedTemplate =
         findByIdAndStatus(nextStepTemplateToExecute.getId(), StepStatus.TEMPLATE);
 
-    // NOTE: a step template with no condition mapper and a payload used to be short-circuited here
-    // after its first execution, to avoid re-creating the same non-mapper batch on every scheduling
-    // cycle. That guard predates per-target expansion and is now redundant: expandTargetBatches
-    // (see below) expands every INJECT_EXECUTION step — payload or not — into one batch per scope
-    // target, and the per-target hash dedup a few lines down (committedTargetHashes/commitHashes)
-    // already prevents an already-executed (combo, target) pair from being re-created. Keeping
-    // the old template-level guard here caused a regression: a payload step without a mapper
-    // condition would only ever execute once for a single asset and be silently skipped for every
-    // other asset in scope, while sibling injector-contract steps (hasPayload() == false) kept
-    // expanding correctly.
 
     ActionStep actionStep =
         factoryAction(persistedTemplate.getStepAction(), persistedTemplate.getId());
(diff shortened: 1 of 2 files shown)


openaev-api/src/main/java/io/openaev/service/chaining/StepService.java
```diff
@@ -113,13 +113,16 @@ public List<Step> createReadySteps(
...
+    // NOTE: a step template with no condition mapper and a payload used to be short-circuited here
+    // after its first execution, to avoid re-creating the same non-mapper batch on every scheduling
+    // cycle. That guard predates per-target expansion and is now redundant: expandTargetBatches
+    // (see below) expands every INJECT_EXECUTION step — payload or not — into one batch per scope
+    // target, and the per-target hash dedup a few lines down (committedTargetHashes/commitHashes)
+    // already prevents an already-executed (combo, target) pair from being re-created. Keeping
+    // the old template-level guard here caused a regression: a payload step without a mapper
+    // condition would only ever execute once for a single asset and be silently skipped for every
+    // other asset in scope, while sibling injector-contract steps (hasPayload() == false) kept
+    // expanding correctly.
```
guillaumejparis:
```suggestion
```

guillaumejparis (approved):

PR #7108, opened 2026-07-31
fix(attack-path): cluster endpoint overflow + missing maven modules in Docker build (#7108)

Seb-MIGUEL:
## Summary

- **Attack path (causal-chain view)**: a depth column with more distinct endpoints than the cap (>4) rendered one node per host with no collapse, producing a long unreadable vertical stack (e.g. 56 endpoints from a /24 scope). Mirrors the existing per-type finding cluster (`CHAIN_FINDINGS_MAX_PER_TYPE`) with a new `CHAIN_ENDPOINTS_MAX_PER_DEPTH = 4` cap, collapsing the overflow into a single `apEndpointCluster` "+N" node. Clicking it reveals another `ENDPOINT_BATCH_SIZE` (10) hosts at a time rather than dumping the whole overflow at once.
- **Docker build**: the `Dockerfile`'s `api-builder` stage never `COPY`s `openaev-annotation-processor` and `openaev-maven-plugin` into the build context, even though the root `pom.xml` declares both as reactor modules — so a local `docker build` fails Maven's reactor resolution. Added the two missing `COPY` lines.

## Test plan

- [x] `yarn check-ts` passes with 0 errors
- [x] Verified visually: a simulation with 56 endpoints at one depth now shows 4 individual + one "+52" cluster node instead of a 56-node wall
- [x] Verified click-to-expand reveals 10 more endpoints per click, and the cluster count decrements accordingly
- [x] `docker build --target app` succeeds from source after the Dockerfile fix

- [x] Fixed a review finding: hidden endpoints in the collapsed cluster now route their findings through the cluster node for causal-edge purposes, so a downstream consumer of a hidden host's finding no longer loses its causal edge

### DIFF
--- a/openaev-front/src/admin/components/simulations/simulation/attack_path/attack-path-flow-helpers.ts
+++ b/openaev-front/src/admin/components/simulations/simulation/attack_path/attack-path-flow-helpers.ts
@@ -1411,6 +1411,9 @@ const CHAIN_EP_BLOCK_MIN = 120; // minimum height of one endpoint block (endpoin
 const CHAIN_FIND_HALF = 28; // half of AP_FINDING_SIZE (56), to centre a finding node on its row
 const CHAIN_FINDINGS_MAX_PER_TYPE = 4; // a type with more than this on an endpoint collapses into a "+N"
 // cluster (click to expand), so a heavy endpoint (e.g. 24 portscans) stays a few rows tall, not a column.
+const CHAIN_ENDPOINTS_MAX_PER_DEPTH = 4; // a depth column with more distinct endpoints than this collapses
+// the overflow into a single "+N" endpoint cluster (click to expand), so a step reaching dozens of hosts
+// stays a few blocks tall instead of one node per host in a long unreadable vertical stack.
 
 interface ChainStep {
   injectorId: string;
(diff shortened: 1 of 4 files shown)


openaev-front/src/admin/components/simulations/simulation/attack_path/attack-path-flow-helpers.ts
```diff
@@ -1841,6 +1843,67 @@ export const buildCausalChainFlow = (
 
       cursorY = blockTop + h + CHAIN_STEP_GAP;
     }
+
+    // The collapsed overflow: one "+N" endpoint cluster carrying every hidden host at this depth, wired
+    // to whichever injector(s) reached them. Its own findings stay hidden until expanded — mirrors the
+    // per-type finding cluster above (collapse hides detail behind a click, not just a count).
+    if (hiddenAssetIds.length > 0) {
```
laugiov, line 1864: Causal edges are dropped for hidden endpoints in the collapsed cluster (chain can silently break at scale). 
The visible endpoints populate causalSourceByFinding (:1807/:1812) so their findings can anchor a forward causal edge. 
The new if (hiddenAssetIds.length > 0) block (:1850) creates the +N node and injector→cluster edges but never populates causalSourceByFinding for the hidden endpoints' findings. 
In the forward pass, a downstream step consuming a key produced by a hidden host resolves causalSourceByFinding.get(fid) ?? fid (:1954) to fid, whose node was never placed → filtered out, no edge. And since matched = true (:1948) was already set, the if (!matched) dependsOn fallback (:1984) is skipped — so the consumer ends up with no causal edge and no fallback, disconnected until the cluster is expanded. 
This is inconsistent with the per-type finding cluster, which routes through the cluster node via causalSourceByFinding (:1807, exactly what the map comment at :1626 describes). 
Suggested fix inside the block: hiddenAssetIds.forEach(epId => (assetFindings.get(epId) ?? []).forEach(fid => causalSourceByFinding.set(fid, epClusterId))); so the hidden findings route their causal fan-in through the +N cluster node.

Seb-MIGUEL: Good catch, thanks. Fixed in 906e9b817: the `if (hiddenAssetIds.length > 0)` block now routes each hidden endpoint's findings to `epClusterId` via `causalSourceByFinding`, exactly as suggested, before building the injector→cluster edges. This aligns the behavior with the per-type finding cluster.

laugiov (approved):

Issue #6768, opened 2026-07-18
feat(fds): wire Filigran Design System color tokens into the OpenAEV theme


SamuelHassine:
## Use case

OpenAEV's MUI theme files (`openaev-front/src/components/ThemeDark.ts` / `ThemeLight.ts`) carry hardcoded hex values that have drifted from the Filigran Design System (FDS) reference palette (e.g. the `#00f1bd` literal that is 1-2 hex digits off `--color-filigran-tonic-primary`), and several fundamental theme slots (background, paper, accent, secondary/EE) diverge from the FDS values already validated on the OpenCTI pilot. There is also no body/html background gradient at all (flat fill only), unlike OpenCTI's proven two-stop pattern, and `background.secondary` was declared in the theme type but never assigned (resolved to `undefined` at runtime).

## Proposed solution

Wire the FDS color tokens into the OpenAEV theme, following the OpenCTI pilot precedent:

- Add a generated token bridge (`fds-tokens.generated.ts` + `.meta.json` sidecar) produced by the design-system repo's `generate:mui-bridge` script, plus `fds-migration/` tracking docs and a conformity-check script.
- Wire 9 fields in both modes to FDS tokens: `primary`, `secondary`, `EE_COLOR`, `gradient.main`, `xtmhub.main`, `accent`, `paper`, `background.default`, `background.secondary` (+ dark `nav` for free).
- Wire the body/html two-stop background gradient (ISO OpenCTI) and stop hiding it behind opaque content backgrounds in `private/Index.tsx` / `public/Index.tsx`.
- Keep deliberately-unwired decisions documented (severity palette, light nav, `text.secondary`, feedback colors - wave 2).

## Additional information

Scope arbitrated and checkpoint-validated on 2026-07-13 (visual review, live `getComputedStyle` verification, cross-product comparison with OpenCTI). Targets the `design-system/current` integration branch, not `main`.

PR #7150, opened 2026-08-02
feat(nav): replace the left menu with the design system Navbar

### LINKED ISSUE #7223: feat(fds): replace the left navigation with the design-system Navbar
## Use case

The Filigran Design System (FDS) is being rolled out across the XTM product line so that shared components are maintained once instead of once per product. The left navigation is the first component migrated in OpenAEV, and this migration also serves as the pilot that produced the implementation playbook used for the following products.

Today OpenAEV maintains its own navigation: its own selection logic, its own collapsed-state persistence, its own tenant switcher, and MUI menu items rather than real links. Every navigation fix has to be made again in each product.

## Current workaround

Each product keeps its own implementation and they drift apart.

## Proposed solution

Replace the hand-written left navigation with the `Navbar` component from `@filigran/design-system`, with strict iso-functionality: same entries, same permissions, same submenus, same persisted collapsed state.

What it brings:
- entries become real anchors, so Ctrl-click and middle-click open a new tab
- the tenant switcher is recomposed on the library's `Menu`, with real anchors and a proper Enterprise gate
- full keyboard accessibility and correct ARIA semantics come from the library

Out of scope: the top bar and other menus are unchanged. The MUI token bridge is tracked separately in #6768.

## Additional information

The library is a private git dependency, so this also wires the authentication needed to install it in CI.

Yes — the PR is already open and reviewed.

sandyghs:
## What changed, in plain terms

OpenAEV's left navigation is no longer OpenAEV's. It is now the `Navbar`
component of `@filigran/design-system`, consumed straight from the library as a
pinned git dependency. The menu entries, the permissions and feature flags that
filter them, the collapsed/expanded preference and where it is stored — all
unchanged. What changed is who owns the pixels.

Concretely, for a user: the rail is 180px expanded and 48px collapsed (it was
55px), rows are real links so Ctrl/Cmd-click and middle-click open a section in
a new tab, collapsed rows show tooltips and open their submenus as flyouts, the
product header is the library's `ProductSwitcher`, and on a viewport narrower
than 1024px the navigation now starts collapsed unless the user has already
expressed a preference.

## Why this pull request exists

This is the **pilot** for consuming the design system inside a product. It is
deliberately single-subject — navigation only — because its real job is to
find out what integrating the library actually costs, and to write that down
for the next product to follow. Two deliverables of equal rank:

1. this working integration, with green CI;
2. `fds-migration/IMPLEMENTATION-PLAYBOOK.md`, a step-by-step procedure written
   for a developer with none of this context.

**The green install job is the central proof.** Until this pull request, nobody
had ever installed the library from a product's CI.

## Library pin

`ad108751b6611c0a4f613bd0f1dcc15358821c81` — head of the library's `main`.

## What is in here

**Consuming the library.** The dependency is a pinned git dependency under the
mandatory `@filigran/design-system@` alias, with the repository allow-listed in
`.yarnrc.yml`, and the library stylesheet imported at the application entry
point.

**CI.** Two fixes, both prerequisites for the above:
- the workflows did not trigger on `design-system/current` at all, so nothing
  on this branch was ever tested;
- installing a private git dependency needs credentials. `FDS_GIT_TOKEN` is
  wired through `git config url.…insteadOf` ahead of every `yarn install`. A
  composite action, `setup-fds-auth`, holds the mechanism (token passed through
  `env:`, with a guard that fails loudly on an empty token) and is called from
  the four build composite actions; the three Alpine-based workflows, which
  cannot call a composite action, inline the same two lines. Nine install sites
  in total.

**The navigation itself.** A product adapter (`AppNavbar`, `useNavbarState`,
`nav-menu-model`) feeds the real menu, the real permissions and the real flags
into the library component. The legacy `leftmenu/` implementation is deleted,
along with the documented gradient bug it carried. The tenant switcher is
recomposed on library pieces (`NavbarItem asChild` + `Chip` + `Icon`).

**Host prerequisites** (`static/css/design-system-host.css`). Two rules.

- One is a permanent host concern, not a compensation: the application's menu
  icons come from MUI and render at 24px, while the design system's rows are
  built around 16px glyphs. Scoped to the navigation.
- One is the single surviving compensation: the library's overlay layer sits at
  `z-50`, below MUI's modal layer, so flyouts opened from the rail were painted
  under dialogs. It carries all three obligations this pull request requires of
  any compensation — a comment saying why, an entry in `LIBRARY-FEEDBACK.md`,
  and a testable removal condition (delete it once the library's overlays use a
  token rather than a hard-coded `z-50`).

Every other compensation written during this migration has been removed rather
than shipped; the table in the playbook lists each one with the condition that
retired it.

## Assumed divergences

- The tenant switcher's dropdown is recomposed on the library's `Menu` instead
  of the legacy MUI `Popover`. Its rows stay real anchors (`MenuItem asChild`),
  so middle-click and Ctrl-click still open a tenant in a new tab.
- Submenu entries no longer carry a tooltip of their own. The legacy menu
  wrapped every sub-item in one, duplicating a label that is already visible.
- Its Enterprise-Edition chip is filled where the legacy one was outlined —
  design decision — and uses a provisional tone until the library's own
  Enterprise-Edition tone lands.
- The navigation keeps OpenAEV's MUI icons rather than the library's lucide
  set. The icon slots take `ReactNode` by design, so nothing is lost; a
  convergence would be its own design mission.
- The logo's home destination is a plain anchor, so the tenant basename is
  prefixed by hand — the library applies no router basename (reported).
- A group row lights up while any of its children is the current page, whereas
  the legacy menu lit it only on an exact path match. The library derives the
  trigger's state from a descendant's `aria-current`; this is the library's
  behaviour, accepted rather than compensated.

## One compensation already removed

The pin includes the `ProductSwitcher` evolution, so the product-side collapsed
logo workaround is **deleted** rather than shipped: `logoCollapsed` and
`logoHref` replace an `` that swapped its own source on the collapse state
and forced `width: 100%` / `object-position: center` to survive a 126px slot
centred on a 48px rail. Re-measured after removal: collapsed logo 28x28 at
x=10 inside the rail (it used to sit at x=-39, off-screen), expanded 126x28 at
x=16. The full compensation-to-removal cycle is written up in the playbook as
the worked example of a pin bump.

## Verification

- Types, lint, translations, unit tests (15), production build: green on the
  reviewed commit, and every check on this pull request is green — including
  the install job, which genuinely clones and builds the library at the pin.
- Every element of the navigation was compared against the library's own
  documentation site running at this exact pin, by reading **computed styles**
  in a real browser, in both themes and both rail states. Every measured value
  now matches, or has a documented cause. The two preflight defects above were
  found that way — they were invisible to code review, since the markup and the
  class names are identical on both sides.
- Verified by hand on a running instance: both themes, both rail states,
  tooltips, flyouts, Ctrl-click, the switcher, narrow viewports, reload.

## Known, out of scope

- The MUI token bridge (`fds-tokens.generated.ts`) is stale: the library's
  `theme.css` has moved since it was generated in July. Regenerating restyles
  the whole application and needs its own review, so it is not riding along
  here.
- Only weight 400 of the product's fonts is imported, so bold is synthesised.
  Pre-existing, separate ticket.

## Related issues

* Closes #7223
* Relates to #6768 — the MUI token bridge is a neighbouring but distinct
  subject; this pull request does not close it.

## Feedback to the library

`fds-migration/LIBRARY-FEEDBACK.md` — 14 entries, each stating what the product
needed, what the library offers today, and what the product had to do instead.
Seven are already resolved by library releases, and each of those carries the
evidence of its removal.


.github/workflows/core-ci.yml
```diff
@@ -2,11 +2,14 @@ name: "[Core] CI"
 
 on:
   push:
-    branches: [main]
+    # design-system/current is the long-lived integration branch for the
+    # Filigran design-system migration. Without it here, no CI ever ran on
+    # that branch nor on any PR targeting it.
+    branches: [main, design-system/current]
```
RomuDeuxfois, line 12: nitpick: can be good to put a warn to remove that at this end :)

.github/workflows/deploy-design-system.yml
```diff
@@ -33,8 +33,19 @@ jobs:
           npm install -g corepack
         shell: alpine.sh --root {0}
       - name: Build into Alpine
+        env:
+          FDS_GIT_TOKEN: ${{ secrets.FDS_GIT_TOKEN }}
         run: |
           export JAVA_HOME=/usr/lib/jvm/default-jvm/
+          # openaev-front consumes the private XTM-Foundation/filigran-design-system
```
RomuDeuxfois, line 45: **todo**: why not using ./.github/actions/setup-fds-auth direclty ?

.github/workflows/deploy-testing-xtm-one.yml
```diff
@@ -33,8 +33,19 @@ jobs:
           npm install -g corepack
         shell: alpine.sh --root {0}
       - name: Build into Alpine
+        env:
+          FDS_GIT_TOKEN: ${{ secrets.FDS_GIT_TOKEN }}
         run: |
           export JAVA_HOME=/usr/lib/jvm/default-jvm/
+          # openaev-front consumes the private XTM-Foundation/filigran-design-system
```
RomuDeuxfois, line 45: **todo**: why not using ./.github/actions/setup-fds-auth direclty ?

.github/workflows/test-feature-branch.yml
```diff
@@ -31,8 +31,19 @@ jobs:
           npm install -g corepack
         shell: alpine.sh --root {0}
       - name: Build into Alpine
+        env:
+          FDS_GIT_TOKEN: ${{ secrets.FDS_GIT_TOKEN }}
         run: |
           export JAVA_HOME=/usr/lib/jvm/default-jvm/
+          # openaev-front consumes the private XTM-Foundation/filigran-design-system
```
RomuDeuxfois, line 43: **todo**: why not using ./.github/actions/setup-fds-auth direclty ?

openaev-front/src/admin/components/nav/LeftBar.tsx
```diff
@@ -197,15 +197,13 @@ const LeftBar = () => {
...
   entries.push(
-    {
-      userRight: settingsItems.some(item => item.userRight),
-      items: settingsItems,
-    },
-  );
-  const bottomEntries = [
+    // "Getting Started" sits right under "Settings" and scrolls with the rest
+    // of the menu. The group is always visible because that row is: "Settings"
+    // still hides on its own when the user has no settings right.
```
RomuDeuxfois: nitpick: Ia wrote a lot of comments, but most of them aren't relevant.

openaev-front/src/admin/components/nav/LeftBarHeader.tsx
```diff
@@ -128,112 +56,83 @@ const LeftBarHeader: FunctionComponent<Props> = ({ navOpen }) => {
...
+    <ProductSwitcher
+      label={t('Filigran products')}
+      logo={(
+        <img
+          src={theme.logo}
+          alt=""
+          // Inline, not utility classes: the product has no Tailwind build, so
+          // only the utilities the library's own components emit exist at
+          // runtime — `object-contain` is not one of them, and a class here
+          // silently stretched the logo to the slot's width.
```
RomuDeuxfois: todo: comment are not usefull at all

openaev-front/src/components/common/menu/navbar/MadeByFiligran.tsx
```diff
@@ -0,0 +1,78 @@
...
+ * stays aligned with the design system when those tokens move.
+ *
+ * Collapsed, it behaves like the header's logo: the label disappears and only
+ * the Filigran emblem remains. The emblem is not a separate asset — the
+ * wordmark SVG starts with it, so a 12px square box with `object-cover` and a
+ * left origin crops the lettering away. That is what the legacy row did, and
+ * it keeps a single asset to maintain. The accessible name is carried by
+ * `aria-label` and is therefore identical in both states.
+ */
+const MadeByFiligran: FunctionComponent<Props> = ({ collapsed }) => {
```
RomuDeuxfois, line 15: qs: Why isn't this part of the internal library?

openaev-front/src/components/common/menu/navbar/nav-menu-model.ts
```diff
@@ -0,0 +1,30 @@
+import { type ReactElement } from 'react';
+
+export interface NavMenuEntries {
```
RomuDeuxfois, line 3: QS: Why isn't this part of the internal library?

openaev-front/src/components/common/menu/navbar/NavbarRowContent.tsx
```diff
@@ -0,0 +1,51 @@
...
+ * composed by the consumer. Every row here IS a router `Link` (so Ctrl/⌘-click
+ * and "open in new tab" work), which means every row goes through `asChild` and
+ * therefore has to re-declare the anatomy the library would otherwise own.
+ *
+ * The class names below are the library's own — see `NavbarItem.tsx` /
+ * `NavbarSubmenu.tsx` — reproduced here rather than invented. Reported upstream:
+ * see fds-migration/LIBRARY-FEEDBACK.md, "NavbarItem has no link destination".
+ */
+
+const iconSpanStyle = { display: 'inline-flex' } as const;
```
RomuDeuxfois, line 18: QS: Why isn't this part of the internal library?

openaev-front/src/components/common/menu/navbar/useNavbarState.ts
```diff
@@ -0,0 +1,63 @@
+import { useCallback, useState } from 'react';
+
+import { MESSAGING$ } from '../../../../utils/Environment';
```
RomuDeuxfois, line 3: QS: Why isn't this part of the internal library?

RomuDeuxfois:
**tests**
1. no hand on collapse

- actual
{B49AE423-FEAD-4764-8113-CBDF547C760A}
{4ADD7F9B-A371-4AF8-9E79-B12A034C0D86}
{B441119F-2F97-410E-8910-A0B81647E996}

- expected
{D0D67EFA-A4B0-48B6-BB84-E56A419607E8}

3. we lost the current item on reload (already the case)

{3250AB79-863B-4E51-BDCE-E7CDCC295E23}

reload page

{5F5F8B92-0334-441F-AB09-137220CDB039}

sandyghs:
> **tests**
> 
> 1. no hand on collapse
> 
> * actual
> 
> {B49AE423-FEAD-4764-8113-CBDF547C760A} {4ADD7F9B-A371-4AF8-9E79-B12A034C0D86} {B441119F-2F97-410E-8910-A0B81647E996}
> * expected
> 
> {D0D67EFA-A4B0-48B6-BB84-E56A419607E8}
> 3. we lost the current item on reload (already the case)
> 
> {3250AB79-863B-4E51-BDCE-E7CDCC295E23}
> reload page
> 
> {5F5F8B92-0334-441F-AB09-137220CDB039}

no hand on collapse:
- Good catch, thank you, this is a real regression. 
- The cause is in the design system component, not in this PR, so I fixed it at the source rather than patching it here: XTM-Foundation/filigran-design-system#84, merged. The product will pick it up at the next pin bump.

we lost the current item on reload :
- Noted, thank you. As you point out it predates the migration — the legacy left bar behaved the same way, and OpenCTI solves it by persisting the open submenu in localStorage while OpenAEV keeps it in volatile state.
- I'd rather not fold it into an iso-functionality PR. It's recorded as a follow-up on the product side, and I've also logged the library-side option (a submenu that opens itself when it contains the current page) with the two constraints we established, so nobody has to re-derive them.

.github/workflows/deploy-design-system.yml
sandyghs: Fair question, and the fact that you asked it three times means the answer was missing where it's needed — it's now written in all three files.

The reason: these jobs run with `shell: alpine.sh`, and an Alpine shell cannot execute a composite action's `shell: bash` step. So the two auth lines are duplicated deliberately. The comment says so explicitly, including the warning that "cleaning up" the duplication would break all three.

.github/workflows/deploy-testing-xtm-one.yml
sandyghs: Same here:
The reason: these jobs run with `shell: alpine.sh`, and an Alpine shell cannot execute a composite action's `shell: bash` step. So the two auth lines are duplicated deliberately. The comment says so explicitly, including the warning that "cleaning up" the duplication would break all three.

.github/workflows/test-feature-branch.yml
sandyghs: Same here:
The reason: these jobs run with `shell: alpine.sh`, and an Alpine shell cannot execute a composite action's `shell: bash` step. So the two auth lines are duplicated deliberately. The comment says so explicitly, including the warning that "cleaning up" the duplication would break all three.

openaev-front/src/admin/components/nav/LeftBar.tsx
sandyghs: You're right, and I've followed you on this. The density that helped document the migration process is not the density you want when reading the code afterwards.

Cleanup pass done across the adapter: 153 comment lines down to 85, a 44% reduction. The rule I applied: a comment stays if it explains a constraint you cannot guess from the code, and goes if it paraphrases what the line already says. All the docblocks that restated the signature below them are gone.

Examples of what I kept, so you can judge the criterion: "no Tailwind build here, so geometry is inline — a class here silently stretched the logo" (an invisible constraint that actually caused a bug), and "the collapsed row is 48px but its content box is 46px, the library reserves 2px for the selection indicator" (an unguessable number).

openaev-front/src/admin/components/nav/LeftBarHeader.tsx
sandyghs: Agreed, same pass — the docblocks that restated the signature are removed. Only the two comments carrying non-obvious constraints remain (the cropped wordmark used as the collapsed emblem, and the optical offset a collapsed row needs).

.github/workflows/core-ci.yml
sandyghs: Good call, added. The trigger is now marked TEMPORARY with its date (2026-08-02) and a testable removal condition: once `git ls-remote --heads origin design-system/current` returns nothing, the trigger goes.

It's the same discipline we apply to the compensations in this PR — every temporary thing carries why it exists and what makes it removable.

openaev-front/src/components/common/menu/navbar/MadeByFiligran.tsx
sandyghs: I completely agree. Logged as a library candidate.

openaev-front/src/components/common/menu/navbar/nav-menu-model.ts
sandyghs: This one stays in the product. It's the shape of *this* product's menu — its routes, its permissions, its feature flags. And there's no generic contract to publish: the library's Navbar is composed through children, not driven by data, so it has no menu model to own.

openaev-front/src/components/common/menu/navbar/NavbarRowContent.tsx
sandyghs: This one stays in the product — and more than that, it should disappear rather than move up. It only exists because `asChild` makes the library's `icon`/`showIcon` props inert, so we have to rebuild the row body by hand. Moving it into the library would freeze the workaround.

The gap is filed with the library. The day `NavbarItem` accepts a destination directly, this file goes away.

openaev-front/src/components/common/menu/navbar/useNavbarState.ts
sandyghs: Product side. It's persistence and broadcasting, tied to this product's storage key and message bus, and consumed by its TopBar. The library's Navbar is already a controlled component — that's the right boundary: the library owns the widget, the product owns where the state lives. Moving this up would force the library to pick a storage key on behalf of every host.

Issue #7077, opened 2026-07-30
feat(output-types): add a new output processor type: ActionOutput, and check filter in conditions related to contains.


savacano28:
## Use case

Add a new output processor type: ActionOutput, and check filter in conditions related to contains.

PR #7155, opened 2026-08-03
fix: remove the isfindings filter to process chained outputs (#7077)

savacano28:
### Proposed changes

* Remove the isfindings filter to process chained outputs.

### Testing Instructions

1. Create an paylaod with action_output
2. User this payload in a chaining scenario
3. Add new action to be triggered when first action with action_output push structrued data into opeaev.
4. Check you should have all executions
image
image
image
image

### Related issues

* Related to #7077

### Checklist

### Further comments

### DIFF
--- a/openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
+++ b/openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
@@ -1361,8 +1361,8 @@ public class InjectExecutionStep implements ActionStep {
 
     if (inject.getPayload().isPresent()) {
       Set<OutputParser> outputParsers = structuredOutputUtils.extractOutputParsers(inject);
-      outputParsers.stream()
-          .flatMap(outputParser -> outputParser.getContractOutputElements().stream())
+      injectorContractContentUtils
+          .getAllContractOutputs(outputParsers, false)
           .forEach(
               out ->
                   typeMappings.put(
--- a/openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
+++ b/openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
@@ -1361,8 +1361,8 @@ public class InjectExecutionStep implements ActionStep {
 
     if (inject.getPayload().isPresent()) {
       Set<OutputParser> outputParsers = structuredOutputUtils.extractOutputParsers(inject);
-      injectorContractContentUtils
-          .getAllContractOutputs(outputParsers)
+      outputParsers.stream()
+          .flatMap(outputParser -> outputParser.getContractOutputElements().stream())
           .forEach(
               out ->
                   typeMappings.put(
--- a/openaev-api/src/main/java/io/openaev/utils/injector_contract/InjectorContractContentUtils.java
+++ b/openaev-api/src/main/java/io/openaev/utils/injector_contract/InjectorContractContentUtils.java
@@ -53,11 +53,22 @@ public class InjectorContractContentUtils {
    * @return list of contract output elements
    */
   public List<ContractOutputElement> getAllContractOutputs(Set<OutputParser> outputParsers) {
+    return getAllContractOutputs(outputParsers, true);
+  }
+
+  /**
+   * Retrieves contract output elements from output parsers, optionally keeping only finding
+   * outputs.
+   *
+   * @param outputParsers the set of output parsers to inspect
+   * @param onlyFindings whether to keep only outputs flagged as findings
+   * @return list of contract output elements
+   */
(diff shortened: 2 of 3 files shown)


Seb-MIGUEL (commented):
Quick review — looks good overall, small and well-targeted fix.

**What it does**
`buildTypeMappingsFromInject` (payload-based branch) was going through `InjectorContractContentUtils.getAllContractOutputs(Set<OutputParser>)`, which filters on `ContractOutputElement::isFinding` — a UI-only flag for computing findings, unrelated to chaining. Non-finding outputs (e.g. `action_output`, used purely to pass data to the next chained step) were silently dropped from the type mapping, so a chained action triggered on that output never got its input types resolved. Inlining the flatMap without the filter fixes that.

**Correctness**
- Verified `ActionOutput` is registered as `PRIMITIVE` / `PrimitiveType.ActionOutput` in `ChainingOutputType`, so the new test's assertions are correct and the fix produces the expected mapping.
- Nice catch on the asymmetry: the `else if` branch just below (contract-based inject, `getAllContractOutputs(InjectorContract)`) was never filtered by `isFinding` in the first place — this change makes the payload-based branch consistent with it, which is the right call given the doc comment on the shared helper explicitly says the filter is "related to flag in the UI to compute findings," not chaining eligibility. Chainability itself is still gated correctly downstream via `ChainingTypeRegistry` (`NOT_CHAINABLE` kind), so removing the filter here doesn't risk exposing non-chainable fields.
- `AgentExecutionProcessingHandler` still uses the filtered `getAllContractOutputs(Set<OutputParser>)` for findings dispatch — that's a different concern (real findings extraction) and correctly untouched by this PR.

**Test coverage**
Good regression test (`given_payloadOutputMarkedAsNonFinding_should_stillBuildTypeMapping`) that pins down exactly the bug being fixed.

**Nits (non-blocking)**
- The inlined `outputParsers.stream().flatMap(...).forEach(...)` duplicates the flatMap logic that already lives in `InjectorContractContentUtils.getAllContractOutputs(Set<OutputParser>)` (minus the filter). Not a big deal — there's precedent for this inline pattern elsewhere (`InjectsFileExport`) — but an alternative worth considering would be a boolean/overload param on the shared helper (e.g. `getAllContractOutputs(outputParsers, onlyFindings)`) so both call sites stay in sync going forward instead of two independently-maintained versions of the same traversal.
- Checklist says "I wrote test cases for the relevant use case" is unchecked, but the diff does add one — might just want to tick that box.

No blocking issues from me. 🚀

Seb-MIGUEL (commented):
Re-reviewed from a staff/security angle after the updates — nice work addressing the earlier feedback.

**Changes since last pass**
- The inlined flatMap was replaced with a proper `getAllContractOutputs(outputParsers, onlyFindings)` overload on `InjectorContractContentUtils`, with the old 1-arg method delegating to it with `true` — exactly preserves the previous (filtered) behavior for every other caller (`AgentExecutionProcessingHandler` still gets findings-only, unchanged). No regression there.
- Test now explicitly sets `outputElement.setFinding(false)` per Copilot's non-blocking suggestion, instead of relying on the implicit default — good, makes the test's intent unambiguous.

**Security assessment**
- No new authz/tenant-scoping concern: this is pure in-memory type-mapping logic over data already scoped to the current inject/tenant; no new DB access, no new deserialization, no user-controlled input parsed here.
- The one thing worth being deliberate about (not a blocker, more a "know what we're widening"): setting `onlyFindings=false` here means any payload output field with `isFinding=false` — not just `action_output` — now becomes chainable into the next step, wherever the payload branch is used. That said, this isn't a *new* capability introduced by this PR: the contract-based branch just below (`getAllContractOutputs(InjectorContract)`) was never filtered by `isFinding` in the first place, so isFinding=false fields were already chainable through that path. This PR makes the two paths consistent rather than opening a new one.
- Adjacent systemic note for awareness (out of scope for this PR, not asking for changes here): once a field is chainable, its resolved value is substituted into the next inject's command template via plain `String.replace("#{key}", value)` in `ExecutableInjectService.replaceArgumentsByValue` — no escaping. That's pre-existing behavior for every already-chainable finding type (credentials, hostnames, etc.), so this PR doesn't introduce it, it just means a couple more output types can now feed that same path. Worth a separate look at some point (e.g. if a scanned target can influence action_output content, does that content ever get used unescaped in a *shell* command downstream), but not something this PR needs to solve.

**Verdict**: no blockers from a security or staff standpoint. Clean, minimal, consistent fix with a real regression test. LGTM from me — leaving the formal approve to a maintainer.

Seb-MIGUEL (approved):
Approving after the staff/security re-review above — no blockers, feedback from the first pass was addressed cleanly (onlyFindings overload + explicit test intent). Nice fix.

Issue #7109, opened 2026-07-31
feat(platform): implement safe-mode configuration and visual banner


antoinemzs:
## Use case
Introduce the "safe-mode", meaning no background jobs are running, and a visual signal is up on the frontend.

This is a run mode that allows the platform to be partially available while there are potential performance issues or resources over-consumption.

This allows to send a signal to customers, of which the platform is experiencing issues, that the filigran team is proactive in dealing with them.

## Current workaround
N/A

## Proposed solution
1. Configuration: add `openaev.run-mode` configuration with two options: `normal` (default), `safe`
2. When the configuration is set to `safe`, no background jobs run (globally disable quartz jobs)
3. When the configuration is set to `safe`, display a banner on the frontend (all pages) signifying that the background processing is disabled, with a link to a documentation page.
4. Docs: add documentation page on the run modes, describing `normal` as "all features are enabled and running" and `safe` as "background processing is disabled" with a list of impacted features (based on the disables jobs in the quartz engine)

PR #7159, opened 2026-08-03
fix(engine-sync): implement stagger with proper quartz engine scheduler (#7112)

### LINKED ISSUE #7112: fix(engine): engine sync concurrency cap permanently starves models sharing the same trigger phase
## Summary

Since 3.260731.0 (which introduced the `engine.indexing-max-concurrent-models` concurrency cap, default 3), some platforms only sync a fixed subset of exactly 3 search-engine models after a restart. All other models are never indexed for the whole lifetime of the JVM: their cursors in `indexing_status` never advance, their indices stay stale (or empty after a forced reindex), and nothing is written to the logs at default level.

Observed on `main-ff` after the expectation reindex migration: `openaev_expectation-inject-000001` was recreated empty, the cursor stayed at epoch, and only `asset`, `security-platform` and `attack-pattern` ever logged "Executing engine sync for model" (verified over 6 hours of logs). A thread dump showed all Quartz workers idle, ruling out stuck executions.

## Root cause

Three defects compound in `EngineSyncExecutionJob` / `SelfConfiguredPlatformJob`:

1. **All model triggers share the same phase.** `getTrigger()` builds every trigger with the same 15-second simple schedule and an implicit start time of "now", during the same registration loop. All 13 triggers therefore fire at the exact same instant, every 15 seconds, forever.
2. **Quartz fires simultaneous triggers in a deterministic order.** Ties on `nextFireTime` and priority are broken by the trigger key, which is a random name generated once at boot. Combined with (1), the first `indexing-max-concurrent-models` models in that fixed order win the semaphore permits on every pass, and the remaining models are skipped on every pass - permanently. A restart re-rolls the random trigger keys, which is why restarting "fixed" some platforms and not others.
3. **Starvation is invisible.** The skip path logs at `debug` only, so a model that never wins a permit produces no log output at all at default level.

Additionally, trigger registration runs in the bean constructor while the Spring context is still initializing, and derives the model list from a live `getBeansOfType(Handler.class)` lookup - correct completeness of that lookup is only guaranteed once the whole context is up.

## Fix

- Stagger each model trigger's start time by `index % 15` seconds so every model owns its own phase within the interval; simultaneous contention on the semaphore no longer occurs in steady state.
- Track consecutive skipped passes per model and log a `warn` every 20 consecutive skips (5 minutes), so starvation can never be silent again.
- Defer trigger registration to `SmartInitializingSingleton.afterSingletonsInstantiated()` and log the registered job list at startup, making registration deterministic and auditable.

## Impact

Any platform on 3.260731.0 can, after any restart, permanently stop indexing an arbitrary subset of models (10 of 13 in the observed case), including `expectation-inject` right after the forced reindex migration - leaving dashboards empty. Workaround until this fix: restart the API pod until all 13 models appear in "Executing engine sync for model" logs.

antoinemzs:
### Proposed changes

* Remove the custom semaphore-based concurrency cap and rely on a dedicated Quartz scheduler (bounded daemon thread pool) to enforce `engine.indexing-max-concurrent-models` for the engine sync jobs.
* Stagger each model's trigger by one second (plus a sub-second random jitter) and give each trigger a distinct priority, so thread-capped fires cannot deterministically starve the same models (#7112).
* Use the next-with-existing-count misfire policy with a configurable threshold: a temporarily starved model runs as soon as a thread frees up instead of being skipped, without stacking repeated jobs for the same model.
* New configuration knob `engine.indexing-misfire-threshold-ms` / `ENGINE_INDEXING_MISFIRE_THRESHOLD_MS` (default 120000), documented in `configuration.md` and `application.properties`; `engine.indexing-max-concurrent-models` is clamped to >= 1 in `EngineConfig`.
* Review fixes: milliseconds naming aligned with `connect-timeout-ms` / `CONNECT_TIMEOUT_MS`, registration logging restored in `SelfConfiguredPlatformJob`, scheduler instance name is now caller-provided in `CustomSchedulerFactoryFactory` (Quartz registers schedulers JVM-wide by name), and unit tests cover per-model registration, stagger phases, cap clamping and failure propagation.

### Testing Instructions

1. Engine sync fires correctly, all models eventually synced.
2. `EngineSyncExecutionJobTest` covers registration (one staggered job per model, distinct priorities, scheduler start), the concurrency-cap clamp, failure propagation and the all-tenants transaction scope.

### Related issues

* Closes #7112


openaev-api/src/main/java/io/openaev/scheduler/jobs/engine_sync/EngineSyncExecutionJob.java
```diff
@@ -150,33 +95,28 @@ protected List<JobDetail> getJobDetails() {
...
-    SimpleScheduleBuilder _15_seconds =
-        simpleSchedule().withIntervalInSeconds(SYNC_INTERVAL_SECONDS).repeatForever();
+  protected Trigger getTrigger() {
+    SimpleScheduleBuilder _15_seconds = simpleSchedule().withIntervalInSeconds(15).repeatForever();
 
     return newTrigger()
-        .startAt(Date.from(Instant.now().plusSeconds(jobIndex % SYNC_INTERVAL_SECONDS)))
-        .withSchedule(_15_seconds.withMisfireHandlingInstructionNextWithRemainingCount())
+        .withSchedule(_15_seconds.withMisfireHandlingInstructionIgnoreMisfires())
         .build();
```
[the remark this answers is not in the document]
antoinemzs: Disagree: the jobs that would misfire are stacked for immediate execution as soon as a thread frees up. All models have a guaranteed processing time, best effort on a 15s basis.

openaev-api/src/main/java/io/openaev/scheduler/SelfConfiguredPlatformJob.java
```diff
@@ -1,66 +1,32 @@
...
-   * stagger the start time of each trigger: triggers built with the same schedule and the same
-   * start instant all fire at the exact same moment forever, which starves models behind a
-   * concurrency cap (see {@code EngineSyncExecutionJob}).
-   *
-   * @param jobIndex position of the job definition in the {@link #getJobDetails()} list
-   */
-  protected abstract Trigger getTrigger(int jobIndex);
+  @PostConstruct
+  void register() throws SchedulerException {
+    this.registerTriggers();
```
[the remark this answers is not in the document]
antoinemzs: the relevant classes are in the classpath at boot time, so there is no chance for missing a model class.

docs/docs/deployment/configuration.md
```diff
@@ -158,9 +158,10 @@ Each OpenCTI connection is scoped to an OpenAEV tenant, identified by its UUID (
...
+|:-----------------------------------------|:-----------------------------------------|:--------------|:-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| engine.indexing-grace-window-seconds     | ENGINE_INDEXING_GRACE_WINDOW_SECONDS     | 60            | Indexing cursor grace window in seconds. The cursor persisted after each indexing round never gets closer to wall-clock than this window, so rows committed late by long write transactions are still indexed. Must exceed the longest expected write transaction. |
+| engine.indexing-misfire-threshold-millis | ENGINE_INDEXING_MISFIRE_THRESHOLD_MILLIS | 120000        | Threshold in milliseconds after which a model synchronisation is declared having misfired and should be rescheduled..                                                                                                                                              |
```
savacano28: * rescheduled.
* This property can be configurable? Should we add indexing-misfire-threshold-millis into properties file?
* Try to keep same naming already in place for ms? => indexing-misfire-threshold-milliseconds

openaev-model/src/main/java/io/openaev/config/EngineConfig.java
```diff
@@ -93,6 +93,13 @@ public static class Defaults {
      */
     public static final int INDEXING_MAX_CONCURRENT_MODELS = 3;
 
+    /**
+     * Default threshold in milliseconds after which a model synchronisation is declared having
+     * misfired and should be rescheduled. If the engine sync threadpool is temporarily exhausted,
+     * this will avoid stacking repeated jobs for syncing the same model over and over.
+     */
+    public static final long INDEXING_MISFIRE_THRESHOLD_MILLIS = 120000L;
```
savacano28: THRESHOLD_MS as CONNECT_TIMEOUT_MS

openaev-api/src/main/java/io/openaev/scheduler/SelfConfiguredPlatformJob.java
```diff
@@ -1,66 +1,33 @@
...
+
   private void registerTriggers() throws SchedulerException {
-    List<JobDetail> jobDetails = getJobDetails();
-    for (int i = 0; i < jobDetails.size(); i++) {
-      scheduler.scheduleJob(jobDetails.get(i), getTrigger(i));
+    int delaySeconds = 0;
+    for (JobDetail jd : getJobDetails()) {
+      scheduler.scheduleJob(jd, getTrigger(delaySeconds++));
     }
-    log.info(
```
savacano28, line 60: Could we keep the logs?

savacano28 (dismissed):

docs/docs/deployment/configuration.md
SamuelHassine: Done in the latest push:

- Fixed the double period ("rescheduled..").
- The property was already configurable through the EngineConfig binding; it is now also listed as a commented entry in application.properties next to engine.indexing-max-concurrent-models, so it is discoverable like the other engine tuning knobs.
- Renamed to match the naming already in place for milliseconds (engine.connect-timeout-ms, engine.socket-timeout-ms): the property is now engine.indexing-misfire-threshold-ms / ENGINE_INDEXING_MISFIRE_THRESHOLD_MS, backed by EngineConfig#indexingMisfireThresholdMs.

openaev-model/src/main/java/io/openaev/config/EngineConfig.java
SamuelHassine: Renamed in the latest push: the constant is now INDEXING_MISFIRE_THRESHOLD_MS (same convention as CONNECT_TIMEOUT_MS / SOCKET_TIMEOUT_MS), with the field renamed to indexingMisfireThresholdMs and the property to engine.indexing-misfire-threshold-ms.

openaev-api/src/main/java/io/openaev/scheduler/SelfConfiguredPlatformJob.java
SamuelHassine: Restored in the latest push: SelfConfiguredPlatformJob#registerTriggers logs the number of registered jobs and their keys again ("Registered {} job(s) for {}: {}"), so startup registration stays auditable.

SamuelHassine (approved):
Full review and fix pass completed. The branch was rebased onto latest main (safe-mode gate from #7109 preserved), all review comments from @savacano28 are implemented (ms naming convention, configurable misfire threshold listed in application.properties, registration logs restored), the scheduler factory now takes a caller-provided unique instance name, and unit tests were restored for per-model registration, stagger phases, cap clamping and failure propagation. All 47 checks are green and the branch is a single signed commit.

SamuelHassine:
## Review-and-fix summary

**Rebase**: squash-rebased onto latest `main` (the branch was 8 commits and a merge behind). No supersession: the safe-mode `@ConditionalOnProperty(openaev.run-mode)` gate added to `EngineSyncExecutionJob` by #7109 is preserved. The branch is now a single GPG-signed commit (`b7743dd`) with `Co-authored-by: Antoine MAZEAS` preserving attribution.

**Review comments addressed** (all 3 open threads replied and resolved):

- Misfire threshold renamed to the ms convention already in place (`CONNECT_TIMEOUT_MS` style): `INDEXING_MISFIRE_THRESHOLD_MS`, `indexingMisfireThresholdMs`, `engine.indexing-misfire-threshold-ms` / `ENGINE_INDEXING_MISFIRE_THRESHOLD_MS`; docs table fixed (double period) and a commented entry added to `application.properties` next to the other engine tuning knobs.
- Registration logging restored in `SelfConfiguredPlatformJob#registerTriggers` ("Registered {} job(s) for {}: {}").

**Independent review fixes**:

- `CustomSchedulerFactoryFactory`: the scheduler instance name is now a required parameter instead of a hardcoded `OpenAEV_EngineSync_Scheduler` (Quartz registers schedulers JVM-wide by name; a second caller reusing the name would silently get the first scheduler and its thread pool). Unused `get()` / `get(int)` overloads removed.
- Tests: permit-era names/assertions updated (`Releases the permit...` -> `Propagates a sync failure without preventing the next pass`), the clamp test now asserts the factory receives `1` for a non-positive cap, and the registration/stagger tests deleted by this PR were restored (one staggered job per model, distinct trigger priorities, scheduler start, no double-start). `EngineSyncExecutionJobTest`: 6/6 green locally.

**CI**: all 47 checks green on the new head; PR is `MERGEABLE` / `CLEAN`.

**Note for maintainers**: @savacano28's change request predates the fix push; a re-review or dismissal of the stale review is the only remaining gate.

Issue #6926, opened 2026-07-24
feat(chaining): allow linking inputs to many primitive types


savacano28:
## Use case

Ch 3: In action creation we can link the input arguments to various primitives types.

PR #7160, opened 2026-08-03
fix(chaining): resolve dropped/duplicated combinations when mappers share a primitive type (#6926)

mergenhan:
## Problem

When linking two or more action input fields to overlapping primitive types (e.g. two `text`-typed inputs both fed by the same scope variable pool), the chaining engine did not generate all expected combinations. Some combinations were silently dropped while others were re-executed as spurious duplicates.

Example: two inputs both linked to `{key, pass}` should produce 4 combinations (`key/key`, `key/pass`, `pass/key`, `pass/pass`), but only 3 were executed, with one combination duplicated and the two "swapped" cross-combinations (`key/pass` / `pass/key`) colliding into one.

## Root cause

`ConditionService`'s combination hashing/dedup logic (`buildExecutionBatches` -> `tryAddBatch`) collapsed each combination into a single `Map<String,String>` keyed by **primitive type name**, not by mapper/field identity. When two mapper fields share a source type, "swapped" combinations produce an identical collapsed map and therefore an identical hash - one gets deduplicated as if it were the other.

A second, related issue existed downstream in `InjectExecutionStep.applyMapping`, which re-derived each mapper's value from that same shared, type-keyed input map when building the final inject content - so even once a combination correctly ran, its content could still resolve to the wrong (shared) value for two same-typed fields.

## Fix

- `ConditionService`: hashing and per-mapper value resolution now use a new **position-based identity** (`comboIdentity`), keyed by each mapper's index within the ordered mapper list, instead of by type name. This can never collide across different mapper positions, regardless of shared types. The existing type-keyed map (`toComboMap`) is kept only for building `ExecutionBatch#inputString()`, so the JSON format/consumers relying on it (e.g. filter conditions) are unaffected.
- `InjectExecutionStep#applyMapping`: now prefers each mapper's own already-resolved value before falling back to the shared type-keyed lookup, so inject content generation can't reintroduce the same collapsing bug downstream.

Both changes are minimal, contained to the two files below, and preserve all existing public method signatures/JSON formats.

## Tests

- Added `given_twoMappersSharingSameSourceType_should_generateAllCombinationsWithoutDuplicates`: asserts 2 mappers sharing a type pool produce all 4 distinct combinations with unique hashes.
- Added `given_alreadyExecutedSwappedCombo_should_stillGenerateItsDistinctMirrorCombo`: asserts a pre-executed cross-combination doesn't block its distinct mirror combination.
- Updated one existing test's hardcoded hash to the new position-based scheme (hash format intentionally changed, not a regression).
- Full `ConditionServiceTest` suite: 119/119 passing, no regressions.

Fixes OpenAEV-Platform/filigran-private#251

### DIFF
--- a/openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
+++ b/openaev-api/src/main/java/io/openaev/api/chaining/InjectExecutionStep.java
@@ -1117,25 +1117,30 @@ public class InjectExecutionStep implements ActionStep {
    * @param inputValues the source JSON containing the values to map
    */
   private void applyMapping(ObjectNode contentNode, Condition mapping, JsonNode inputValues) {
+    String targetJsonKey = mapping.getKey();
+
+    // Prefer the value already resolved for this specific mapper during chaining (see
+    // ConditionService#toResolvedMapper). This is required when several mappers share the same
+    // key type(s): looking the value up from `inputValues` by type alone can't tell them apart and
+    // would collapse them onto the same shared source value. Values only known once the batch is
+    // expanded per target (e.g. "_target"-derived fields) are not resolved at that stage, so they
+    // still fall through to the inputValues lookup below.
+    if (targetJsonKey != null && mapping.getValue() != null && !mapping.getValue().isBlank()) {
+      contentNode.set(targetJsonKey, TextNode.valueOf(mapping.getValue()));
+      return;
+    }
+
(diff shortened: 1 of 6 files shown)


savacano28:
Review:
* Code: OK
* Tests: OK
image
image
image
image
image
image
image
image

savacano28 (approved):

PR #7161, opened 2026-08-03
fix(attack-path): make expectation table scrollable, reconstruct injector command line, paginate findings/executions (#7161)

Seb-MIGUEL:
## What

- The security-platform expectation table (Prevention/Detection/Vulnerability) had a horizontal scroll container with no visible scrollbar, leaving the "Detection time"/"Alerts" columns unreachable on Chromium/WebKit at the drawer's default width. The scrollbar thumb now always renders.
- Each alert's `link` field (already resolved by `AttackPathSecurityPlatformResolver`, but dropped in the frontend mapping) is now threaded through to the alerts popover, rendered as a real clickable link.
- The Terminal view had no visibility into what a network injector (NetExec, Nmap…) actually sent — it has no `command` on the DTO (only resolved for Command-payload-backed injects). It's now reconstructed from the injector's own execution trace (always redacted with `***`) by un-redacting the recognized flags using the inject's resolved field values, applying our own partial-reveal masking (password/hash) instead of a blanket `***`.
- The endpoint/injector panel's Findings groups and Executions list now have numbered pagination with a page-size selector (5/10/25/50, default 10), replacing the unbounded/fixed-5-per-page display.

## Test plan

- [ ] Open an execution's Terminal view for a network-injector action (e.g. NetExec SMB) and confirm the "Command" block shows the reconstructed line with the username visible and the password partially masked.
- [ ] Open an execution's Result tab, confirm the platform table's scrollbar is visible and the Alerts column is reachable; click an alert count and confirm a link opens.
- [ ] Open an endpoint/injector panel with many findings/executions and confirm pagination + page-size selector work.

### DIFF
--- a/openaev-api/src/main/java/io/openaev/service/attackpath/AttackPathGraphService.java
+++ b/openaev-api/src/main/java/io/openaev/service/attackpath/AttackPathGraphService.java
@@ -342,7 +342,9 @@ public class AttackPathGraphService {
         maskSecrets(e.getTerminalOutput(), secrets),
         // A network injector (NetExec, Nmap…) never has its own `command` snapshot; reconstruct one
         // server-side instead of leaving the client to fetch the raw inject content itself.
-        (injectId != null && e.getPayloadId() == null && (e.getCommand() == null || e.getCommand().isBlank()))
+        (injectId != null
+                && e.getPayloadId() == null
+                && (e.getCommand() == null || e.getCommand().isBlank()))
             ? maskSecrets(injectorCommandLine(injectId), secrets)
             : null);
   }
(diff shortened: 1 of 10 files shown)


guillaumejparis:
I see that you are doing masking in the frontend, we need to do it in backend otherwise it is kind of useless because you can easily retrieve the values from the request

Seb-MIGUEL:
Good catch, you're right — thanks for flagging it.

Fixed in 4fc5ef443: the reconstruction and partial masking now happen entirely server-side in `AttackPathGraphService#injectorCommandLine`, the same way the existing `command`/`terminalOutput` snapshot fields are already masked (reusing `PrimitiveValueMaskingUtils`). The new `injectorCommandLine` field on `AttackPathExecutionDetailDTO` carries the already-masked string — the raw password/hash is never sent to the client at all now.

The frontend no longer fetches the inject's raw content (`getInjectSimple` is gone) or does any masking itself; it just renders `detail.injectorCommandLine` as-is.

guillaumejparis (approved):

PR #7186, opened 2026-08-05
fix(injector-contract): dedupe attack patterns/tags/domains on the search endpoint (#7189)

### LINKED ISSUE #7189: Injector contracts search returns duplicate attack patterns/tags/domains
## Bug

The "Attack patterns" column on the injector contracts list shows the same MITRE ATT&CK technique listed twice (e.g. \`[T1018] Remote System Discovery, [T1018] Remote System Discovery\`), for contracts that legitimately have only one mapped technique.

## Root cause

\`InjectorContractService.buildCommonInjectorContractContext()\` aggregates \`domains\`, \`attackPatterns\`, and \`tags\` via three independent LEFT JOINs under a single \`GROUP BY\` that covers none of them. When a contract has more than one row in one of those collections (e.g. several tags), the cartesian product duplicates every value from the others.

Confirmed the underlying data is correct: \`injectors_contracts_attack_patterns\` has exactly one row per real mapping — only the join's cartesian-product expansion at query time was wrong.

## Fix

See linked PR.

Seb-MIGUEL:
## Summary
`buildCommonInjectorContractContext()` aggregates domains, attack patterns, and tags via three independent LEFT JOINs under a single `GROUP BY` that covers none of them. When a contract has more than one row in one of those collections (e.g. several tags), the cartesian product duplicates every value from the others — visible in the UI as the same MITRE ATT&CK technique listed twice in the "Attack patterns" column.

The underlying data was never wrong (confirmed directly: `injectors_contracts_attack_patterns` has exactly one row per real mapping); only the join's cartesian-product expansion was.

**Backend**: dedupe the three aggregated arrays in the two mappers that read them (`mapFull`, `mapThreatArsenal`) rather than fighting Hibernate's Criteria API for a `DISTINCT` inside `array_agg` (no such option exists in this Hibernate version's `arrayAgg` overloads — only `countDistinct` is special-cased).

**Frontend**: `InjectorContracts.jsx` rendered `injector_contract_attack_patterns` directly with no dedup, unlike the sibling kill-chain-phases block just above it which already wraps its derived list in a `Set`. Added the same `Set`-based dedup as a second line of defense.

Verified against a rebuilt image: scanned 5000 real contracts via the search endpoint (the one the frontend list page uses) — zero duplicates in attack_patterns/tags/domains, including contracts with several tags that previously exhibited the bug.

Fixes #7189

## Test plan
- [x] `mvn install -DskipTests -pl openaev-api -am` — compiles clean
- [x] New regression test (`searchDoesNotDuplicateAttackPatternsWhenContractHasSeveralTags`) reproducing the exact trigger condition (one attack pattern + two tags on the same contract) — passes in CI
- [x] End-to-end: scanned 5000 real contracts post-fix, zero duplicates
- [x] Full CI suite green (API Tests, E2E chrome/webkit, Spotless, etc.)

### DIFF
--- a/openaev-api/src/main/java/io/openaev/rest/injector_contract/InjectorContractService.java
+++ b/openaev-api/src/main/java/io/openaev/rest/injector_contract/InjectorContractService.java
@@ -746,12 +746,32 @@ public class InjectorContractService implements DependenciesManager {
         tuple.get("payload_type", String.class),
         tuple.get("collector_type", String.class),
         tuple.get("injector_contract_injector_type", String.class),
-        tuple.get("injector_contract_attack_patterns", String[].class),
-        tuple.get("injector_contract_tags", String[].class),
-        tuple.get("injector_contract_domains", String[].class),
+        dedupePreservingOrder(tuple.get("injector_contract_attack_patterns", String[].class)),
+        dedupePreservingOrder(tuple.get("injector_contract_tags", String[].class)),
+        dedupePreservingOrder(tuple.get("injector_contract_domains", String[].class)),
         tuple.get("injector_contract_updated_at", Instant.class),
         tuple.get("payload_execution_arch", Payload.PAYLOAD_EXECUTION_ARCH.class),
         injectorNames);
+  }
+
+  /**
+   * Deduplicates an array of IDs while preserving first-seen order.
+   *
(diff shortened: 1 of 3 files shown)


savacano28 (approved):
Code: OK
Tests:
* Threat arsenal page: OK
* Attack path: Injector contract: OK

PR #7192, opened 2026-08-05
fix(attack-path): truncate a finding panel's list items to one line (#7204)

### LINKED ISSUE #7204: fix(attack-path): long finding values render as a wall of text in the category panel
## Bug

In the attack path view, opening a finding category panel (the contextual side panel listing the findings of one category) with a long value renders that value as a wall of wrapped text spanning dozens of lines. A captured nmap XML report surfaced as an "Action output" finding, for example, fills the whole panel with one item.

## Expected behavior

Each list item shows its value on a single ellipsised line, the same way the graph's own finding node labels already render the same value. The full (masked) value stays available via the item's title tooltip and the detail panel.

## Component

`openaev-front/src/admin/components/simulations/simulation/attack_path/CategoryFindingsPanel.tsx` renders the value with `wordBreak: 'break-all'` instead of `noWrap`.

Seb-MIGUEL:
## Summary

- A long output value (e.g. a captured nmap XML report shown as an "Action output" finding) rendered as a wall of wrapped text spanning dozens of lines in the finding category panel's item list, instead of a single-line, ellipsised label like the graph's own finding nodes already show for the same kind of value.
- Switched the item's `Typography` to `noWrap` (single line + ellipsis) instead of `wordBreak: 'break-all'`. The full (masked) value is unchanged - still available via the existing `title` tooltip and the detail panel.

## Test plan

- [x] `eslint` clean on the changed file.
- [x] `yarn check-ts` clean on the full frontend.

Closes #7204

### DIFF
--- a/openaev-front/src/admin/components/simulations/simulation/attack_path/CategoryFindingsPanel.tsx
+++ b/openaev-front/src/admin/components/simulations/simulation/attack_path/CategoryFindingsPanel.tsx
@@ -186,9 +186,8 @@ const CategoryFindingsPanel = ({
                     flex: 1,
                   }}
                   >
-                    {/* A single line, ellipsised (never the multi-line wall of text a long output value —
-                        e.g. a captured nmap XML report — would otherwise render as, same as the graph's
-                        own node label): the full value stays available in the title tooltip below. */}
+                    {/* One ellipsised line even for huge values (e.g. a raw action output); the
+                        title tooltip keeps the full (masked) value, like the graph's node labels. */}
                     <Typography variant="body2" noWrap title={maskedValue}>{maskedValue}</Typography>
                     {endpointName && (
                       <Typography variant="caption" color="text.secondary" noWrap title={endpointName} sx={{ display: 'block' }}>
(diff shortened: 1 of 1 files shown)


SamuelHassine (approved):
Reviewed the full component, not just the diff. The `noWrap` switch is correct: the value `Typography` sits in a `minWidth: 0` flex parent so the ellipsis actually engages, it matches the sibling endpoint caption and the graph node labels, and the `title` tooltip only ever receives the already-masked value (no secret leak for credentials/sid/password_policy findings). `eslint` and `yarn check-ts` are clean, all quality-gates checks and every codecov status are green. LGTM.

SamuelHassine:
## Review summary

- Independent review of the full `CategoryFindingsPanel.tsx`: the `noWrap` change is correct (the value's flex parent has `minWidth: 0`, so the ellipsis engages), consistent with the sibling endpoint caption and the graph's node labels, and the tooltip keeps only the masked value so nothing sensitive is exposed.
- Follow-up commit 5c5ab05: condensed the added JSX comment (it narrated the change, used non-ASCII dashes and wrongly said the tooltip sits "below" the value).
- Issue link fixed: the PR previously referenced #7189, which is a different bug (injector contracts duplicate attack patterns). Created #7204 for this UI bug and updated the title and body (`Closes #7204`).
- Checks: `eslint` and `yarn check-ts` clean locally; all quality-gates jobs, CodeQL, signed-commits and every codecov patch/project status pass on the final head. The earlier codecov frontend project failure (9% vs 80% target) was from a stale partial upload and cleared once the full suite ran with carryforward flags.
- No test added on purpose: this is a one-property presentational change, the repo has no attack-path component tests, and asserting an MUI class would be a throwaway assertion.
- No unresolved review threads; Copilot reviews had no actionable comments.

PR #7206, opened 2026-08-05
fix(attack-path): only spend the live camera budget on real growth (#7205)

### LINKED ISSUE #7205: fix(attack-path): live graph stops re-framing during a run — the initial load consumes the whole camera budget
## Description

During a running simulation, the Attack path graph used to re-frame itself (fit-to-view) as new nodes were drawn. Since #7099 it never re-frames on live growth: the run draws itself progressively off-screen and the user has to pan/zoom by hand (or hit the fit control) to see what arrived.

## Root cause

#7099 replaced the ReactFlow `fitView` camera with the hand-rolled `canvas/AttackPathCanvas.tsx`, and introduced a framing budget `AP_MAX_LIVE_FITS = 2` (`SimulationAttackPath.tsx:61`) whose documented intent is:

> During a live run, the **first structural deltas** re-frame the whole graph at most this many times ("unzoom 1-2 times max"); after that the camera switches to pursuit and follows the newest nodes.

The budget is spent by whatever bumps `structuralNonce` — but the **initial load** bumps it twice before any live delta arrives:

| bump | source | effect |
|---|---|---|
| 1 | `useAttackPathLiveGraph.ts:181` — collapsed snapshot seed | `liveFitCount = 1` |
| 2 | `useAttackPathLiveGraph.ts:225` — causal/full graph merged in | `liveFitCount = 2` → `setPursuitActive(true)` |
| 3+ | the first (and every) real live delta | pursuit already active → no re-fit, ever |

`AttackPathCanvas.tsx:240` (`if (firstFit || !pursuitActiveRef.current)`) then suppresses every growth re-fit for the rest of the run. So live growth gets **zero** re-fits instead of the intended 1-2.

Note both initial bumps happen on a normal small run: `fullEligible` is true whenever the run is absent from the summary list or under the size ceiling (`SimulationAttackPath.tsx:220-228`).

## Fix

Only spend the framing budget on **live** growth — a bump that actually carried new node ids. The seed and the full-graph merge carry no batch, so they still frame the graph on load but no longer graduate the camera to pursuit before the run has produced anything.

## Reproduction

Covered by a regression test on the existing `SimulationAttackPath` harness (it captures the `fitRequest` prop handed to the canvas): after the initial load settles, a live delta introducing an endpoint must bump `fitRequest`. Before the fix it stays put (`expected 2 to be greater than 2`).

Seb-MIGUEL:
## Summary

During a running simulation the Attack path graph stopped re-framing itself as new nodes were drawn — the run drew itself progressively off-screen and the user had to pan/zoom by hand.

`AP_MAX_LIVE_FITS = 2` (`SimulationAttackPath.tsx:61`) is documented as a budget for *"the first structural **deltas**"*, but it was spent by anything that bumped `structuralNonce` — and the **initial load** bumps it twice before any delta arrives:

| bump | source | effect |
|---|---|---|
| 1 | `useAttackPathLiveGraph.ts:181` — collapsed snapshot seed | `liveFitCount = 1` |
| 2 | `useAttackPathLiveGraph.ts:225` — causal/full graph merged in | `liveFitCount = 2` → `setPursuitActive(true)` |
| 3+ | the first (and every) real live delta | pursuit already active → no re-fit, ever |

`AttackPathCanvas.tsx:240` (`firstFit || !pursuitActiveRef.current`) then suppresses every growth re-fit for the rest of the run, so live growth got **zero** re-fits instead of the intended 1-2. Both initial bumps happen on any normal small run (`fullEligible` is true when the run is absent from the summary list or under the size ceiling, `SimulationAttackPath.tsx:220-228`).

**Fix:** only spend the budget on bumps that actually carried new node ids. The seed and the full merge carry no batch — they still frame the graph on load, but no longer graduate the camera to pursuit before the run has produced anything.

Fixes #7205

## Test plan

- [x] Regression test added on the existing `SimulationAttackPath` harness (it captures the `fitRequest` prop handed to the canvas). Verified it **fails without the fix** (`AssertionError: expected 2 to be greater than 2`) and passes with it.
- [x] The same test also pins the intended hand-over, so the fix restores re-framing without disabling pursuit: delta 1 re-frames, delta 2 re-frames, delta 3 does not (pursuit owns the camera).
- [x] `yarn vitest run src/__tests__/.../attack_path/` — 100 tests / 4 files pass.
- [x] `yarn eslint` on both changed files, and `yarn check-ts`, both clean.

## Not covered by this PR

The other half of the reported symptom — the live view only catching up after ~45 s instead of ~3 s — is a **separate** issue and is deliberately left out. Mechanism: `useAttackPathLiveGraph.ts:339-345` picks the 45 s safety net whenever `isStreamHealthy()` is true, but that helper (`useDataLoader.js:364-369`) only measures the shared stream's 1 s ping — never whether `attack-path-version` nudges actually arrive. So any nudge-delivery failure silently degrades the cadence to 45 s while the LIVE badge stays green.

Changing that policy would contradict the existing, deliberate `given_aHealthyStream_should_pollOnlyAsASafetyNet` test (spec 003 FR6), and would mask the nudge failure rather than fix it — so it needs its own issue and a root-cause pass on why the nudges stop (candidates: `StreamApi.listenAttackPathVersion` is a `@TransactionalEventListener` without `fallbackExecution`, and the `streamExecutor` pool's `DiscardOldestPolicy`, which `AttackPathVersionService`'s own javadoc flags as a flooding risk).

### DIFF
--- a/openaev-front/src/__tests__/admin/components/simulations/simulation/attack_path/SimulationAttackPath.test.tsx
+++ b/openaev-front/src/__tests__/admin/components/simulations/simulation/attack_path/SimulationAttackPath.test.tsx
@@ -540,4 +540,55 @@ describe('SimulationAttackPath live delta updates', () => {
     expect(screen.getByText('Reconnecting…')).toBeTruthy();
     expect(screen.getByTestId('attack-path-flow')).toBeTruthy();
   });
+
+  it('re-frames the camera when a live delta grows the graph, then hands over to pursuit', async () => {
+    // One growth delta per tick, each introducing a distinct endpoint so every one is real growth.
+    const growthDelta = (version: number, host: string) => ({
+      data: {
+        sinceVersion: version - 1,
+        newVersion: version,
+        attackPathNodes: [{
+          id: `NODE_ENDPOINT|${host}`,
+          type: 'ASSET',
+          label: host,
+          hostname: host,
+          ref: host,
+          findingCounts: {},
+        }],
+        attackPathEdges: [{
+          edgeId: `edge-nmap-${host}`,
+          edgeSourceId: 'NODE_INJECTOR|nmap',
+          edgeTargetId: `NODE_ENDPOINT|${host}`,
+          type: 'EDGE_EXECUTIONS',
+          count: 1,
+        }],
+      },
+    });
+
+    setup();
+    // Both initial reads must settle: the collapsed seed AND the causal overlay merged in after it
+    // each bump the structural nonce, and neither is live growth — they are the initial framing.
+    await mounted();
+    await mounted();
+    const afterInitialLoad = mocks.flowProps.current?.fitRequest ?? 0;
+
+    // A live delta introduces an endpoint. It lands outside the framed viewport, so the camera must
+    // re-frame — otherwise the run draws itself off-screen until the user pans by hand.
+    mocks.fetchAttackPathGraphDelta.mockResolvedValue(growthDelta(11, 'host-a'));
+    await tick();
+    const afterFirstGrowth = mocks.flowProps.current?.fitRequest ?? 0;
+    expect(afterFirstGrowth).toBeGreaterThan(afterInitialLoad);
+
+    // The framing budget is AP_MAX_LIVE_FITS (2) LIVE deltas, so the second one still re-frames.
+    mocks.fetchAttackPathGraphDelta.mockResolvedValue(growthDelta(12, 'host-b'));
+    await tick();
+    const afterSecondGrowth = mocks.flowProps.current?.fitRequest ?? 0;
+    expect(afterSecondGrowth).toBeGreaterThan(afterFirstGrowth);
+
+    // Budget spent: the camera now CHASES the newest nodes at the current zoom instead of pulling
+    // back on every delta — so no further re-fit, and the new nodes are handed to pursuit instead.
+    mocks.fetchAttackPathGraphDelta.mockResolvedValue(growthDelta(13, 'host-c'));
+    await tick();
+    expect(mocks.flowProps.current?.fitRequest ?? 0).toBe(afterSecondGrowth);
+  });
 });
--- a/openaev-front/src/admin/components/simulations/simulation/attack_path/SimulationAttackPath.tsx
+++ b/openaev-front/src/admin/components/simulations/simulation/attack_path/SimulationAttackPath.tsx
@@ -380,11 +380,18 @@ const SimulationAttackPath = ({ scenarioExerciseIds, scenarioId, hideLaunchCta =
           nonce: structuralNonce,
         });
       } else {
-        liveFitCount.current += 1;
         setFitNonce(n => n + 1);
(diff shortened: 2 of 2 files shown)


SamuelHassine (approved):
Independent review done on the full files (SimulationAttackPath.tsx live-fit/pursuit state machine, useAttackPathLiveGraph.ts seed/merge/delta paths, AttackPathCanvas.tsx fit gate, and the delta store). The fix is correct and minimal: the two initial-load structural bumps (collapsed seed at useAttackPathLiveGraph.ts:181, full merge at :225) never set a batch, so gating the budget spend on newNodeIds.length > 0 is exactly right, and the else branch still bumps fitNonce so the mount-time framing is preserved. The regression test drives the real hook and component (only the fetch layer and canvas are mocked) and pins both halves: growth deltas 1 and 2 re-frame, delta 3 hands over to pursuit. Verified locally: eslint clean on both files, check-ts clean, attack_path vitest suite 100 tests / 4 files green. Approving.

SamuelHassine:
Full review-and-fix pass completed on this PR.

**Review scope**: full reads of SimulationAttackPath.tsx (AP_MAX_LIVE_FITS budget + pursuit state machine), useAttackPathLiveGraph.ts (seed at :181, full merge at :225, delta poll and batch lifecycle), AttackPathCanvas.tsx (the firstFit || !pursuitActiveRef gate at ~:240 and the fitRequest/pursuitRequest consumers), attack-path-delta-store.ts (what populates newNodeIds), and the test harness.

**Findings**: no must-fix or should-fix issues; no code changes were needed.

- Verified the seed and full-merge nonce bumps genuinely carry no batch, so the new newNodeIds.length > 0 gate spends the camera budget only on real growth, and the initial load still frames the graph (the else branch keeps bumping fitNonce unconditionally).
- Edge-only structural deltas (no new node ids) now re-frame without spending budget, which is correct: the graph bounds do not grow without new nodes, and pursuit would have nothing to chase anyway.
- One nit worth noting for the record, not fixed here to keep the diff minimal across the three concurrent attack-path PRs: the hook's batch is cleared only by its 500 ms TTL (BATCH_TTL_MS), never on simulation switch, resync, or full merge. A growth delta landing less than 500 ms before one of those load bumps can therefore overspend one budget unit (pursuit engages one delta early). The window is sub-second and the degradation is graceful (still strictly better than the pre-fix behaviour of zero re-fits), so it can ride along in a later cleanup if it ever shows up in practice.

**Docs**: the Docs Reviewer bot flagged that no documentation was modified - correct outcome for an internal camera-budget bugfix; the Documentation gap check passed.

**Checks**: eslint (both changed files), check-ts, and the attack_path vitest suite (100 tests / 4 files) all pass locally; CI is fully green including codecov patch/project and the required quality-gates / Quality Status and Artifact & Image Sanity Check. The commit is signed and verified, the title matches the repo convention, and the body links Fixes #7205. Scope note respected: the 45 s catch-up cadence issue stays out of this PR as stated in the description.

PR #7211, opened 2026-08-05
fix(chaining): keep a field's defined value when a type is also linked (#7210)

### LINKED ISSUE #7210: fix(chaining): defined value is dropped when an output type is linked on the same field
## Description

In the chaining **Configure action** drawer, a payload/contract argument field can be given a **defined value** (a literal the user types) and/or be **linked to an upstream output type** (bound to a primitive/complex type produced by a previous action). These are meant to coexist — the defined value should keep participating as one more candidate alongside the linked type's resolved pool.

Actual: they behave as either/or. When the user types a defined value and *then* links a type, the defined value is silently ignored.

## Root cause (frontend only)

- `ConfigureActionDetail.tsx` rendered the defined-value input only when there was **no** link, so linking a type hid it.
- `ComponentStepperDrawer.tsx` built the step `MAPPER` conditions inline and never set `condition_value`.
- A helper written for exactly this — `mapFieldLinksToStepConditions` (carries `inject_content[key]` into `condition_value`) — existed but was **dead code**.

The backend already supports both: `ConditionService#resolveMapperPairs` (~line 1230) adds the mapper's defined value as an extra candidate alongside the type-resolved pool, "never silently discarded just because the mapper is no longer MappingType.DEFAULT". So this is purely a UI/serialization bug.

## Fix

- Show the defined-value input unless the field is **auto-linked** (system-managed fields like targeted-asset still hide it), so a defined value stays editable alongside a manually linked type.
- Serialize conditions through `mapFieldLinksToStepConditions` so `condition_value` is sent; align its keyTypes fallback with the previous inline behaviour.

Seb-MIGUEL:
## Summary

In the chaining **Configure action** drawer, a payload/contract argument could be given a **defined value** OR bound to an upstream **output type**, but not both - once a type was linked the value was silently ignored.

- `ConfigureActionDetail.tsx` rendered the defined-value input only when there was no link, so linking a type hid it.
- `ComponentStepperDrawer.tsx` built the step `MAPPER` conditions inline and never set `condition_value`, so the typed value never reached the backend.
- A purpose-built helper `mapFieldLinksToStepConditions` (carries `inject_content[key]` into `condition_value`, comment: "instead of being dropped once a primitive type is linked") existed but was **dead code**.

The backend already supports both together - `ConditionService#resolveMapperPairs` adds the mapper's defined value as one more candidate alongside the linked type's resolved pool, "never silently discarded just because the mapper is no longer MappingType.DEFAULT". So the fix is frontend-only.

## Changes

- Show the defined-value input whenever the field is **not auto-linked** (auto-linked system fields like `targeted-asset` still hide it), so a value stays available alongside a manually linked type.
- Serialize step conditions through `mapFieldLinksToStepConditions` (removing the inline mapping that dropped `condition_value`), and align its keyTypes fallback with the previous inline behaviour (`'text'` for the degenerate no-type link, so the backend never gets empty keyTypes and drops the step's execution batches).
- Only carry **scalar** defined values (string/number/boolean) into `condition_value`: an array (multi-cardinality select) or object would otherwise be coerced into a meaningless string ("a,b", "[object Object]") and injected as a bogus runtime candidate.

Fixes #7210

## Test plan

- [x] Unit tests on `mapFieldLinksToStepConditions` (6): defined value carried into `condition_value` alongside its type; `condition_value` omitted when no value; `'text'` keyTypes fallback for a link with no type; numeric value coerced to string; non-scalar content (array/object) skipped; unlinked fields produce no conditions. 6/6 pass.
- [x] `yarn eslint` (changed files) + `yarn check-ts` clean.
- [ ] Manual: type a defined value on a payload argument, then link an output type - both persist; at run time the defined value participates as an extra candidate.

### DIFF
--- a/openaev-front/src/__tests__/admin/components/chaining/logic/drawer/ConfigureActionDetail.utils.test.ts
+++ b/openaev-front/src/__tests__/admin/components/chaining/logic/drawer/ConfigureActionDetail.utils.test.ts
@@ -1,0 +1,75 @@
+import { describe, expect, it } from 'vitest';
+
+import { mapFieldLinksToStepConditions } from '../../../../../../admin/components/chaining/logic/drawer/ConfigureActionDetail.utils';
+import type { ActionDetailData } from '../../../../../../admin/components/chaining/logic/types';
+
+// Minimal ActionDetailData carrying only what mapFieldLinksToStepConditions reads.
+const data = (over: Partial<ActionDetailData>): ActionDetailData => ({
+  inject_title: 'a',
+  inject_injector_contract: 'c',
+  inject_assets: [],
+  inject_asset_groups: [],
+  inject_teams: [],
+  inject_all_teams: false,
+  inject_documents: [],
+  inject_content: {},
+  inject_field_links: {},
+  contract_fields: [],
+  ...over,
+});
+
+describe('mapFieldLinksToStepConditions', () => {
+  it('carries a linked field’s defined value into condition_value alongside its type', () => {
+    // The field "target" is linked to the "host" output type AND has a defined value typed in.
+    const conditions = mapFieldLinksToStepConditions(data({
+      inject_content: { target: '10.0.0.5' },
+      inject_field_links: {
+        target: {
+          outputTypes: ['host'],
+          localScope: false,
+        },
+      },
+    }));
+
+    expect(conditions).toHaveLength(1);
+    expect(conditions[0]).toMatchObject({
+      condition_type: 'MAPPER',
+      condition_key: 'target',
+      condition_key_types: ['host'],
+      condition_value: '10.0.0.5', // the defined value is NOT dropped once a type is linked
+      condition_mapping_type: 'GLOBAL',
+    });
+  });
+
+  it('omits condition_value when the field has no defined value', () => {
+    const conditions = mapFieldLinksToStepConditions(data({
+      inject_content: {},
+      inject_field_links: {
+        target: {
+          outputTypes: ['host'],
+          localScope: true,
+        },
+      },
+    }));
+
+    expect(conditions[0].condition_value).toBeUndefined();
+    expect(conditions[0].condition_mapping_type).toBe('LOCAL');
+  });
+
+  it('falls back to a text key type for a link that carries no output type', () => {
+    // Degenerate case: a link with no output types must never serialize empty keyTypes, or the
+    // backend drops the whole step's execution batches.
+    const conditions = mapFieldLinksToStepConditions(data({
+      inject_content: { target: 'literal' },
+      inject_field_links: {
+        target: {
+          outputTypes: [],
+          localScope: false,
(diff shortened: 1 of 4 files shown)


SamuelHassine (approved):
Full review done: read the complete changed files plus the chaining-engine instructions/skill, and independently verified the backend claims this fix rests on. `ConditionService#resolveMapperPairs` does add the mapper's defined value as one more candidate for non-DEFAULT mappers (deduped against the pool), and empty keyTypes on a dynamic mapper does make `prepareInputsForStepExecution` return an empty batch list, so the 'text' fallback is required. The revived `mapFieldLinksToStepConditions` helper's output is field-by-field identical to the inline mapping it replaces (same ordering, temporary ids, keyTypes fallback and mapping type) with `condition_value` as the only delta, so no regression for existing saved workflows. One hardening pushed on top: only scalar defined values are carried into `condition_value` (an array or object would have been coerced into "a,b" / "[object Object]" and injected as a bogus runtime candidate), plus 3 more unit tests pinning the helper contract. All 6 tests, eslint, check-ts and the full CI matrix are green. LGTM.

SamuelHassine:
Review-and-fix pass summary:

- Independent review of all 4 files (full files, not just hunks), against the chaining-engine instructions and review skill.
- Backend claims verified in `ConditionService`: `resolveMapperPairs` keeps the mapper's defined value as an extra candidate for LOCAL/GLOBAL mappers, and missing keyTypes flips `hasMissingDynamicValues`, dropping every execution batch for the step - both exactly as the PR body states. `ConditionMapper` persists `condition_value` unconditionally, so nothing is rejected server-side.
- Dead-code revival checked: the helper's output matches the previous inline mapping field-by-field (ordering, temporary ids, 'text' keyTypes fallback, mapping type); the only behavioural delta is the added `condition_value`, which is the fix itself.
- Visibility predicate checked: only payload `targeted-asset` fields are in the auto-linked set (read-only links), so hiding their input still holds; argumentType-based default links keep the value input available, which is consistent with the fix since the backend never ignores the value for a non-DEFAULT mapper.
- Fix pushed (142f5c5, signed): carry only scalar defined values into `condition_value` - arrays (multi-cardinality selects) and objects were being String()-coerced into bogus candidates like "a,b". Added 3 unit tests (number coercion, non-scalar skipped, unlinked fields ignored); 6/6 pass, eslint + check-ts clean.
- Copilot reviewed all 4 files and generated no comments; no review threads to resolve.
- Docs Reviewer bot note: no docs update needed - this is an internal Configure-action drawer bugfix with no user-facing workflow or configuration change; the documentation gap check also passed.
- CI: all checks green including the required `quality-gates / Quality Status` and `Artifact & Image Sanity Check`; all codecov patch/project contexts pass. PR is MERGEABLE/CLEAN.
- PR body re-flowed to ASCII and updated to reflect the final state; title already matches the repo convention.

PR #7213, opened 2026-08-05
fix(chaining): forbid manual action->event links and label event nodes as events (#7212)

### LINKED ISSUE #7212: fix(chaining): logic page allows manual action→event links, and event nodes are labeled like triggers
Two issues on the scenario/simulation **Logic** page (the chaining graph).

## 1. Manual action → event link should not be possible

A user can drag from an **action** node's connector handle onto an **event** node and it creates a link. There is no manual action→event relationship — that link exists only as the automatic, informational **inferred** edge ("this action produces output an event listens on"). The only manual association is the other way: gate an action with an event (**event → action**, via the action's `step_condition_ids`).

Root cause: `LogicGraph.linkNodes` accepted the connection in **either** drag order, so an action→event drag was turned into a gating edge; and `GraphActionCard` exposed a connector handle to start that drag.

## 2. Event node reads like a trigger

The event node (`kind 'trigger'` internally) titled itself from the trigger's technical **condition-field keys** — or the literal "Trigger" — whenever the event had no name, so an unnamed event looked like a trigger. The rest of the chaining UI already calls these **Event** (drawer title, creation form, warning banner).

## Fix

- Remove the action card's connector handle and accept only the **trigger→action** direction in `linkNodes`.
- Title the event node from its name, falling back to **"Untitled event"** (not the condition fields), and change the node eyebrow from "Trigger" to "Event". The listened-on fields remain in the tooltip's "Listens on" row.

Seb-MIGUEL:
## Summary

Three Logic-page issues (chaining graph), all about the action/event relationship and event labeling, plus review follow-ups.

### 1. Manual action -> event link is no longer possible

`LogicGraph.linkNodes` accepted the connection in either drag order, so dragging from an action's connector handle onto an event created a gating edge. There is no manual action -> event relationship - it exists only as the automatic, informational inferred edge; gating goes the other way (event -> action, via `step_condition_ids`).

- Removed the action card's connector handle (`GraphActionCard`) and its `onConnectStart` prop.
- `linkNodes` now accepts only trigger -> action; any other combination is a defensive no-op.

### 2. The action "+" (add a downstream event) also removed

The inline "+" under an action ("Add a trigger fed by this action") let a user create an event from an action - the same action -> event relationship, via a button. Removed it and its now-dead plumbing end to end (`GraphActionCard` button, `LogicGraph` handler/prop, `Logic` `onAddTriggerAfterAction` handler + `prefillEventFields` state threaded to the event drawer, and - review follow-up - the `prefillFields` prop and `buildPrefilledGroups` helper left in `EventCreationForm`, plus the two orphaned i18n keys in all locales). The event "+" (add an action gated by this trigger, event -> action) is kept, as is the global add-component button.

### 3. Event node reads like a trigger

The event node (`kind 'trigger'` internally) titled itself from the trigger's technical condition-field keys - or literally "Trigger" - when unnamed, so an unnamed event looked like a trigger. The rest of the chaining UI already says Event.

- Title from the event name, falling back to "Untitled event" (new key added to all 9 locales).
- Node eyebrow changed from "Trigger" to "Event". Listened-on fields stay in the tooltip's "Listens on" row.
- Review follow-up: "Untitled event" is properly translated in all 8 non-English locales, and the pre-existing untranslated "Untitled action" key was translated in the same pass.

Fixes #7212

## Test plan

- [x] Component test `GraphTriggerCard.test.tsx`: shows the event name when present; falls back to "Untitled event" (not the condition-field labels) when unnamed; tooltip eyebrow reads "Event", never "Trigger". Verified the title tests fail without the fix and pass with it.
- [x] `yarn check-ts`, `yarn eslint` (touched files), and `yarn i18n-checker` all clean; all 9 lang JSON files parse.
- [x] Manual: on the Logic page - an action can no longer be linked to an event (no drag handle, no "+"); dragging from an event onto an action still gates it; the event "+" still adds a gated action. An unnamed event shows "Untitled event" with an "Event" eyebrow.

## Related

Second of the chaining QA PRs (first: #7211, defined-value + type-link). A separate, larger follow-up will restore the per-MITRE-tactic column layout with events placed left of their gated actions (regressed in #7099).


openaev-front/src/admin/components/chaining/logic/logic-graph/GraphActionCard.tsx
```diff
@@ -315,71 +309,14 @@ const GraphActionCard = ({
           </Box>
         )}
 
-        {!readOnly && onConnectStart && (
```
heditar, line 318: Should it be still possible to gate an action?

openaev-front/src/utils/lang/fr.json
```diff
@@ -3466,6 +3466,7 @@
   "Unsaved edits are reflected instantly": "Les modifications non enregistrées sont reflétées instantanément",
   "Unsubscribe from notifications on this entity": "Se desabonner des notifications de cette entite",
   "Untitled action": "Untitled action",
+  "Untitled event": "Untitled event",
```
heditar: no translation

openaev-front/src/admin/components/chaining/logic/logic-graph/GraphActionCard.tsx
SamuelHassine: Good question - yes, gating an action is still fully possible; what was removed is only the action-initiated direction. Verified end to end in LogicGraph.tsx: (1) drag-to-gate still works - the drag starts from the event node's connector handle (GraphTriggerCard passes kind 'trigger' to handleConnectStart) and the drop is resolved with elementFromPoint against the action card's wrapper div (data-node-id / data-node-kind="action"), so linkNodes receives trigger->action and creates the gating edge; the action card never needed its own handle to be a drop target. (2) The event's inline "+" ("Add an action gated by this trigger") still opens the drawer with linkToEventId set, so the new action is created already gated. (3) The global add-component button is untouched. The direction guard in linkNodes also cannot be defeated by drag order: the endpoints are passed positionally (drag source first) and a connect drag can only start from an event node, so action->event can never be produced.

openaev-front/src/utils/lang/fr.json
SamuelHassine: Good catch - fixed in ab2da52. "Untitled event" is now properly translated in all 8 non-English locales (fr "Evenement sans titre" with the proper accents in the file itself, plus de/es/it/ja/ko/ru/zh following each locale's existing noun conventions). The adjacent pre-existing "Untitled action" key had the same problem in the same locales, so it was translated in the same commit. Also removed the two i18n keys orphaned by this PR's UI removals ("Add a trigger fed by this action" and "Drag onto a trigger to gate this action") from all 9 locales. yarn i18n-checker is clean.

SamuelHassine (approved):
Reviewed the full changed files (LogicGraph, GraphActionCard, GraphTriggerCard, Logic, ComponentStepperDrawer, the new component test, all 9 lang files) against the chaining-engine instructions. The direction guard is correct and cannot be defeated by drag order (a connect drag can only start from an event node and endpoints are positional), and every remaining path to gate an action (event handle drag, event "+", global add) was verified working in code. Review feedback is addressed in ab2da52: proper translations for "Untitled event" and the pre-existing "Untitled action" in all 8 non-English locales, orphaned i18n keys removed, dead prefill plumbing in EventCreationForm removed, stale JSDoc refreshed, and the "Event" eyebrow pinned in the component test. All checks green. LGTM.

SamuelHassine:
Review-and-fix summary:

- Independent review of the full changed files plus callers/callees, with the chaining-engine instructions applied. Verified the core question from review: gating an action still works after removing the action-side handle and "+" - via the event node's connector-handle drag (the drop resolves against the action card's data-node-id wrapper, so no action-side handle is needed), via the event's inline "+", and via the global add-component button. The trigger->action-only guard in linkNodes cannot be bypassed by drag order because a connect drag can only start from an event node and the endpoints are passed positionally.
- Pushed ab2da52 addressing both review comments and review follow-ups: translated "Untitled event" and the pre-existing untranslated "Untitled action" in de/es/fr/it/ja/ko/ru/zh (each following the locale's existing noun conventions); removed the two i18n keys orphaned by the UI removals ("Add a trigger fed by this action", "Drag onto a trigger to gate this action") from all 9 locales; removed the dead prefillFields prop and buildPrefilledGroups helper from EventCreationForm (last caller was removed with the action "+"); refreshed stale JSDoc on both graph cards; added a tooltip test pinning the "Event" eyebrow.
- Checks run locally: GraphTriggerCard component tests (3/3), yarn check-ts, yarn eslint on touched files, yarn i18n-checker, JSON parse of all 9 lang files - all clean. CI is fully green including codecov.
- Both review threads are replied to and resolved; title normalized to ASCII and body updated to reflect the final state.

heditar (approved):

PR #7251, opened 2026-08-06
docs(fds): point the implementation playbook at its new home in the library

sandyghs:
Companion to XTM-Foundation/filigran-design-system#93, which receives the files.

## Why

The implementation playbook was written here, by this repository's pilot. The
second product pilot then found **no reference to it anywhere in its own
repository** and had to be told where it was. A document you have to announce is
not findable.

It now lives next to the library it describes — the one repository every pilot
already has to look at:
[`process/PRODUCT-IMPLEMENTATION-PLAYBOOK.md`](https://github.com/XTM-Foundation/filigran-design-system/blob/main/process/PRODUCT-IMPLEMENTATION-PLAYBOOK.md).

## What this PR does

`fds-migration/IMPLEMENTATION-PLAYBOOK.md` becomes a ~28-line pointer: where the
document went, why, what travelled with it, and what stayed.

Two companions leave with it, because they only make sense beside it and a
second copy here would drift invisibly until the next pilot followed the stale
one:

- `fds-migration/PLAYBOOK-DEFECTS.md` — the playbook's own defect register;
- `fds-migration/artifacts/ci-design-system-secret.test.ts` — the CI credential
  guard products copy into their own suite. Nothing in this repository runs it
  (it was never copied into `openaev-front`'s suite, only stored here), so
  removing it changes no check. It is now exercised by the library's CI against
  a fixture, which is more than it ever had here.

`fds-migration/LIBRARY-FEEDBACK.md` **stays**. It records what this product found
missing in the library while integrating it — this product's observation,
belonging to this product.

The rest of `fds-migration/` is untouched: migration state, component and token
mapping, implementation log, all specific to this repository.

## References updated

Two, both found by grepping the whole repository for the old path:

| File | Was | Now |
| --- | --- | --- |
| `.github/actions/setup-fds-auth/action.yml` | `See fds-migration/IMPLEMENTATION-PLAYBOOK.md.` | points at `process/PRODUCT-IMPLEMENTATION-PLAYBOOK.md` in the design-system repository |
| `fds-migration/LIBRARY-FEEDBACK.md` | `see the pin-bump section of IMPLEMENTATION-PLAYBOOK.md` | same section, new location named |

`fds-migration/AGENTS.md` is generated from the design-system repository and
carries no reference to the playbook, so it needs no change here.

## Not in scope

The `opencti` pointer. A pull request is under review there; its pointer comes
separately, after that one lands.


sandyghs:
cleaning documentation files

PR #7270, opened 2026-08-07
fix(chaining): allow an action to be gated by several events (#7266)

### LINKED ISSUE #7266: fix(chaining): an action gated by two events is buildable in the UI but makes the scenario unlaunchable
Launching a scenario fails with a toast — nothing runs:

```
New step (TEMPLATE): Only 1 condition can be first parent
```

The scenario is perfectly valid as far as the UI is concerned: one of its actions is gated by **two events**, and the logic map displays it as such (`2 trigger(s)` on the action card).

## The UI allows it, the launch refuses it

**Linking is unbounded.** `step_condition_ids` is a list, the graph appends to it:

```js
updateStep(actionId, buildStepUpdate(action, [...action.step_condition_ids, triggerId]))
```

and `ConditionService.linkExistingConditionsToStep` links every id it is given as a root, with no cardinality check:

```java
for (String conditionRootId : conditionRootIds) {
  Condition root = findConditionRootById(conditionRootId);
  linkToStep(root, step, true);
  conditionRepository.save(root);
}
```

**Copying at launch refuses it.** `StepService.copyStepConditionTemplate` collects the roots linked to the step and throws as soon as more than one is not a MAPPER:

```java
if (rootConditions.size() > 1) {
  long nonMapperRootCount =
      rootConditions.stream().filter(c -> c.getType() != ConditionType.MAPPER).count();
  if (nonMapperRootCount > 1) {
    throw new IllegalArgumentException("New step (TEMPLATE): Only 1 condition can be first parent");
  }
}
```

So a state the logic map lets you build — and renders without any warning — makes the scenario permanently unlaunchable. There is no hint in the UI that a second trigger is illegal, and the error surfaces only at launch, phrased in terms of "conditions" and "first parent" rather than events.

## Not a regression from the recent chaining work

- The validation dates back to the original chaining step CRUD (#4824), refined by "reuse shared event/condition when copying step templates" — nothing recent touched it.
- #7235 (lock banner / edit after reset) is front-only: `Logic.tsx`, `LogicReadOnlyBanner.tsx`, `Index.tsx`, `SimulationLogic.tsx`.
- #7237 (preserve a preserved event's subtree) does not affect the root count: the copy filters roots on `getConditionParent() == null`, so the leaves it keeps linked are never counted as roots.

## Which way should it be fixed?

The copy loop immediately below the guard **already iterates over every root** and copies each subtree, and the engine evaluates gating per root tree (`[Chaining] Filter tree NOT satisfied: rootId=…`). So the guard looks stricter than the rest of the stack, which suggests **allowing several event roots** rather than forbidding them.

That needs confirming against the engine's semantics for several gating trees on one step (AND across trees? OR?) before removing the check. If multiple gating events are *not* meant to be supported, then the fix belongs in the UI instead: refuse a second trigger on an action, and say so at link time rather than at launch.

Either way, the current state — buildable in the UI, fatal at launch — should not persist.

Seb-MIGUEL:
Fixes #7266.

Launching a scenario failed with a toast and nothing ran:

```
New step (TEMPLATE): Only 1 condition can be first parent
```

The scenario was valid as far as the UI is concerned: one action was gated by **two events**, rendered on the card as `2 trigger(s)`.

### Why allowing it is the coherent fix

Everything in the stack except one guard already supports several gating events:

- **Linking is unbounded.** `step_condition_ids` is a list, the graph appends to it (`[...action.step_condition_ids, triggerId]`), and `linkExistingConditionsToStep` links every id as a root with no cardinality check.
- **The engine already ANDs across trees.** `ConditionService.evaluateFilterConditions` iterates the conditions and returns `false` on the first unsatisfied tree — so "all events must be satisfied" is the existing, implemented semantics.
- **The copy already handles N roots.** The loop in `copyStepConditionTemplate` iterates `rootConditions` and copies each subtree.

Only the cardinality guard immediately above that loop rejected more than one non-MAPPER root, so a state the UI happily builds made the scenario permanently unlaunchable.

### Change

Drop that guard. Deliberately left in place:

- the **"at least one root"** check in the same method;
- `findRootConditionInput` — an **Event API** tree is one event, so exactly one root is right;
- the single-non-MAPPER-root rule for a tree authored **inline** with a step (`ConditionService.createConditionTree`), which is not how the UI links existing events, and which two existing tests pin.

### Verification

Unit test added, run **with and without** the change:

| | result |
|---|---|
| without | `IllegalArgument New step (TEMPLATE): Only 1 condition can be first parent` — the reported error, exactly |
| with | passes; both trees copied (4 conditions), roots `SMB_FOUND` and `SHARE_FOUND` |

`253 tests, 0 failures` across `StepServiceTest`, `StepServiceIntegrationTest`, `ConditionServiceTest`, `StepServiceScenarioIntegrationTest`, `WorkflowServiceTest`. Spotless applied.

### Worth a reviewer's eye

The AND-across-events semantics is **inherited, not chosen here** — it is what `evaluateFilterConditions` already does. If the product intent for two events on one action is OR, this PR makes that wrong behaviour reachable instead of blocked, and the combination logic needs its own change. Flagging it explicitly rather than assuming.

### DIFF
--- a/openaev-api/src/main/java/io/openaev/service/chaining/StepService.java
+++ b/openaev-api/src/main/java/io/openaev/service/chaining/StepService.java
@@ -575,17 +575,17 @@ public class StepService {
           "New step (TEMPLATE): At least 1 condition must be a root (no parent)");
     }
 
-    // Allow one event/filter root plus any number of mapper roots.
-    // This happens when a step has one linked event root and one or more action mappings.
-    if (rootConditions.size() > 1) {
-      long nonMapperRootCount =
-          rootConditions.stream().filter(c -> c.getType() != ConditionType.MAPPER).count();
-      if (nonMapperRootCount > 1) {
-        throw new IllegalArgumentException(
-            "New step (TEMPLATE): Only 1 condition can be first parent");
-      }
-    }
-
+    // Any number of event/filter roots is accepted, plus any number of mapper roots.
+    //
+    // The logic map lets an action be gated by SEVERAL events - `step_condition_ids` is a list, the
+    // graph appends to it, and linkExistingConditionsToStep links every id as a root - and the
+    // engine
+    // already ANDs across gating trees (evaluateFilterConditions returns false on the first
+    // unsatisfied tree). The loop below already copies every root and its subtree. This guard was
+    // the
+    // only thing rejecting more than one event root, so a scenario built that way in the UI became
+    // permanently unlaunchable, failing at copy time with "Only 1 condition can be first parent".
+    //
     // Full set of the source workflow's non-MAPPER conditions, to resolve children by parent id
     // independent of conditions_steps linkage (Event-API links only the root to the step).
     List<Condition> allSourceConditions =
--- a/openaev-api/src/test/java/io/openaev/service/chaining/StepServiceTest.java
+++ b/openaev-api/src/test/java/io/openaev/service/chaining/StepServiceTest.java
@@ -1299,6 +1299,77 @@ class StepServiceTest {
   class CopyStepConditionTemplateFields {
 
     @Test
+    void given_stepGatedByTwoEvents_should_copyBothTreesInsteadOfThrowing() {
+      // Arrange — an action gated by TWO events, which the logic map lets you build
+      // (step_condition_ids is a list) and the engine already ANDs across.
+      Workflow sourceWorkflow = new Workflow();
+      sourceWorkflow.setId("source-workflow-id");
+
+      Step sourceStep = new Step();
+      sourceStep.setId("source-step-id");
+      sourceStep.setWorkflow(sourceWorkflow);
+
+      Workflow targetWorkflow = new Workflow();
+      targetWorkflow.setId("target-workflow-id");
+
+      Step targetStep = new Step();
+      targetStep.setId("target-step-id");
+      targetStep.setWorkflow(targetWorkflow);
+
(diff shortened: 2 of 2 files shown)


guillaumejparis:
Can you try to solve this in frontend ? No reason to change backend here

Seb-MIGUEL:
Understood — moved to the frontend in #7274, which refuses the second event at link time with an explicit message. This PR is back to draft; close it if you prefer the front-only route as the definitive answer.

For the record, two things the front-only fix does not cover, so the decision is made with them in sight:

1. **Data already in that state stays unlaunchable.** There is at least one such scenario on `main-ff` right now; it will keep failing on Launch with `Only 1 condition can be first parent` until someone manually removes the extra event. #7274 only prevents new occurrences.
2. **The backend message stays cryptic**, and the API path is unguarded: it speaks of "conditions" and "first parent" rather than naming the action and its events.

Also worth knowing, since it shaped the original approach: the rest of the stack already supports several gating events — `linkExistingConditionsToStep` links every id as a root without any cardinality check, `evaluateFilterConditions` ANDs across gating trees, and the copy loop iterates every root. The guard removed here was the only element disagreeing. So "one event per action" is a product decision rather than a technical constraint — fine by me, just making it explicit so it is not re-discovered later.

Happy to add a clearer backend error message (naming the action, speaking of events) to #7274 without changing any behaviour, if that suits.

Issue #265, opened 2023-04-04
Triggering the inject now change the date of the inject


SamuelHassine:
## Description

Triggering the inject now changes the date of the inject, so when you stop / reset the exercise, all the injects are changed.

We should be able to trigger injects "now" without impacting the scenario.

RomuDeuxfois:
We have a reference between inject and exercice (ID), so when we change a property on inject we change this inject inside the exercice directly.

We can have a new property to handle manual trigger date to dissociate the two behaviors.

Do you have any other idea on how to do that ?

PR #7278, opened 2026-08-07
fix(chaining): credential finding and chokepoint highlighting is not working (#5048)

mergenhan:
## Fixes

Fixes OpenAEV-Platform/filigran-private#265

Related to #5048 (attack path view epic - not closed by this PR)

## Problem

When selecting a Chokepoint or a Discovered Credential finding in the Attack Path graph, the highlighted path did not match the selection - only the endpoint node itself got marked, with no path/links highlighted in blue and nothing dimmed, unlike every other finding type.

## Root causes

1. **Chokepoint click had no path highlighting at all.** It only set `selected: true` on the endpoint node - no upstream/downstream path walk, no dimming of unrelated nodes/edges (finding clicks already did this).
2. **Credential findings couldn't resolve their canonical graph node.** The drawer's finding value is masked server-side (`"user:****"`), but the graph's raw finding node value is unmasked (`"user:pass"`). The lookup used strict equality instead of the existing `findingValuesMatch()` helper (already used correctly in `highlightGraphFinding`/`openFindingFromGraph`), so the match silently failed and the credential's path never highlighted.
3. **Fixing both surfaced a further, chain-mode-specific bug**: the causal graph dedupes identical actions/finding values shared by multiple hosts into a single node (e.g. two hosts both exposing port 445, or the same step run against several hosts). An unrestricted path walk through those shared nodes bled into every other endpoint sharing the same recon result or action - highlighting the whole graph instead of just the selected path.

## Fix

- Added the same upstream/downstream path-set computation + dimming used for finding highlights to the chokepoint (endpoint-only) focus branch.
- Replaced the strict-equality canonical node lookup for drawer-clicked credential findings with `findingValuesMatch()`.
- Added a guard so the path walk (both for chokepoint and finding highlights) never marks another endpoint's own chain node as selected, while still allowing genuinely shared upstream nodes (injector/finding) to light up.
- **Producing-execution attribution for same-username credentials** (second commit): two credentials findings sharing a username but differing only by password/hash resolved to whichever category-page entry sorted first, so the wrong finding's producing action could be shown. Producing executions are now resolved from the full graph's exact execution->finding links (matched by the finding's own node id) FIRST, with the ambiguous category-page match kept only as a fallback.

## Review follow-up

- The upstream/downstream path walks (both the ones added here and the pre-existing ones they mirror) iterated a magic number of passes (8/6/3) over the edge list. They now go through a shared `expandPathSet()` helper that expands to a fixpoint (terminates because the set only grows, bounded by the graph's node ids), with walk-specific guards expressed as a `blocked()` predicate. The two deliberately shallow child walks keep an explicit cap via the named `AP_CHILD_WALK_PASSES` constant. Unit tests added for the helper.
- Documented on `findingValuesMatch()` why only credentials get the username-half comparison: the backend masks ONLY the credentials category in drawer DTO values; `password_policy`/`sid` are display-masked in the front, are not drawer categories, and resolve by exact node id instead.

## Testing

- `yarn check-ts` - 0 errors
- `yarn eslint` on the three touched files - clean
- `yarn vitest run src/__tests__/admin/components/simulations/simulation/attack_path/` - 6 files, 124/124 passed (includes 6 new `expandPathSet` tests, among them a 20-hop chain a fixed 8-pass cap would have truncated)
- Verified against live API data from a real chaining scenario (2-endpoint causal chain with shared injectors/findings) to confirm the endpoint-leak fix scopes correctly
- Manually verified in a locally rebuilt/redeployed instance: chokepoint clicks and credential-finding clicks now highlight only the correct path and dim everything else

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/openaev-front/src/admin/components/simulations/simulation/attack_path/SimulationAttackPath.tsx
+++ b/openaev-front/src/admin/components/simulations/simulation/attack_path/SimulationAttackPath.tsx
@@ -1030,7 +1030,7 @@ const SimulationAttackPath = ({ scenarioExerciseIds, scenarioId, hideLaunchCta =
       // not just its type cluster (the backend escapes `\`/`|`, so a share value never matches a
       // rebuilt `NODE_FINDING|type|value`, hence matching on typeFindings+value like highlightGraphFinding).
       const canonicalId = (fullDto?.attackPathNodes ?? [])
-        .find(n => n.type === 'FINDING' && (n.typeFindings ?? '') === (item.type ?? '') && (n.value ?? n.label) === (item.value ?? ''))?.id;
+        .find(n => n.type === 'FINDING' && (n.typeFindings ?? '') === (item.type ?? '') && findingValuesMatch(item.type ?? '', n.value ?? n.label ?? '', item.value ?? ''))?.id;
       if (!chainMode) {
         // Non-chain path-focus view: the finding's own node only exists once its type cluster is
         // expanded (see the auto-expand effect below), so leave selectedFindingId null here — the
@@ -1762,32 +1762,69 @@ const SimulationAttackPath = ({ scenarioExerciseIds, scenarioId, hideLaunchCta =
     // Focused finding-path view: keep the focused scope, and let clicks on findings/clusters
     // highlight the exact sub-path (injector -> endpoint -> selection).
     if (pathFinding) {
-      // Endpoint focus (no specific finding, e.g. a chokepoint click) with nothing sub-selected: only
-      // the focused endpoint is "selected" (blue); its injectors, edges and finding clusters keep their
-      // verdict colour (green/orange/red) so blue means "what I picked", not "the whole subgraph".
+      // Endpoint focus (no specific finding, e.g. a chokepoint click) with nothing sub-selected: mirror
+      // the finding/injector highlight below — walk the endpoint's FULL incoming path (every injector
+      // and finding cluster that reached it) and light it blue, dimming everything off that path. Same
+      // visual language as picking a finding: blue means "the path to what I picked".
       if (!pathFinding.type && !selectedInjectorId && !selectedFindingId) {
         // The endpoint's own node id in the scoped chain graph is `chain-ep|depth|<raw id>`, not the
         // raw id itself (that's the non-chain buildFindingPathFlow layout's scheme) — match either.
         const isFocusedEndpoint = (nodeId: string) => nodeId === pathFinding.endpointNodeId
           || (nodeId.startsWith('chain-ep|') && nodeId.endsWith(`|${pathFinding.endpointNodeId}`));
+        // A DIFFERENT endpoint's own instance in the scoped chain graph (any `chain-ep|...` node that
+        // isn't this one). The causal chain dedupes identical actions/finding VALUES across hosts into
+        // ONE shared node (e.g. two hosts both exposing port 445 collapse to a single "port|445" node,
+        // and a step run against several hosts is one shared injector node) — walking through those
+        // shared hops is correct (they genuinely are on this endpoint's path too), but must never carry
+        // the walk INTO another endpoint's own node, or every host sharing any recon result/action with
+        // this one lights up as if it were on the path (nothing left to dim, an all-blue graph).
+        const isOtherEndpoint = (nodeId: string) => nodeId.startsWith('chain-ep|') && !isFocusedEndpoint(nodeId);
+        const focusedNodeIds = baseFlow.nodes.filter(n => isFocusedEndpoint(n.id)).map(n => n.id);
+        const pathSet = new Set<string>(focusedNodeIds);
+        // Walk UPSTREAM from the endpoint: every injector/finding/endpoint that eventually leads into
+        // it (mirrors the selectedInjectorId/selectedFindingId upstream walks a few branches below).
+        for (let pass = 0; pass < 8; pass += 1) {
+          for (const e of baseFlow.edges) {
+            if (e.source && e.target && pathSet.has(e.target) && !pathSet.has(e.source) && !isOtherEndpoint(e.source)) {
+              pathSet.add(e.source);
+            }
+          }
+        }
+        // Walk DOWNSTREAM too, scoped from the endpoint itself: its own finding clusters hang off it as
+        // children, so without this pass they'd render dimmed even though they belong to this chokepoint.
+        const down = new Set<string>(focusedNodeIds);
+        for (let pass = 0; pass < 3; pass += 1) {
+          for (const e of baseFlow.edges) {
+            if (e.source && e.target && down.has(e.source) && !down.has(e.target) && !isOtherEndpoint(e.target)) {
+              down.add(e.target);
+              pathSet.add(e.target);
+            }
+          }
+        }
         return {
-          nodes: baseFlow.nodes.map(n => ({
-            ...n,
-            selected: isFocusedEndpoint(n.id),
-            data: {
-              ...(n.data ?? {}),
-              dimmed: false,
-            },
-          })),
-          edges: baseFlow.edges.map(e => ({
-            ...e,
-            selected: false,
-            data: {
-              ...(e.data ?? {}),
-              count: e.data?.count ?? 1,
-              dimmed: false,
-            },
-          })),
+          nodes: baseFlow.nodes.map((n) => {
+            const selected = pathSet.has(n.id);
+            return {
+              ...n,
+              selected,
+              data: {
+                ...(n.data ?? {}),
+                dimmed: !selected,
+              },
+            };
+          }),
+          edges: baseFlow.edges.map((e) => {
+            const selected = pathSet.has(e.source) && pathSet.has(e.target);
+            return {
+              ...e,
+              selected,
+              data: {
+                ...(e.data ?? {}),
+                count: e.data?.count ?? 1,
+                dimmed: !selected,
+              },
+            };
+          }),
         };
       }
       const injectorIds = new Set(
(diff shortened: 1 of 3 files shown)


mergenhan:
Review note — PR #7278 (fix #265)

What was broken: Clicking a Chokepoint or a Discovered Credential finding in the Attack Path graph didn't highlight the path — only the endpoint node got marked, no blue links, nothing dimmed.

Why, and what changed (3 separate but related bugs):

1. Chokepoint had no highlight logic at all. Unlike finding clicks (which already compute a path set and dim everything off it), the chokepoint branch only did  selected: true  on the endpoint node — literally no path/dim computation existed. Added the same upstream/downstream path-walk pattern already used for findings.
2. Credential findings couldn't match their own graph node. The drawer shows credential values masked ( user:•••• , done server-side for security), but the graph's underlying finding node still holds the raw value ( user:pass ). The code that resolves "which graph node does this drawer item correspond to" used strict  ===  instead of the existing  findingValuesMatch()  helper (which already knows how to compare masked vs. raw credential values, and is used correctly elsewhere in the same file) — so credentials silently never matched, path stayed empty. One-line fix: use the existing helper.
3. Fixing 1 and 2 exposed a real, deeper bug: in chain mode, when two hosts share an identical action or finding value (e.g. both expose port 445, or the same recon step ran against both), the graph merges them into one shared node. The path walk (both mine and the pre-existing finding-highlight one) crossed through that shared node and pulled in the other host's own chain nodes too — so selecting one endpoint lit up all endpoints. Fix: the walk still passes through shared injector/finding nodes (correct — they ARE on the path), but is now blocked from ever marking another endpoint's own node as selected.

How I verified it: pulled the real graph JSON via the API for an actual 2-endpoint chaining scenario with shared steps/findings, replayed the exact path-walk logic against it outside the app to confirm node-by-node which ones would light up before/after the fix, then confirmed visually in a locally rebuilt instance. Existing test suite (12 tests) +  tsc  both stay green — no behavioral change to the working finding types, only the previously-broken ones.

openaev-front/src/admin/components/simulations/simulation/attack_path/SimulationAttackPath.tsx
```diff
@@ -1762,32 +1762,69 @@ const SimulationAttackPath = ({ scenarioExerciseIds, scenarioId, hideLaunchCta =
...
+        // and a step run against several hosts is one shared injector node) — walking through those
+        // shared hops is correct (they genuinely are on this endpoint's path too), but must never carry
+        // the walk INTO another endpoint's own node, or every host sharing any recon result/action with
+        // this one lights up as if it were on the path (nothing left to dim, an all-blue graph).
+        const isOtherEndpoint = (nodeId: string) => nodeId.startsWith('chain-ep|') && !isFocusedEndpoint(nodeId);
+        const focusedNodeIds = baseFlow.nodes.filter(n => isFocusedEndpoint(n.id)).map(n => n.id);
+        const pathSet = new Set<string>(focusedNodeIds);
+        // Walk UPSTREAM from the endpoint: every injector/finding/endpoint that eventually leads into
+        // it (mirrors the selectedInjectorId/selectedFindingId upstream walks a few branches below).
+        for (let pass = 0; pass < 8; pass += 1) {
```
savacano28: why this number? pass < 8?

openaev-front/src/admin/components/simulations/simulation/attack_path/SimulationAttackPath.tsx
```diff
@@ -1030,7 +1030,7 @@ const SimulationAttackPath = ({ scenarioExerciseIds, scenarioId, hideLaunchCta =
       // not just its type cluster (the backend escapes `\`/`|`, so a share value never matches a
       // rebuilt `NODE_FINDING|type|value`, hence matching on typeFindings+value like highlightGraphFinding).
       const canonicalId = (fullDto?.attackPathNodes ?? [])
-        .find(n => n.type === 'FINDING' && (n.typeFindings ?? '') === (item.type ?? '') && (n.value ?? n.label) === (item.value ?? ''))?.id;
+        .find(n => n.type === 'FINDING' && (n.typeFindings ?? '') === (item.type ?? '') && findingValuesMatch(item.type ?? '', n.value ?? n.label ?? '', item.value ?? ''))?.id;
```
savacano28, line 1038: Question: findingValuesMatches only works for credentials, what about other masked findings like PasswordPolicy (in backend we are masking: keys, tokens, and passwords)?

mergenhan:
**Additional fix applied:** Credential highlighting in the graph was working correctly, but for two different credentials sharing the same username (yet having different passwords/hashes), the "action that generated this credential" information could be attributed to the wrong credential—this was because the drawer's category page matching logic (`findingValuesMatch`) compared only usernames (since passwords are masked server-side). Now, an exact (unmasked) node-ID match is attempted first; the old (ambiguous) fallback method is used only if that fails. `tsc` is clean, and the relevant test suite passed 12/12.

openaev-front/src/admin/components/simulations/simulation/attack_path/SimulationAttackPath.tsx
SamuelHassine: Good catch - there was no principled reason for 8: it mirrored the pre-existing walks in this memo, which all used hardcoded pass caps (8, 6 or 3 depending on the branch), and for full-path walks the cap was arbitrary and would silently truncate the highlight on chains deeper than the cap.

Fixed in 0f80bc7: all the walk loops now go through a shared expandPathSet() helper (attack-path-flow-helpers.ts) that expands the set until a full pass adds nothing. The fixpoint always terminates because the set only grows and is bounded by the graph's node ids, so no arbitrary cap is needed. The only walks that keep a cap are the two deliberately SHALLOW child walks (a selection's own expanded children), which now use the named AP_CHILD_WALK_PASSES constant with a comment explaining why they must stay shallow. Unit tests for the helper are included (including a 20-hop chain that a fixed cap of 8 would have truncated).

openaev-front/src/admin/components/simulations/simulation/attack_path/SimulationAttackPath.tsx
SamuelHassine: Good question. The backend masks drawer/category DTO values ONLY for the credentials category (AttackPathGraphService: maskValue is set only when category == credentials, and maskCredential keeps the username half in the clear - which is exactly the half findingValuesMatch compares). The keys/tokens/passwords masking you mention happens elsewhere: in the reconstructed command line and terminal output (maskSecrets / PrimitiveValueMaskingUtils), not in finding item values.

The other secret finding types (password_policy, sid) are masked in the FRONT for display only (maskFindingValue); their DTO/graph values stay raw. They are also not drawer categories (CATEGORY_OF_TYPE has no entry for them), so no item of those types ever reaches this matcher - and a fully-masked value could not be value-matched anyway, since there is no plaintext half left to compare. That is why the follow-up commit on this PR resolves producing executions by the finding's exact node id FIRST and only falls back to the value match. I documented all of this on findingValuesMatch in 0f80bc7 so the constraint is explicit for the next masked type someone adds.

SamuelHassine (approved):
Reviewed the full changed file (not just the diff), the callers of every touched function, and the backend masking path (AttackPathGraphService) backing the drawer DTOs. The three fixes are correct and well-scoped: the chokepoint focus branch now mirrors the finding-highlight path walk, the drawer credential lookup uses the existing findingValuesMatch() helper consistently, and the other-endpoint guard correctly stops the walk at another endpoint's own chain node while letting genuinely shared upstream nodes light up. The second commit's exact-node-id-first resolution for producing executions removes a real same-username attribution bug.

Review follow-up applied in 0f80bc7: the path walks now expand to a fixpoint through a shared expandPathSet() helper instead of magic pass caps (answering the "why pass < 8?" thread), the deliberately shallow child walks keep an explicit AP_CHILD_WALK_PASSES cap, findingValuesMatch() now documents why only credentials get the username-half comparison (the backend masks only the credentials drawer category; password_policy/sid resolve by exact node id), and the helper has unit tests. check-ts, eslint and the 124 attack-path tests are green locally; CI is fully green.

SamuelHassine:
## Review-and-fix summary (automated pass)

**What this PR does**: fixes Attack Path graph highlighting for chokepoint clicks (previously no path walk/dimming at all) and drawer-clicked credential findings (masked drawer value vs raw graph value never matched under strict equality), adds a guard so chain-mode path walks never bleed into another endpoint's own chain node through deduped shared nodes, and (second commit) resolves a finding's producing executions by its exact node id first so two same-username credentials no longer show each other's producing action.

**Independent review findings applied** (commit 0f80bc7):

- The upstream/downstream path walks (the two added here and the pre-existing ones they mirror) iterated a magic number of passes (8/6/3). They now expand to a FIXPOINT through a shared `expandPathSet()` helper in `attack-path-flow-helpers.ts` (terminates because the set only grows, bounded by the graph's node ids); walk-specific guards became a `blocked()` predicate. The two deliberately shallow child walks keep an explicit cap via the named `AP_CHILD_WALK_PASSES` constant. Unit tests added (6, including a 20-hop chain a fixed 8-pass cap would truncate).
- Duplicated focused/other-endpoint predicates hoisted to a single definition in the highlight memo.
- `findingValuesMatch()` now documents why only credentials get the username-half comparison: the backend masks ONLY the credentials category in drawer DTO values (`AttackPathGraphService`); `password_policy`/`sid` are display-masked in the front, are not drawer categories, and resolve by exact node id.

**Review threads**: both of @savacano28's questions ("why pass < 8?" and "what about other masked findings like PasswordPolicy?") are answered inline and resolved; the first is answered in code (no more magic pass cap), the second in the `findingValuesMatch()` documentation.

**Follow-up (out of scope for this PR, backend change)**: `AttackPathFindingItemDTO` carries no finding node id, so a DRAWER click on one of two same-username credentials can still resolve the other credential's canonical graph node (the drawer rows are visually identical anyway since the secret half is masked). Graph clicks are exact by node id since the second commit. Adding the finding's node id to the drawer DTO would make drawer picks exact too.

**Verification**: `yarn check-ts` clean, `yarn eslint` clean on touched files, `yarn vitest run` on the attack_path suites 124/124 green locally. CI fully green on the final head (a136ce8 + 0f80bc7), commits signed, no merge conflicts (mergeable/CLEAN). Not merging - left for a maintainer.

PR #7280, opened 2026-08-07
feat(scope): capability and grant holder cannot grant a capability or grant they don't hold (#7197) (#7198)

### LINKED ISSUE #7197: feat(scope): US.2 — Tenant-side: capability holder cannot grant a capability they don't hold
## Use case

Splitting user/group/role management into its own capability (US.1) removes the BYPASS-specific escalation path, but doesn't by itself stop a broader class of over-granting: today, nothing structurally prevents a caller who holds "Manage tenant users, groups and roles" from creating/updating a role with capabilities beyond their own, or assigning a user to a group whose role carries capabilities the assigning caller doesn't hold. This story defines that rule generically — not scoped to BYPASS or any single capability — so it holds for any capability, current or future.

**AC1: Given** a caller holding "Manage tenant users, groups and roles" (or equivalent role-management capability)

**When** they attempt to create or update a tenant role with `role_capabilities` containing any capability the caller does not themselves hold

**Then** the request is rejected with HTTP 400, no partial role is created/updated, and the error identifies which capability the caller lacks and cannot grant

**AC2: Given** the same caller

**When** they create or update a tenant role with `role_capabilities` limited to capabilities they themselves hold

**Then** the request succeeds, unchanged from current behavior — no over-restriction of legitimate role management

**AC3: Given** a caller attempting to assign a user to a tenant group whose role carries capabilities the caller does not hold

**When** that assignment is attempted via API or UI

**Then** it is rejected — group/role assignment is subject to the same holder-capability check as role create/update, not just role definition itself

**AC4: Given** the seeded Administrators role/group

**When** its members perform role or group/role-assignment operations

**Then** they are unaffected by this guard, since they hold the full tenant-scoped capability set — this rule constrains non-admin capability holders, not the seeded admin path

### LINKED ISSUE #7198: feat(scope): US.3 — Platform-side: capability holder cannot grant a capability they don't hold
## Use case

Same generic no-escalation rule as US.2, applied to the platform side: a caller managing platform roles/groups should not be able to create or update a platform role with capabilities beyond their own, nor assign a user to a platform group whose role carries capabilities the caller doesn't hold. This closes the equivalent structural gap on the platform path (PlatformRoleService / PlatformRoleApi), independent of whatever tenant-side fix ships in US.1/US.2.

**AC1: Given** a caller holding "Manage platform users, groups and roles" 

**When** they attempt to create or update a platform role with `role_capabilities` containing any capability the caller does not themselves hold

**Then** the request is rejected with HTTP 400, no partial role is created/updated, and the error identifies which capability the caller lacks and cannot grant

**AC2: Given** the same caller

**When** they create or update a platform role with `role_capabilities` limited to capabilities they themselves hold

**Then** the request succeeds, unchanged from current behavior — no over-restriction of legitimate platform role management

**AC3: Given** a caller attempting to assign a user to a platform group whose role carries capabilities the caller does not hold

**When** that assignment is attempted via API or UI

**Then** it is rejected — group/role assignment is subject to the same holder-capability check as role create/update, not just role definition itself

**AC4: Given** the platform's seeded/root administrative role

**When** its members perform role or group/role-assignment operations

**Then** they are unaffected by this guard, since they hold the full platform-scoped capability set — this rule constrains non-root platform capability holders, not the seeded bootstrap path

damgouj:
### Proposed changes

* Add a guard to block privilege escalation

### Related issues

* Closes https://github.com/OpenAEV-Platform/openaev/issues/7197
* Closes https://github.com/OpenAEV-Platform/openaev/issues/7198


openaev-api/src/main/java/io/openaev/processor/datapack/V20260805_Observer_and_manager_users.java
```diff
@@ -100,7 +100,7 @@ private void ensureRolesAndGroups() {
   }
 
   private Role createRoleIfMissing(String roleName, Set<Capability> capabilities) {
-    return roleService.createRole(roleName, roleName, capabilities);
+    return tenantRoleService.createRole(roleName, roleName, capabilities);
```
damgouj, line 103: question: not createRoleInternal here ?

damgouj (commented):
Tests OK, nice improvements, thanks !

Just a question but you can approve and merge after if you want :)

RomuDeuxfois (approved):

PR #7285, opened 2026-08-07
fix(attack-path): same-username credentials could show the wrong producing action when highlighted (#7278)

mergenhan:
## Problem

Follow-up finding while validating #7278 (chokepoint/credential highlighting): when a **credential finding** is clicked in the Attack Path graph, the node/path highlighting itself is correct (keyed by the finding's own unique node id), but the **"producing execution" resolution** (which drives the Result/Terminal + execution-feed scoping) could point at the wrong finding.

## Root cause

`highlightGraphFinding` / `openFindingFromGraph` tried the drawer's category-page match (`findingValuesMatch()`) first. For the `credentials` type, that helper compares **only the username**, because the drawer's finding value is masked server-side (`"user:••••"`) — the secret can never be compared client-side.

So two credential findings sharing the same username but differing only by password/hash resolve identically in the drawer (`"user : ••••••"` both), and the category-page `.find()` silently returns whichever one sorts first — attributing the wrong producing action to the finding actually clicked.

## Fix

Resolve the producing executions from the **full graph's exact, unmasked execution→finding links**, matched by the finding's own node id (or canonical id) FIRST. Fall back to the ambiguous category-page match only when that exact lookup finds nothing (e.g. the full graph isn't loaded yet). This mirrors the exact-match approach already used for chain-mode finding clicks.

## Testing

- `yarn tsc --noEmit -p .` — 0 errors
- `yarn vitest run src/__tests__/admin/components/simulations/simulation/attack_path/SimulationAttackPath.test.tsx` — 4/4 passed

## Note for reviewers

This does not change any node/path highlighting behavior (already correct per #7278) — it only fixes which execution gets attributed as the "producing action" for a clicked credential finding when two credentials share a username.

### DIFF
--- a/openaev-front/src/admin/components/simulations/simulation/attack_path/SimulationAttackPath.tsx
+++ b/openaev-front/src/admin/components/simulations/simulation/attack_path/SimulationAttackPath.tsx
@@ -1162,16 +1162,14 @@ const SimulationAttackPath = ({ scenarioExerciseIds, scenarioId }: SimulationAtt
     const applyExec = (ids: string[]) => setHighlightedExecutionIds(new Set(ids));
     const matchIn = (items: AttackPathFindingItemDTO[]) =>
       items.find(it => it.endpointKey === endpointKey && (it.type ?? '') === type && findingValuesMatch(type, it.value ?? '', value));
-    const loaded = matchIn(findingsPage?.items ?? []);
-    if (loaded) {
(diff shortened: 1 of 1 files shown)


mergenhan:
Yanlışlıkla yeni PR olarak açılmış; fix zaten #7278'in branch'ine (fix/attack-path-chokepoint-credential-highlight) commit olarak eklendi. Kapatılıyor.

Issue #105, opened 2021-11-24
Implement more SMS platforms in the player


SamuelHassine:
## Problem to Solve

Implement more SMS platforms in the player

## Current Workaround

{ Please describe how you currently solve or work around this problem, given OpenEx's limitation. }

## Proposed Solution

{ Please describe the solution you would like OpenEx to provide, to solve the problem above. }

## Additional Information

{ Any additional information, including logs or screenshots if you have any. }

Issue #22, opened 2019-04-23
Migrate, update and finish documentation


SamuelHassine:
## Problem to Solve

Update the current documentation, migrate it to Docusaurus.

## Current Workaround

None.

## Proposed Solution

- Write the documentation

## Additional Information

None.

Issue #24, opened 2019-04-23
Implement unit and integration tests


SamuelHassine:
## Problem to Solve

It can be useful to have test since we have more developers on the project.

## Current Workaround

None.

## Proposed Solution

- Implement tests

## Additional Information

None.

Issue #18, opened 2019-04-23
Export a whole exercise (and re-import it!)


SamuelHassine:
## Problem to Solve

It will be very useful to share scenarios with other customers and import the scenarios they shared.

## Current Workaround

None.

## Proposed Solution

- Add the features in the API.
- Allow to import & export scenarios in the UI.

## Additional Information

None.

PR #7305, opened 2026-08-09
chore: adopt the design system Header for the admin top bar

sandyghs:
Adopts the design system `Header` for the admin top bar — the third product pilot after the two `Navbar` ones, and the first to take a container **and** its entire interior.

## What ships

The bar is `Header` + `HeaderGroup` from `@filigran/design-system`, and **every control inside it is a library component**: `SearchField`, `Button` (`ia`/`tertiary`, which paints the AI gradient this bar used to hand-roll with `backgroundClip`), `IconButton`, `Menu`, `Tooltip`, `Chip`, `Badge`, `Spinner`, `ProgressBar`, `Text`. The rail keeps the `Navbar` from pilot 1 and gains the tenant switcher on the library `Menu`.

Structure, after Samuel's review and Sandy's arbitration:

```
[ search 200–500 ]        [ AI cluster gap 8 ] 16 │ 16 [ platform actions gap 8 ]
```

The rule is `HeaderGroup separatorBefore` (library PR #117), not a `<div>`.

## Pin

`@filigran/design-system` at **`35a476849ba72d48cacae2568643f0b5638bc468`** (head of the library's `main`).

Adopted along the way: `Header`/`HeaderGroup` (+ `separatorBefore`), `SearchField` (+ its 16px magnifier fix and its clear cross), `Menu`, `Chip` with the `ee` severity, `Badge`, `Spinner` (`xl` tier), `ProgressBar` (+ its `tone` axis), `Text`, `--fds-z-overlay`, `--fds-header-height`.

## Numbers

| | |
|---|---|
| Search field window | **200–500px** (design decision, round 4) |
| Bar height | 68px, from the library |
| Separator | 1px @ 50%, 36px tall, **16px clear on each side**, measured in both themes |
| Unread badge / running counter | dot 8px / counter **20px**, both announced (`aria-describedby`) |
| Spinner | **32px** on the bar button (`xl`, 4.00px clearance on the glyph), 16px per row (`sm`) |
| Progress bars | 328×**4px**, determinate, `aria-valuenow`, **coloured by status** (`tone`) |
| Tests | **650** front unit tests (43 files) |
| MUI controls left in bar + rail | **0** of 155 rendered elements, asserted per SYMBOL by `expectNoMuiControls` — not a hand-kept list, and mutation-proven |

## Rounds

1. Bar adopted; rail fixed to the viewport; MUI token bridge regenerated.
2. Interior converted to library controls (the scope rule: *where the library ships a component, the pilot uses it*).
3. Clear cross and magnifier fixed upstream, then bumped in.
4. Search window narrowed to 200–500; **57px of bar overflow at 768px accepted** (non-default state, on a shell that already carries a 1400px page floor).
5. Narrative comments compressed to one-line markers (223 → 47 lines), prose moved to `fds-migration/`.
6. Review (Romuald): the current-page highlight was silently lost — a layered utility losing to the variant's own `bg-transparent`; magic numbers named.
7. Review (Sandy): the last MUI survivors converted — the EE tag, both badges, the Ariane EE marker.
8. Badge announcements proved to reach assistive technology (they did not: the badge sat in an `aria-hidden` slot and the test asserted `textContent`).
9. Separator, `Spinner`, `ProgressBar`, `Text`.
10. This bump: the `xl` spinner tier, the `ProgressBar` `tone` axis, and the guard inverted to bite per symbol. The `Badge` default moved to red in the library — visible here, with no product change, and left for per-site arbitration.

Overflow at 768px with the rail expanded by hand is now **85px** (57 → 69 with the EE chip, → 85 with the separator's clear space), accepted each time and recorded in the log.

## What the library still owes this bar

**Exactly one, and it is the only exemption in the guard:** `Popover`. PR #105 added the primitive but kept it internal (`require()` returns `undefined` for it), so the bulk-operations panel stays a MUI `Popover`, and with it the `Box` wrappers laying its rows out. Dated exemption, removal condition = the library exports a `Popover`. Everything else #22 ever listed — `Divider`, `Badge`, `Progress` — has shipped and is adopted.

One gap inside a shipped component, named rather than worked around: `ProgressBar` has no `warning` tone. Nothing in this bar needs one today; a status that did would keep the default rather than borrow a neighbouring family.

## Verification

`yarn check-ts` · `eslint --max-warnings 0` · **650** vitest · `yarn i18n-checker` · `yarn build` — all green, and CI green on the head.

**Migration conformity: 12 checks, 0 issues, exit 0** — and truthfully so. The MUI token bridge was regenerated at this pin: library PR #116 touched `theme.css`, which moved the bridge's provenance hash out of date (`sha256:e2eb556…` → `sha256:70ff37b…`). **No token value changed** — the regenerated file differs by that one hash line and nothing else. Worth knowing for the next pilot: `bridge-freshness` compares the bridge against the *sibling library checkout*, not against the pinned commit, so a checkout left behind reports OK on a stale bridge. It did here, for two bumps, until the checkout was moved to the pin.

Checkpoints were re-taken on every bump on a **cold Vite cache**: the pre-bundle survives a pin change and will happily serve a build without the new exports. Accessible names and descriptions are measured in Chromium's own accessibility tree, not inferred from the DOM.

One CI re-run to declare: `f4b5153d3` attempt 2 — `API Tests (6)` failed on Maven Central returning **HTTP 429 Too Many Requests** while resolving `spring-boot-starter-parent:pom:3.5.16`, before any test ran; an infrastructure flake with no relation to this diff, green on retry with no code change.

Docs live in `fds-migration/`: `IMPLEMENTATION-LOG.md` (per-session journal, including every design arbitration) and `LIBRARY-FEEDBACK.md` (what the library was missing, with removal conditions).


openaev-front/src/admin/components/nav/TopBar.tsx
```diff
@@ -96,133 +85,113 @@ const TopBar: FunctionComponent = () => {
...
-            alignItems: 'center',
-            marginTop: bannerHeightNumber,
-            height: '100%',
-            minHeight: 68,
-            paddingLeft: theme.spacing(3),
-            paddingRight: theme.spacing(3),
-            display: 'flex',
-            justifyContent: 'space-between',
-            background: `linear-gradient(90deg, ${alpha(gradientStart, 0.9)} 0%, ${alpha(gradientEnd, 0.9)} 100%)`,
+            minWidth: 200,
```
RomuDeuxfois: **todo**: Are we agree with hardcoding things here ?

openaev-front/src/admin/components/nav/BulkOperationsIndicator.tsx
```diff
@@ -72,46 +73,54 @@ const BulkOperationsIndicator: FunctionComponent = () => {
...
-          )}
-        </IconButton>
+              >
+                <Badge
+                  badgeContent={runningCount}
+                  color="primary"
+                  overlap="circular"
+                  sx={{
+                    '& .MuiBadge-badge': {
+                      fontSize: 10,
```
RomuDeuxfois: **todo**: Are we ok to harcoding things in this component (next lines too) ?

RomuDeuxfois:
Tests:
- I have different background here
image
image
- When I am on the notification page, the notification btn is not hilight
image
image


New changes (just checking if it's ok) :
image

openaev-front/src/admin/components/nav/TopBar.tsx
sandyghs: The **values** are a design decision, the **form** was fair criticism — fixed in 484b83976.

The 200px floor and 500px ceiling were chosen by Sandy (design lead) in review round 4, replacing the `550px / 50% / 680px` window this pilot first shipped. They are recorded in [`fds-migration/IMPLEMENTATION-LOG.md`](../blob/sandyghs-miniature-enigma/fds-migration/IMPLEMENTATION-LOG.md) (entry 2026-08-11), together with the consequence that was accepted with them: at a 768px viewport with the navigation rail expanded *by hand*, the bar overflows its own box by 57px. That was measured, presented, and accepted as a non-default state — the shell already carries a pre-existing 1400px page floor, so global responsive behaviour belongs to the Layout chantier, not to this pilot.

What changed in 484b83976 is that they are no longer literals at the call site. They now sit in `navbarConstants.ts`, beside the rail widths, which is already the single home for the geometry several files have to agree on:

```ts
export const TOP_BAR_SEARCH_MIN_WIDTH = 200;
export const TOP_BAR_SEARCH_MAX_WIDTH = 500;
```

Two things I deliberately did **not** do:

- I did not move the window onto the `SearchField` instance. The component forwards `className` to its wrapper but spreads its remaining props — `style` included — onto the inner `<input>`, so a width passed there sizes the text box instead of the field. The group is the product's own layer and is the correct place; the field spans it through `fullWidth`. Filed upstream as feedback #18.
- I did not make the test import the constants. It still asserts the literal `200px` / `500px`, because an assertion that reads its expectation from the code under test proves nothing.

openaev-front/src/admin/components/nav/BulkOperationsIndicator.tsx
sandyghs: Named in 484b83976. Checking first was worth it, because these numbers have two different stories and only one of them is ours:

**Pre-existing** — the badge font size, its 16px minimum, and the 32px spinner with its 2px ring are byte-identical to the legacy bar at `0d58172f2` (`git show 0d58172f2:openaev-front/src/admin/components/nav/BulkOperationsIndicator.tsx`). This PR only moved them into the library's icon slot. Named for readability, values untouched.

**New here, and it was a real latent bug** — the centring offsets. The library's icon slot is not a positioning context (the MUI `IconButton` it replaced was), so this component now positions the spinner absolutely. That arrived as two literal `-16px`, silently coupled to the 32px size: change the size and the ring goes off-centre with nothing to catch it. Now derived:

```ts
const SPINNER_SIZE = 32;
const SPINNER_CENTRING_OFFSET = `-${SPINNER_SIZE / 2}px`;
```

On the wider question — the `Badge` and the `Popover` staying MUI is an **interim compensation, not a preference**. The library ships no `Badge` at all and its nearest surfaces (`Menu` for command rows, `Dialog` for modals) do not fit a progress panel. That is filed as [feedback #22](../blob/sandyghs-miniature-enigma/fds-migration/LIBRARY-FEEDBACK.md), and the site carries a one-line marker naming the removal condition:

```
// FDS-WORKAROUND #22: badge and spinner stay MUI in a positioned span — remove when the library ships a Badge — see fds-migration/LIBRARY-FEEDBACK.md
```

So when the library ships `Badge`, this compensation and these constants go away together.

sandyghs:
Thanks — the notification one was a real bug, and it was bigger than the bell. Details per point.

### 1. Notification button not highlighted — fixed in 46f2ee3cb

Reproduced, and it was **not only the bell**: all four `TopBarIconLink` controls (triggers, notifications, agents, and the external CTEM shortcut, which has no current-page notion) failed the same way. The account menu was fine because it is the library's own `IconButton`.

The cause was not the route predicate — `aria-current="page"` was already correct on every route, so the defect was purely visual and nothing failed. The control asked for the selected state through the class list, and the class silently lost:

| | offset in the library's compiled stylesheet |
|---|---|
| `.bg-filigran-brand-primary-transparency` | 53927 |
| `.bg-transparent` | 54564 |

`iconButtonVariants({ priority: 'tertiary' })` already emits `bg-transparent`. Both rules are in the **same cascade layer** with the **same specificity**, so source order decides and `bg-transparent` wins — computed background `rgba(0, 0, 0, 0)` on every route. The library's own `IconButton` escapes this because it merges its classes with tailwind-merge, which *drops* the conflicting `bg-transparent`; this component composes its own class list, so it cannot rely on that.

Fix: paint the selected background inline, next to the glyph colour that is already inline for the neighbouring cascade-layer reason (feedback #24), using the library's own token — never a literal:

```ts
const SELECTED_BACKGROUND = 'var(--color-filigran-brand-primary-transparency)';
```

Measured in the running product afterwards, per route: each control highlights on exactly its own route and on no other, with `color(srgb 0.258824 0.792157 1 / 0.1)` — the same computed value the library's `IconButton` produces. Covered by five new assertions in `TopBarIconLink.test.tsx`; the two background ones fail on the parent commit and pass here.

### 2. "I have different background here" — I could not reproduce it, and I would rather ask than guess

Two hypotheses checked and **both ruled out from the code**, not merely unobserved:

- **Gradient angle.** Not it. The legacy bar used `linear-gradient(90deg, …)` and the library `Header` also paints `linear-gradient(90deg, …)`. Same angle, so there is nothing to align.
- **A tenant colour landing on the wrong surface.** The two surfaces in your screenshots — the navigation rail and the right-hand secondary menu — are painted from theme tokens this PR does not touch. `RightMenu`'s only functional change here is its top offset (`68` → `var(--fds-header-height, 68px)`); its background is still `theme.palette.background.nav`. And the regenerated token bridge leaves the relevant values identical at both ends: `--bg-elevation-heading-layer-0` and `--bg-elevation-default-layer-0` are `#070d18` dark / `#f2f2f3` light before **and** after.

Measured on my instance at `484b83976`, defaults, both themes, on `/admin/settings/security/users` and `/admin/settings/customization`: rail, page and right menu all coherent with those tokens (dark: all three `rgb(7, 13, 24)`).

There **is** one real delta this PR introduces, but on the **bar**, not on those panels: the legacy bar pre-faded its own gradient stops at 90%, whereas the library paints its glass layer at Figma's 94% (measured `::before` opacity `0.94`). So the bar is very slightly less transparent than before. That is known and recorded (feedback #17).

Could you tell me which surface you mean and your config — **theme**, whether the tenant has a **custom background / navigation / paper colour** set (your first screenshot shows that panel, all `Défaut`), and your **browser zoom**? Your two screenshots are 309×649 and 426×600, so I am not certain I am looking at the same surface you are. If it is the bar, the 90% → 94% above is the explanation and it is a deliberate design-system value; if it is the rail or the right menu, I have not managed to make it differ and I would like your exact state to chase it properly.

### 3. "New changes (just checking if it's ok)"

Yes, intended. The search field is bounded to a **200px floor and a 500px ceiling** — a design decision by Sandy in review round 4, replacing the wider `550–680px` window this pilot first shipped. It is recorded in [`fds-migration/IMPLEMENTATION-LOG.md`](../blob/sandyghs-miniature-enigma/fds-migration/IMPLEMENTATION-LOG.md), entry 2026-08-11, with the accepted trade-off spelled out. The values are now named constants in `navbarConstants.ts` (484b83976), following your other comment.

---

Head is `484b83976`. I have left your two inline threads open for you to resolve.

RomuDeuxfois:
@sandyghs please be more concise on your points :)

1. Fix
2. We have sometime gradient and sometime not, so I can't say if it's ok or not.. I let you decide it.
image

sandyghs:
> @sandyghs please be more concise on your points :)
> 
> 1. Fix
> 2. We have sometime gradient and sometime not, so I can't say if it's ok or not.. I let you decide it.
> 
> image

yes sorry I’ll be more concise from now on!
2. The navbar PR only concerned the main left navbar – hence the gap you found! I’ll sort that out in a separate PR.

RomuDeuxfois (approved):

PR #7318, opened 2026-08-10
fix(attack-path): removed password_policy masking from the attack path view (#5048)

impolitepanda:
### Proposed changes

* Align the viewing of password policies from the attack path view to match the one in the finding screens
* Password policies are now therefore unmasked

### Testing Instructions

1. Create a simulation with an inject generating a password output
2. Launch the simulation then check the attack path view, the password policy shouldn't be masked when it appears.

### Related issues

* Relates to https://github.com/OpenAEV-Platform/filigran-private/issues/266

### Checklist

### DIFF
--- a/openaev-front/src/admin/components/simulations/simulation/attack_path/attack-path-flow-helpers.ts
+++ b/openaev-front/src/admin/components/simulations/simulation/attack_path/attack-path-flow-helpers.ts
@@ -1180,14 +1180,14 @@ export const buildFindingPathFlow = (
 // resolved through FILTER_TO_FINDING_TYPES, defaulting to the type itself so new types work with no code.
 export type AttackPathFindingFilter = 'endpoints' | string;
 
-// Finding types whose value is a captured secret; masked by default in the UI (spec §14). Revealing
-// them is an explicit, permission-gated action handled by the Result/Terminal increment.
-export const SENSITIVE_FINDING_TYPES = new Set(['credentials', 'password_policy', 'sid']);
+// Finding types whose value is a captured secret; masked by default in the UI (spec §14). Only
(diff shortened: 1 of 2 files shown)


camrrx (approved):

openaev-front/src/admin/components/simulations/simulation/attack_path/attack-path-flow-helpers.ts
```diff
@@ -1180,14 +1180,14 @@ export const buildFindingPathFlow = (
 // resolved through FILTER_TO_FINDING_TYPES, defaulting to the type itself so new types work with no code.
 export type AttackPathFindingFilter = 'endpoints' | string;
 
-// Finding types whose value is a captured secret; masked by default in the UI (spec §14). Revealing
-// them is an explicit, permission-gated action handled by the Result/Terminal increment.
-export const SENSITIVE_FINDING_TYPES = new Set(['credentials', 'password_policy', 'sid']);
+// Finding types whose value is a captured secret; masked by default in the UI (spec §14). Only
+// credentials and sid are masked here; revealing remains a permission-gated action.
+export const SENSITIVE_FINDING_TYPES = new Set(['credentials', 'sid']);
```
savacano28, line 1185: I think we should ask to Product: If we need this filter because currently in findings we dont mask any finding type and also we could check why in the lines 1195-1201 we have also other filter for credentials.

impolitepanda: Done here: https://github.com/OpenAEV-Platform/filigran-private/issues/266#issuecomment-5242381293

I'll change the PR content to relates instead of close for the linked issue, this way we can still merge this one.

PR #7329, opened 2026-08-11
fix(rbac): lenient scope filtering on role capability update (#7321)

### LINKED ISSUE #7321: fix: tenant role update fails when role holds stale platform-scoped capabilities
## Bug

An admin with `BYPASS` cannot uncheck **"Manage Assessment"** on a tenant role, the update fails. Unchecking **"Delete assets"** on another role works fine.

## Root cause

`Capability.validateForTenantRole()` / `validateForPlatformRole()` throw `IllegalArgumentException` if **any** capability in the submitted set is out of scope for the target role type (tenant vs platform). The role in question already holds one or more `PLATFORM`-scoped capabilities (e.g. `ACCESS_PLATFORM_SETTINGS`, `ACCESS_TENANTS`, `ACCESS_PLATFORM_USERS_GROUPS_AND_ROLES`) — likely leaked in through a front-end capability picker that didn't filter by `scope=TENANT` before this dual-scope separation existed.

Because the whole submitted `Set<Capability>` is validated atomically, **any** update to that role fails — regardless of which capability the user is actually toggling. This explains why "Delete assets" works (clean role) and "Manage Assessment" doesn't (polluted role).

## Fix (this PR)

Defense-in-depth at the persistence boundary, both directions:
- `RoleService` (tenant-scoped roles): silently filter out non-`TENANT` capabilities before persisting, with a `log.warn` when a stripped capability was present — never let it hard-block an unrelated update.
- `PlatformRoleService` (platform-scoped roles): mirror fix, filter out non-`PLATFORM` capabilities.
- Migration to purge already-persisted out-of-scope rows in `roles_capabilities` for both role kinds.

## Follow-up (not in this PR)

- Front-end root cause: the tenant role capability picker must call `GET /api/capabilities?scope=TENANT` (not the unscoped/platform+tenant list) so this can never be re-introduced from the UI. Tracked separately.

RomuDeuxfois:
## Problem

Closes #7321.

Updating a **tenant** role (e.g. unchecking "Manage Assessment") fails with `IllegalArgumentException` whenever that role already holds a **platform-only** capability, and vice-versa for platform roles. This is legacy data left over from the dual-scope split (`V5_04__Merge_roles_and_groups_dual_scope`): roles/groups became nullable-tenant without a migration cleaning up capabilities that no longer matched their role's scope.

`Capability.validateForTenantRole()` / `validateForPlatformRole()` reject the **whole payload** as soon as **one** capability is out of scope — so a polluted role can never be updated again, including by the very update that would remove the offending capability. This explains why "Manage Assessment" breaks and "Delete assets" doesn't: it's not the capability being toggled, it's whatever stale capability the role already carries.

### DIFF
--- a/openaev-model/src/main/java/io/openaev/database/model/Capability.java
+++ b/openaev-model/src/main/java/io/openaev/database/model/Capability.java
@@ -10,7 +10,9 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
 
+@Slf4j
 public enum Capability {
 
   // Superuser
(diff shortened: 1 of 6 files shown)


RomuDeuxfois:
@copilot fix CI

openaev-model/src/main/java/io/openaev/database/model/Capability.java
```diff
@@ -569,4 +596,48 @@ private static void validateScope(Set<Capability> capabilities, CapabilityScope
...
+    return filterScope(capabilities, CapabilityScope.TENANT);
+  }
+
+  /**
+   * Fail-open counterpart of {@link #validateScope}: splits the capabilities on {@code
+   * requiredScope}, returns the valid ones in a new mutable set and warns once about those dropped.
+   * Never throws, so the caller must treat the returned set - possibly empty - as the authoritative
+   * one.
+   */
+  private static Set<Capability> filterScope(
```
camrrx, line 626: Nice to have: clean up the DB to avoid having roles linked to the wrong capabilities, in case a platform capability is somehow assigned to a tenant role.

camrrx (approved):

PR #7330, opened 2026-08-11
fix(credentials): add credentials capability (#5536)

MarineLeM:
### Proposed changes

* Add "credentials" capability

### Tests :

- Existing DB startup → default roles (Observer, Manager) have credential permissions.
- New tenant creation → default roles (Observer, Manager) have credential permissions.
- New DB startup → default roles created with credential permissions.
- credential_asset flag disabled → credential permissions hidden in the capabilities form.

### Related issues

* Chunk-3 #5536


MarineLeM:
Add a migration to include the credential capability in the default roles (Observer, Manager) on all existing tenants

RomuDeuxfois (approved):

PR #7357, opened 2026-08-11
feat(health): probe dependencies in the background and expose them on Prometheus (#7356)

### LINKED ISSUE #7356: feat(health): expose pg/es/s3 used size on /api/health
### Context

`/api/health` currently only reports connectivity to its dependencies. There is no way to know, from the platform itself, how much storage each backing service actually uses.

### Proposal

Return the storage used by each dependency on `/api/health`, behind an opt-in `details=true` parameter so the default probe response (and its cost) stays unchanged:

- `pg_used_size` — total size of the PostgreSQL database
- `es_used_size` — size of the Elasticsearch/OpenSearch indexes, **primaries only** (real data volume, not the filesystem footprint inflated by replicas)
- `s3_used_size` — total size of the objects stored in the bucket

### Constraint

The endpoint is polled very frequently (load balancer probes). Computing these metrics is expensive — especially the object storage one, which has to walk the whole bucket listing — so the values must be cached and refreshed periodically rather than computed on every call.

sbocahu:
Closes #7356

## What

`GET /api/health` used to probe PostgreSQL, RabbitMQ and the object storage **synchronously on every call**, and kept its own single-flight cache for the storage sizes. Load balancers poll it far more often than dependencies change state, so a degraded dependency turned every probe into a blocked request thread.

Dependency I/O now happens in exactly one place, on a schedule. Both `/api/health` and the Prometheus scrape are pure readers of the resulting in-memory snapshot.

```
                    ┌───────────────────────┐
   @Scheduled  ───►  DependencyProbeService   ── the only component doing dependency I/O
                    └──────────┬────────────┘
                               ▼
                    ┌───────────────────────┐
                    │ DependencyHealthStore  │  in-memory snapshot
                    └──────┬────────┬───────┘
                           ▼        ▼
            /api/health (503 logic)   DependencyHealthMetrics ──► /actuator/prometheus
```

## Metrics

| Metric | Labels | Meaning |
|---|---|---|
| `openaev_dependency_up` | `dependency=postgresql\|rabbitmq\|s3\|engine` | `1` up, `0` down |
| `openaev_dependency_probe_duration_seconds` | `dependency=...` | Latency of the last probe |
| `openaev_storage_used_bytes` | `storage=postgresql\|engine\|s3` | Used size as of the last storage probe |

Never-probed values are exported as `NaN`, not `0` — "unknown" and "down" must not look alike on a dashboard.

## Schedules

| Property | Default | Probes |
|---|---|---|
| `openaev.healthcheck.connectivity-probe-interval` | `PT10S` | Connectivity of all four dependencies |
| `openaev.healthcheck.storage-probe-interval` | `PT4H` | Storage sizes (walks the whole bucket listing) |

Both run on the shared 20-thread `threadPoolTaskScheduler`, so a hanging dependency starves no other scheduled task.

## `/api/health` semantics

Unchanged response contract. It now answers **503** when a liveness dependency is down, was never probed, or was last probed more than **3 intervals** ago — a scheduler that stopped probing can no longer vouch for the platform. The 503 body names the failing dependencies.

The analytics engine now gets a **real connectivity probe** (`EngineService#ping()`, a cheap client ping rather than the index-stats call, implemented for Elasticsearch and OpenSearch). It is exported as a metric but **deliberately kept out of the 503**: an engine outage degrades analytics only and must not pull instances out of the load balancer. This resolves the "engine is never checked" gap raised on the previous revision of this PR, without changing liveness semantics.

## ⚠️ Security change: `/actuator/**` is no longer `permitAll()`

`management.endpoints.web.exposure.include` is now `prometheus` (was empty), and the scrape is served on the **public API port**. Since `/actuator/**` used to be `permitAll()`, exposing it as-is would publish JVM internals, Hikari pool saturation and dependency state to anyone.

`ActuatorSecurityConfig` adds a dedicated `@Order(1)` stateless filter chain requiring:

```
Authorization: Bearer ${openaev.metrics.key}
```

compared in constant time. **The key has no default**, so while `OPENAEV_METRICS_KEY` is unset every `/actuator/**` request is rejected with a 401 — enabling the endpoint can never silently publish the metrics.

> **Reviewers:** this is a behaviour change for any deployment that had uncommented the exposure line and scraped `/actuator/**` anonymously. Those scrapers must now send the key.

Prometheus scrape config:

```yaml
scrape_configs:
  - job_name: openaev
    metrics_path: /actuator/prometheus
    authorization:
      credentials: <OPENAEV_METRICS_KEY>
```

### Follow-up worth considering (not in this PR)

Serving the scrape on a dedicated, unpublished `management.server.port` would make the network the primary control instead of the application config. Worth noting for reviewers: Spring Boot's `ServletManagementChildContextConfiguration$ServletManagementContextSecurityConfiguration` imports the parent `springSecurityFilterChain` into the management child context, so the scrape key would keep applying on a separate port — the two controls compose rather than replace each other. Left out here because it needs a deployment change and a rework of `ActuatorSecurityTest` onto a real connector.

## Bonus fix: the SPA catch-all was swallowing `/actuator/*`

`HomeApi`'s catch-all excluded `api`, `login`, `logout`, `oauth2`, `saml2`, `assets`, `static` and `swagger-ui` — but **not `actuator`**. Any unmapped `/actuator/*` path answered `200 text/html` with the React index instead of `404`.

This is pre-existing (reproducible on `main` today) and it is what made a stale deployment indistinguishable from a working scrape during testing. It is not a data leak — it serves `index.html` — but it returns the wrong status and hides misconfiguration. Fixed by adding `actuator$` to the negative lookahead, with a regression test.

## Configuration

`openaev.healthcheck.storage-usage-cache-duration` (introduced earlier **in this same PR**, never released) is replaced by the two probe intervals above, plus the new `openaev.metrics.key`. All three documented in `docs/deployment/configuration.md`.

## Tests

- `DependencyHealthStoreTest` — liveness reasons: not probed / stale / down, and the engine exclusion.
- `DependencyProbeServiceTest` — per-dependency up/down recording, one failing dependency does not stop the others, failures without a message fall back to the exception type, storage sizes best-effort. Also carries over the lombok `@Qualifier`-copy regression guard, now that `DependencyProbeService` is the class holding the short-timeout `MinioClient`.
- `DependencyHealthMetricsTest` — `NaN` when unprobed, `1`/`0`, latency and byte sizes on a `SimpleMeterRegistry`.
- `HealthCheckServiceTest` — 503 on down / never probed / stale, 200 with the engine down.
- `ActuatorSecurityTest` — no key → 401, wrong key → 401, right key → 200 with the metrics present, and an unexposed actuator path → 404 rather than the SPA.

## Checklist

- [x] `mvn spotless:check`
- [x] `mvn compile` (`openaev-model` + `openaev-api`) and `mvn test-compile`
- [x] Unit tests green (24 health tests)
- [x] ArchUnit suite green (40 tests, no new baseline violation)
- [x] Rebased on latest `main`
- [x] Documentation updated (`docs/deployment/configuration.md`)
- [ ] **Integration tests not run locally** — `ActuatorSecurityTest` and `HealthCheckApiTest` need PostgreSQL/MinIO/Elasticsearch/RabbitMQ; relying on CI


sbocahu:
Two fixes pushed after testing on a deployed environment, where `/api/health` returned **503 on every call with no log output at all**.

### Root cause

`MinioService.isTenantPathExists()` probed the object storage with:

```java
client.statObject(... .object(getTenantPath("")) ...);   // key = "<tenant-uuid>/"
```

A tenant path is a **key prefix, not an object**. `statObject` only succeeds if a zero-byte directory marker happens to exist at that exact key — never the case on a freshly provisioned bucket, and never on plain S3, which has no directory markers at all. Result:

```
io.openaev.service.exception.HealthCheckFailureException: FileStorage check failure
  caused by io.minio.errors.ErrorResponseException: Object does not exist
```

The probe failed permanently and the load balancer pulled the instance out of rotation.

This only stayed hidden because MinIO tolerates the call as soon as *any* file exists under the prefix — which is true in dev and CI, but not in a new environment.

### Fix 1 — probe with a listing (`df88107`)

Replaced with a one-key listing on the tenant prefix. It still proves connectivity, credential validity and bucket access, but succeeds when the tenant has no files yet. Renamed to `checkTenantPathAccessible`, since a method named `isTenantPathExists` that intentionally passes when the path does *not* exist would invite someone to "fix" it straight back.

Added a regression test in `MinioServiceTest` that points `TenantContext` at a random tenant UUID (guaranteed empty) and asserts the check does not throw — it fails against the previous implementation.

### Fix 2 — make the failure observable (`5384f07`)

The failure was completely silent, which is why it took a deployment to find. Two things combined:

- `logging.level.root=fatal` and `logging.level.io.openaev=error` — nothing below ERROR is emitted.
- The 503 is a `ResponseStatusException`, which Spring resolves via `ResponseStatusExceptionResolver` at **DEBUG**. `RestBehavior` has ~25 exception handlers but none for `ResponseStatusException`, so nothing else logged it.

A health probe that can fail forever without emitting a single line is undebuggable, so the failing dependency is now logged at ERROR with its cause.

### Note for reviewers

Fix 1 changes probe semantics: the file-storage check no longer requires the tenant path to contain anything. That is intentional — the previous behaviour made "tenant has never stored a file" indistinguishable from "object storage is down". Happy to split these two commits into their own PR if you would rather keep this one purely additive.

openaev-api/src/main/java/io/openaev/rest/health_check/dto/HealthCheckDetailsOutput.java
```diff
@@ -0,0 +1,27 @@
+package io.openaev.rest.health_check.dto;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.swagger.v3.oas.annotations.media.Schema;
+
```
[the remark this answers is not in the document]
sbocahu: Fixed in d91abccf6: the DTO now lives in `io.openaev.api.health_check.dto`, matching the existing pattern where legacy `rest` controllers import their DTOs from `io.openaev.api.*` (e.g. `AtomicTestingApi`, `TenantUserApi`). The controller itself stays in `rest` as it is pre-existing.

docs/docs/deployment/configuration.md
```diff
@@ -42,6 +42,7 @@ Here are the configuration keys, for both containers (environment variables) and
...
 | openaev.admin.encryption_salt               | OPENAEV_ADMIN_ENCRYPTION_SALT               | ChangeMe              | Encryption salt used for encrypting sensitive data in database. Must be at least 8 bytes long. Encryption key and salt are used to generate a 256bit encryption key for encrypting purpose |
 | openaev.healthcheck.key                     | OPENAEV_HEALTHCHECK_KEY                     | ChangeMe              | The key to use in the health check endpoint (/api/health)                                                                                                                                  |
+| openaev.healthcheck.storage-usage-cache-duration | OPENAEV_HEALTHCHECK_STORAGE_USAGE_CACHE_DURATION | PT4H      | Validity of the storage usage metrics returned by `/api/health?details=true` (ISO-8601 duration). They are cached because computing them walks the object storage listing and queries the engine |
```
RomuDeuxfois:
```suggestion
| openaev.healthcheck.storage-usage-cache-duration | OPENAEV_HEALTHCHECK_STORAGE_USAGE_CACHE_DURATION | PT4H      |  Max age of the storage usage metrics returned by /api/health?details=true (ISO-8601 duration, e.g. PT30M). No effect on the health check itself. Once expired, the next call with details=true recomputes them synchronously, walking the whole object storage listing, so low values make some calls slow rather than refreshing anything in the background |
```

openaev-model/src/main/java/io/openaev/service/ElasticService.java
```diff
@@ -1230,6 +1230,26 @@ public ObjectMapper getObjectMapper() {
...
+          elasticClient
+              .indices()
+              .stats(s -> s.index(engineConfig.getIndexPattern()).metric("store"));
+      var all = stats.all();
+      if (all == null || all.primaries() == null || all.primaries().store() == null) {
+        return null;
+      }
+      return all.primaries().store().sizeInBytes();
+    } catch (IOException | ElasticsearchException e) {
+      log.warn("Unable to retrieve engine used size", e);
```
RomuDeuxfois: **TODO**: throw an exception here, the caller handles it correctly.

openaev-model/src/main/java/io/openaev/service/OpenSearchService.java
```diff
@@ -1298,6 +1299,33 @@ public ObjectMapper getObjectMapper() {
...
+                .path("_all")
+                .path("primaries")
+                .path("store")
+                .path("size_in_bytes");
+        if (size.isNumber()) {
+          return size.asLong();
+        }
+      }
+    } catch (IOException e) {
+      log.warn("Unable to retrieve engine used size", e);
```
RomuDeuxfois: TODO: throw an exception here, the caller handles it correctly.

openaev-model/src/main/java/io/openaev/service/MinioService.java
```diff
@@ -189,16 +190,56 @@ public void deleteDirectoryInTenantPath(String directory) {
...
+   * <p>A tenant path is only a key prefix, not an object: statting it fails whenever the tenant has
+   * no file yet, and plain S3 has no directory markers at all. A one-key listing proves access
+   * without requiring anything to be stored, and succeeds on an empty tenant.
+   *
+   * <p>Takes the client as a parameter so callers (e.g. health checks) can pass one configured with
+   * short timeouts.
    */
-  public void isTenantPathExists(MinioClient client) throws Exception {
-    client.statObject(StatObjectArgs.builder().bucket(bucket()).object(getTenantPath("")).build());
+  public void checkTenantPathAccessible(MinioClient client) throws Exception {
```
RomuDeuxfois: **todo**: rename to checkStorageAccessible since this method do much more now. Thanks for the improvment.

openaev-model/src/main/java/io/openaev/service/MinioService.java
```diff
@@ -189,16 +190,56 @@ public void deleteDirectoryInTenantPath(String directory) {
...
+   *
+   * @return the sum of the object sizes in bytes
+   */
+  public long computeUsedSize() {
+    long usedSize = 0L;
+    for (Result<Item> result : listObjects("", false)) {
+      try {
+        usedSize += result.get().size();
+      } catch (Exception e) {
+        log.warn("Error while computing the size of a stored object", e);
```
RomuDeuxfois: TODO: throw an exception here, the caller handles it correctly.

openaev-api/src/main/resources/application.properties
```diff
@@ -551,6 +551,9 @@ openaev.expectation.manual.default-score-value=50
 # a unique random string used when calling the health endpoint
 # Example of a call : /api/health?health_access_key=ChangeMe
 openaev.healthcheck.key=ChangeMe
+# validity of the storage usage metrics returned by /api/health?details=true (ISO-8601 duration).
+# Computing them walks the object storage listing and queries the engine, so they are cached.
```
RomuDeuxfois:
```suggestion
# Max age of the /api/health?details=true storage metrics (ISO-8601). On expiry the next such
# call recomputes them synchronously so low values just make calls slow.
```

openaev-api/src/main/java/io/openaev/rest/health_check/HealthCheckApi.java
```diff
@@ -68,9 +78,23 @@ public ResponseEntity<?> healthCheck(
       healthCheckService.runHealthCheck();
     } catch (HealthCheckFailureException e) {
       String message = String.format("Health check failure : %s", e.getMessage());
+      // ResponseStatusException is resolved by Spring at DEBUG, invisible at our ERROR log level:
+      // without this the probe returns 503 with no trace of which dependency actually failed.
```
RomuDeuxfois: **TODO**: remove this comment, it's not useful.

openaev-api/src/main/java/io/openaev/rest/health_check/HealthCheckApi.java
```diff
@@ -68,9 +78,23 @@ public ResponseEntity<?> healthCheck(
...
+      log.error(message, e);
       throw new ResponseStatusException(
           HttpStatusCode.valueOf(HttpStatus.SERVICE_UNAVAILABLE.value()), message);
     }
-    return new ResponseEntity<>("success", HttpStatus.OK);
+    if (!details) {
+      // Default response kept as-is: probes already rely on this plain body.
```
RomuDeuxfois: TODO: remove this comment, it's not useful.

openaev-api/src/main/java/io/openaev/rest/health_check/HealthCheckApi.java
```diff
@@ -58,7 +67,8 @@ public void setHealthCheckKey(
         @ApiResponse(responseCode = "503", description = "Service is not running properly")
       })
   public ResponseEntity<?> healthCheck(
```
RomuDeuxfois: **TODO**: returning two different types depending on the parameter isn't great.
Could we always return the new HealthCheckDetailsOutput DTO and annotate it with
@JsonInclude(NON_NULL), so the detail fields are omitted from the JSON when
details aren't requested? With that in place, ResponseEntity<?> can be tightened
to ResponseEntity<HealthCheckDetailsOutput>, and the Javadoc updated accordingly.

openaev-api/src/main/java/io/openaev/service/HealthCheckService.java
```diff
@@ -50,14 +115,69 @@ protected void runRabbitMQCheck() throws HealthCheckFailureException {
...
+
+  /**
+   * Dedicated MinIO client for probes, built once and reused.
+   *
+   * <p>It is a private instance so the short probe timeouts never mutate the client injected by
+   * Spring. It is cached because rebuilding it per probe means a fresh OkHttp connection pool — so
+   * a new TCP (and TLS) handshake on every call, and with AWS role authentication a fresh
+   * credentials lookup on IMDS/STS, which dominates the endpoint response time.
+   */
+  private MinioClient healthCheckMinioClient() {
```
RomuDeuxfois: **todo**: could be better to have a new bean in MinioDriver

@Bean
public MinioClient healthCheckMinioClient() {
  MinioClient client = getMinioClient();
  client.setTimeout(PROBE_TIMEOUT_MS, PROBE_TIMEOUT_MS, PROBE_TIMEOUT_MS);
  return client;
}

And use it here
@Qualifier("healthCheckMinioClient")
private final MinioClient healthCheckMinioClient;

openaev-model/src/main/java/io/openaev/database/repository/HealthCheckRepository.java
```diff
@@ -10,4 +10,16 @@
...
+
+  /**
+   * Total disk space used by the current PostgreSQL database, in bytes.
+   *
+   * <p>Platform-wide metric exposed by the health check endpoint: it is intentionally not
+   * tenant-scoped (there is no per-tenant physical storage in PostgreSQL, data of all tenants share
+   * the same database).
+   *
+   * @return the database size in bytes
+   */
```
RomuDeuxfois: **TODO**: drop these comments, they add nothing.

RomuDeuxfois (approved):

openaev-api/src/main/java/io/openaev/rest/health_check/HealthCheckApi.java
```diff
@@ -69,9 +79,21 @@ public ResponseEntity<?> healthCheck(
       healthCheckService.runHealthCheck();
     } catch (HealthCheckFailureException e) {
       String message = String.format("Health check failure : %s", e.getMessage());
+      log.error(message, e);
       throw new ResponseStatusException(
           HttpStatusCode.valueOf(HttpStatus.SERVICE_UNAVAILABLE.value()), message);
```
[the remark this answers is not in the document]
sbocahu: Fixed in 6ca8f8999: the failure is now logged at WARN without the stack. `DependencyProbeService` already logs the cause when the probe observes it, so this line only needs to mark that a request was answered with a 503.

openaev-api/src/main/resources/application.properties
```diff
@@ -580,12 +588,13 @@ pyroscope.profile.export.timeout=300s
...
-# Enable actuator -> disabled by default
-# once activated the api is by default open without authentication
-management.endpoints.web.exposure.include=
-#management.endpoints.web.exposure.include=metrics,prometheus
+# Actuator: only the Prometheus scrape is exposed, and it is served on the API port, so it is
+# rejected with a 401 unless openaev.metrics.key is set and sent as "Authorization: Bearer <key>".
+# Empty key (the default) denies every actuator request.
+management.endpoints.web.exposure.include=prometheus
 management.endpoint.health.show-details=always
 management.server.port=8080
```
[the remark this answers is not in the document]
sbocahu: Fixed in 6ca8f8999: `management.server.port` is now `${server.port}`, so the scrape follows the API port that `ActuatorSecurityConfig` guards. A hardened deployment can still override it to a dedicated, unpublished port.

openaev-api/src/main/java/io/openaev/health/DependencyHealthMetrics.java
```diff
@@ -0,0 +1,76 @@
...
+  private final MeterRegistry meterRegistry;
+  private final DependencyHealthStore store;
+
+  @PostConstruct
+  public void registerGauges() {
+    for (PlatformDependency dependency : PlatformDependency.values()) {
+      Gauge.builder("openaev.dependency.up", store, s -> upValue(s, dependency))
+          .tag(DEPENDENCY_TAG, dependency.getLabel())
+          .description("1 when the last connectivity probe of the dependency succeeded, 0 when not")
+          .register(meterRegistry);
```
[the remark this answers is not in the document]
sbocahu: Fixed in 6ca8f8999: the description now reads `1 when the last connectivity probe of the dependency succeeded, 0 when it failed, NaN when it has never been probed`.

openaev-api/src/test/java/io/openaev/service/MinioServiceTest.java
```diff
@@ -142,4 +143,20 @@ void should_delete_directory_in_tenant_path() throws Exception {
...
+    // -- ARRANGE --
+    String previousTenant = TenantContext.getCurrentTenant();
+    TenantContext.setCurrentTenant(UUID.randomUUID().toString());
+
+    // -- ACT & ASSERT --
+    try {
+      assertDoesNotThrow(() -> minioService.checkStorageAccessible());
+    } finally {
+      TenantContext.setCurrentTenant(previousTenant);
+    }
```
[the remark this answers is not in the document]
sbocahu: Fixed in 6ca8f8999: the test now reads `TenantContext.hasCurrentTenant()` first and calls `clearCurrentTenant()` when no tenant was set, instead of restoring the default-tenant fallback.

docs/docs/deployment/configuration.md
```diff
@@ -42,6 +42,9 @@ Here are the configuration keys, for both containers (environment variables) and
...
+| openaev.healthcheck.connectivity-probe-interval | OPENAEV_HEALTHCHECK_CONNECTIVITY-PROBE-INTERVAL | PT10S         | Interval between two background connectivity probes of the dependencies (ISO-8601 duration). `/api/health` serves the result of the last probe and never contacts a dependency itself; it answers 503 once a required dependency (database, RabbitMQ, file storage) is down or has not been probed for 3 intervals |
+| openaev.healthcheck.storage-probe-interval  | OPENAEV_HEALTHCHECK_STORAGE-PROBE-INTERVAL  | PT4H                  | Interval between two background computations of the storage sizes returned by `/api/health?details=true` and exported as `openaev_storage_used_bytes` (ISO-8601 duration). Computing them walks the whole object storage listing and queries the engine cluster, so low values are costly |
+| openaev.metrics.key                         | OPENAEV_METRICS_KEY                         | *empty*               | Scrape key for `/actuator/prometheus`, sent by the scraper as `Authorization: Bearer <key>`. While empty (the default), every `/actuator/**` request is rejected with a 401                 |
```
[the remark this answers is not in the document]
sbocahu: Partly fixed in 6ca8f8999.

The env var names were indeed wrong: they now use underscores (`OPENAEV_HEALTHCHECK_CONNECTIVITY_PROBE_INTERVAL`, `OPENAEV_HEALTHCHECK_STORAGE_PROBE_INTERVAL`), matching Spring relaxed binding.

The `<key>` escaping is not applied: it sits inside a code span, and Markdown neither parses HTML nor decodes entities there, so `<key>` is already safe while `&lt;key&gt;` would render literally as `&lt;key&gt;`.

openaev-api/src/main/java/io/openaev/config/ActuatorSecurityConfig.java
```diff
@@ -0,0 +1,101 @@
...
+  private static final String BEARER_PREFIX = "Bearer ";
+
+  @Value("${openaev.metrics.key:}")
+  private String metricsKey;
+
+  @Bean
+  @Order(1)
+  public SecurityFilterChain actuatorFilterChain(HttpSecurity http) throws Exception {
+    http.securityMatcher(ACTUATOR_URI)
+        .csrf(AbstractHttpConfigurer::disable)
```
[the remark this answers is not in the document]
sbocahu: Stale alert: this was flagged on a revision where the actuator chain called `csrf(AbstractHttpConfigurer::disable)`. CSRF has been left enabled since d53a77653 — a scrape is a GET, which Spring Security exempts anyway, so disabling it would only widen the surface if a writable endpoint were ever exposed on this chain. The alert should clear on the next CodeQL run.

openaev-api/src/main/java/io/openaev/health/DependencyProbeService.java
```diff
@@ -0,0 +1,108 @@
...
+
+  private final RabbitmqService rabbitmqService;
+  private final EngineService engineService;
+
+  /**
+   * Connectivity of every dependency. Each probe is a single round trip on a client configured with
+   * short timeouts, so the whole pass stays bounded even when a dependency is down.
+   */
+  @Scheduled(fixedDelayString = "${openaev.healthcheck.connectivity-probe-interval:PT10S}")
+  public void probeConnectivity() {
```
laugiov: Hello Sebastien,
The four probes share one `@Scheduled(fixedDelay)` pass, and `fixedDelay` never
overlaps. So while `ping()` (line 54) waits on the engine, the other three are
not refreshed.

Engine socket timeout is 60s (`EngineConfig.java:107`), staleness budget is 30s
(`HealthCheckService.java:28`). A hung engine makes `postgresql`, `rabbitmq` and
`s3` go stale, and `/api/health` returns 503. `PlatformDependency:13-17` says
this must never happen.

Could each dependency get its own `@Scheduled`, like the storage probe already
has?

sbocahu: Hello Laurent, good catch — this is real, and I verified the whole chain: `fixedDelay` never overlaps a task with itself, the engine socket timeout is 60s (`EngineConfig:107`) and the staleness budget is 30s (`PT10S` × `STALE_PROBE_INTERVALS = 3`). So a hung `ping()` did stale out `postgresql`, `rabbitmq` and `s3` and 503 an otherwise healthy instance — exactly what `PlatformDependency:13-17` forbids.

Fixed in 3084473fb: each dependency now has its own `@Scheduled`, as you suggested. They all stay on `connectivity-probe-interval`, so the staleness threshold and the configuration surface are unchanged; only the coupling is removed. Added a test that probing RabbitMQ alone leaves the other three untouched, so the single-pass version cannot come back unnoticed.

One thing your comment made me notice: the class javadoc claimed every probe ran "on a client configured with short timeouts". That is true for RabbitMQ and MinIO, but false for the engine, which reuses the shared client — the javadoc now says so explicitly. Giving the engine its own fast-fail client would mean touching both `ElasticService` and `OpenSearchService`; happy to open a follow-up issue if you think it is worth it, since the engine gauge can still lag by a socket timeout even though it gates nothing.

laugiov: Thanks for the change @sbocahu !
Fix looks right :)

On the engine client: I would not bother. 
Now that nothing is coupled to it, the only cost is that the engine gauge can be up to one socket timeout late.
That is the same order as a Prometheus scrape interval, and an alert on it would need a longer for: window anyway. Not worth a second client in both services.

laugiov (approved):

PR #7362, opened 2026-08-12
fix(chaining): resolve target dns resolution for attack path

EvaE-Filigran:
### Proposed changes

* Resolve `#{var}` argument template to its actual value before displaying the DNS resolution target on the map

### Testing Instructions

1. Create a scenario with chaining, set a DNS resolution payload's hostname from an argument (e.g. `#{var}`)
2. Launch the simulation
3. Check the map displays the resolved hostname value instead of `#{var}`

### Related issues

* #5048

### Checklist


impolitepanda:
tested OK !
image

impolitepanda (approved):

PR #7373, opened 2026-08-12
fix(capabilities): order the capability tree from CapabilityGroup

johanah29:
### Proposed changes

* The group order of the roles capability tree followed the `Capability` declaration order, because the roots were grouped as they came. The roots are now sorted on their group, so the `CapabilityGroup` declaration order is the single source of truth — reordered to put Dashboards / Reportings / Findings first.
* The tree no longer goes through the normalized store on the front side: the `referential` entity maps are Immutable maps and do not preserve the API order. `useCapabilities` reads `/api/capabilities` directly and hands over the backend order untouched.

### Testing Instructions

1. Settings > Security > Roles, create or open a role: the groups follow the `CapabilityGroup` declaration order (Bypass, Dashboards, Reportings, Findings, Assessment, ...), for platform roles as well as tenant roles.
2. Same order on the role detail page, and after a tenant switch.

### Related issues

*

### Checklist

### Further comments

Rebuilding the order on the front, from the id list returned by normalizr, was the other option. It was dropped because it stores a property of the data in component state and only works around the store, while the capability tree is static reference data read by three forms — cheaper to keep it out of the referential entirely.

`CapabilityTreeBuilderTest` now asserts the root order for every scope; both new assertions fail without the sort.


openaev-api/src/main/java/io/openaev/api/capabilities/CapabilityTreeBuilder.java
```diff
@@ -12,6 +12,23 @@
  */
 public final class CapabilityTreeBuilder {
 
+  private static final List<CapabilityGroup> UI_GROUP_ORDER =
```
RomuDeuxfois: **todo**: why not reorganize CapabilityGroup and use the direct order from there ?

johanah29: Only reorganizing CapabilityGroup does not work. We need to apply an explicit sort

RomuDeuxfois: I agree, but why your mechanism is working on UI_GROUP_ORDER  and not direclty with CapabilityGroup ?

PR #7375, opened 2026-08-12
fix(injection): unify value normalisation in CommandArgumentBinder (#7366)

### LINKED ISSUE #7366: fix(command): unify value normalisation and strengthen tests in CommandArgumentBinder
`CommandArgumentBinder` renders argument values differently depending on the target engine, and the differences are not intentional.

**Control characters are normalised inconsistently.** The cmd path replaces carriage returns and line feeds with spaces, the sh and powershell paths keep them, and literal mode keeps everything. `sanitize` drops other control characters for all engines. The handling should be uniform and decided once, not per engine.

**Some values cannot be represented safely in every declaration form.** The cmd declaration relies on a surrounding quoted form that not every value can sit inside. The binder should validate that a value is representable for the selected engine and fail explicitly when it is not, rather than rendering something the engine will parse differently than intended.

**Tests assert the wrong property.** `CommandArgumentBinderTest` checks the shape of the rendered command: its prefix, its suffix, the absence of newlines. It does not assert that a value stayed inside its own declaration, which is the property that actually matters. The parameterised fixtures already include values that would exercise this, so the assertions are what need work.

Scope: `openaev-api/src/main/java/io/openaev/utils/command/CommandArgumentBinder.java` and its test.

laugiov:
Closes #7366.

`CommandArgumentBinder` handled argument values differently depending on the target engine, and the differences were not intentional. This makes the handling uniform and gives the tests the ability to see it.

## Behaviour changes

- **Line separators are removed from every argument value, on every engine.** Previously cmd replaced CR and LF with a space, sh and powershell kept them, and literal mode kept everything. Now `sanitize` removes CR, LF and the Unicode line separators (U+0085, U+2028, U+2029) once, for all engines. Tab is deliberately kept: it separates nothing in any supported shell.
- **A cmd argument value containing a double quote is now refused**, with an error naming the argument. The `set "VAR=value"` form has no escape for a quote inside it, so such a value could not be declared as written. Refusing is verifiable; guessing at an escape is not. Other engines are unchanged, they quote with `'` and can escape it.
- **Placeholder substitution now reads the template once.** It previously ran one pass per bound argument over the already-substituted output, so a value could be re-read as a placeholder naming another argument, and the result depended on the order arguments happened to be bound.

## Tests

The suite goes from 55 to 152 cases.

- A property test per engine over 20 000 generated values, drawn from an alphabet of every metacharacter, escape, quote, separator and line terminator the engines give meaning to. The seed is fixed, so a failure names a value that reproduces exactly.
- Containment is asserted instead of shape: the assertions check that a value stayed inside its own declaration and that the command part is untouched, rather than checking the rendered string looks a certain way.
- Both directions are pinned. Refusing everything would satisfy a containment property trivially, so acceptance is asserted too.
- The characters kept on purpose (tab, byte order mark) have their own tests, so the decision is visible rather than accidental.

## Documentation

The class javadoc claimed a placeholder inside a wider quoted string was left unsubstituted. It is substituted, and the reference then lands inside quotes the binder does not control. Corrected, and pinned by a test.

### DIFF
--- a/openaev-api/src/main/java/io/openaev/utils/command/CommandArgumentBinder.java
+++ b/openaev-api/src/main/java/io/openaev/utils/command/CommandArgumentBinder.java
@@ -25,10 +25,13 @@ import lombok.extern.slf4j.Slf4j;
  *   <li>cmd — {@code set "OAEV_ARG_TARGET=a; whoami" & echo "!OAEV_ARG_TARGET!"}
  * </ul>
  *
- * <p><b>Template authoring rule</b>: do not wrap a placeholder in quotes yourself ({@code
- * "#{arg}"}). The binder owns the quoting; directly adjacent quotes around a placeholder are
- * detected and removed, but a placeholder embedded in a wider single-quoted literal cannot be
- * resolved (it stays literal — safe, but not substituted).
+ * <p><b>Template authoring rule</b>: do not quote a placeholder yourself ({@code "#{arg}"}), and do
+ * not put one inside a wider quoted string. The binder owns the quoting. Quotes directly wrapping a
+ * placeholder are detected and removed, but a placeholder sitting inside a wider quoted string is
+ * still substituted, and the reference then lands inside quotes the binder does not control. The
+ * command stays structurally safe, yet it does not do what the template says: under {@code sh} the
+ * reference is single-quoted and never expands, so the command handles the variable name instead of
+ * the value.
  *
  * <p>Not thread-safe: create one instance per rendered command.
  */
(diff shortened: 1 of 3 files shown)


Dimfacion:
Seems to be working well and tested OK on linux and windows. Just a small remark but when there is an exception due to the use of a double quote, we do have the error on the logs (which is what we want :
image
However, from a user perspective, I have no idea why it crashed and how to fix things :
image
Maybe it would be nice to add a trace in the error of the payload explaining that an argument cannot be represented ?

PR #7377, opened 2026-08-12
feat(logs): log tenant access denials with caller and request context

RomuDeuxfois:
## Why

When there is a failure to register from the injector caused by tenants logic, we are not able to identify the issue. This PR is adding logs when the TENANT_ACCESS_DENIED is triggered that could help us to investigates these kind of issues


heditar:
tested

heditar (approved):

Issue #5045, opened 2026-03-02
feat(chaining): logic creation


mergenhan:
As a Scenario Builder, 
I want to match an event onto an action 
so that it can triggers visually.

PR #7395, opened 2026-08-13
fix(chaining): update targets on action when adding an asset (#5045)

camrrx:
### Proposed changes

*Adding an asset in the allowlist after configuring the logic is not updating the “targets” so the action will not be triggered. 

### Testing Instructions

2. First add an action 
3. Then add an asset in the allowlist
4. The action should updat the "targets" on the logic map (frontend) and on the steps table in step_data column (backend)

### Related issues

* Related to #5045

### Checklist

### Further comments


savacano28 (approved):
Tested ok:
Test: add action and later add scope rule: target
Image
Image

Test: Remove target 
Image
Image

PR #7397, opened 2026-08-13
fix(connectors): keep a deleted connector deleted while its container still runs

RomuDeuxfois:
## Problem

Deleting a connector managed by the Integration Manager did not stick: the row came back a few seconds later, orphaned this time, and could not be removed for as long as its container lived.

The deletion only consulted the owning instance status, and that status is an *intention*, not a fact — `requestedStatus` is set to `stopping` the moment the user clicks stop, while the composer tears the container down asynchronously. So the container was usually still alive when the delete went through, and its next registration heartbeat (every ~40s, and the endpoint is an upsert) recreated the connector row. The owning instance had been deleted along with it, so what came back was an unmanaged connector — which is also why the UI then offers "Migrate" on something the user believed deleted.

### DIFF
--- a/openaev-api/src/main/java/io/openaev/service/connector_instances/ConnectorInstanceService.java
+++ b/openaev-api/src/main/java/io/openaev/service/connector_instances/ConnectorInstanceService.java
@@ -515,11 +515,14 @@ public class ConnectorInstanceService {
         HeartbeatWindow.forInstance(instance, instance.getCatalogConnector().getContainerType());
     if (lastHeartbeat != null && lastHeartbeat.isAfter(Instant.now().minus(heartbeatWindow))) {
       throw new BadRequestException(
-          "The connector "
+          "The "
+              + instance.getCatalogConnector().getContainerType().name().toLowerCase(Locale.ROOT)
+              + " "
               + connectorId
               + " is still running (last heartbeat "
               + lastHeartbeat
-              + "): stop it before deleting its instance");
+              + "): stop it before deleting its instance "
+              + instance.getId());
     }
   }
 
--- a/openaev-api/src/main/java/io/openaev/service/connector_instances/ConnectorInstanceService.java
+++ b/openaev-api/src/main/java/io/openaev/service/connector_instances/ConnectorInstanceService.java
@@ -21,10 +21,13 @@ import io.openaev.rest.connector_instance.dto.CreateConnectorInstanceInput;
 import io.openaev.rest.exception.BadRequestException;
 import io.openaev.service.EndpointService;
 import io.openaev.service.connectors.ConnectorOrchestrationService;
+import io.openaev.service.connectors.HeartbeatWindow;
 import io.openaev.service.exception.ConnectorStatusException;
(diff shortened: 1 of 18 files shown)


RomuDeuxfois:
@copilot fix

openaev-api/src/main/java/io/openaev/service/connector_instances/ConnectorInstanceService.java
```diff
@@ -458,6 +451,53 @@ public void deleteById(String id) throws ConnectorStatusException {
...
+   */
+  private void throwIfConnectorStillPinging(
+      ConnectorInstancePersisted instance, String connectorId) {
+    if (connectorId == null) {
+      return;
+    }
+    ConnectorCompositeId compositeId =
+        ConnectorCompositeId.of(connectorId, instance.getTenant().getId());
+    Instant lastHeartbeat =
+        switch (instance.getCatalogConnector().getContainerType()) {
```
damgouj, line 505: suggestion: maybe it will change here because injectors V2 tenants PR is going to be merged

openaev-api/src/test/java/io/openaev/rest/InjectorApiTest.java
```diff
@@ -803,4 +808,49 @@ void given_missingInjectorImage_should_return404() throws Exception {
       mvc.perform(get(INJECT0R_URI + "/nonexistent-type/image")).andExpect(status().isNotFound());
     }
   }
+
+  @Nested
+  @DisplayName("Deleting a managed injector")
```
damgouj, line 797: suggestion: maybe it will change here because injectors V2 tenants PR is going to be merged

openaev-front/src/admin/components/integrations/common/ConnectorPopover.tsx
```diff
@@ -56,12 +57,16 @@ const ConnectorPopover = ({ connectorInstanceId, connectorName, disabled = false
...
-  const canDeleteInstance = instance?.connector_instance_requested_status === 'stopping'
+  // Deletable once nothing can bring it back: a stop asked for or done, and no recent heartbeat.
+  // The stop is only ever requested, so the container may well still be registering. Same rule
+  // server-side.
+  const stopRequestedOrEffective = instance?.connector_instance_requested_status === 'stopping'
     || instance?.connector_instance_current_status === 'stopped';
+  const stillPinging = connector?.isExternal === true && isHeartbeatFresh(connector?.updatedAt);
+  const canDeleteInstance = stopRequestedOrEffective && !stillPinging;
+  const deleteDisabledMessage = stopRequestedOrEffective
+    ? 'Wait until the connector stops pinging before deleting it'
```
damgouj, line 68: question: the user needs to wait what ? Not sure it's clear enough ;)

damgouj (changes requested):
I can test it after the rebase if you want

RomuDeuxfois:
@damgouj PR available for a second round.

damgouj (approved):

Issue #5049, opened 2026-03-02
feat(chaining): event creation


mergenhan:
As a Scenario Builder, 
I want to create an Event 
So that actions triggered automatically when conditions match.

PR #7406, opened 2026-08-13
fix(chaining): fix event filtering (#5049)

heditar:
### Proposed changes

* ConditionUtils.matchesAnyLeafCondition was not checking the type of the value which was causing some issue on events using multiple type

### Testing Instructions

1. Create a scenario with one action taking an input (with local scope) and link the input port type
2. Gate the action created by an event with two conditions: port = 445 and host. is not null
3. In the scope variable add 3 values port:5 port:445 host:localhost
4. Run the scenario and check the acvtion is not executed on other port that your default value and port 445

Here is a test scenario:
[test 3_2026-08-13T14_45_48.076613Z_(with_variable_values & with_scope_definition).zip](https://github.com/user-attachments/files/31033666/test.3_2026-08-13T14_45_48.076613Z_.with_variable_values.with_scope_definition.zip)

### Checklist

### Further comments


EvaE-Filigran:
Test fail i still have bad executions, when an event should filter "host:localhost and port 445" as local only, I have result from global pool and mix of value from local pool that should not be executed.

heditar:
Clarified the expectation with @EvaE-Filigran 
- if 3 ports in scope: 445,8000,8080
- an action that echo the port is filter by an event with a condition port=445, if the variable in the action is local it should execute once with port 445 if it is global it should execute with the three port


This PR fix an issue when if an event had a condition port=445 and "host is not null"  then the action was executed for the three ports with global (as expected) and local (not expected)

impolitepanda (approved):
Tested OK for my part.

However, I believe @EvaE-Filigran 's remark are pertinent and we should study it a bit more next week.

So far, I don't think it's a blocker for release.

Issue #37, opened 2019-04-23
Be able to deploy OpenEx without using Composer


SamuelHassine:
## Problem to Solve

In some case, command line is not available on the PHP web hosting environment or packages are not able to be downloaded.

## Current Workaround

None.

## Proposed Solution

* Create a full package with dependencies

## Additional Information

None.

Issue #36, opened 2019-04-23
Administrator toggle field is not working


SamuelHassine:
## Description

The "Administrator" toggle field is not working in the admin interface.

## Environment

N/A

## Reproducible Steps

1. Edit a user on the admin interface
2. Try to switch him to administrator

## Expected Output

The user is administrator.

## Actual Output

Nothing.
 
## Additional information

None.

PR #7427, opened 2026-08-14
feat(paper): adopt the library Paper and its header in detail surfaces

sandyghs:
## What changes

Container surfaces move from the MUI `Paper` to the design-system `Paper`, and
the three titled wrappers every detail screen is built from now use the
library's own header instead of drawing their own.

- **13 container surfaces** converted across the lessons screens.
- **`Section`, `InformationGrid`, `SectionBlock`** pass their title to the
  library's `title` slot — 106 usages in 33 files. Their container becomes a
  one-row grid so the panel fills its row.
- **Login panel** moves to elevation layer 1, and a customer theme now colours
  the surface border as well as its background.
- **A list's last row** no longer draws a divider on top of the Paper's own
  border.
- **A render guard** ships with the conversion: 11 tests pinning the header
  wiring, which neither of the two conformity checks can see.

## Why

The product drew its own section headers — different font, size, weight, casing
and letter-spacing from the library's. Adopting the library slot removes that
divergence at the source rather than in 106 call sites, and makes every later
header change a library change.

## Where to read the detail

Everything measured lives in the migration record, versioned in this PR:

- `fds-migration/PAPER-GAP-INVENTORY.md` — §13 for this wave (typography table,
  the flex trap, the column fix, the accepted truncation gap), §14 for the
  census rule that keeps floating surfaces out of the perimeter.
- `fds-migration/LIBRARY-FEEDBACK.md` — the two gaps opened upstream: the header
  renders on the presence of `title` rather than its value (#36), and a
  truncated title offers no way to read the full text (#37).

## Known and accepted

Every visual the conversion gives up, with its site. Each was arbitrated; none
is a defect to be reported back.

- **A title truncates.** One title out of fourteen, at the narrowest track a
  `DetailSections` can produce, loses 16px to an ellipsis. The full text stays
  in the accessibility tree; the library exposing it on hover is upstream #37.
  Site: `components/common/detail/EntityDetailCommon.tsx`.
- **A vertical mission gradient is dropped.** The library Paper paints no
  gradient. The accent border colour stays, and it is what marks plan mode.
  Site: `admin/components/autonomous/AutonomousOutcome.tsx`.
- **A sheen gradient and a local radius are dropped.** The surface now carries
  the library's radius; everything else on the card is unchanged. Site:
  `admin/components/workspaces/custom_dashboards/CustomDashboardReactLayout.tsx`.
- **A hover shadow is dropped.** The library draws no shadow. The lift and the
  tinted border stay, so the affordance survives. Site:
  `admin/components/getting_started/GettingStartedScenarios.tsx`.
- **The login panel GAINS a border.** Not a loss but the same kind of decision:
  moving the panel to elevation layer 1 puts it in the same family as every
  other panel in the product, and the border is what that family draws. Site:
  `public/components/login/Login.tsx`.
- **One `padding={0}` does not render 0.** A global unlayered product class,
  `.paper` in `static/css/index.css`, carries `padding: 20px` and beats the
  library's layered utility whatever the order. The surface renders the 20px it
  rendered before, so it is iso — but by accident, and the prop is the loser of
  that cascade. Measured and recorded in PAPER-GAP-INVENTORY §5.9. Site:
  `admin/components/integrations/injectors/InjectorPage.tsx`.

## Not in this PR

`DetailHero` and the clickable cards keep their MUI surface: their surface is a
real router link, so converting would break cmd-click. They wait for the
library's `Card`.


openaev-front/src/admin/components/lessons/scenarios/Lessons.tsx
```diff
@@ -214,11 +215,12 @@ const Lessons: FunctionComponent<Props> = ({
           </Can>
         </header>
         {lessonsCategories.length === 0 ? (
```
RomuDeuxfois: not ideal

Image

Current https://main-ff.oaev.staging.filigran.io/2cffad3a-0001-4078-b0e2-ef74274022c3/admin/scenarios/ba7a1cad-cffc-49b5-9866-085cfe59459b/lessons

openaev-front/src/admin/components/lessons/simulations/LessonsCategories.jsx
```diff
@@ -137,15 +138,21 @@ const LessonsCategories = ({
               }}
               >
                 <Typography sx={SECTION_LABEL_SX}>{t('Questions')}</Typography>
+                {/* padding=0 (iso): row dividers stay edge to edge — PAPER-GAP-INVENTORY §6.1. */}
```
RomuDeuxfois: seems weird to me but it's just a feeling

Image

openaev-front/src/public/components/login/Login.tsx
```diff
@@ -115,12 +115,20 @@ const Login = () => {
         {consentOk && (
           <>
             {isLocal && !reset && (
+              // The login panel is an elevation-1 surface, like every other
```
RomuDeuxfois, line 112: Are we ok with this new sceen ?

Image

what we have now

Image

RomuDeuxfois:
Tests:
- this divider is new, is that normal ?
image
- right menu is darker
image

openaev-front/src/public/components/login/Login.tsx
sandyghs: Intentional — it was painting an off-scale colour in dark mode, it now uses the token like every other surface.
The two lines under the logo are the configured login message (Settings → Policies), so they come from the DB, not from this PR — I think your two screenshots are from different instances. 
What exactly bothers you there?

openaev-front/src/admin/components/lessons/simulations/LessonsCategories.jsx
sandyghs: Code or render? If it's the JSX comment pointing at a doc section, happy to change how we write those.

sandyghs:
> Tests:
> 
> * this divider is new, is that normal ?
> 
> image - right menu is darker image

- New divider is from the new header and yes it's normal :)
- The submenus get handled properly in phase 2, with the secondary navbars.

openaev-front/src/admin/components/lessons/scenarios/Lessons.tsx
sandyghs: Good catch! that was a miss on our side. Fixed.

openaev-front/src/admin/components/lessons/simulations/LessonsCategories.jsx
RomuDeuxfois: The rendering with the colors. But I can't explain why.. Just a feeling.

RomuDeuxfois (approved):

PR #7428, opened 2026-08-14
feat(rbac): extract tenant users, groups and roles as an independent capability triad (#7196)

### LINKED ISSUE #7196: feat(scope): US.1 — Extract tenant "users, groups & roles" management as an independent capability triad
## Use case

MANAGE_TENANT_SETTINGS currently bundles tenant user, group, and role management together with unrelated administrative powers (collectors, injectors, tag rules, attack patterns, tenant settings generally). This story extracts tenant user/group/role management into its own independently-grantable capability triad, mirroring the pattern already used on the platform side ("Access / Manage / Delete platform users, groups and roles").

**AC1: Given** the current `MANAGE_TENANT_SETTINGS` capability bundles tenant user/group/role management together with other tenant-admin powers

**When** this story is implemented

**Then** three new independent capabilities exist: "Access tenant users, groups and roles," "Manage tenant users, groups and roles," "Delete tenant users, groups and roles" — no longer reachable via `MANAGE_TENANT_SETTINGS` alone

**AC2: Given** a tenant role that previously held `MANAGE_TENANT_SETTINGS` only, to reach user/group/role management

**When** this change ships

**Then** that role no longer grants user/group/role management unless it's explicitly given one of the new triad capabilities — no silent capability loss or gain, this must be called out as a breaking change requiring a migration note for any existing role relying on the old bundled behavior

**AC3: Given** the new triad capabilities

**When** they are added to `Capability.java`

**Then** they follow the same structural pattern as the confirmed platform-side triad (naming, scope declaration, `checkable` behavior) — no ad hoc/divergent shape

**AC4: Given** `MANAGE_TENANT_SETTINGS` after this change

**When** inspected

**Then** it retains only the capabilities unrelated to user/group/role management (collectors, injectors, tag rules, attack patterns, tenant settings) — user/group/role management is fully removed from it, not duplicated in both places

### LINKED ISSUE #7399: feat(rbac): migrate existing tenant roles to the new users/groups/roles capability triad
## Context
Splitting tenant user/group/role management out of TENANT_SETTINGS is a breaking change for any existing tenant role that relied on ACCESS/MANAGE/DELETE_TENANT_SETTINGS to reach user/group/role management. This issue ships a data migration that grants the equivalent new triad capability to every role that already holds the corresponding TENANT_SETTINGS capability, so no tenant silently loses (or gains) user/group/role management access.

## Acceptance Criteria
- [ ] A Flyway migration inserts the new triad capability into `roles_capabilities` for every existing tenant role that already holds the corresponding TENANT_SETTINGS-tier capability, following the same 3-tier cascade pattern (ACCESS→ACCESS, MANAGE→ACCESS+MANAGE, DELETE→ACCESS+MANAGE+DELETE)
- [ ] The migration only targets tenant-scoped roles (`tenant_id IS NOT NULL`), never platform roles
- [ ] The migration is idempotent (`ON CONFLICT ... DO NOTHING` or equivalent) and safe to run on a database that already has the new capabilities granted
- [ ] A migration note documenting this breaking change is added to `docs/docs/administration/users-and-rbac.md`

### Red test (write first)
- `MigrationTest#roleWithOnlyAccessTenantSettingsGainsOnlyAccessTriad` — before: a seeded tenant role holding only `ACCESS_TENANT_SETTINGS` has zero rows in `roles_capabilities` for any of the three new triad capabilities; after: the migration runs and the role has exactly `ACCESS_TENANT_USERS_GROUPS_AND_ROLES` added, and `MANAGE_.../DELETE_...` are still absent. Covers AC1.
- `MigrationTest#platformRoleWithSettingsTierCapabilityUntouched` — before: a seeded platform role (`tenant_id IS NULL`) holding a platform-tier settings capability has zero rows for the new tenant triad capabilities; after: running the migration, the platform role still has zero rows for the new triad — no rows were inserted for it. Covers AC2.
- `MigrationTest#migrationIsIdempotent` — before: running the migration once on a role with `MANAGE_TENANT_SETTINGS` produces exactly 2 new `roles_capabilities` rows (ACCESS+MANAGE of the new triad); after: running the migration a second time on the same database produces 0 additional rows (row count for that role/capability pair stays at 2, no duplicates, no constraint violation). Covers AC3.

### Additional test cases
- Migration test: a tenant role with `MANAGE_TENANT_SETTINGS` ends up with `ACCESS_TENANT_USERS_GROUPS_AND_ROLES` + `MANAGE_TENANT_USERS_GROUPS_AND_ROLES`
- Migration test: a tenant role with `DELETE_TENANT_SETTINGS` ends up with all three new triad capabilities
- Migration test: a platform role with a platform-tier settings capability is untouched
- Migration test: running the migration twice does not create duplicate `roles_capabilities` rows
- Migration test: a tenant role with none of the TENANT_SETTINGS capabilities receives none of the new triad capabilities

### Sensitive Path considerations
- **Cross-tenant isolation test to add:** the migration operates on `roles_capabilities` filtered by `tenant_id IS NOT NULL`, but does not otherwise scope by a specific tenant — it must run once, globally, across all tenants. Add a test seeding two tenants (A and B) each with a role holding `MANAGE_TENANT_SETTINGS`, and assert after migration that tenant A's role only gains the triad capability on tenant A's own `roles_capabilities` row (correct `tenant_id`/`role_id` pairing) — i.e. the migration must never cross-assign a tenant-B role's new capability grant to a tenant-A role or vice versa.
- **Fail-open vs fail-closed:** this issue is a one-time data migration, not a runtime resolution checkpoint — "fail-open/closed" in the request-handling sense doesn't directly apply. The relevant equivalent here: the migration must fail the deployment (migration failure blocks startup, per Flyway's default behavior) rather than partially apply and silently leave some roles ungranted. Confirmed as the expected Flyway default; no override to a "continue on error" mode should be introduced.
- **Audit-logging point:** open question for the Implementer — this migration performs bulk grants of a sensitive capability (user/group/role management access) outside the normal role-edit UI/API path. Confirm whether the existing audit-log mechanism captures direct SQL/migration-driven `roles_capabilities` inserts, or only inserts made through the application's role-update service layer. If the audit log only hooks the service layer, this migration's grants will be invisible to audit history — flag this gap explicitly to the Implementer/reviewer rather than assuming coverage.

## Definition of Done
- [ ] Migration merged, applies cleanly on top of the previous schema version
- [ ] Verified against a database snapshot containing pre-existing tenant roles with each of the three TENANT_SETTINGS tiers
- [ ] Documentation updated with the breaking-change/migration note

### LINKED ISSUE #7400: feat(rbac): gate tenant users, groups and roles UI on the new capability triad
## Context
The tenant users, groups, and roles admin screens (components across `users/tenant_users`, `groups/tenant_groups`, and `roles/tenant_roles`) currently gate their create/edit/delete actions on the `TENANT_SETTINGS` CASL subject. This issue rewires that gating to the new dedicated capability triad so the frontend permission model matches the backend split.

## Acceptance Criteria
- [ ] A new `TENANT_USERS_GROUPS_AND_ROLES` entry is added to `SUBJECTS` in `permissions/types.ts`
- [ ] All `Can`/`ability.can(...)` checks currently using `SUBJECTS.TENANT_SETTINGS` for tenant user, group, or role screens (create buttons, popovers, update/delete actions) are switched to the new subject
- [ ] Group-role/user-group management screens (group manage roles/users, role forms, etc.) use the new subject consistently with their backend-enforced capability
- [ ] `MANAGE_TENANT_SETTINGS`-gated actions unrelated to users/groups/roles (e.g. collectors, injectors, tag rules) are left untouched
- [ ] i18n labels for the three new capabilities are added to every locale file under `utils/lang/`

### Red test (write first)
*Depends on #7402 (new subject/capability must exist on the backend/permission model before the frontend can gate on it).*
- `TenantUsersList.test.tsx#hidesCreateButtonForSettingsOnlyUser` — before: a user with only `MANAGE_TENANT_SETTINGS` (no new triad capability) currently SEES the create-user button, because the component still gates on `SUBJECTS.TENANT_SETTINGS` (this is the real current behavior — the test asserting the button is hidden must fail against today's code); after rewiring: the same user no longer sees the create-user button, only a user with `MANAGE_TENANT_USERS_GROUPS_AND_ROLES` does. Covers AC2 and AC4.
- `TenantRolesList.test.tsx#gatesDeleteOnNewDeleteCapability` — before: a user with `DELETE_TENANT_SETTINGS` but not `DELETE_TENANT_USERS_GROUPS_AND_ROLES` currently sees the delete-role affordance (current behavior); after: delete affordance is hidden for that user and only shown when `DELETE_TENANT_USERS_GROUPS_AND_ROLES` is held. Covers AC2/AC3.
- `i18n-lint.test.ts#tenantUsersGroupsAndRolesLabelsPresentInAllLocales` — before: locale files under `utils/lang/` have no key for the three new capability labels (lookup returns undefined for at least the non-English locales); after: every locale file resolves a non-empty label for all three capability keys. Covers AC5.

### Additional test cases
- Component test: a user with only `ACCESS_TENANT_USERS_GROUPS_AND_ROLES` sees the tenant users/groups/roles lists but no create/edit/delete affordances
- Component test: a user with `MANAGE_TENANT_USERS_GROUPS_AND_ROLES` sees create/update affordances but not delete
- Component test: a user with `DELETE_TENANT_USERS_GROUPS_AND_ROLES` sees all affordances
- Component test: a user with only `MANAGE_TENANT_SETTINGS` (and none of the new triad) no longer sees tenant user/group/role create/edit affordances
- Snapshot/i18n test: the three new capability labels render correctly in at least English and French

### Sensitive Path considerations
- **Cross-tenant isolation test to add:** frontend CASL gating reflects capabilities already resolved and scoped server-side for the current session's tenant, so this UI layer itself cannot leak another tenant's data by construction — the isolation guarantee lives in the backend resolution (#7402) and API enforcement (#7401). Add a component-level test confirming the frontend never renders tenant-B-scoped role/group/user data when the active session's ability object is tenant-A scoped (i.e. no client-side merging of abilities across tenants), even though the primary isolation boundary is backend-enforced.
- **Fail-open vs fail-closed:** CASL's `ability.can(...)` returns `false` for any subject/action not explicitly granted (default-deny), so an unresolved or missing capability on the ability object results in the affordance being hidden — this is inherently fail-closed and requires no additional code. Confirm during implementation that no `ability.can` check here is inverted (e.g. checking `cannot` and defaulting to show) which would silently flip this to fail-open.
- **Audit-logging point:** N/A — this issue only changes UI affordance gating; it performs no grant/revoke/delete action itself. The underlying delete/role-change actions are performed via the existing backend API endpoints, whose audit-logging is covered by #7401's audit-log test.

## Definition of Done
- [ ] All identified components updated and passing lint/type-check
- [ ] No regression in existing TENANT_SETTINGS-gated screens unrelated to users/groups/roles
- [ ] i18n keys present in all locale files with no missing-translation warnings

### LINKED ISSUE #7401: test(rbac): add end-to-end coverage for the tenant users/groups/roles capability split
## Context
Introducing a new capability triad and removing user/group/role management from TENANT_SETTINGS needs dedicated regression coverage — both to prove the split is enforced end-to-end at the API layer and to confirm the existing capability/role audit logging correctly records grants and revokes of the new capabilities.

## Acceptance Criteria
- [ ] An API-level capability test suite verifies that tenant USER/GROUP/GROUP_ROLE endpoints enforce the new triad (403 without the matching capability, success with it) at each of the three tiers
- [ ] A test verifies `MANAGE_TENANT_SETTINGS` alone no longer grants access to tenant user/group/role endpoints
- [ ] An audit-logging test verifies that granting or revoking one of the new triad capabilities on a role is recorded by the existing audit log mechanism
- [ ] Regression test confirms `DELETE_TENANT_SETTINGS` and its siblings still work correctly for every remaining (non-users/groups/roles) resource under tenant settings

### Red test (write first)
*This is a cross-cutting regression/e2e suite spanning the already-merged slices from #7402, #7400, #7399 — kept as its own issue per convention (not folded into a single slice's DoD) since it exercises the interaction of all three.*
- `TenantUserApiIT#manageTenantSettingsAloneReturns403OnUserCreate` — before: with today's code, a role holding only `MANAGE_TENANT_SETTINGS` gets a 2xx creating a tenant user (current real behavior, since USER/WRITE still resolves under TENANT_SETTINGS); after #7402/#7399/#7400 land: the same request returns 403, and only a role holding `MANAGE_TENANT_USERS_GROUPS_AND_ROLES` gets 2xx. Covers AC2.
- `TenantGroupApiIT#tripleTierEnforcementAcrossAllThreeCapabilities` — before: no endpoint-level check exists yet for the new triad, so a request from a role holding zero triad capabilities and zero TENANT_SETTINGS capabilities returns whatever the current default is (test asserts 403 and confirms this is already true, then asserts progressively that ACCESS-only still 403s on write, MANAGE grants write, DELETE grants delete) — before wiring, the ACCESS-only-role-attempts-write case incorrectly succeeds if the endpoint still checks the old TENANT_SETTINGS subject; after: it correctly 403s. Covers AC1.
- `AuditLogIT#grantingNewTriadCapabilityIsAudited` — before: granting `MANAGE_TENANT_USERS_GROUPS_AND_ROLES` to a role produces zero matching audit log entries (capability doesn't exist yet / isn't wired to the audit hook); after: exactly one audit entry is created naming the capability, the role, and the acting admin. Covers AC3.

### Additional test cases
- Create a role with only `ACCESS_TENANT_USERS_GROUPS_AND_ROLES`, assign to a user, confirm the user can list tenant users/groups/roles but not create/update/delete
- Same test pattern for `MANAGE_TENANT_USERS_GROUPS_AND_ROLES` (create/update allowed, delete forbidden) and `DELETE_TENANT_USERS_GROUPS_AND_ROLES` (all allowed)
- Negative test: a role with only `MANAGE_TENANT_SETTINGS` receives 403 on tenant user/group/role write endpoints
- Audit log test: granting `MANAGE_TENANT_USERS_GROUPS_AND_ROLES` to a role produces an audit entry naming the capability
- Regression test: a role with `DELETE_TENANT_SETTINGS` can still manage tags, tag rules, attack patterns, collectors, injectors, and other retained TENANT_SETTINGS resources
- **Cross-tenant isolation test (added per Sensitive Path check):** create identically-named/identically-capabilitied roles in tenant A and tenant B (e.g. both named "Support" with `MANAGE_TENANT_USERS_GROUPS_AND_ROLES`); assert a user assigned tenant A's role can never list/create/update tenant B's users, groups, or roles via the API, and that capability resolution never matches across tenants by role name or capability name alone — only by the tenant-scoped role assignment.

### Sensitive Path considerations
- **Cross-tenant isolation test to add:** see the dedicated cross-tenant isolation test added to "Additional test cases" above — this is the primary end-to-end place in the batch where this must be proven at the API layer (definitions in #7402 and UI gating in #7400 rely on this backend guarantee).
- **Fail-open vs fail-closed:** this issue's own AC1/AC2 imply the endpoints must deny (403) whenever the required capability is absent or unresolved — i.e. fail-closed. Add an explicit test for the resolution-error edge case if one exists in the codebase (e.g. a role referencing a capability that fails to resolve/deserialize) to confirm the endpoint denies rather than defaults to allow; if no such error path currently exists in the codebase, state this as an open question for the Implementer rather than fabricating a scenario.
- **Audit-logging point:** covered directly by AC3 and the `AuditLogIT#grantingNewTriadCapabilityIsAudited` red test above — extend the same test (or add a sibling) to also cover **revoking** a triad capability from a role, and confirm role **deletion** itself (if the role held one of the new triad capabilities) is captured by the existing audit mechanism, since deletion is a sensitive action per this issue's own AC scope.

## Definition of Done
- [ ] All new tests pass in CI
- [ ] No existing test suite regresses
- [ ] Coverage report shows the new capability branches exercised

### LINKED ISSUE #7402: feat(rbac): add tenant users, groups and roles capability triad
## Context
Tenant user, group, and role management is currently reachable only through the broader ACCESS/MANAGE/DELETE_TENANT_SETTINGS capabilities, which also gate unrelated tenant-admin surfaces (collectors, injectors, tag rules, attack patterns, etc.). This issue introduces three new independently-grantable capabilities — Access / Manage / Delete tenant users, groups and roles — modeled on the existing platform-side triad, and removes the USER/USER_GROUP/GROUP_ROLE resource pairs from the TENANT_SETTINGS triad so the two capability sets no longer overlap.

## Acceptance Criteria
- [ ] `ACCESS_TENANT_USERS_GROUPS_AND_ROLES`, `MANAGE_TENANT_USERS_GROUPS_AND_ROLES`, `DELETE_TENANT_USERS_GROUPS_AND_ROLES` are added to `Capability.java` with the same structural pattern as `ACCESS/MANAGE/DELETE_PLATFORM_USERS_GROUPS_AND_ROLES` (parent chaining, `TENANT` scope, `checkable=true`, `hidden=false`)
- [ ] A new `TENANT_USERS_GROUPS_AND_ROLES` entry is added to `CapabilityGroup.java`
- [ ] The `USER`, `USER_GROUP`, `GROUP_ROLE` resource/action pairs are removed from `ACCESS_TENANT_SETTINGS`, `MANAGE_TENANT_SETTINGS`, `DELETE_TENANT_SETTINGS` and moved to the new triad (READ/SEARCH → ACCESS, WRITE/CREATE → MANAGE, DELETE → DELETE)
- [ ] No resource/action pair is duplicated between the new triad and `TENANT_SETTINGS` (each pair belongs to exactly one capability)
- [ ] `Capability.of(resource, action)` resolves USER/USER_GROUP/GROUP_ROLE pairs to the new triad, not to TENANT_SETTINGS

### Red test (write first)
*Note on sequencing: because the three capabilities are new enum constants, write them as minimal structural stubs (enum value + scope/flags only, no resolution wiring) first — that stub is not "production logic", it just makes the symbols compilable — then write and confirm-red the tests below before wiring `Capability.of(...)` resolution and the TENANT_SETTINGS removal.*
- `CapabilityGroupTest#tenantUsersGroupsAndRolesGroupExists` — before: `CapabilityGroup.valueOf("TENANT_USERS_GROUPS_AND_ROLES")` throws `IllegalArgumentException` (group does not exist); after: the group exists and contains the three new capabilities in ACCESS→MANAGE→DELETE order. Covers AC2.
- `CapabilityTest#tenantScopedCapabilityCountIncludesNewTriad` — before: `Capability.allTenantScoped().size()` equals the current baseline count N; after: equals N+3 once the three new capabilities exist with `TENANT` scope, `checkable=true`, `hidden=false`. Covers AC1.
- `CapabilityTest#userReadPairNoLongerResolvesToTenantSettings` — before: `Capability.of(RESOURCE_USER, ACTION_READ) == ACCESS_TENANT_SETTINGS` is currently `true` (today's real behavior — confirm this test fails when asserting the target outcome); after: `Capability.of(RESOURCE_USER, ACTION_READ) == ACCESS_TENANT_USERS_GROUPS_AND_ROLES` and `!= ACCESS_TENANT_SETTINGS`. Covers AC3 and AC5.
- `CapabilityTest#noDuplicateResourceActionPairsAcrossTenantCapabilities` — before: once the new triad pairs are defined but the TENANT_SETTINGS removal hasn't landed yet, iterating all TENANT-scope capabilities' resource/action pairs finds USER/USER_GROUP/GROUP_ROLE duplicated across both the new triad and TENANT_SETTINGS (test fails); after: no duplicates remain. Covers AC4.

### Additional test cases
- `Capability.of(USER, READ)` returns `ACCESS_TENANT_USERS_GROUPS_AND_ROLES`; `Capability.of(USER, WRITE)` returns `MANAGE_TENANT_USERS_GROUPS_AND_ROLES`; `Capability.of(USER_GROUP, DELETE)` returns `DELETE_TENANT_USERS_GROUPS_AND_ROLES`
- `Capability.resolveWithParents(Set.of(DELETE_TENANT_USERS_GROUPS_AND_ROLES))` returns the full parent chain (DELETE → MANAGE → ACCESS)
- `Capability.allTenantScoped()` includes the three new capabilities
- Existing `ACCESS_TENANT_SETTINGS`/`MANAGE_TENANT_SETTINGS`/`DELETE_TENANT_SETTINGS` no longer resolve USER/USER_GROUP/GROUP_ROLE pairs (`Capability.of(USER, READ)` no longer returns `ACCESS_TENANT_SETTINGS`)
- `CapabilityTreeBuilder.buildTree(CapabilityScope.TENANT)` includes a `TENANT_USERS_GROUPS_AND_ROLES` category node with the three capabilities nested under it in parent order

### Sensitive Path considerations
- **Cross-tenant isolation test to add:** capability *definitions* here are global enum constants (not per-tenant), but the actual grant lives in `roles_capabilities` keyed by role+tenant. Add a test creating identically-capabilitied roles in two different tenants (A and B) and asserting resolution/enforcement for a tenant-B user never returns or is granted by a tenant-A role's grant of the new triad — the isolation boundary to test is the role/tenant relationship, not the enum itself.
- **Fail-open vs fail-closed:** not stated in this issue's current text (it only changes capability *definitions and resolution*, not a runtime enforcement checkpoint). Per convention this must be fail-closed on resolution error. Open question for the Implementer: confirm the existing PLATFORM triad's error-path behavior (what happens if `Capability.of(...)` or `resolveWithParents(...)` throws/returns empty during a live permission check) and ensure the new TENANT triad follows the identical fail-closed behavior — this issue's own precedent code was not found to state this explicitly.
- **Audit-logging point:** N/A directly in this issue — it only defines the capability model and does not itself perform a grant/revoke or delete action. The grant/revoke audit-logging test is covered by #7401. Open question for the Implementer: confirm the existing audit-log mechanism logs new enum values generically (by capability name) with no per-capability special-casing required.

## Definition of Done
- [ ] Code merged with unit tests passing
- [ ] No behavior change for any capability outside the tenant users/groups/roles and tenant settings triads
- [ ] Reviewed against the platform-side triad for structural parity (naming, scope, checkable flag)

RomuDeuxfois:
Implements US.1 (#7196) across its four chunks, one commit each.

| Commit | Issue | Scope |
|:---|:---|:---|
| `c1faa3b77` | #7402 | the capability triad and its resolution |
| `5b88a90dc` | #7399 | migration of existing tenant roles, breaking-change doc |
| `e0c162146` | #7400 | frontend gating and i18n |
| `08a4551da` | #7401 | end-to-end coverage |

## What changes

Managing a tenant's users, groups and roles used to come with the tenant settings capabilities, which also cover collectors, injectors, tag rules, attack patterns and organizations. Granting someone the right to create a user therefore also granted them the right to deploy a connector.

`ACCESS/MANAGE/DELETE_TENANT_USERS_GROUPS_AND_ROLES` now own the `USER`, `USER_GROUP` and `GROUP_ROLE` resources, following the platform-side triad: `TENANT` scope, parent chaining, checkable and visible. Moving the pairs had to be atomic — the lookup map is built with `toUnmodifiableMap`, so a pair declared twice fails class initialization rather than a test.

Existing roles are migrated: each tenant role receives the new capability matching the tenant settings tier it already holds (Access → Access, Manage → Access+Manage, Delete → all three). Platform roles are excluded, the insert is idempotent, and no role gains or loses access on upgrade.

The frontend follows the split, and applies the house rule **reading rights hide a screen, create/update/delete rights only grey out their action**. Access to the screens now accepts the new capability — `useSecurityScope` exposes the tenant settings and tenant users flags separately, so Organizations, Sessions and Policies stay where they belong.

Two gating bugs surfaced on the way and are fixed here: `RolePopover` gated Delete on `MANAGE`, and `GroupPopover`'s Delete entry had no check of its own.


RomuDeuxfois:

openaev-front/src/admin/components/settings/groups/tenant_groups/TenantGroupsTab.tsx
```diff
@@ -27,6 +27,9 @@ import {
 } from './tenantGroups.queryable';
 
 const TenantGroupsTab = () => {
+  const ability = useContext(AbilityContext);
+  // MANAGE is greyed out rather than hidden: the affordance stays discoverable.
```
damgouj, line 31: question: is product OK with this?

openaev-front/src/utils/lang/zh.json
```diff
@@ -1,18 +1,22 @@
...
   "Discard the current attack path and re-author it from the beginning. Everything in the current logic is wiped.": "Discard the current attack path and re-author it from the beginning. Everything in the current logic is wiped.",
   "How should the AI build?": "How should the AI build?",
   "Keep the current attack path and continue from it - the orchestrator refines and extends what is already there. An AI-built scenario reopens its full reasoning history.": "Keep the current attack path and continue from it - the orchestrator refines and extends what is already there. An AI-built scenario reopens its full reasoning history.",
+  "MANAGE_TENANT_USERS_GROUPS_AND_ROLES": "管理租户用户、群组和角色",
   "Rebuild from scratch": "Rebuild from scratch",
   "Rebuild the logic": "Rebuild the logic",
   "Refine": "Refine",
   "Refine the existing logic": "Refine the existing logic",
   "Refine the logic": "Refine the logic",
+  "TENANT_USERS_GROUPS_AND_ROLES": "租户用户、群组和角色",
```
damgouj: suggestion: I think you should execute the translation yourself, there are missing keys (and some not well translated but not your dev)

damgouj (changes requested):
Tests:
- I think not this dev but I have only the new capa ACCESS_USERS_ROLES_GROUPS but I can see all this
Image
- When I have the TENANT_SETTINGS_CAPA, I can see the users, roles and groups menu
Image
- Tests from a fresh database, migration not tested

damgouj (approved):

PR #7501, opened 2026-08-18
fix(chaining): do not allow the duplication of the variable (#7502)

### LINKED ISSUE #7502: fix(chaining): scope - Variable - Duplicate scope variable (same key + type) shows raw DB constraint error instead of a form validation message
### Description
When creating a second scope variable with the same **key** and **type** as an existing one, the backend rejects the request with a raw database constraint violation:

> "Error applying constraint uk_scope_variable_key_type_workflow" -> 409

This is a low-level SQL/DB error leaking into the UI instead of a clear, actionable validation message.

### Steps to reproduce
1. Create a scope variable with key `X` and type `Y`.
2. Create a second scope variable with the same key `X` and same type `Y`.
3. Observe the raw constraint error `uk_scope_variable_key_type_workflow (409)` instead of a helpful message.

### Expected behavior
The form should proactively validate (or gracefully catch the 409) and display a clear, user-friendly message such as:

> "A variable with this key and type already exists. Please change the name or the type."

### Actual behavior
The raw DB constraint name and HTTP status are surfaced directly to the user, which is confusing and not actionable.

### Impact
- Poor UX: users don't understand what went wrong or how to fix it.
- Exposes internal implementation details (constraint name) to end users.

### Suggested fix
- Catch the `uk_scope_variable_key_type_workflow` constraint violation (409) on the backend and map it to a clear business error.
- On the front-end, surface this as an inline form validation message (e.g., on the key/type fields) prompting the user to change the name or the type, rather than displaying the raw error.

Image

camrrx:
### Proposed changes
* Add a zod validation if the key and type pair already exists

### Testing Instructions

1. Create a variable with a key and a type
2. Try to create a new variable with the same key and the same type from previous one 

### Related issues

* Closes #7502

### DIFF
--- a/openaev-front/src/admin/components/chaining/ScopeVariableCreateDialog.tsx
+++ b/openaev-front/src/admin/components/chaining/ScopeVariableCreateDialog.tsx
@@ -80,7 +80,7 @@ const ScopeVariableCreateDialog = ({
         const pairKey = `${normalizedKey}::${data.scope_variable_type}`;
         if (existingVariablePairs.has(pairKey)) {
           const message
-            = 'A variable with this key and type already exists. Please change the name or the type.';
+            = 'A variable with this key and type already exists. Please change the key or the type.';
           context.addIssue({
             code: 'custom',
             path: ['scope_variable_key'],
--- a/openaev-front/src/admin/components/chaining/ScopeVariableCreateDialog.tsx
+++ b/openaev-front/src/admin/components/chaining/ScopeVariableCreateDialog.tsx
@@ -7,8 +7,8 @@ import {
   DialogTitle,
 } from '@mui/material';
 import { useTheme } from '@mui/material/styles';
-import { useMemo } from 'react';
-import { FormProvider, useForm } from 'react-hook-form';
+import { useEffect, useMemo } from 'react';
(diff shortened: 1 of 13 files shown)


openaev-front/src/admin/components/chaining/ScopeVariableCreateDialog.tsx
```diff
@@ -46,17 +56,44 @@ const ScopeVariableCreateDialog = ({ open, onClose, onSubmit }: ScopeVariableCre
...
+          { message: t('Type is required') },
+        ),
+        scope_variable_value: z.string().min(1, { message: t('Value is required') }),
+        scope_variable_description: z.string().optional(),
+      })
+      .superRefine((data, context) => {
+        const normalizedKey = data.scope_variable_key.trim();
+        const pairKey = `${normalizedKey}::${data.scope_variable_type}`;
+        if (existingVariablePairs.has(pairKey)) {
+          const message
```
savacano28, line 82: Currently, We use variables based on their type, not their name. Could you check what happens when saving two variables with the same type? Also, I think we should just use 'key' instead of 'name', since we use both words for the same thing right now :)

savacano28 (approved):
Tested OK!

PR #7504, opened 2026-08-18
feat(audit-logging): phase 2 - US.5 - logging of agent inactive status (#7034)

### LINKED ISSUE #7034: feat(audit-logging): Phase 2 - US.5 - Agent updates
GIVEN an Agent disconnect due to Time out , THEN the disconnection event is recorded in the audit log
GIVEN an Agent IP changes , THEN the IP change is recorded in the audit log ( to verify -> OK ).

damgouj:
### Proposed changes

* Log when an agent becomes inactive (1 hour timeout)

### Testing Instructions

1. Install an agent and stop it
2. Log inactive after 1 hour (or update the last_seen column)
3. Start again the agent
4. Log active when registered again

### Related issues

* Closes #7034

### Checklist


openaev-api/src/main/java/io/openaev/scheduler/jobs/AgentInactivityMonitorJob.java
```diff
@@ -0,0 +1,92 @@
...
+
+          logger.logEvent(
+              AuditEvent.builder()
+                  .eventType(EventType.EXECUTION)
+                  .eventScope(AuditEventScope.COVERAGE_GAP)
+                  .eventStatus(EventStatus.WARNING)
+                  .resourceType(ResourceType.AGENT)
+                  .resourceId(agent.getId())
+                  .message(
+                      "Agent '%s' became inactive (no heartbeat for 1 hour)"
```
Dimfacion, line 85: **nitpick :** Should ACTIVE_THRESHOLD_MILLIS be configurable ? If so, this message might need some parameters instead of a fixed '1 hour'

openaev-api/src/main/java/io/openaev/service/autonomous/CapabilityResolverService.java
```diff
@@ -162,10 +162,7 @@ private ArsenalInventory buildArsenalInventory() {
       }
     }
 
-    int activeAgents =
-        (int)
-            agentRepository.countByLastSeenAfter(
-                Instant.ofEpochMilli(now - AgentHelper.ACTIVE_THRESHOLD));
+    int activeAgents = (int) agentRepository.countByStatus(AgentStatus.ACTIVE);
```
Dimfacion, line 165: **praise :** well seen !

openaev-model/src/main/java/io/openaev/database/model/Agent.java
```diff
@@ -125,9 +125,15 @@ public enum DEPLOYMENT_MODE {
 
   @JsonProperty("agent_active")
   public boolean isActive() {
-    return new AgentHelper().isAgentActiveFromLastSeen(this.getLastSeen());
+    return this.status == AgentStatus.ACTIVE;
   }
 
+  @AuditStateIgnore
```
Dimfacion, line 131: **issue :** I don't think it should be set to AuditStateIgnore as we want to monitor and have a log if the status has changed during a call

Dimfacion (changes requested):

Dimfacion:
It works for traces when the agent goes from ACTIVE to INACTIVE but when going from INACTIVE to ACTIVE, while I do have a trace, it shows incorrectly the status as INACTIVE in the audit logs.

openaev-api/src/main/java/io/openaev/scheduler/jobs/AgentInactivityMonitorJob.java
damgouj: It's not since 2 years and it can't be because we don't manage the ping from some external executors. It could bring some false results for customers if we let them do what they want with this THRESHOLD

openaev-model/src/main/java/io/openaev/database/model/Agent.java
damgouj: Well seen, fix to see it in the audit log but it's also already checked with the isActive attribute

openaev-api/src/main/java/io/openaev/utils/object/ObjectRedactionUtils.java
```diff
@@ -66,6 +66,9 @@ private ObjectRedactionUtils() {}
   private static final Set<ResourceType> USER_ENTITY_TYPES =
       Set.of(ResourceType.USER, ResourceType.PLATFORM_USER, ResourceType.PLAYER);
 
+  /** Non-sensitive fields removed from audit payloads to avoid noisy diffs. */
+  private static final Set<String> AUDIT_FIELDS_TO_REMOVE = Set.of("asset_status");
```
Dimfacion: **nitpick :** Maybe it would be better to have an annotation to not audit log a specific field ? A bit like it's done with @AuditStateIgnore, it should be doable to create a very similar filter and apply it on the objectMapper used in AccessControlAuditLogAspect ? Even better if we can have this kind of config :


```
@Configuration
public class AuditObjectMapperConfig {

  @Bean("auditObjectMapper")
  public ObjectMapper auditObjectMapper(ObjectMapper base) {
    ObjectMapper mapper = base.copy();
    mapper.setFilterProvider(...);
    return mapper;
  }
}
```


And then call the object mapper with a qualifier ?
What do you think of this idea ?

Dimfacion (approved):
Retested, all good ! Thank you for your work on this :)

PR #7520, opened 2026-08-19
feat(chaining): event input validation (#5049)

camrrx:
### Proposed changes

* Add zod validation for event input 
* Filter operator depending on the type

### Testing Instructions

1. Create chaining simulation 
2. Add an event
3. Test different type and different operator 
3.1 If the type is text : operator coud not be "greater than" for example 
3.2 If the type is number : there is an error zod message validation if the user try to enter a text message

### Related issues

* Link to #5049


EvaE-Filigran:
Tests OK for type port and number.
- Only number can be set
- Condition type Greater / Less /Than only available for type number

EvaE-Filigran (approved):

PR #7530, opened 2026-08-20
feat(rbac): UI improvement for capability triad visibility and restrictions (#7199)

### LINKED ISSUE #7199: feat(scope): US.5 — UI improvement for capability triad visibility and restrictions

- [ ]  Capabilities the current user does not hold are displayed as disabled/greyed out everywhere capabilities can be viewed or edited.
- [ ]  A user cannot add a capability they do not hold to any role.
- [ ]  A user cannot remove or toggle a capability they do not hold from an existing role.
- [ ]  When opening a role that already contains capabilities the current user does not hold, those capabilities remain visible but disabled/greyed out.
- [ ]  Example: when opening the Admin role, capabilities not held by the current user are visible in grey and cannot be clicked, added, removed, or changed.

### LINKED ISSUE #7200: feat(scope): US.6 — UI error when assigning unauthorized roles or users

- [ ]  When assigning a role to a group, if the role contains at least one capability the current user does not hold, the assignment is blocked.
- [ ]  When adding a user to a group, if the group has at least one role containing a capability the current user does not hold, the assignment is blocked.
- [ ]  The backend restriction remains the source of truth; this story only adds the UI feedback layer for already-covered authorization cases.
- [ ]  When the operation is blocked, the UI displays: “You can’t do this operation because you don’t have access to the required capabilities.”
- [ ]  The error is shown consistently for both role-to-group assignment and user-to-group assignment flows.

johanah29:
### Proposed changes

Make the capability model visible and self-enforcing in the UI: a user can never grant, through a role or a group, a capability they do not hold themselves. The API already validates this, the UI now explains it.

**Role editor**

* A capability the current user does not hold is locked: greyed, checkbox disabled, lock icon with a tooltip. It stays removable from a role that already carries it, so an existing role can be narrowed but never widened.
* A capability group whose whole subtree is locked is greyed as a group.
* Saving is blocked while a restricted capability is still selected, and the error names the offending capabilities with their translated labels.
01-role-capabilities-locks
02-capabilities-security-section

**Group → roles**

* A role carrying capabilities the current user lacks is locked in the picker: row disabled, lock icon, tooltip.
* A restricted role already attached to the group stays detachable. It can be removed, never added.
* The submit is disabled while a restricted role is still selected, and a header lock lists them.
03-group-manage-roles

**Group → users**

* When the group's own roles carry capabilities the current user lacks, membership is read-only: a callout at the top of the picker states it and names the missing capabilities, every row is frozen and the submit disabled.
image

### Technical notes

**A single scope table behind a context.** Everything the tenant and platform role APIs do differently now lives in one table (`RoleScopeContext`). A `RoleScopeProvider` publishes it and `useRoleScope()` reads it, so the role components take a scope from the tree instead of a handful of injected functions.

**Deduplication of the tenant / platform fork.**
A lot of Files were identical apart from constants.

**One role shape.** Both scopes now return `RoleOutput`, so the `ManageableRole` projection and the four adapter functions that mapped one shape onto the other are gone.

### Testing Instructions

1. Create a role holding only the tenant Users/Groups/Roles triad, attach it to a group of the tenant, and put a test user in that group.
2. Log in as that user, go to Settings → Security → Roles → Create, then open the Capabilities tab: everything the user does not hold is locked, and a group with no grantable capability under it is greyed as a whole.
3. Tick a restricted capability that a role already carries, then save: the form refuses and names the capabilities to remove.
4. Go to Settings → Security → Groups, open a group's `Manage roles`: restricted roles are locked, an attached restricted role can still be unticked, and the submit stays disabled while one is selected.
5. Open the same group's `Manage users`: if its roles carry capabilities the user lacks, the read-only callout is shown and nothing can be changed.
6. Repeat with `?scope=platform` on an EE platform, using a platform role and a platform group.

### Related issues

* Related #7199


corinnekrych:
@RomuDeuxfois my 2 cents UI proposals / suggestion:
Poposal1: disable title of the node so it's visually easier to spot the capabilities you can change
Before(left) vs After(right)
Screenshot 2026-08-31 at 14 46 42
Proposal2: the tooltip on disabled could extend the lock item to give it a broaden scope of howvering

RomuDeuxfois:
@corinnekrych I finished my changes.
I go with a big refacto, the main changes are in the screenshots.

openaev-front/src/admin/components/settings/roles/RoleScopeContext.ts
```diff
@@ -0,0 +1,62 @@
...
+  /** Resolves attached role ids: a selected role may be absent from the current search page. */
+  find: (roleIds: string[]) => Promise<RoleOutput[]>;
+  /** Own local storage entry, so both lists keep their own sort and filters. */
+  storageKey: string;
+  detailUrl: (roleId: string) => string;
+}
+
+export const ROLE_SCOPES = {
+  TENANT: {
+    scope: 'TENANT',
```
corinnekrych: neat: have an enum or type alias for TENANT and PALTFORM
```
export const RoleScope = {
  TENANT: 'TENANT',
  PLATFORM: 'PLATFORM',
} as const;

export type RoleScope = (typeof RoleScope)[keyof typeof RoleScope];
```

RomuDeuxfois: Done in f886fe259. I kept `CapabilityScope` derived from the generated back-end types and added the value form next to it in `utils/permissions/types.ts`, following the `ACTIONS` / `SUBJECTS` convention of that file (`RoleScope` as a name would have clashed with the interface just above):

```ts
export const CAPABILITY_SCOPES = {
  TENANT: 'TENANT',
  PLATFORM: 'PLATFORM',
} as const satisfies Record<CapabilityScope, CapabilityScope>;
```

The `satisfies` breaks the build if the back end ever adds a scope. Raw literals replaced in the 7 files of this PR that had them (`RoleScopeContext`, `Roles`, `useSecurityScope`, `GroupDetail`, `RoleDetail`, `PlatformGroupManage{Roles,Users}`). `SecurityMenu`, `Users`, `Sessions`, `UserDetail` and `Index` still spell them out but are outside this PR's diff — happy to convert them too if you prefer.

corinnekrych (approved):

PR #7537, opened 2026-08-20
feat(audit-logging): phase 2 - US.2 - System & Integration Activity Logging  (#6383)

### LINKED ISSUE #6383: feat(audit-logging): Phase 2 - US.2 - System & Integration Activity Logging
## Use case

As a CISO, I want all system-level events (integration changes, config modifications, background job failures) to be logged, so that I can audit platform health and detect unauthorized changes.

## Acceptance criteria
- GIVEN an integration is added/modified/removed, WHEN the change occurs, THEN an audit event is recorded with before/after values
- GIVEN a system config is changed, WHEN saved, THEN the old and new values are logged with actor and timestamp
- GIVEN a user clicks "Install" on an integration from the catalog, WHEN the action is confirmed, THEN an audit event is recorded with the actor, timestamp, and the integration name/identifier
- GIVEN a user clicks "Start" on an installed integration, WHEN the action is triggered, THEN an audit event is recorded with the actor, timestamp, and integration state transition (e.g. stopped → running)
- GIVEN  non critical health requests are emitted by  any integration , WHEN they are processed, THEN they are filtered out from audit logs (they remain visible in application/access on UI.
- GIVEN an integration changes state in the catalog (installed, uninstalled…  started, stopped), WHEN the state change occurs, THEN an audit event capturing the before and after state, actor, and timestamp is recorded

damgouj:
### Proposed changes

* Manage audit logging for XTM composer, delete unnecessary logs for integrations

### Testing Instructions

1. Delete, update, create, deploy,... connectors from the catalog
2. Check each action has 1 log
3. Check same logs aren't repeated during the tests

### Related issues

* #6383 

### Checklist

### DIFF
--- a/openaev-model/src/main/java/io/openaev/database/model/Collector.java
+++ b/openaev-model/src/main/java/io/openaev/database/model/Collector.java
@@ -6,6 +6,8 @@ import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import io.hypersistence.utils.hibernate.type.json.JsonType;
+import io.openaev.database.audit.AuditStateCapturable;
+import io.openaev.database.audit.AuditStateIgnore;
 import io.openaev.database.audit.ModelBaseListener;
 import io.openaev.jsonapi.BusinessId;
 import jakarta.persistence.*;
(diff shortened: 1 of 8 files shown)


openaev-model/src/main/java/io/openaev/database/model/Collector.java
```diff
@@ -94,6 +98,7 @@ public class Collector extends BaseConnectorEntity implements TenantIdBase {
   @ManyToOne(fetch = FetchType.LAZY)
   @JoinColumn(name = "collector_security_platform")
   @JsonProperty("collector_security_platform")
+  @AuditStateIgnore
```
Dimfacion, line 101: **question :** not entirely sure on why we ignore changes on the security_platform. Can you share the reasoning here ?

Dimfacion:
Silence, at last ;)
Tested ok, looks good to me, I'm just waiting for an answer on one of my comments and we'll be good to go

openaev-model/src/main/java/io/openaev/database/model/Collector.java
damgouj: Because it's only set at startup and if I keep it, the log is always written

Dimfacion (approved):

PR #7552, opened 2026-08-21
feat(ocsf): parser for OCSF inputs (#7764)

### LINKED ISSUE #7764: feat(ocsf): implement generic parser for schema version 1.x
## Use case
Prowler will be sending OCSF Detection Finding class objects as its output, so OpenAEV needs to be able to read this output and map to its internal data model.

## Current workaround
No workaround; OCSF is not a supported data format as API input

## Proposed solution
Two-fold
1. Create a capability to generate Java source code to represent the complete OCSF schema at a given version
2. Expose this code in OpenAEV to deserialise OCSF inputs into POJOs, for further mapping

Design goals:
* 1:1 matching with the OCSF spec at the requested schema version (as of writing, 1.9.0)
* Side-by-side cohabitation of different parser versions, e.g. the parser for v1.9.0 and the future parser for version 2.0.0 (not announced); client code may choose which parser version to pick after a cursory lookup of the input objects in JSON form
* Cached source OCSF schema source in java resources for quick re-generation; allowed by Apache 2.0 license.

antoinemzs:
### Proposed changes

* Generation of complete OCSF objects graph model through maven plugin
* Backend-side parser helper

### Testing Instructions

1. Parser example in unit tests

Also:

From the repository root:
```shell
antoinemzs@filibook:~/filigran/openbas$ mvn clean install -pl openaev-maven-plugin -DskipTests=true && mvn openaev:generate-ocsf-parser -Dversion=1.8.0 && mvn spotless:apply
```
This command will generate the parser source code for OCSF schema version 1.8.0 through the maven plugin. The git diff will show the extent of the generated code.

### Related issues

* Closes #7764


openaev-api/src/test/java/io/openaev/ocsf/OcsfFilterTest.java
```diff
@@ -0,0 +1,53 @@
...
+
+    List<OcsfClassDetectionFinding> sf = filter.filterOcsfClassDetectionFindings(trace);
+
+    assertThat(sf)
+        .hasSize(3)
+        .allSatisfy(
+            finding ->
+                assertThat(finding)
+                    .satisfies(
+                        f -> assertThat(f.getMetadataField().getVersionField()).isEqualTo("1.5.0"))
```
laugiov: CI is red on this assertion. getVersionField() returns OcsfDatatypeStringT, and OcsfDatatype.equals() rejects any object of a different class, so isEqualTo("1.5.0") can never pass. 

The API Tests (remaining) shard fails on this line and on nothing else. Either compare .getValue(), or add @JsonValue as suggested on OcsfDatatype and assert on the unwrapped value.

openaev-ocsf/src/main/java/io/openaev/ocsf/schema/OcsfDatatype.java
```diff
@@ -0,0 +1,48 @@
...
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.openaev.ocsf.parsing.OcsfSerialisable;
+import java.util.Objects;
+import lombok.Getter;
+
+@Getter
+public abstract class OcsfDatatype<T> implements OcsfSerialisable {
+  private final T value;
```
laugiov: Round-trip serialisation is broken. toOcsf() (line 46) emits the wrapper rather than the value, so metadata.version comes back as {"value":"1.5.0"} and severity_id as {"value":5}. The output is not valid OCSF, which defeats the purpose of OcsfSerialisable.

Adding @JsonValue on the value getter fixes it. I verified locally: the same fixture then round-trips to "1.5.0" and 5, and deserialisation keeps working unchanged. Worth locking in with a test that re-serialises the prowler fixture and diffs against the source JSON.

Minor, same class: toString() is not overridden, so logs and debuggers show the identity hash instead of the value.

openaev-ocsf/src/main/java/io/openaev/ocsf/schema/v190/OcsfConverter.java
```diff
@@ -0,0 +1,1841 @@
+package io.openaev.ocsf.schema.v190;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+public class OcsfConverter {
+
+  private com.fasterxml.jackson.databind.ObjectMapper mapper = new ObjectMapper();
```
laugiov: The generated mapper leaves FAIL_ON_UNKNOWN_PROPERTIES enabled, so any field the schema does not know aborts parsing. Verified: {"class_uid":2004,"severity_id":1,"vendor_custom_field":"x"} throws UnrecognizedPropertyException.

OCSF producers are expected to add their own fields, so a single Prowler or CrowdStrike release adding one attribute would break ingestion in production, with nothing catching it beforehand. With the feature disabled the same payload parses cleanly.

Please emit the mapper with the feature off, and add a test that feeds an unknown field so this cannot regress.

openaev-ocsf/src/main/java/io/openaev/ocsf/schema/OcsfDatatype.java
```diff
@@ -0,0 +1,48 @@
...
+    this.value = value;
+  }
+
+  /**
+   * Introspect and validate the format of the underlying data fragment. Some OCSF data types have
+   * validation rules (regexp based).
+   *
+   * @return true if the underlying data fragment clears the validation rules
+   */
+  protected boolean validate() {
```
laugiov: validate() is protected and has no caller anywhere in the repository, so the six generated regex validators (IP, MAC, email, UUID, datetime, bytestring) never run. Either expose a public entry point and call it during parsing, or stop generating the overrides. Right now the validation machinery is inert but reads as if it were active, which is the worst of both.

openaev-maven-plugin/src/main/java/io/openaev/ocsf/parser/GenerateParser.java
```diff
@@ -0,0 +1,62 @@
...
+    return basedir == null ? defaultBaseDir : basedir;
+  }
+
+  public void execute() throws MojoExecutionException {
+    Path path = Paths.get(getFinalBaseDir().getAbsoluteFile().toURI());
+    String subLocation = "src/main/java";
+    PluginContext ctx =
+        new PluginContext(
+            path.resolve("openaev-maven-plugin/src/main/resources"),
+            path.resolve("openaev-model").resolve(subLocation));
```
laugiov: Not a defect, but a decision that is free to make now and expensive to reverse later, so flagging it here on the line that picks the destination.

This sends 59.8k lines of generated code into openaev-model, the module every other module depends on. Nothing consumes them yet: OcsfFilterTest is the only reference to io.openaev.ocsf in the entire repository. So we grow the foundation jar and the review surface for a feature that is not wired in. Once other code starts importing from there, moving it becomes a cross-module refactor rather than a one-line change.

Two options that both keep the work intact:

- a dedicated openaev-ocsf module, so the core stays small and schema bumps never touch openaev-model;
- or generation into target/generated-sources on the generate-sources phase, which also drops this PR from 188k lines to roughly 2.5k of actually reviewable code.

Whichever we pick, a CI job that regenerates and asserts an empty diff would be worth adding. I ran -Dversion=1.9.0 followed by spotless:apply and got a byte-identical tree. 
That reproducibility is the strongest property of this generator and nothing currently protects it.

antoinemzs: generating the source code is rather expensive, because it depends on network calls to the OCSF schema server, with the possibility of it being down, slow... additionally, we need the source code to be available in the IDE at development time (before build), as it needs to be statically referenced to allow for autocomplete, and general linking in the IDE. The `generate-sources` phase doesn't seem to be practical for this purpose.

>a dedicated openaev-ocsf module

yes good thinking, this would make the parser its own jar that we could then reference from multiple places without having to load the whole model jar

=> new openaev-ocsf module created

openaev-api/src/test/java/io/openaev/ocsf/OcsfFilterTest.java
antoinemzs: asserting on equal `OcsfDatatypeStringT` instances

openaev-ocsf/src/main/java/io/openaev/ocsf/schema/OcsfDatatype.java
antoinemzs: fixed

openaev-ocsf/src/main/java/io/openaev/ocsf/schema/v190/OcsfConverter.java
antoinemzs: done; although OCSF does not advise to extend the schema that way, see: https://github.com/ocsf/ocsf-docs/blob/main/articles/patching-core-using-extensions.md

openaev-ocsf/src/main/java/io/openaev/ocsf/schema/OcsfDatatype.java
antoinemzs: I'm keen on the public method (it was an oversight to be honest), but I wonder whether it's desirbale for the parser to blow up on either serialisation or deserialisation just because of a format issue.

I'd be more interested in a method that can validate the structure on demand (using the public method). What do you think?

openaev-maven-plugin/src/main/java/io/openaev/ocsf/parser/GenerateParser.java
antoinemzs: @laugiov I've added a CI step https://github.com/OpenAEV-Platform/openaev/actions/runs/34823465580/job/103910085401

antoinemzs:
@laugiov ready for round 2

Issue #155, opened 2022-07-12
Introduce injects expectations and scoring


SamuelHassine:
## Problem to Solve

Be able to have injects expectations in the platform:

* Read a media article
* Specific flag / input to be entered by a player
* Upload a specific file (such as a press release, a situation report, etc.)

Expectations can be:

* Automatically validated / confirmed when this is a flag or just a read article
* Validated / confirmed by the animation team when it is a file or a specific reply

Expectations should have:

* An expiration time
* A specific associated score to compute lessons learned or scoring

Issue #44, opened 2019-04-23
Be able to postpone an entire exercise


SamuelHassine:
## Problem to Solve

Allow a user to postpone a entire exercise in a simply way by maintaining proportionally the time step between injects and without having to postpone manually one inject after the other.

## Current Workaround

None.

## Proposed Solution

- Create a feature that allows to change the start time or date of the exercise in a simplified way

## Additional Information

None.

Issue #43, opened 2019-04-23
Update to Symfony 5


SamuelHassine:
## Problem to Solve

Symfony 3 is obsolete, update to Symfony 4.

## Current Workaround

None.

## Proposed Solution

Update the project and its structure.

## Additional Information

None.

PR #7565, opened 2026-08-22
feat(fds): adopt the library Combobox and Select across the product selection fields

sandyghs:
**DO NOT MERGE** until the design review signs off.

Adopts the library `Combobox` and `Select` across the product's selection
fields, and fixes the visual defects the review surfaced on the way. Pinned on
library `cd4b8ee` (#155: controlled `open`, `startIcon`, `adornment`).

107 files. The wave grew well past its opening scope, so the sections below are
ordered by what a reviewer needs to judge rather than by when it landed.

## Scope

| family | sites | state |
|---|---|---|
| MUI `Autocomplete` → `Combobox` | 41 of 44 | converted |
| MUI `Autocomplete` in `ToolBar` | 3 | settled — out of scope |
| MUI `Select` family → `Select` | all remaining product sites | converted |

The `Select` conversion was not in the original scope. It was added because a
page mixing the two families reads as two different form languages, and the
review asked for the whole field surface at once. It lands over 11 commits,
grouped by what made each group awkward rather than by directory.

## Deliberate UX changes — please read these two before the diff

**The create button moved.** On the threat arsenal list, `Créer` used to sit in
the pagination row. The card view makes that row wider than the list view,
because it puts the sort select there: measured at a 1400px viewport, 47px of
the button was visible in list view and **0 in card view, with no horizontal
scroll to reach it**. It now sits in the page header beside the export/import
icons — one instance, rendered in both views, 8px from the icon group. Covered
by `threatArsenal-list-header.spec.ts`.

**The list toolbar is no longer one row at every width.** It carried selection,
search, filters, sort, page size and pagination side by side; at 1400px the
pagination arrows sat at x=1471, off-window, with no horizontal scroll to reach
them. It now has one fixed break point at 1600px:

- below 1600px it stacks — what acts on the whole list, then what narrows it;
- at or above 1600px everything shares one line;
- the filter chips keep a line of their own either way, and that line is absent
  when no filter is applied.

The break point is a media query. Nothing measures anything at runtime and no
state depends on the width — five adaptive attempts were built and discarded
before this, each failing the same way: a measurement whose result depended on
the layout it produced. Above the break point the row wrapper drops to
`display: contents`, so its groups become items of the toolbar itself; `order`
puts the pagination last and `margin-left: auto` holds it at the edge. The
groups shrink rather than wrap, which is why the tight case fits: on the arsenal
list at 1600px the three groups use 993px of the 1106px available.

Both are product UX decisions taken by the design lead, not side effects of the
migration. Both are shared surfaces — the toolbar is used by 62 pages — so the
one-line container also wraps: a page carrying heavier slots falls back to two
rows instead of running past its container.

That safety net is load-bearing, not theoretical. Above the break point the card
view can still wrap, because it carries the sort select the list view does not:
measured in CI, at 1600px the card view wraps to two rows (44px offset) while
the list view sits on one. Nothing overflows and no horizontal scroll appears in
either case — the arrangement is the one-line one, the container simply needs a
second row for it. A single break point for both views was preferred over a
second mechanism.

## Visual defects fixed, each traced to its own cause

- **Orphan MUI labels, 5 sites.** A converted field kept its MUI `InputLabel`,
  which is a floating label (`position:absolute; transform:scale(.75)`), so it
  rendered at a different size and offset from the library's own. All five now
  use the library label; the six drawer rows share one 26px label-to-field
  offset.
- **Chip row alignment.** `TagsFilter` laid its row out with `float:left` and a
  `5px` top margin, which offset the chips 3px from their neighbours. The row is
  now a centred flex row; all three chips land on centre 545.
- **Toolbar top row height, 52 → 36px.** Two causes, each fixed where it is set:
  MUI's `TablePagination` wraps its content in a `Toolbar` carrying its own
  `minHeight`, and its captions are `<p>` elements carrying the browser's 1em
  vertical margins; separately the select-all `Checkbox` carries its own
  padding. Click targets after the change: checkbox 36×36, pagination arrows
  32×32 — both above the 24×24 target-size minimum, so nothing was traded for
  the height.
- **Filter popover surfaces.** Rendered per mode from the design node: the
  intent is separation of the panel from the page, not one absolute ordering.
  The node is authored in dark, where the panel sits above its fields; in light
  the hierarchy inverts, so the two modes take different tokens.

## Notable in the Combobox wave

- `AutocompleteField`'s `renderOption` prop only ever tested its result for
  `null` — a caller's node was discarded. It is now `hideOption`, which is what
  it always meant. Three callers follow; the wrapper's `variant` prop goes with
  it, the library field having one style.
- Three `renderOption={option => option.label}` are removed. The library floors
  an option row at 48px whenever `renderOption` is passed, whatever it returns,
  so an identity one cost 16px per row for nothing.
- The chips-only token inputs use `open={false}` with **no** `onOpenChange`: no
  panel is mounted at all, and `Enter` / `ArrowDown` keep their native meaning
  in the input. Verified at the pointer.
- `FilterChipPopoverInput` no longer drives selection from the option row. The
  library owns the row; the new selection is diffed against the old one. Two
  visible consequences: `Enter` now reaches the library's handler, and a rapid
  double click toggles twice instead of being absorbed.
- On the four chip-feeding fields the name moved from the label to the
  placeholder, so the field reads as one thing feeding a token list.

## Library gaps measured here, reported, not worked around

Recorded in `fds-migration/LIBRARY-FEEDBACK.md`:

- **#39** — `Combobox` needs a cause on every callback; one product site
  expresses that cause as an event's presence.
- **#42** — the option-row floor should be `min-h-8`, so a site pays only for
  what it renders. The 48px floor applies to 12 sites rendering one glyph plus
  one line of text; the product cannot avoid it without dropping the glyph.
- **#43** — `Select` renders no wrapper of its own, so its label and trigger
  become siblings of whatever contains them. The most serious of the three.
- **#44** — `ComboboxLabel` has no `required`, so the two field families expose
  the required marker differently, and therefore expose different accessible
  names.
- **#45** — `Select` and `Tab`. Filed as a pointer only: the defect was measured
  by another session on OpenCTI and belongs to OpenCTI PR #17884. Nothing here
  compensates for it.

## Known and deliberately not addressed in this PR

- The 3 `ToolBar` `Autocomplete` sites — settled as out of scope.
- `Select` and `Tab` interaction — accepted for a later pass, tracked as gap
  #45 pointing at OpenCTI PR #17884.
- The new `labelPosition` axis is not wired to any product site yet.
- Above the break point the tab order does not follow the visual order: one DOM
  order cannot serve two arrangements whose reading orders differ, and the
  stacked arrangement — the common widths — is the one kept concordant.
  Duplicating the pagination group per arrangement would fix it but needs 13
  unscoped test locators re-scoped first, so it is deferred, not adopted.
- Between roughly 1396px and 1600px the list view stacks the toolbar without
  needing to. One threshold for both views was preferred over a second
  mechanism; the band renders the stacked arrangement, which is the reviewed
  one.
- Symmetrically, just above the break point the card view wraps rather than
  fitting one line, for the same reason — one threshold, two views of different
  widths. Measured, no overflow either way.
- Per-field height grew by 9px against the MUI original, accepted by the design
  lead.

## Verification

- Typecheck and lint clean.
- Unit suite: authoritative result is the CI run on this head. Locally
  `src/__tests__/utils/Cron.test.tsx` fails 81 tests, reproduced on the base
  commit and independent of timezone; that file is not touched here.
- E2E: the toolbar break point is pinned on both sides (1600 and 1599) in both
  views, by the wrapper's computed `display` rather than by the number of visual
  lines; the chips row is asserted present only while a filter is applied.
- The toolbar was measured at 1280 / 1400 / 1599 / 1600 / 1601 / 1711 / 1920 in
  both views, ascending and descending — 26 measurements, no overflow inside the
  bar at any width, no group wrapping internally, and the break point landing
  exactly on 1600.
- Served bytes checked on a running dev server for the library slots and for the
  break point; `leadingIcon` returns 0 occurrences, confirming the rename to
  `startIcon`.


RomuDeuxfois:
**Tests:**
- Asset creation
Tags field is in fullwidth, good for me. But, subcategory, platform and architecture fields are not in fullwidth and seems to be 'lost' in the drawer.
image
- Action creation
Expectations fields are overlapping on label/placeholder
image
image
- Action creation
Weird spacing on domain field
image
- Filters
Filters are bigger now
image
Tag filter has a big space
image
Dropdown list is smaller
image
- Scenario creation is not aligned
Before
image
Now
image
- Scenario with chaining in Attack Path tab is not aligned
image
- Widget creation
Label in dropdown are smaller
image
- Profil dropdown
Dropdown is large
image

sandyghs:
@RomuDeuxfois 

Quick follow-up: the oversized option rows you flagged are fixed since the lib pin bump (#176 upstream) — rows are back to 32px everywhere, including multi-select with checkboxes. The 12px option text stays: that's the design system's own type scale. Feel free to re-check on the branch.

RomuDeuxfois:
Tests:
- I still have a weird spacing here on Action creation
image
- Scenario with chaining in Attack Path tab is not aligned (Live btn)
image

sandyghs:
**Live button.** Fixed in 7fa9ec05d — the view group had been converted with
`size="sm"` (24px) while that row sits on a 36px control height. Now `size="md"`,
36px, matching the row.

**Spacing on Action creation.** Deferred: we can't reach that screen. It is behind
the Enterprise Edition licence, which we are not enabling here yet — turning it on
hides the EE chips, and those still have to be migrated. Recorded for the wave
after the EE chip migration. Nothing else from your list is left open.

PR #7570, opened 2026-08-24
fix(session): add a platform-scoped session capability (#7575)

### LINKED ISSUE #7575: fix(session): session capabilities cannot be granted and the platform sessions API has no capability of its own
## Description

`MANAGE_SESSIONS` was introduced with `hidden=true`. `CapabilityTreeBuilder` filters hidden capabilities out of the tree served to the role editor, so the capability exists but no role can ever be granted it: the Sessions screen and the four `SessionApi` endpoints are only reachable by a BYPASS user.

On top of that, `PlatformSessionApi` (`/api/platform-sessions`, platform-wide) is gated on the very same `ResourceType.SESSION`, whose capability is `TENANT`-scoped. `PlatformRoleService.assertCanAssignCapabilities` refuses tenant-scoped capabilities on a platform role, so no platform role can be granted access to the platform-wide listing — there is no platform-scoped capability for sessions at all.

The front follows the same gap: `Sessions.tsx` and the menu entry are gated on `TENANT_SETTINGS` / `PLATFORM_USERS_GROUPS_AND_ROLES`, and there is no `SESSIONS` subject in `permissions/types.ts`, so `parserRbac` silently drops the capability even when a user holds it.

## Reproducible steps

1. Settings > Security > Roles, create or edit a tenant role.
   → no session-related capability anywhere in the list.
2. Grant that role everything the editor offers, assign it to a user, log in as them.
3. Settings > Security > Sessions.
   → `GET /api/sessions` answers 403.
4. Same on the platform scope: no platform role can be given access to `GET /api/platform-sessions`.

## Expected output

Session management is grantable through a capability visible in the role editor, with the tenant listing and the platform-wide listing gated separately so a platform role can hold the platform one.

## Actual output

No role can be granted session access; both APIs are BYPASS-only, and the platform-wide API has no platform-scoped capability to check.

RomuDeuxfois:
### Proposed changes

`MANAGE_SESSIONS` was `hidden`, so no role could ever be granted it, and `/api/platform-sessions` was gated on the same tenant-scoped capability, which a platform role is not allowed to hold. Both listings were BYPASS-only.

* `MANAGE_SESSIONS` is no longer hidden and now appears in the tenant role editor
* New `MANAGE_PLATFORM_SESSIONS` / `ResourceType.PLATFORM_SESSION` gates the platform-wide listing, so the two scopes are grantable independently

Aside, but it fell out of the same screen: the listing carries `session_user_name`, resolved back-end side in one projection query. The front used to load every tenant user (or fetch platform users id by id) just to display a name.

Closes #7575


heditar:
Tested OK: granted the capa and tested api with and without the capa + UI

heditar (approved):

PR #7582, opened 2026-08-24
fix(chaining): rebuild imported action TTPs from the target injector contract (#7577)

### LINKED ISSUE #7577: fix(chaining): We should add action's TTP when import any scenario
## Description

with the import/export we can't export action's TTP informations or we can't import action's TTPs informations. 

## Environment

1. main-ff and demo

## Reproducible steps

1. create a logic for chaining scenario with different TTPs
2. export scenario
3. import scenario to another platform
4. see all actions are the Other TTP

## Expected output

All actions should have their existing TTPs

## Actual output

All actions looks TTPless

Image

Image

heditar:
Fixes #7577

## Problem

After exporting a chaining scenario and importing it (on another instance), every action was displayed as TTP-less and grouped under the "Other" tactic.

## Fix

`V1_DataImporter.sanitizateStepData()` now rebuilds both fields from the injector contract so the step is imported with the correct ttp

A contract that cannot be resolved on the target tenant is a no-op: such a step is already discarded upfront by `evaluateChainingStepResolvability`.

## How to Tests

- Export a scenario from one main with different ttp 
- Import it locally -> if you go to the logic page it should be organise by TTP and not have all the actions in 'Other'
-  You can use the folowing scenario for test
[test hedi ttp _2026-08-24T15_20_05.521363936Z_(no_variable_values & no_scope_definition).zip](https://github.com/user-attachments/files/31382604/test.hedi.ttp._2026-08-24T15_20_05.521363936Z_.no_variable_values.no_scope_definition.zip)

### DIFF
--- a/openaev-api/src/main/java/io/openaev/importer/V1_DataImporter.java
+++ b/openaev-api/src/main/java/io/openaev/importer/V1_DataImporter.java
@@ -1926,7 +1926,12 @@ public class V1_DataImporter implements Importer {
     if (result.injectorContract() != null) {
       return result.injectorContract();
     } else {
-      log.warn("An error has occurred when importing the payload: {}", result.payload().getName());
+      log.warn(
+          "Payload '{}' (id={}) was created but no injector contract could be synchronised for it:"
+              + " no payload-capable injector is registered in tenant {}",
+          result.payload().getName(),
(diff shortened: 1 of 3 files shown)


GaetanSantucci (approved):
tested, it works

Issue #7491, opened 2026-08-18
feat(audit-logging): Phase 2 - US.7 - Enhancements for token lifecycle


RomuDeuxfois:
## Acceptance Criteria

**AC1 — DB change: retain masked token history**

Given a token is renewed, the previous token value is not overwritten — it's retained in the DB as a masked/starred reference (not the actual token value).

**AC2 — Token creation logged**

Given a user creates a token, an audit log entry is written with `token_id`, creator, timestamp.

**AC3 — Token renewal logged with masked reference**

Given a token is renewed, the audit log entry records `token_id`, user, timestamp, and the masked/starred reference — not the raw value.

PR #7585, opened 2026-08-24
feat(audit-log): introduce soft delete of tokens for audit (#7491)

Dimfacion:
### Proposed changes

* Adding soft delete of tokens
* Throwing a new audit log event whenever we create a new token

### Testing Instructions

1. Create a user
2. Check that the audit log event for the token creation was created
3. Renew your own token
4. Check that the audit log event for the token creation was created
5. Check that the tokens are still visible in the database wwith masked value

### Related issues

* Related to #7491
*

### Checklist


.github/instructions/security.instructions.md
```diff
@@ -21,6 +21,7 @@ Every REST endpoint must have `@AccessControl`. See the annotation in `io.openae
    - Open for READ (like Player): add to `RESOURCES_OPEN`
    - Sub-resource (like Inject): add to `RESOURCES_USING_PARENT_PERMISSION`
    - Standard capability-based: no change (handled by `Capability.of()` lookup)
+5. When applicable, use tokens from the table token with token_delete_at being NULL.
```
RomuDeuxfois: **question:** why is this here? Items 1-4 are the checklist for declaring a new resource type, and filtering soft-deleted tokens is unrelated to that.

openaev-model/src/main/java/io/openaev/database/specification/TokenSpecification.java
```diff
@@ -8,4 +8,8 @@ public class TokenSpecification {
   public static Specification<Token> fromUser(String userId) {
```
RomuDeuxfois: **todo:** `active()` is opt-in, and it has already been missed: `ConnectorInstanceService` calls `findAll(fromUser(...))` with no `.and(active())`. After a renew, a connector can be provisioned with the masked `[RENEWED:<uuid>]` value. `findAll(Specification)` also emits no `ORDER BY`, so which of the two rows comes back is non-deterministic.

Rather than composing the filter at each call site, I'd drop this class and keep the whole token lifecycle in `TokenRepository`, where the filter is part of the method name and can't be silently omitted.

`ConnectorInstanceService`:
```java
Token token = tokenRepository
    .findFirstByUserIdAndDeletedAtIsNullOrderByCreatedAsc(currentUser().getId())
    .orElseThrow(() -> new IllegalArgumentException("No token found for current user"));
```

`MeApi.tokens()` then needs one derived method:
```java
List<Token> findByUserIdAndDeletedAtIsNullOrderByCreatedAsc(String userId);
```

After that `TokenSpecification` has no callers and `JpaSpecificationExecutor<Token>` can go too.

openaev-model/src/main/java/io/openaev/database/repository/UserRepository.java
```diff
@@ -213,13 +213,16 @@ List<User> findAllByIdInAndTenantId(
...
               + "LIMIT 1",
       nativeQuery = true)
   Optional<User> findByTokenAndTenantId(
       @Param("token") String token, @Param("tenantId") String tenantId);
 
-  @Query("SELECT u FROM User u JOIN Token t ON u.id = t.user.id WHERE t.value = :token")
+  @Query(
+      "SELECT u FROM User u JOIN Token t ON u.id = t.user.id "
+          + "WHERE t.value = :token AND t.deletedAt IS NULL")
   Optional<User> findByToken(@Param("token") String token);
```
RomuDeuxfois, line 226: **todo:** remove this query and resolve through the token repository in `UserService`. A `User` repository shouldn't encode token lifecycle.

```java
public Optional<User> findByToken(@NotBlank String token) {
  return tokenRepository.findByValueAndDeletedAtIsNull(token).map(Token::getUser);
}
```

openaev-model/src/main/java/io/openaev/database/repository/UserRepository.java
```diff
@@ -213,13 +213,16 @@ List<User> findAllByIdInAndTenantId(
               + "JOIN tokens t ON us.user_id = t.token_user "
               + "JOIN users_tenants ut ON us.user_id = ut.user_id "
               + "WHERE t.token_value = :token "
+              + "AND t.token_deleted_at IS NULL "
               + "AND ut.tenant_id = :tenantId "
               + "LIMIT 1",
       nativeQuery = true)
   Optional<User> findByTokenAndTenantId(
```
RomuDeuxfois, line 220: **todo:** move it to `TokenRepository` so all token-validity SQL lives in one file. The native query isn't needed either.

```java
@Query("SELECT t.user FROM Token t JOIN t.user.tenants tn "
     + "WHERE t.value = :value AND t.deletedAt IS NULL AND tn.id = :tenantId")
Optional<User> findActiveTokenUserInTenant(@Param("value") String value,
                                           @Param("tenantId") String tenantId);
```

**Goal:** manage the token lifecycle inside `TokenRepository`.

openaev-model/src/main/java/io/openaev/database/repository/TokenRepository.java
```diff
@@ -18,16 +18,17 @@ public interface TokenRepository
   @NotNull
   Optional<Token> findById(@NotNull String id);
 
-  Optional<Token> findByValue(String value);
+  Optional<Token> findByIdAndDeletedAtIsNull(String id);
+
+  Optional<Token> findByValueAndDeletedAtIsNull(String value);
```
RomuDeuxfois: **todo:** this has no production caller

openaev-model/src/main/java/io/openaev/database/repository/TokenRepository.java
RomuDeuxfois, line 13: **todo:** to keep unfiltered access out of reach, extend `Repository<Token, String>` instead of `CrudRepository` and declare only the methods you want.

That path does legitimately need to see soft-deleted rows, so give it an explicitly named accessor rather than the ambiguous one:

```java
/** Bootstrap only: resolves the admin token even when soft-deleted, so it can be restored. */
@Query("SELECT t FROM Token t WHERE t.id = :id")
Optional<Token> findByIdIncludingDeleted(@Param("id") String id);
```

openaev-front/src/admin/components/profile/Index.jsx
```diff
@@ -26,7 +26,9 @@ const Index = () => {
     user: helper.getMe(),
     tokens: helper.getMeTokens(),
   }));
-  const onRenew = tokenId => dispatch(renewToken(tokenId));
+  const onRenew = async (tokenId) => {
```
RomuDeuxfois, line 29: **todo:** evict the revoked token from the store instead of working around it in the component.

```js
export const renewToken = tokenId => dispatch => postReferential(schema.token, '/api/me/token/refresh', { token_id: tokenId })(dispatch)
  .then((data) => {
    dispatch({
      type: Constants.DATA_DELETE_SUCCESS,
      payload: {
        type: 'tokens',
        id: tokenId,
      },
    });
    return data;
  });
```

openaev-api/src/main/java/io/openaev/rest/user/MeApi.java
```diff
@@ -135,12 +136,18 @@ public Token renewToken(@Valid @RequestBody RenewTokenInput input) {
...
       throw new AccessDeniedException("You are not allowed to renew this token");
     }
-    token.setValue(UUID.randomUUID().toString());
-    return tokenRepository.save(token);
+
+    token.setDeletedAt(Instant.now());
+    token.setValue("[RENEWED:%s]".formatted(token.getId()));
+    tokenRepository.save(token);
+
+    return userService.createUserToken(user);
```
RomuDeuxfois: **todo:** this belongs in a service. `renewToken` does entity lookup, ownership check, revocation and re-issuance directly in the controller, and the soft-delete + masking rules now live in two places (here and `UserService.createUserToken`). Moving it next to `createUserToken` keeps the whole token lifecycle in one class.

openaev-api/src/main/java/io/openaev/rest/user/MeApi.java
```diff
@@ -156,6 +163,6 @@ public List<TenantOutput> myTenants() {
   @Transactional
   @AccessControl(skipRBAC = true)
   public List<Token> tokens() {
-    return tokenRepository.findAll(fromUser(currentUser().getId()));
+    return tokenRepository.findAll(fromUser(currentUser().getId()).and(active()));
```
RomuDeuxfois: **todo:** `findAll` shouldn't be reachable directly. Once `TokenSpecification` is gone, replace this with a derived method and drop `JpaSpecificationExecutor<Token>` from the repository, so no future call site can bypass the filter:

```java
return tokenRepository.findByUserIdAndDeletedAtIsNullOrderByCreatedAsc(currentUser().getId());
```

openaev-api/src/main/java/io/openaev/service/UserService.java
```diff
@@ -476,7 +483,40 @@ public Token createUserToken(User user, String discreteToken) {
...
+    if (auditLogger == null) {
+      return;
+    }
+    User actor = currentUserOrNull();
+    String tokenUserId = createdToken.getUser() != null ? createdToken.getUser().getId() : null;
+    Map<String, Object> contextData = new LinkedHashMap<>();
+    contextData.put("token_id", createdToken.getId());
+    contextData.put("token_user_id", tokenUserId);
+    contextData.put("actor_user_id", actor != null ? actor.getId() : null);
+    contextData.put("timestamp", createdToken.getCreated());
```
RomuDeuxfois: **question:** `contextData.put("timestamp", createdToken.getCreated())` carries the token's creation date, but `AuditEvent` already has its own event timestamp. Which one is this meant to be? If it's the token's, `token_created_at` would be clearer and consistent with the three other keys.

openaev-api/src/main/java/io/openaev/service/UserService.java
```diff
@@ -476,7 +483,40 @@ public Token createUserToken(User user, String discreteToken) {
...
+    logTokenCreated(createdToken);
+    return createdToken;
+  }
+
+  /**
+   * Emits an audit event for a token creation.
+   *
+   * @param createdToken the token that was created
+   */
+  private void logTokenCreated(Token createdToken) {
```
RomuDeuxfois, line 534: **suggestion:** a renewal currently emits a `TOKEN`/`CREATE` for the new token and nothing for the revoked one, so "when was token X revoked, and by whom" can't be answered from a `resource_type=TOKEN` query.

openaev-api/src/main/java/io/openaev/rest/user/MeApi.java
```diff
@@ -135,12 +136,18 @@ public Token renewToken(@Valid @RequestBody RenewTokenInput input) {
...
+        tokenRepository
+            .findByIdAndDeletedAtIsNull(input.getTokenId())
+            .orElseThrow(ElementNotFoundException::new);
     if (!user.equals(token.getUser())) {
       throw new AccessDeniedException("You are not allowed to renew this token");
     }
-    token.setValue(UUID.randomUUID().toString());
-    return tokenRepository.save(token);
+
+    token.setDeletedAt(Instant.now());
```
RomuDeuxfois: **issue (blocking):** this soft-deletes whatever token id the caller owns including `ADMIN_TOKEN_UUID` and `InitAdminCommandLineRunner` never undoes it, which permanently breaks `OPENAEV_ADMIN_TOKEN`.

Fix in the runner:
```java
if (hasText(this.adminToken)) {
  ...
  token.setValue(this.adminToken);
  token.setDeletedAt(null);
}
this.tokenRepository.save(token);
```

openaev-api/src/main/java/io/openaev/utils/object/ObjectRedactionUtils.java
```diff
@@ -50,7 +52,14 @@ private ObjectRedactionUtils() {}
 
   /** Sensitive-like fields that are explicitly allowed and therefore not Hashed. */
   private static final Set<Pattern> ALLOWED_SENSITIVE_FIELDS_REGEX_TO_HASH =
```
RomuDeuxfois, line 54: **nitpick:** worth a comment on why these three escape the broad `.*token.*` rule above.

RomuDeuxfois:
**Why keep the row after a renew?**

The audit event on token creation is a good addition. It's the soft delete I'm unsure about.

After a renew, the old row keeps the id, the owner, the two dates, and a value masked to [RENEWED:<id>]. The audit event already has the id, the owner, the actor and the creation date. So the row only adds the revocation date and it can't say who did it, while the event can.

And the question we'd keep it for: "this token value showed up somewhere, whose was it?" neither side can answer. The row hides the value, and in the audit trail it gets hashed and then replaced by the redaction placeholder.

What would change my mind: do we need the old token ids in the database itself, separately from the audit log? If yes, the soft delete makes sense and my inline comments cover the rest.

**On security**

This touches the table that authenticates every API call, and the rule "a revoked token must never come back" is easy to forget: a .and(active()) to add, or five words in a @Query. Drop them and it still compiles, CI stays green, no test fails.

It already happened twice on this branch and by the person who knows the feature best.

Before shipping a new auth concept I'd want:

- a test that a revoked token can no longer log in. Nothing covers it today
- findById comes unfiltered from CrudRepository, so the unsafe call is the default one
- an audit event on revocation
- maybe a purge ?

.github/instructions/security.instructions.md
Dimfacion: Sorry, I misread when adding this rule. Fixing it

openaev-model/src/main/java/io/openaev/database/repository/UserRepository.java
Dimfacion: I understand your logic but I feel odd that TokenRepository returns a User object. As a dev, if I want to find a way to get a user from his token, I would rather look at UserRepository, no ?

openaev-api/src/main/java/io/openaev/service/UserService.java
Dimfacion: Whenever you call the api to do a renewal, there are two events that are emitted. A manuel one for the creation of the new token and the automatic one regarding the call to /api/me/token/refresh. That second one already answer the question of whom renewed the token. I can add a new manual event stating information regarding the revokation of the token but then, it would mean that a call to this endpoint would result in 3 different event being raised. Isn't that a bit much ?

Issue #7500, opened 2026-08-18
fix(finding): finding page data analysis


mergenhan:
## Description

In the Finding page we should have making for sensitive data

## Reproducible steps

Steps to create the smallest reproducible scenario:
1. add a credential in the finding or discover a credential from a simulation
2. check the finding page
3. see the plaintext credential

## Expected output

credentials should be masked

## Actual output

credentials are not masked

savacano28:
@mergenhan Maybe we could extend this feature to also cover:
* AsreproastableAccountOutputProcessor.java  /  KerberoastableAccountOutputProcessor.java :  **username ,  host, hash**                                                                      
* PasswordPolicyOutputProcessor.java :  **key ,  value  can disclose policy details**  
?

mergenhan:
We don't need to mask password policy. It's only text from as policy
About the AsreproastableAccountOutputProcessor and KerberoastableAccountOutputProcessor, do we are sure that it produce hash? If it produce hash we should mask has but i think we're masking all hashed in this PR. @savacano28 @heditar

savacano28:
These output parsers were created at Product's request. We could now clean up the hashes for asreproastable and Kerberoastable accounts, if that is the case.

mergenhan:
@savacano28, I see your point no we don't need to remove it. Because we can make a relation account to hash but we don't need to mask these accounts.

PR #7587, opened 2026-08-24
feat(finding): derive sensitive finding masking from contract output type (#7500)

heditar:
### Proposed changes

Credentials found during simulations were shown in cleartext on the Findings page. The API now masks them: `admin:mo******`. The account name stays readable, only the secret is hidden. The database keeps the cleartext, which deduplication, correlation and the attack path all need.

A value is masked because of what it is — a `Password`, a `Hash`, a `Key` — read from the type it was extracted as. No new column, no migration, no flag to set in each output processor.

* `SensitiveValueMaskingUtils` (new) — one masking rule for the whole platform, replacing two older ones.
* `FindingApi` / `FindingMapper` / `FindingService` + `FindingOutput` (new) — finding endpoints no longer serialize the JPA entity, which is what exposed the cleartext.
* `AttackPathGraphService` — the graph masks server-side. It used to send the cleartext and mask it in the browser.
* Frontend — finding views retyped, client-side masking removed.
* Tests and `docs/docs/usage/evaluate/findings/findings.md`.

### Testing Instructions

1. Start the services: `cd openaev-dev && podman compose up -d openaev-dev-pgsql openaev-dev-minio openaev-dev-elasticsearch openaev-dev-rabbitmq`
2. Build with JDK 21: `mvn install -DskipTests -Dspotless.check.skip=true`
3. Backend: `mvn -pl openaev-api test -Dtest='FindingApiTest,FindingServiceTest,SensitiveValueMaskingUtilsTest,SensitiveValueCompositionConsistencyTest,ConsumedFindingKeyDTOTest,WorkflowServiceTest,AttackPath*Test' -Dsurefire.failIfNoSpecifiedTests=false`
4. Frontend: `cd openaev-front && yarn check-ts && yarn lint`
5. Run a simulation producing a `Credentials` finding. On **Findings**, the value reads `admin:mo******` in the list, the detail page and the attack path drawer. `SELECT finding_value FROM findings` still returns the cleartext.

### Related issues

* Related #7500

### Checklist

### Further comments

Masking happens on the way out of the API only. Two surfaces are deliberately left out and tracked separately: reports and dashboards still read the cleartext from Elasticsearch (#7892), and the finding search still queries the cleartext column, so a secret can be confirmed by searching it (#7893).


openaev-api/src/main/java/io/openaev/output_processor/FindingCapableOutputProcessor.java
```diff
@@ -44,7 +60,8 @@ public final void process(
         this::toFindingValue,
         this::toFindingAssets,
         this::toFindingTeams,
-        this::toFindingUsers);
+        this::toFindingUsers,
+        this.sensitive);
```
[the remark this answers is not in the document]
heditar: Good catch on the argument order, but this is pre-existing on `main` and not introduced here — this PR only appends the trailing `sensitive` flag to the `generateFindings(...)` call:

```java
// main
findingService.generateFindings(
    executionContext, contractOutputContext, structuredOutputNode,
    this::validate, this::toFindingValue, this::toFindingAssets,
    this::toFindingTeams, this::toFindingUsers);
```

It is also latent today: `linkFinding(...)` only assigns when the extracted list is non-empty, and **no concrete processor overrides `toFindingUsers` / `toFindingTeams`** (verified across `io.openaev.output_processor`), so both lambdas resolve to the base-class defaults returning `Collections.emptyList()`. Nothing is ever swapped at runtime.

That said it is a real trap for the first processor that implements either method, so it deserves its own fix rather than being smuggled into a redaction PR. I'll open a separate issue/PR for it so the change comes with a test covering user/team association.

heditar:
### PR risk classification
High Risk — This PR changes security-relevant core logic (redaction of sensitive finding values across API, service, mapper and output processors) and ships a database migration adding finding_is_sensitive, so it affects both critical data processing and existing data.

heditar:
### PR risk classification
High Risk — This PR touches security-relevant logic (redaction of sensitive finding values) across 19 files, including the finding API, service, output processors and mappers. It also adds a database migration introducing a new column, so it affects both core business logic and existing data.

heditar:
### PR risk classification
High Risk — This PR spans 19 files and introduces security-relevant logic (redaction of sensitive finding values in the API) together with a database migration adding a new column, so it affects both core data processing and existing data.

heditar:
### PR risk classification
High Risk — This PR introduces security-relevant logic (redaction of sensitive finding values in the API) across ~19 files, including a database migration adding a new column, changes to output processors, finding service/mapper/writer and API contracts. It touches core data processing and alters persisted data and exposed payloads, so the blast radius is broad.

heditar:
### PR risk classification
High Risk — This PR spans ~20 files of core finding business logic and introduces security-relevant redaction of sensitive values in the REST API, plus a database migration that adds and backfills `findings.finding_is_sensitive`, so it has a direct impact on existing data.

heditar:
### PR risk classification
High Risk — This PR spans 19 files, adds a database migration and implements security-relevant logic (redaction of sensitive finding values in the API and output processors), impacting both existing data and how sensitive information is exposed.

heditar:
### PR risk classification
High Risk — This PR touches security-relevant logic (redaction of sensitive finding values in the API) across 19 files and includes a database migration adding a new finding flag, so both core data processing and existing data are impacted.

heditar:
### PR risk classification
High Risk — This PR spans 19 files and introduces security-relevant redaction of sensitive finding values across API, service, output processors and mappers, together with a database migration adding a new column, so both critical logic and existing data are impacted.

heditar:
### PR risk classification
High Risk — This PR touches security-relevant logic (redaction of sensitive finding values) across 19+ files, including all output processors, the finding model/serializer and a database migration, so it affects core data processing and existing stored data.

heditar:
### PR risk classification
High Risk — This PR spans 19 files and adds security-relevant redaction of sensitive finding values in the API, along with a database migration adding a new column, so it touches core business logic and has a direct impact on existing data.

openaev-api/src/main/java/io/openaev/rest/finding/FindingApi.java
```diff
@@ -28,13 +28,13 @@ public class FindingApi extends RestBehavior {
   // -- CRUD --
 
   @GetMapping({FINDING_URI + "/{id}", TENANT_FINDING_URI + "/{id}"})
-  @Transactional
+  @Transactional(readOnly = true)
   @AccessControl(
       resourceId = "#id",
       actionPerformed = Action.READ,
       resourceType = ResourceType.FINDING)
   public ResponseEntity<Finding> finding(@PathVariable @NotNull final String id) {
```
savacano28: Maybe we should return a DTO and apply the redacted value during this generation?

openaev-api/src/main/java/io/openaev/rest/finding/FindingService.java
```diff
@@ -39,6 +48,57 @@ public class FindingService {
...
+   *   <li>{@code admin:motdepasse} becomes {@code ad******:mo******}
+   *   <li>{@code Sup3rS3cret} becomes {@code Su******}
+   *   <li>{@code abcd} becomes {@code ******}
+   * </ul>
+   *
+   * @param value the cleartext finding value
+   * @param sensitive whether the finding holds sensitive material
+   * @return the value as-is when the finding is not sensitive, its redacted form otherwise
+   */
+  public static String redact(final String value, final boolean sensitive) {
```
savacano28, line 73: We have a class with some mask rules defined, used to mask scope variables. Maybe we could centralize this masking service?
You can check: PrimitiveValueMaskingUtils.
The rules were already defined by Product (e.g., the masking fragment length was defined for passwords, hashes, and usernames).

openaev-api/src/main/java/io/openaev/output_processor/FindingCapableOutputProcessor.java
```diff
@@ -17,13 +18,26 @@ public abstract class FindingCapableOutputProcessor extends AbstractOutputProces
...
       FindingService findingService) {
+    this(type, technicalType, fields, findingService, false);
+  }
+
+  protected FindingCapableOutputProcessor(
+      ContractOutputType type,
+      ContractOutputTechnicalType technicalType,
+      List<ContractOutputField> fields,
+      FindingService findingService,
+      boolean sensitive) {
```
savacano28, line 37: Maybe it would be useful to have this param: sensitive at the output processor level?

openaev-api/src/main/java/io/openaev/rest/finding/FindingApi.java
heditar: Yes good idea, I will update it

openaev-api/src/main/java/io/openaev/output_processor/FindingCapableOutputProcessor.java
heditar: I don't think that is necessary, and in the case we want to go with this solution we can update it

openaev-api/src/main/java/io/openaev/rest/finding/FindingService.java
heditar: The logic is different and safer in the new code, maybe @mergenhan can help us with that. The masking logic in this PR comes from Antoine's brainstorm

note: `FindingService.redact` vs the existing `PrimitiveValueMaskingUtils`

| | `PrimitiveValueMaskingUtils.maskForDisplay` | `FindingService.redact` |
|---|---|---|
| **Rule** | keep prefix N + **suffix N**, mask the middle | split on `:`, keep a 2-char head **per part** |
| **Lengths** | Password (1,1), Hash (3,3), Key (3,3) | (2, 0) |
| **Mask** | `*` repeated **= length hidden** | fixed `"******"` (6 chars) |
| **Value too short** | `"*".repeat(length)` | full `MASK` |
| **Composite value** | no notion — value is atomic | splits on `:` and masks each part |
| **null / blank** | null returned as-is | null and blank returned as-is |
| **Round-trip guard** | `isMaskedRepresentationOfCurrentValue` | ❌ none |
| **Unrecognized input** | cleartext (fail-open) | `sensitive=false` → cleartext (fail-open) |

heditar: I checked with Murat it is ok to stay it like that for now until the work with legal is done to clarify how we handle findings

savacano28: It'd be nice to centralize at least these masking rules for now, even if they end up changing later.

savacano28:
Tests:
I tested the masking behavior for findings of type credential and found some issues:

✅ Working as expected:
* The server response for fetching findings correctly masks the credential values.
* The front-end displays the masked values as expected.

❌ Issues found:
* Search bypasses masking : Even though the value isn't displayed (masked) in the UI, I can still search for the specific (unmasked) value in the search bar and get results. The masking isn't enforced at the search level.
image
image

* Reports expose unmasked values: When generating a report for this scenario, the credential values appear unmasked in the report output.
image

* Masking in Attack path and findings is not the same:
image
image

savacano28 (changes requested):

heditar:
note: cette branche a été force-pushée depuis `heditar-finding-sensitivity-derived-from-type` et porte désormais l'approche « sensibilité dérivée du `ContractOutputType` » (identique à #7869), et non plus la colonne `finding_is_sensitive` décrite ci-dessus. Le titre et la description de cette PR ne correspondent plus à son diff.

L'état précédent est sauvegardé sur la branche `heditar-sensitive-finding-redaction-backup` (`3dddf0fd5`), rien n'est perdu. Pour restaurer :

```
git fetch origin
git reset --hard origin/heditar-sensitive-finding-redaction-backup
git push --force-with-lease origin HEAD:heditar-sensitive-finding-redaction
```

heditar:
> Tests: I tested the masking behavior for findings of type credential and found some issues:
> 
> ✅ Working as expected:
> 
> * The server response for fetching findings correctly masks the credential values.
> * The front-end displays the masked values as expected.
> 
> ❌ Issues found:
> 
> * Search bypasses masking : Even though the value isn't displayed (masked) in the UI, I can still search for the specific (unmasked) value in the search bar and get results. The masking isn't enforced at the search level.
> 
> image image
> * Reports expose unmasked values: When generating a report for this scenario, the credential values appear unmasked in the report output.
> 
> image
> * Masking in Attack path and findings is not the same:
> 
> image image

I centralized the logic and now all the screen use the same logic. I am opening 2 issues one for the report and one for the findings as the PR is already quite complexe.

savacano28:
Code: OK
Tests:
* Execution: OK  
* Export/import: OK
* Variables: OK. 
   Note: Variable values are still exposed in the attack path.
  image
  image
* Findings: OK. 
Question: Is this level of masking expected for small values?
  image
  image

savacano28 (approved):

PR #7598, opened 2026-08-25
fix(multi-tenancy): implement tenant auto-assign groups for user creation and updates (#7548)

### LINKED ISSUE #7548: fix(multitenancy): auto-assign ignores tenant default groups when a user is created from the platform screen
## Description

Default group auto-assignation is only applied for the scope of the
screen the creation was triggered from, and the platform screen never looks at the tenants
attached to the new user.

- Creation from a **tenant** screen
  → the tenant's auto-assign groups are applied. Works.
- Creation from the **platform** screen
  → only *platform-scoped* auto-assign groups are applied. The auto-assign groups of the
  tenants selected in the form are never applied.

Consequence: a user created from the platform screen and attached to tenant T lands in T
with no group at all (unless a platform-scoped auto-assign group exists), so no access.

## Reproducible steps

1. In tenant `T`, Settings > Security > Groups: create a group with "Auto assign" enabled
   (tenant-scoped group).
2. Settings > Security > Users **inside tenant `T`**: create user `A`.
   → `A` is a member of the auto-assign group. OK.
3. Platform > Users: create user `B` and select tenant `T` in the "Tenants" field.
   → `B` has no group from `T`.
4. Same result when attaching `T` afterwards by editing `B` from the platform screen.

## Expected output

Creation (and tenant attachment) from the platform screen assigns platform-scoped auto-assign
groups **plus** the auto-assign groups of every attached tenant.

## Actual output

Only platform-scoped auto-assign groups are assigned; tenant auto-assign groups are skipped.

GaetanSantucci:
This pull request enhances the user group assignment logic in the `UserService` to ensure that users are automatically assigned to all applicable platform and tenant-scoped auto-assign groups both on creation and update. It also adds comprehensive integration and unit tests to verify this behavior.

**User group auto-assignment improvements:**

* Refactored group assignment logic by introducing a private `syncAutoAssignGroups` method in `UserService`, which ensures that a user is assigned to all relevant platform and tenant auto-assign groups without removing existing group memberships. This method is now called during both user creation and update

### Testing Instructions

1. In tenant T, Settings > Security > Groups: create a group with "Auto assign" enabled (tenant-scoped group).
2. Settings > Security > Users inside tenant T: create user `A` → `A` is a member of the auto-assign group. OK.
3. Platform > Users: create user `B` and select tenant T in the "Tenants" field → `B` must have the group that was auto-assigned.
4. Same result when attaching T afterwards by editing `B` from the platform screen.

## Expected output
Creation (and tenant attachment) from the platform screen assigns platform-scoped auto-assign
groups plus the auto-assign groups of every attached tenant.

### Related issues

* Closes #7548 

### Checklist

### DIFF
--- a/openaev-api/src/main/java/io/openaev/service/UserService.java
+++ b/openaev-api/src/main/java/io/openaev/service/UserService.java
@@ -40,6 +40,7 @@ import jakarta.persistence.criteria.CriteriaBuilder;
 import jakarta.validation.constraints.NotBlank;
 import jakarta.validation.constraints.NotNull;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -172,7 +173,8 @@ public class UserService {
     if (StringUtils.hasLength(password)) {
       user.setPassword(this.encodeUserPassword(password));
     }
-    syncAutoAssignGroups(user);
+    // Creation enters every scope at once: platform, plus each tenant attached in the input.
+    assignAutoAssignGroups(user, user.getTenants().stream().map(Tenant::getId).toList(), true);
     User savedUser = userRepository.save(user);
     this.createUserToken(savedUser, token);
     return savedUser;
@@ -241,7 +243,14 @@ public class UserService {
         new ArrayList<>(
             referenceResolver.resolve(
                 input.tenantIds(), Tenant.class, tenantRepository::countByIdIn)));
-    syncAutoAssignGroups(existing);
+    // Only tenants the user just joined trigger auto-assignment: re-applying it to tenants he
+    // already belonged to would restore groups deliberately removed from within those tenants.
+    List<String> attachedTenantIds =
+        existing.getTenants().stream()
+            .map(Tenant::getId)
+            .filter(tenantId -> !oldTenantIds.contains(tenantId))
+            .toList();
+    assignAutoAssignGroups(existing, attachedTenantIds, false);
     User savedUser = userRepository.save(existing);
     // Evict cache for old tenants (removed memberships) and new tenants (added memberships)
     List<String> newTenantIds = input.tenantIds() != null ? input.tenantIds() : List.of();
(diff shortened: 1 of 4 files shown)


openaev-api/src/main/java/io/openaev/service/UserService.java
```diff
@@ -548,4 +548,27 @@ public static PreAuthenticatedAuthenticationToken buildAuthenticationToken(
   public Optional<User> findByEmailIgnoreCase(String email) {
     return userRepository.findByEmailIgnoreCase(email);
   }
+
+  /**
+   * Ensures the user holds every default-assign group applicable to their current scope:
+   * platform-scoped auto-assign groups, plus the auto-assign groups of each tenant they are
+   * attached to. Never removes an existing group membership.
+   */
+  private void syncAutoAssignGroups(User user) {
```
RomuDeuxfois: **Edge case**: remove a user from a tenant group that has default assignation enabled, then update that user's phone number right after. The removed group is silently added back. This sync should only run when a tenant is attached or detached.

openaev-api/src/main/java/io/openaev/service/UserService.java
```diff
@@ -548,4 +548,27 @@ public static PreAuthenticatedAuthenticationToken buildAuthenticationToken(
...
+    Specification<Group> spec = GroupSpecification.defaultUserAssignablePlatform();
+    for (Tenant tenant : user.getTenants()) {
+      spec = spec.or(GroupSpecification.defaultUserAssignableTenant(tenant.getId()));
+    }
+    List<Group> applicableGroups = groupRepository.findAll(spec);
```
RomuDeuxfois, line 594: **nitpick**: can be good to use a service instead of a repository here

openaev-api/src/main/java/io/openaev/service/UserService.java
```diff
@@ -171,9 +172,7 @@ private User createUser(User user, String password, String token) {
     if (StringUtils.hasLength(password)) {
       user.setPassword(this.encodeUserPassword(password));
     }
-    List<Group> assignableGroups =
-        groupRepository.findAll(GroupSpecification.defaultUserAssignablePlatform());
-    user.setGroups(assignableGroups);
+    syncAutoAssignGroups(user);
```
RomuDeuxfois: **nitpick**: assignDefaultTenantGroups seems to duplicate your sync logic in TenantUserService line 71.

RomuDeuxfois (approved):

PR #7606, opened 2026-08-25
fix(challenge): translate challenge's files from jsx to tsx

camrrx:
### Proposed changes

* Translate challenge's files from jsx to tsx

### Testing Instructions

1. Try all the actions related to challenges

### Related issues

* Closes #ISSUE-NUMBER


openaev-front/src/admin/components/components/challenges/ChallengeForm.tsx
```diff
@@ -0,0 +1,492 @@
...
+                  <DeleteOutlined />
+                </IconButton>
+              </ListItem>
+            ))}
+          </List>
+          {flagFields.length === 0 && (
+            <Typography variant="body2">
+              {t('At least one flag is required for a challenge.')}
+            </Typography>
+          )}
```
GaetanSantucci: We need to improve the required fields, because they're required but not visible unless we expand/unfold the flag section.

camrrx: Yes, as discussed, I added a default flag that cannot be deleted because we need at least one flag to validate the form. We can still add additional flags if needed.

GaetanSantucci (approved):
test ok

PR #7629, opened 2026-08-27
fix(rabbitmq): bound a publish so it cannot hold a transaction open (#7630)

RomuDeuxfois:
### Proposed changes

`Executor.executeExternal` publishes inside the inject's transaction, and nothing bounded that publish. The shared `ConnectionFactory` carried no timeout at all, and a broker under a memory or disk alarm accepts the connection then stops draining its sockets, so `basicPublish` blocks with no timeout that applies — a socket write has no write timeout in Java. The job thread parks, the transaction keeps its row locks, the Hikari pool drains and the platform stops answering, session store included. Four occurrences in four days, ~21 h of downtime.

* `RabbitmqDriver.createConnectionFactory()` gets a connection timeout, a handshake timeout and a heartbeat — covers an unreachable broker
* `RabbitmqService.publish()` runs on a bounded executor and gives up after `openaev.rabbitmq.publish-timeout-ms` (30 s) — covers a blocked one

On timeout the worker stays parked, since the write cannot be interrupted, but the caller returns: the transaction rolls back and the inject fails instead of the platform. The pool is capped at 8 threads so a stuck broker strands a bounded number of them.

This is a containment fix. Publishing outside the transaction entirely is `issue-external-executor-phased-transactions.md`.

### Checks done

`RabbitmqServicePublishTest` (new, 3 cases: broker that never answers, healthy broker, broker error surfaced rather than swallowed as a timeout), `ExecutorTest` and `HealthCheckServiceTest` pass — 11 tests. The timeout case pins the point that matters: the caller comes back on its own while the mocked broker is still hanging.


RomuDeuxfois:
Closes #7630

RomuDeuxfois:
Superseded by #7631, which carries the same work on a correctly named branch. This PR's head ref was detached by a branch rename and can no longer be updated.

Issue #7630, opened 2026-08-27
fix(rabbitmq): a publish that never returns holds the inject transaction open and freezes the platform


RomuDeuxfois:
## Description

Sending an inject to RabbitMQ happens inside its database transaction. If the broker stops answering the send never returns, so the transaction stays open and holds a database connection. Once the pool is empty the platform stops responding, login included, since sessions live in that same database.

Four times in four days, ~21 h of downtime.

## Reproducible steps

1. Make the broker accept connections but stop reading them. A memory or disk alarm does it.
2. Dispatch an inject on an external injector.

## Expected output

The send gives up, the inject fails, the platform is unaffected.

## Actual output

`basicPublish` blocks forever — Java has no socket write timeout. Users get `HikariPool-1 - Connection is not available`. Restarting does not help: PostgreSQL keeps the dead pod's locks.

## Additional information

`pg_stat_activity` fingerprint: `idle in transaction` on `Client:ClientRead`, last query `findByInjectId`, silent under a millisecond after it returned.

PR #7631, opened 2026-08-27
fix(rabbitmq): bound a publish so it cannot hold a transaction open (#7630)

RomuDeuxfois:
### Proposed changes

Sending an inject to RabbitMQ happens inside its database transaction, and nothing made that send give up. Java has no socket write timeout, so a broker that stops reading parks the thread forever, holding a connection until the pool is empty and the platform stops answering.

* `publish()` runs on a separate thread and gives up after `openaev.rabbitmq.publish-timeout-ms`, 30 s. Only the inject fails.
* Connection timeout 60 s → 10 s.

Contains the damage, does not remove the cause (#7630).

### Checks done

`ExecutorTest`, `HealthCheckServiceTest` and `SimulationInjectApiTest` pass. Nothing tests the timeout itself.


openaev-api/src/main/resources/application.properties
```diff
@@ -130,6 +130,8 @@ openaev.rabbitmq.management-port=15672
 openaev.rabbitmq.queue-type=classic
 # Whether or not the calls to the management plugin of rabbitmq can be insecure
 openaev.rabbitmq.management-insecure=true
+# Hard cap on a single publish, so a broker that stops answering cannot hold a transaction open
+openaev.rabbitmq.publish-timeout-ms=30000
```
gabriel-peze, line 134: Does this new property should be set and describe in the doc ?

openaev-model/src/main/java/io/openaev/service/RabbitmqService.java
```diff
@@ -138,6 +189,25 @@ public void publish(String injectType, String publishedJson)
...
+              return thread;
+            });
+  }
+
+  @PreDestroy
+  void shutdownPublishExecutor() {
+    if (publishExecutor != null) {
+      publishExecutor.shutdownNow();
+    }
+  }
```
gabriel-peze: nitpick : Post and Pre methods should be on top of the file to directly see them at first read.

openaev-model/src/main/java/io/openaev/service/RabbitmqService.java
```diff
@@ -115,6 +132,40 @@ public void publish(String injectType, String publishedJson)
...
     }
 
+    Future<Void> pending =
+        publishExecutor.submit(
+            () -> {
+              doPublish(injectType, publishedJson);
+              return null;
+            });
+    try {
+      pending.get(publishTimeoutMs, TimeUnit.MILLISECONDS);
```
gabriel-peze, line 208: Before using a value from the properties you should verify his value.
If a negative value is set the timeout will be triggered instantly at call.

openaev-model/src/main/java/io/openaev/service/RabbitmqService.java
```diff
@@ -115,6 +132,40 @@ public void publish(String injectType, String publishedJson)
...
+      Throwable cause = ex.getCause();
+      if (cause instanceof IOException ioCause) {
+        throw ioCause;
+      }
+      if (cause instanceof TimeoutException timeoutCause) {
+        throw timeoutCause;
+      }
+      throw new IOException("Failed to publish to RabbitMQ", cause);
+    }
+  }
```
gabriel-peze, line 231: There is no test for the whole process, i think it's a good idea to take the opportinity to add some.

RomuDeuxfois: I added a test, let me know if it's ok for you :)

RomuDeuxfois:
@gabriel-peze PR ready for a second round

gabriel-peze (approved):

PR #7637, opened 2026-08-27
feat(multi-tenancy): differentiate tenant-scoped user creation (#7613)

### LINKED ISSUE #7613: fix(multi-tenancy): a tenant admin creating a user grants them the platform-wide auto-assign groups
## Description

A user created inside a tenant should get that tenant's auto-assign groups only: a tenant administrator manages their own tenant and must not be able to hand out platform-wide groups. Today such a user also lands in every platform group flagged "Auto assign" — groups the tenant administrator can neither see nor manage. When one of them carries platform-level permissions, creating a user inside a tenant grants platform-wide access, decided by someone with no authority over it.

Side effect: the platform administrator's auto-assign group fills up with users they never created and cannot see, since a tenant's users are not listed on the platform screen.

## Reproducible steps

1. Platform > Security > Groups: create a group with "Auto assign" enabled.
2. Enter tenant T > Settings > Security > Users: create user A.
3. Open the platform group's members.

## Expected output

A holds T's auto-assign groups only; the platform group's membership is unchanged.

## Actual output

A is a member of the platform auto-assign group.

GaetanSantucci:
### Proposed changes

* Introduced tenant-aware user creation flow so that user provisioning behavior is explicitly scoped to the current tenant context.
* Differentiated user creation logic between tenant-scoped and non-tenant/global contexts to avoid ambiguous ownership and cross-tenant side effects.
* Updated related service/controller handling to consistently propagate and enforce tenant scope during user creation operations.

### Testing Instructions

1. Prepare at least two tenants in a test environment (e.g., `tenant-a` and `tenant-b`) with appropriate admin/operator roles.
2. From `tenant-a` context, create a new user and verify:
   - user is created successfully,
   - user is attached only to `tenant-a`,
   - no visibility/association appears in `tenant-b`.
3. From `tenant-b` context, create a new user and verify equivalent tenant isolation behavior.
4. (If supported) Execute user creation in a non-tenant/global context and verify expected fallback behavior is preserved.
5. Validate error cases:
   - missing/invalid tenant scope,
   - unauthorized actor attempting tenant-scoped creation,
   - duplicate/conflicting identity in same tenant rules.
6. Run automated tests covering user creation and tenancy boundaries; ensure no regressions in existing user provisioning flows.

### Related issues

* Closes #7613
* Related to multi-tenancy user management hardening

### Checklist

### Further comments

This change focuses on making tenant boundaries explicit at user creation time to reduce accidental cross-tenant coupling and improve predictability in multi-tenant environments.  
The implementation favors explicit scope propagation and validation over implicit inference, which makes the behavior easier to reason about and safer for future extensions.

### DIFF
--- a/openaev-api/src/main/java/io/openaev/service/UserService.java
+++ b/openaev-api/src/main/java/io/openaev/service/UserService.java
@@ -131,7 +131,6 @@ public class UserService {
     return createUser(input, input.tenantIds(), true);
   }
 
-
   @Transactional(rollbackFor = Exception.class)
   public User createTenantUser(UserInput input) {
     return createUser(input, List.of(), false);
--- a/openaev-api/src/main/java/io/openaev/service/UserService.java
+++ b/openaev-api/src/main/java/io/openaev/service/UserService.java
@@ -128,6 +128,16 @@ public class UserService {
 
   @Transactional(rollbackFor = Exception.class)
   public User createUser(UserInput input) {
+    return createUser(input, input.tenantIds(), true);
+  }
+
+
+  @Transactional(rollbackFor = Exception.class)
+  public User createTenantUser(UserInput input) {
+    return createUser(input, List.of(), false);
+  }
+
+  private User createUser(UserInput input, List<String> tenantIds, boolean includePlatformScope) {
     if (!StringUtils.hasLength(input.plainPassword())) {
       throw new IllegalArgumentException("Password is required when creating a user");
     }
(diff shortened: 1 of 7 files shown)


openaev-api/src/main/java/io/openaev/service/UserService.java
RomuDeuxfois, line 171: **TODO**: since every method calling this one are calling attachTenant just after, I am wondering if we need to put a false on createUser, OR do a little refacto and having just two way to create a user: tenant or platform, not internal.

openaev-api/src/main/java/io/openaev/service/UserService.java
```diff
@@ -128,6 +128,15 @@ public long globalCount() {
 
   @Transactional(rollbackFor = Exception.class)
   public User createUser(UserInput input) {
+    return createUser(input, input.tenantIds(), true);
+  }
+
+  @Transactional(rollbackFor = Exception.class)
+  public User createTenantUser(UserInput input) {
```
RomuDeuxfois: **todo**: seems weird to have createTenantUser inside userservice to call just after createUser. Why not call direclty createUser ?

gabriel-peze (approved):

PR #7645, opened 2026-08-28
fix(dev): stop a trailing comment becoming the RabbitMQ hostname (#7644)

### LINKED ISSUE #7644: fix(dev-setup): application-dev.properties.example has a trailing comment that Spring reads as part of the RabbitMQ hostname
### Symptom

A fresh developer following `openaev-dev/README.md` cannot start the backend. It
fails during startup with:

```
Caused by: java.net.UnknownHostException: host.docker.internal # you can also set it to localhost, but in case you wish to have injectors installed locally through your docker containers, you must use
```

The hostname in the exception is the property value **plus the trailing comment**.

### Cause

`openaev-api/src/main/resources/application-dev.properties.example`, line 226:

```properties
openaev.rabbitmq.hostname=host.docker.internal # you can also set it to localhost, but in case you wish to have injectors installed locally through your docker containers, you must use host.docker.internal
```

In a `.properties` file `#` only starts a comment at the beginning of a line.
Everything after `=` is the value, trailing comment included, so the hostname
becomes the whole sentence.

This is the documented first-run path: README step 2 says to copy this example
verbatim, so every new environment hits it.

### Fix

Move the note above the property:

```properties
# Set to localhost for a plain local run. Use host.docker.internal if you want
# injectors installed locally through your Docker containers to reach the broker.
openaev.rabbitmq.hostname=localhost
```

### Scope

Only this line parses a trailing comment into a value today; the rest of the file
puts its notes on their own lines. Worth a quick scan of the other
`*.properties.example` files for the same shape.

sandyghs:
`.properties` only treats `#` as a comment at the start of a line, so the note written after the value was parsed as part of it. A backend started from a fresh copy of this example dies with:

```
Caused by: java.net.UnknownHostException: host.docker.internal # you can also set it to localhost, but in case you wish to have injectors installed locally through your docker containers, you must use
```

README step 2 tells a new developer to copy this file verbatim, so every fresh environment hits it. The note moves above the property, and the value becomes `localhost` — the plain local run the README's minimal start describes.

Found while standing up a dev environment from scratch. Scanned the other `*.properties.example` files for the same shape: this was the only one.

Closes #7644


openaev-api/src/main/resources/application-dev.properties.example
```diff
@@ -223,7 +223,10 @@ executor.openaev.binaries.version=prerelease
 # INJECTORS #
 #############
 openaev.rabbitmq.prefix=openaev
-openaev.rabbitmq.hostname=host.docker.internal # you can also set it to localhost, but in case you wish to have injectors installed locally through your docker containers, you must use host.docker.internal instead and then point this domain to 127.0.0.1 in your /etc/hosts
+# Set to localhost for a plain local run. Use host.docker.internal if you want
+# injectors installed locally through your Docker containers to reach the broker;
+# in that case, point that name at 127.0.0.1 in your /etc/hosts.
+openaev.rabbitmq.hostname=localhost
```
RomuDeuxfois: Developers run OpenAEV with injectors and collectors most of the time so having host.docker.internal in the first place make more sense.

RomuDeuxfois (changes requested):

openaev-api/src/main/resources/application-dev.properties.example
sandyghs: Done — host.docker.internal is back as the default, and the fix now only targets the real bug

RomuDeuxfois (approved):

sandyghs:
superseded by #7565

PR #7648, opened 2026-08-28
fix(scheduler): run inject finalization in its own job (#7647)

### LINKED ISSUE #7647: fix(scheduler): injects stay stuck in PENDING and simulations never close when a dispatch hangs
## Description

The three finalization steps run at the tail of `InjectsExecutionJob.execute()`. When a dispatch hangs that line is never reached, and `@DisallowConcurrentExecution` stops Quartz from starting another cycle, so nothing is finalized again until the API is restarted.

## Expected output

Injects that never reported back are timed out and their simulations close, regardless of another dispatch being stuck.

## Actual output

They stay `PENDING` and `RUNNING` indefinitely, with nothing surfaced as an error. 12 h on staging, noticed the next day.

RomuDeuxfois:
### Proposed changes

The three finalization steps sit at the tail of `InjectsExecutionJob.execute()`. If a dispatch hangs, that line is never reached and `@DisallowConcurrentExecution` blocks every later cycle: injects stay `PENDING`, simulations stay `RUNNING`.


damgouj (approved):
Tested OK for me, you just have a red build ;)

PR #7665, opened 2026-08-28
fix(deps): update the Elasticsearch client to 9 and require an Elasticsearch 9 cluster

RomuDeuxfois:
### Proposed changes

**After this change OpenAEV can no longer run against an Elasticsearch 8 cluster.** The 9.x Java client sends a content type the 8.x server refuses, so every engine call fails and the platform does not start. Existing deployments have to upgrade their cluster to Elasticsearch 9 first — the added migration page walks through it.

* Client 8.19 -> 9.5.1, dev and test images with it
* `ElasticDriver` moves to `Rest5Client`: the 9.x client drops the HttpClient 4 transport, so auth, the trust-all SSL context and preemptive basic auth are rebuilt on HttpClient 5
* OpenSearch deployments are unaffected

### No data migration

Elasticsearch holds nothing PostgreSQL does not — all 13 `Handler` implementations read through a JPA repository, and the full rebuild path is already exercised by migrations like `V4_16__Force_es_reindex_injects`.

An empty 9.x cluster is therefore a valid target: indices, templates and the lifecycle policy are recreated at startup and reindexed. That also sidesteps the 7.x-created indices the Upgrade Assistant flags. The cost is reindex time, not risk.

None of the Elasticsearch 9 breaking changes apply to us.

### DIFF
--- a/openaev-model/src/main/java/io/openaev/driver/ElasticDriver.java
+++ b/openaev-model/src/main/java/io/openaev/driver/ElasticDriver.java
@@ -13,7 +13,10 @@ import co.elastic.clients.elasticsearch.indices.put_index_template.IndexTemplate
 import co.elastic.clients.json.JsonData;
 import co.elastic.clients.json.jackson.JacksonJsonpMapper;
 import co.elastic.clients.transport.ElasticsearchTransport;
-import co.elastic.clients.transport.rest_client.RestClientTransport;
+import co.elastic.clients.transport.TransportUtils;
+import co.elastic.clients.transport.rest5_client.Rest5ClientTransport;
+import co.elastic.clients.transport.rest5_client.low_level.Rest5Client;
+import co.elastic.clients.transport.rest5_client.low_level.Rest5ClientBuilder;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.openaev.config.EngineConfig;
 import io.openaev.database.model.IndexingStatus;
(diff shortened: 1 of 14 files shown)


openaev-model/src/main/java/io/openaev/driver/ElasticDriver.java
```diff
@@ -73,45 +73,56 @@ public void setSearchEngine(EngineContext searchEngine) {
...
+          connectionManager
+              .setMaxConnTotal(config.getMaxConnections())
+              .setMaxConnPerRoute(config.getMaxConnections());
+          if (!config.isRejectUnauthorized()) {
+            // Trust any server certificate, and any mismatch between it and the requested host.
+            // The hostname verifier only gets a say under the CLIENT policy: the default policy
+            // delegates verification to the SSLEngine, which ignores it entirely.
+            connectionManager.setTlsStrategy(
+                ClientTlsStrategyBuilder.create()
+                    .setSslContext(TransportUtils.insecureSSLContext())
```
antoinemzs: Wild legacy stuff! Perhaps nowadays it's better to initialise a SslContext object with our custom trust manager (trusting specific extra certs according to config); example: https://github.com/OpenAEV-Platform/openaev/blob/06847e6a6af2e7fc707e2415243060058caa4460/openaev-api/src/main/java/io/openaev/authorisation/HttpClientFactory.java#L28-L29

`trustManager` here is a bean, so it can be injected in this present singleton.

Note doing so deviates from the previous default (no validation), so TLS-based setups that didn't bother setting up the TLS trust chain between OpenAEV and Elastic will break; still perhaps this major version jump could be the right time to introduce this change.

RomuDeuxfois: Done in last commits.
The driver now builds its SSLContext from the trustManager bean, same as HttpClientFactory.                  
I kept engine.reject-unauthorized=false trusting anything, unchanged.                                                        

Two things to have in mind: RabbitMQSslConfiguration falls back to useSslProtocol(), which installs amqp-client's TrustEverythingTrustManager. And OpenSearchDriver, Minio, etc still read the JVM store only...

So we need to align everyone.

corinnekrych:
I've been testing the PR:
* on fresh dev DB / new container => it's working fine
* tested with a dump DB from `main-ff` to measure how long the indexing could take and if it goes fine. So far:
Screenshot 2026-09-02 at 09 19 36

**~1h40m of actual compute time** for the full `inject` backfill (largest model, 193k rows) on this local dev setup.
Something to discuss with platform team (added to today agenda) do we need a pre-warming procedure to pre-index ES9 for smooth transition?

corinnekrych (changes requested):
Blocking for now, we're still discussing the options for migration

antoinemzs:
Suggest closing: #7745 would answer the issue better for it includes both versions of the ES client; also, index storage update should be done as an admin routine on the ES server itself, not by deleting and rebuilding the indexes from the app, see:https://www.elastic.co/docs/deploy-manage/upgrade/deployment-or-cluster/elasticsearch

PR #7688, opened 2026-08-31
fix(chaining): ask for the scope instead of ending an unscoped autonomous run (#7684)

### LINKED ISSUE #7684: fix(chaining): autonomous Attack Chaining is ending immediately
## Description

When user start any autonomous attack chaining simulation, simulation, created, run and then ended immediately, it's not asking scope, it's not creating any logic. 

## Environment

1. main-ff and Demo 
## Reproducible steps

1. Create a chaining scenario
2. run Autonomous button
3. Wait 3-5 second and it will be completed. 

## Expected output

If you don't have any scope it should ask scope, if you have scope it should create a logic according to your agent. 
## Actual output

Completed directly.

## Additional information

https://github.com/user-attachments/assets/20685721-4701-476c-8975-77d070a5ddfe

heditar:
## Description

Fixes #7684 — an autonomous attack chaining run ended immediately after launch, without ever asking for a scope.

### Root cause

`AutonomousRunService.doStart` engaged the orchestrator even when the run had no perimeter at all. With nothing to act on, the still-empty simulation read `FINISHED` right away and `reconcileWithSimulation` flipped the run to `COMPLETED` a few seconds after the launch. Guards against that premature `FINISHED -> COMPLETED` sync already existed for plan mode and `WAITING_INPUT`, but not for an unscoped live launch.

### Fix

- **`doStart`** — a live run with no scope (no entity target on the run and no allow-list rule on its simulation workflow) is no longer handed to the orchestrator. It parks in `WAITING_INPUT` and emits a `QUESTION` event, so the cockpit chat asks *"Which targets are in scope?"*. `WAITING_INPUT` is already exempt from the reconcile and from the stall watchdog, so the parked run stays alive instead of completing.
- **`addDirective`** — answering that question engages the orchestrator that was never started (it previously only tried to wake a session that did not exist).
- **`applyLiveConfiguration`** — defining the scope in the Scope tab resumes the parked run the same way, so both surfaces answer the question.
- **`WorkflowService.hasAllowlistScopeForSimulation`** — new read used to detect a run with no perimeter (it covers both the launch stepper's rules and the scenario's own scope, which are seeded onto the simulation workflow at creation).

No frontend change: the reasoning panel already renders a pending `QUESTION` with a reply composer while the run is `WAITING_INPUT`.

## Tests

- New `@Nested` group `Launching a live run without a scope` in `AutonomousRunServiceTest` (5 tests): unscoped launch parks and asks + never engages, a run scope target or a workflow allow-list still engages, answering the question engages, a directive on an already engaged run still only wakes it.
- `AutonomousRunServiceTest`, `AutonomousRunServiceScopeTest`, `AutonomousEventServiceTest`, `AutonomousRunAccessControlTest`, `AutonomousRunReconciliationWriterTest`, `WorkflowServiceTest` all pass; `mvn spotless:apply` clean.

## Checklist

- [x] Linked to an issue (#7684)
- [x] Tests added
- [x] Spotless / compile verified locally
- [ ] Manual validation on main-ff / Demo (launch an autonomous chaining scenario with no scope and check the cockpit asks for it)

### DIFF
--- a/openaev-api/src/main/java/io/openaev/service/autonomous/AutonomousRunService.java
+++ b/openaev-api/src/main/java/io/openaev/service/autonomous/AutonomousRunService.java
@@ -162,6 +162,19 @@ public class AutonomousRunService {
   private static final String ENGAGED_EVENT_DATA = "{\"phase\":\"engaged\"}";
 
   /**
+   * Title of the QUESTION event OpenAEV itself raises when a live run is launched with no perimeter
+   * at all. The run parks in WAITING_INPUT on it and the cockpit renders it as a pending question
+   * with a reply composer, so the operator answers the scope in the chat and the run then starts.
+   */
+  private static final String SCOPE_QUESTION_TITLE = "Which targets are in scope?";
+
+  /** Body of {@link #SCOPE_QUESTION_TITLE}. */
+  private static final String SCOPE_QUESTION_BODY =
+      "This run has no perimeter yet, so nothing can be attacked. Tell me which targets I am"
+          + " authorized to attack - assets, asset groups, teams, persons, IPs or hostnames - and I"
+          + " will start right away. You can also define them in the Scope tab.";
+
+  /**
    * Default OpenAEV-enforced run timeout when the launcher does not specify one: 24 hours.
    * Autonomous runs are long-lived (recon, waiting on human-in-the-loop injects, slow exploitation)
    * so the default is far larger than the 1h chained-scenario workflow timeout.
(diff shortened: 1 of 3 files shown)


heditar:
### PR risk classification
Medium Risk — This PR changes backend business logic in 2 production files (`AutonomousRunService`, `WorkflowService`), altering the autonomous run start/state-transition flow, but the scope is narrow (an unscoped-launch guard plus resume paths) and it introduces no data migration or destructive data impact.

heditar:
### PR risk classification
High Risk — This PR changes core platform behavior in the autonomous run lifecycle (run start state machine, WAITING_INPUT parking, and orchestrator engagement on directives/live scope edits), so a regression would directly affect whether runs execute at all. Although only 3 files are touched, the logic sits on a critical execution path rather than being narrowly scoped.

heditar:
### PR risk classification
Medium Risk — This PR changes autonomous-run lifecycle logic in only 2 production files (`AutonomousRunService`, `WorkflowService`) with a narrowly scoped guard on the start/reconcile state transition; it is backed by new unit tests and has no impact on existing stored data.

heditar:
### PR risk classification
Medium Risk — This PR changes backend business logic in 2 source files (autonomous run scoping and workflow scope rules) plus tests, a narrowly scoped behavioral change; no data migrations or security-critical flows are affected.

heditar:
### PR risk classification
Medium Risk — This PR changes autonomous run start behaviour (parking an unscoped live run in WAITING_INPUT and engaging the orchestrator on the operator reply) in 2 production files plus tests; it touches business logic but stays narrowly scoped to the unscoped-launch path with no data migration.

heditar:
### PR risk classification
High Risk — This PR changes core autonomous run lifecycle logic (start/status transitions, directive handling, live configuration) and introduces the attack-perimeter/allow-list gate in WorkflowService, which is security-relevant core platform behavior.

heditar:
### PR risk classification
High Risk — This PR changes core platform behavior: the autonomous run lifecycle/state machine (launch, WAITING_INPUT parking, orchestrator engagement) in AutonomousRunService and WorkflowService, which drives critical execution flows.

heditar:
### PR risk classification
High Risk — This PR changes the autonomous run start state machine and the orchestrator engagement path in `AutonomousRunService`/`WorkflowService`, which is core platform behavior even though only 3 files are touched. A regression here would directly affect how live runs are launched and resumed.

Issue #7625, opened 2026-08-27
fix(scope): Adding an inject from the Threat Arsenal to multiple scenarios at once fails with an Internal Error (HTTP 500)


DeniseMonreale:
<img width="1445" height="926" alt="Image" src="https://github.com/user-attachments/assets/f15076ff-4e8c-4697-b140-5924df99b431" />

Steps to reproduce:
1. Open the Threat Arsenal
2. Select one or more injector contracts 
3. Use "Add to scenario" and select one or multiple target scenarios at once
4. Confirm  the request fails with a 500 Internal Server Error.

Expected behavior: The selected inject(s) should be created and attached to every selected scenario, as when adding to a single scenario.

Impact: blocks a core Threat Arsenal workflow (reusing a template across multiple scenarios at once) with no workaround other than adding the inject to each scenario individually

PR #7695, opened 2026-09-01
fix(threats): add in bulk to scenario (#7625)

gabriel-peze:
### Proposed changes

* Fix the threat arsenal bulk process, allow the injects to be added to a scenario in bulk

### Testing Instructions

1. Go to the threat arsenal page
2. Select one or more injects
3. Click on "Run a test" on the bulk menu
4. Select "Add to an existing scenario"
5. Select one or more scenario
6. You should be redirected to the scenario, and all the selected injects should be added to the scenario

### Related issues

* Related #7625

### DIFF
--- a/openaev-front/src/admin/components/threat_arsenal/bulk/ThreatArsenalScenarioUpdateComponent.tsx
+++ b/openaev-front/src/admin/components/threat_arsenal/bulk/ThreatArsenalScenarioUpdateComponent.tsx
@@ -25,8 +25,8 @@ import {
   type InjectorContractSearchPaginationInput,
   type PageRawPaginationScenario,
   type RawPaginationScenario,
-  type Scenario,
   type ScenarioIdsAndInjectorContractsInputs,
+  type ScenarioSimple,
   type ThreatArsenalAction,
 } from '../../../../utils/api-types';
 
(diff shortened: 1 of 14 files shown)


openaev-front/src/admin/components/threat_arsenal/bulk/ThreatArsenalScenarioUpdateComponent.tsx
```diff
@@ -132,7 +132,7 @@ const ThreatArsenalScenarioUpdateComponent = ({
         injector_contract_ids_to_ignore: isExclusionMode ? Object.keys(deSelectedElements) : [],
       },
     };
-    updateScenariosWithInjectorContracts(inputs).then((result: AxiosResponse<Scenario[]>) => {
+    updateScenariosWithInjectorContracts(inputs).then((result: AxiosResponse<ScenarioSimple[]>) => {
```
damgouj, line 147: nitpick: add ScenarioSimple result for addScenarioWithInjectorContracts too

damgouj (changes requested):
Doesn't seem to work with chained scenarios but maybe it's normal or we could hide chained scenarios from the list to add injects ?
Works well with time-based scenarios thanks !

damgouj (approved):

Issue #6650, opened 2026-07-10
fix(scope): Bulk update on injects doesn't refetch grid — status/healthcheck shown stale ("missing content") until next data exchange


DeniseMonreale:
## 🐛 Bug description

After performing a **bulk update** on injects (e.g. "add targeted assets" via 
bulk action on a multi-select of Shodan/Nmap/Nuclei injects), most of the 
updated injects are incorrectly flagged as **"missing content"**  even 
though the assets have been correctly assigned server-side.

The issue is intermittent: the flag sometimes clears itself later, sometimes 
not, depending on unrelated subsequent data exchanges with the backend.

## 🔍 Root cause (diagnosed live with @GabrielPeze, MarineLeM squad sync)

- The bulk add action does a `POST` to the backend but **does not wait for 
  the response**.
- For performance reasons, the frontend does an **optimistic update**: it 
  injects the selected objects directly into the local grid state instead 
  of waiting for backend data.
- Inject status / "missing content" flag is **not a static field**  it is 
  **computed server-side via healthcheck**, which itself depends on DB state 
  (e.g. assigned assets).
- Because there is no round-trip with the backend after the bulk POST, the 
  healthchecks are **not recalculated**, and the grid keeps showing the 
  **stale status** (missing content) until an unrelated data exchange forces 
  a refresh.

## 🎯 Use case

bulk-adding targeted assets to a batch of injects in most 
injects being flagged "missing content", even though asset assignment is 
confirmed correct server-side. This blocks confident validation of 
expectation-of-result scenarios and creates false negative signals during 
testing.

## 💡 Suggested fix

After the bulk POST completes, **invalidate / refetch the grid** from the 
backend instead of optimistically injecting the local object into the list. 
This ensures the displayed status reflects the actual recalculated 
healthcheck state.

## 🔁 Related  same symptom, different root causes (recurring pattern)

This "missing content" symptom has recurred at least 3 times before, each 
time with a **different root cause**, and has never been tracked as a 
symptom family:

- #1601 bulk update team → status forced "enabled" without validation 
  (closed as wontfix / not reproducible)
- #2250  inject stuck on "Missing Content" after filling Asset/Asset Group, 
  triggered by scenario import (solved in 1.12.0)
- #3814 inject stuck on "missing content" despite valid default values, 
  triggered by import from OpenCTI/XTM Hub + mass edit (closed without fix, 
  not reproducible at the time)
[...]

PR #7699, opened 2026-09-01
fix(scenario): reload injects list after bulk modification (#6650)

gabriel-peze:
### Proposed changes

* Fix the injects status after a modification in bulk, made the list reload once the bulk is validated

### Testing Instructions

1. On a scenario modify one or more injects with the bulk process
2. The list should reload and refresh the injects status

### Related issues

* Related #6650


EvaE-Filigran:
Test OK - update asset that missing content with a bulk action change the error status and perform the action.

EvaE-Filigran (approved):

PR #7702, opened 2026-09-02
fix(chaining): fix phishing executor (#7626)

### LINKED ISSUE #7626: fix(chaining): Phishing email is not working for chaining
## Description

If user add any phishing action for chaining scenario, phishing action give error immediately.
## Environment

1. main-ff
2. https://main-ff.oaev.staging.filigran.io/2cffad3a-0001-4078-b0e2-ef74274022c3/admin/simulations/ff2143a3-7f06-4bd3-aadb-aaee3d1bf142/injects/e8dd9cb4-064b-4a83-9b21-925f7ab090d0?backlabel=Execution&backuri=%2Fadmin%2Fsimulations%2Fff2143a3-7f06-4bd3-aadb-aaee3d1bf142%2Fexecution%2Ftimeline&query=cGFnZT0wJnNpemU9MjAmZmlsdGVyR3JvdXAlNUJtb2RlJTVEPWFuZCZmaWx0ZXJHcm91cFtmaWx0ZXJzXVtdJmtleT1URUFNU19lOGRkOWNiNC0wNjRiLTRhODMtOWIyMS05MjVmN2FiMDkwZDBfZmlsdGVycw%3D%3D

## Reproducible steps

1. create chaining simulation
2. add phishing action
3. add team or user in the scope
4. run simulation
5. see the phishing action give error

## Expected output

Phishing actions should run in the chaining

Image

Image

EvaE-Filigran:
### Proposed changes

* Fix `phishing_results_inject_fk` FK violation: in chaining, a `PhishingResult` was created referencing an `Inject` still uncommitted in the ambient transaction. Added `phishing_result_step` (FK to `steps`, `ON DELETE CASCADE`) so chaining creates the tracking row against the already-persisted `Step` instead, backfilling `inject_id` on first read (`resolveByToken`) once it's committed.
* Threaded the step id through `ExecutableInject` and `InjectExecutionStep` so `PhishingExecutor` can pass it to `createResult`.

### Testing Instructions

1. Run a chaining simulation targeting a Team with at least one member whose email is invalid (bad domain/mailbox), alongside a valid recipient.
2. Confirm the valid recipient's phishing email is sent/tracked normally, with no `phishing_results_inject_fk` violation in the logs and no resend loop retriggered by the bad recipient.
3. Open the email, click the landing page link, and submit the credentials form for the valid recipient; confirm the open/click/submit events are correctly recorded on the `PhishingResult` row (via the tracking pixel, landing page visit, and form submission).
4. Delete the chaining simulation and confirm the corresponding `phishing_results` rows are cascade-deleted.
5. Run the same scenario (steps 1-4) through the classic time-based scheduler (non-chaining) to confirm no regression there, including cascade deletion on simulation delete.

### Related issues

* Related #7626 

### Checklist

### Further comments

The chaining resend-loop risk (unbounded retries re-sending to a whole team) is a separate, broader issue tracked independently — this PR only fixes the FK violation itself.


openaev-api/src/main/java/io/openaev/migration/V6_20260901180000000__Add_step_to_phishing_results.java
```diff
@@ -0,0 +1,40 @@
...
+          """);
+      statement.execute(
+          "CREATE INDEX IF NOT EXISTS idx_phishing_results_step "
+              + "ON phishing_results (phishing_result_step)");
+      statement.execute(
+          """
+          ALTER TABLE phishing_results
+              ADD CONSTRAINT phishing_results_step_or_inject_chk
+                  CHECK (phishing_result_step IS NOT NULL OR phishing_result_inject IS NOT NULL);
+          """);
```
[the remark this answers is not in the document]
EvaE-Filigran: fixed

openaev-api/src/main/java/io/openaev/migration/V6_20260901180000000__Add_step_to_phishing_results.java
```diff
@@ -0,0 +1,59 @@
...
+      statement.executeUpdate(
+          """
+          DO $$
+          BEGIN
+            IF NOT EXISTS (
+              SELECT 1 FROM pg_constraint WHERE conname = 'phishing_results_step_or_inject_chk'
+            ) THEN
+              ALTER TABLE phishing_results
+                ADD CONSTRAINT phishing_results_step_or_inject_chk
+                CHECK (phishing_result_step IS NOT NULL OR phishing_result_inject IS NOT NULL);
```
gabriel-peze, line 53: I think this script should handle the case where phishing_result_step and phishing_result_inject are both NULL.
Because in that case, the script will fail and block the platform at startup.

openaev-api/src/main/java/io/openaev/injectors/phishing/service/PhishingTrackingService.java
```diff
@@ -314,8 +330,58 @@ public void initializeExpectationsAsResisted(@NotBlank final String injectId) {
...
+    }
+    String data = result.getStep().getData();
+    if (data == null) {
+      return;
+    }
+    String injectId = StepService.getField(data, "inject_id");
+    if (injectId == null) {
+      return;
+    }
+    result.setInject(injectRepository.getReferenceById(injectId));
```
gabriel-peze: Since injectId came from an input from the frontend, should we check for the value before using it ?

In that case i think an EntityNotFoundException can be throw if the id doesn't exist in database for example

openaev-api/src/main/java/io/openaev/injectors/phishing/service/PhishingTrackingService.java
```diff
@@ -314,8 +330,58 @@ public void initializeExpectationsAsResisted(@NotBlank final String injectId) {
...
 
+  /**
+   * Resolves a tracking token, backfilling {@link PhishingResult#getInject} from {@link
+   * PhishingResult#getStep} when the row was created before its inject was committed (chaining
+   * execution, see {@code createResult}). The step's {@code data} JSON carries the inject id once
+   * {@code InjectExecutionStep#run} commits it - see {@code InjectExecutionStep#setInjectId}. Once
+   * backfilled, {@code inject} is kept as the queryable reference going forward; {@code step} is
+   * left untouched (still useful to correlate the producing step attempt).
+   */
   public Optional<PhishingResult> resolveByToken(@NotBlank final String token) {
```
gabriel-peze: Since there is some rules and save in the backfillInjectFromStep method, it's not only a resolver anymore, is it possible to review the name of this method ?

gabriel-peze (changes requested):

openaev-api/src/main/java/io/openaev/migration/V6_20260901180000000__Add_step_to_phishing_results.java
EvaE-Filigran: The field phishing_result_inject is already implemented with a FK, so in the current database state it could not be null. Does that answer your question?

openaev-api/src/main/java/io/openaev/injectors/phishing/service/PhishingTrackingService.java
EvaE-Filigran: The injectId came from our system; only the token came from external input.
Fixed by checking whether the inject exists. If not, a log will be displayed, but we do not stop the result persistence.
At this stage, it's possible that an inject has not yet been persisted in case of chaining.

openaev-api/src/main/java/io/openaev/injectors/phishing/service/PhishingTrackingService.java
EvaE-Filigran: changed by resolveAndBackfillByToken

gabriel-peze (approved):

Issue #3208, opened 2025-05-20
fix: inconsistent counting of endpoints in Endpoints list


antoinemzs:
## Description

Total number of items is inconsistent in the Endpoints list page: depending on which page is displayed, the total count may not be stable, eg. 30 on page 1 and down to 24 on page 2.

## Environment

1. OS (where OpenBAS server runs): linux
2. OpenBAS version: 1.16.1
3. OpenBAS client: web / API

## Reproducible Steps

Steps to create the smallest reproducible scenario:
1. Go to Endpoints list with more than 20 items
2. Note the total number advertised
3. Go to next page
4. Total number of items may have changed

## Expected Output

Stable total on all pages

## Actual Output

Inconsistent total

## Additional information
 
https://prerelease.obas.staging.filigran.io/admin/assets/endpoints?query=cGFnZT0wJnNpemU9MjAmc29ydHMlNUIwJTVEJTVCcHJvcGVydHklNUQ9YXNzZXRfbmFtZSZ0ZXh0U2VhcmNoPSZrZXk9YXNzZXQ%3D
 
Page 1:
![Image](https://github.com/user-attachments/assets/d5b801ed-5a35-4618-845a-5c06a402aee3)

Page 2:
![Image](https://github.com/user-attachments/assets/e23f2fa3-3523-42e2-9ada-e61e8bdba16d)

@guillaumejparis @Dimfacion for your information

antoinemzs:
The intention is to run this SQL
```sql
select a.* from assets a
	join agents ag on a.asset_id = ag.agent_asset
where a.asset_type = 'Endpoint'
	and ag.agent_parent is null and ag.agent_inject is null
group by a.asset_id;
```
But the server would somehow runs this
```sql
select a.* from assets a
	join agents ag on a.asset_id = ag.agent_asset
where a.asset_type = 'Endpoint'
	and ag.agent_parent is null and ag.agent_inject is null;
```

antoinemzs:
Related to: https://github.com/spring-projects/spring-data-jpa/pull/2376

antoinemzs:
We could rewrite the endpoint spec to use an (not?) `exists` clause instead of joining on agents, which would sidestep the issue with Spring Data and there being a `GROUP BY` clause.

PR #7703, opened 2026-09-02
fix(endpoints): stabilize search total across pages (#3208)

gabriel-peze:
### Proposed changes

* The count on the endpoints lists was inconsistent, it change to reflect the real amount of endpoints.

### Testing Instructions

1. Go to a list of endpoints (on a technical inject for example)
2. The displayed amount of elements should represent the real amount of endpoints on the platform

### Related issues

* Related #3208


damgouj (changes requested):
Tests KO when you perform an action on the endpoints list (add, delete,...)
Image

Otherwise, the code looks good to me and same for my other tests.

damgouj (approved):

PR #7722, opened 2026-09-02
feat(audit-log): introduce hard delete + insert of tokens when renew and audit (#7491)

Dimfacion:
### Proposed changes

* Introducing a hard delete as we're expecting it would be enough for our current need
*

### Testing Instructions

1. Renew a token
2. Check the audit log is logging a delete and a create of the token

### Related issues

* Related #7585 
*

### Checklist

### DIFF
--- a/openaev-api/src/main/java/io/openaev/rest/user/MeApi.java
+++ b/openaev-api/src/main/java/io/openaev/rest/user/MeApi.java
@@ -29,10 +29,8 @@ import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpSession;
 import jakarta.validation.Valid;
 import java.util.List;
-import java.util.UUID;
 import lombok.RequiredArgsConstructor;
 import org.springframework.http.ResponseEntity;
-import org.springframework.security.access.AccessDeniedException;
 import org.springframework.transaction.annotation.Transactional;
 import org.springframework.web.bind.annotation.*;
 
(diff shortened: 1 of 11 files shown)


openaev-api/src/main/java/io/openaev/runner/InitAdminCommandLineRunner.java
```diff
@@ -62,7 +62,8 @@ public void run(String... args) {
     User adminUser = adminUserOptional.map(this::updateUser).orElseGet(this::createUser);
 
     // Handle admin token
-    Optional<Token> adminToken = this.tokenRepository.findById(ADMIN_TOKEN_UUID);
+    Optional<Token> adminToken =
```
RomuDeuxfois, line 65: **question**: why changing this part ?

openaev-api/src/main/java/io/openaev/rest/user/MeApi.java
```diff
@@ -130,17 +128,7 @@ public User updatePassword(
   @AccessControl(skipRBAC = true, actionPerformed = Action.WRITE, resourceType = ResourceType.USER)
```
RomuDeuxfois, line 128: **question**: since we have a new resourcetype, should we use the resource type token here ?

Dimfacion: I'm unsure.  The only place where it would make sense is on /tokens endpoints and those are all with a skipRBAC = true. Besides, I feel it's more logical to keep the ResourceType to USER on these endpoints because it's always the token of the current user, never a specific token or the one of someone that is changed. 

That being said, we could add it. Like I said, we're always with skipRBAC=true so it wouldn't break the access control and I'm not strongly against the idea if you feel it would be clearer.

openaev-api/src/main/java/io/openaev/runner/InitAdminCommandLineRunner.java
Dimfacion: If the admin token changes or is renewed and the app is restarted, we would look for the admin token, not find it and create a new token linked to the admin meaning that it would have two tokens linked to him and it wouldn't be visible anywhere but in the DB and it would still give someone admin access while being hidden.

With this change, we overwrite the token of the admin user at the start of the app if it exists meaning that you would still have only one admin token.

openaev-api/src/main/java/io/openaev/rest/user/MeApi.java
RomuDeuxfois: Ok, Let's set that part aside for now

openaev-api/src/main/java/io/openaev/runner/InitAdminCommandLineRunner.java
RomuDeuxfois: Good point, thanks for the clarification

RomuDeuxfois (approved):

RomuDeuxfois:
LGTM :)

PR #7724, opened 2026-09-02
test(multi-tenancy): layer 1 add fail-closed runtime detector and baseline gate (#7723)

### LINKED ISSUE #7723: feat(Multi-tenancy): layer 1 runtime fail-closed detector and baseline gate
## Context

As multi-tenancy v2 activates tenant scoping table by table, a read of an active
table that runs without a tenant scope does not leak: it fails closed and returns
zero rows. That is safe for isolation but silently breaks features (empty screen,
no error), and unlike a leak it cannot be detected from the SQL shape. The real
past incidents were on background and lazy-association paths, not HTTP handlers.

This is layer 1 of the defense-in-depth around the TxCtx rollout: a runtime safety
net that catches these silent failures automatically.

## What this delivers

- A test-scope runtime detector at the JDBC level, so it observes every path
  (HTTP, background, lazy associations). It flags any statement the inspector
  rewrote with `can_access_tenant(...)` that ran under an empty scope.
- A per-test baseline gate: it fails a test when a NEW production path reads an
  active table without a scope. Current known-OK sites are recorded in a baseline
  file; test and fixture callers are auto-waived.
- A CI lane that runs the gate.

## Scope and safety

- Test-scope only, no production code change.
- Off by default (`-Dopenaev.failclosed.detector=on`); the full suite is unaffected
  when off (6169 tests unchanged).
- Baseline investigation over the current suite found zero live fail-closed bugs.

## Acceptance criteria

- [ ] Detector flags an unscoped read of an active table; a scoped read is not flagged.
- [ ] Gate fails on a new production fail-closed read; baselined and test sites are waived.
- [ ] Baseline of current known-OK production sites, with reasons.
- [ ] A CI lane runs the gate.
- [ ] Full openaev-api suite green with the detector off.

Part of #6391.

laugiov:
## What

Layer 1 of the defense-in-depth for the TxCtx rollout: a runtime safety net that
catches tenant reads which silently fail closed.

- A test-scope detector at the JDBC level (so it sees HTTP, background and lazy
  paths): it flags any statement the inspector rewrote with `can_access_tenant(...)`
  that ran under an empty tenant scope, which returns zero rows in production.
- A per-test baseline gate: it fails a test when a NEW production path reads an
  active table without a scope. Test and fixture callers are auto-waived; the
  current known-OK production sites are listed in `failclosed-baseline.txt`.

## No standing CI lane (following @corinnekrych's review)

The detector only fires on a table that a test activates, so a standing CI lane on the
tests we already have does not go further than the sharded suite, and it misses the real
regression class: a lazy load of an activated table from an unrelated API (TeamAPI ->
`getInjector()` on activated `injectors`) that no test activates on that path. So there is
no CI lane. The detector is used on demand: the one-time baseline audit (done), and under
the per-path tests the `activate-tenant-table` runbook already requires, where the real
coverage comes from.

## Why

A read of a v2-active table with no scope does not leak, it returns zero rows,
silently. Unlike a leak it is not detectable from the SQL shape, and the real past
incidents were on background / lazy-association paths, not HTTP handlers.

## Scope and safety

- Test-scope only, no production code changed.
- Off by default (`-Dopenaev.failclosed.detector=on`). Full openaev-api suite green
  with it off: 6169 tests, 0 failures.
- Baseline investigation over the suite found 32 gated unscoped reads, all triaged:
  zero live fail-closed bugs (either harmless eager-inits, or production paths that
  are correctly scoped and only reached unscoped by a test calling them directly).

## Run it

`mvn test -pl openaev-api -Dopenaev.failclosed.detector=on`

A new unscoped production read of an active table fails the test that hits it, with a
`[FAILCLOSED]` marker. Fix the scope, or, if intentional, add the `class.method` to
`failclosed-baseline.txt` with a reason.

Closes #7723. Part of #6391.


corinnekrych (commented):
tbh i am not sure we will catch the "lazy-association paths" error. Since the CI is running on test marked as `@SpringBootTest(properties = "openaev.tenant.active-tables=tags")` and for now most of those test are the isolation test we've done to test a API controller layer. This CI will not go much further to what we already have at integration test.

The regression path is very often a lazy loading for an activated table from an API path that comes from a complete different table. ie: from TeamAPI, there is a call to getInjector() that is v2 activated with lazy laoding. We will have created a InjectorApi with @SpringBootTest(properties = "openaev.tenant.active-tables=tags") and this test would run ok on both CI targets (plain sharded integration test target + your new target)
because we do not have a TeamAPI test with the activated-table pointing to inejctor.

Do we need this extra CI check?

laugiov:
@corinnekrych You're right, we don't need it. The lane only sees a table that a test already activates, so it does not go further than the sharded suite, and it misses exactly your case: TeamAPI -> getInjector on activated `injectors`, with no test activating it there.

So we'll drop the standing lane. The detector's real value was the one-time baseline (it found zero live bugs); going forward, the coverage that matters comes from the per-path tests the `activate-tenant-table` runbook already requires, not a CI job.

corinnekrych (approved):

PR #7732, opened 2026-09-02
fix(security): user recreation after deletion keeps old password (#6038)

### LINKED ISSUE #6038: fix(security): user recreation after deletion keeps old password
## Description
When deleting a user and create a new user with the same email as the old one. The old password is still configure on the new account.

## Environment

1. OS (where OpenAEV server runs): kube
2. OpenAEV version: 2.260603.0
3. OpenAEV client: frontend
4. Other environment details: filigran SaaS

## Reproducible Steps

1. connect with an admin account got to /admin/settings/security/users
2. create user a@a.fr with password 123
3. Disconnect and Connect to the user with password 123
4. Disconect reconect to admin and delete the a@a.fr account
5. Create a new account a@a.fr with password 456
6. Disconnect and try to connect with the account a@a.fr and password 456 it doesn't work but 123 will work
Expect to work with 456.

## Actual Output

Wrong password

## Additional information

tested on filigran main instance

damgouj:
### Proposed changes

New workflow:
* Hide the password field,
* When I create a user, if he doesn't exist, send a mail from "reset password" else do nothing,
* The actual behavior is not changed: when a user is created, we create it or add it to the given tenant.

### Testing Instructions

1. Create/update/delete users -> nothing has changed
2. Create a new user -> you get a mail instead of setting the password
3. Create an existing user -> nothing has changed, the user is attached to the tenant

### Related issues

* Related #6038

### Checklist


openaev-api/src/main/java/io/openaev/service/UserService.java
```diff
@@ -153,10 +155,17 @@ public User createUser(UserInput input, UserCreationScope scope) {
             referenceResolver.resolve(tenantIds, Tenant.class, tenantRepository::countByIdIn)));
     // The user's id is generated on save (UUID generator), not before: evict only after
     // persisting, using the saved user's id, or evictForUser is called with a null key.
-    User createdUser = createUser(user, input.plainPassword(), UUID.randomUUID().toString(), scope);
+    String password =
+        StringUtils.hasText(input.plainPassword())
+            ? input.plainPassword()
+            : randomUtils.getRandomAlphanumeric(64);
```
antoinemzs: password column is nullable, so it seems unnecessary to set a password if we lack on in the input, what do you think? A null password enforces the impossibility to perform the User+Password auth challenge.

openaev-api/src/main/java/io/openaev/service/UserService.java
```diff
@@ -329,7 +340,17 @@ public void delete(String userId) {
...
   public void requestPasswordReset(ResetUserInput input) {
-    Optional<User> optionalUser = userRepository.findByEmailIgnoreCase(input.getLogin());
+    requestPasswordReset(input.getLogin(), input.getLang());
+  }
+
+  @Async
+  @TransactionalEventListener(phase = TransactionPhase.AFTER_COMMIT, fallbackExecution = true)
+  public void onUserPasswordSetupRequested(UserPasswordSetupRequestedEvent event) {
+    requestPasswordReset(event.email(), null);
+  }
```
antoinemzs, line 347: should these two paths be reunited? The top one is triggered from an API call, but perhaps they both should go through the audit as well.

openaev-api/src/main/java/io/openaev/service/UserService.java
damgouj: Yes I will fix it, thanks for your advice, I hesitated between setting a null or random password :)

openaev-api/src/main/java/io/openaev/service/UserService.java
damgouj: I am not sure we can reunite the two but maybe you can help me? My idea was: it's async and sent after the hibernate transaction is finished but if you think we don't need the second annotation, I can do only one method if you prefer. Thanks for your opinion and your review @antoinemzs  :) !

antoinemzs: yes, understood, though the API endpoint also operates in a transaction. I'd be keen to have a single trigger for the password reset. Initially, it's an explicit handoff to the background thread, and the addition here is another trigger mode via the event. Perhaps the API endpoint could also emit an event to consolidate the behaviour and benefit from the audit?

that means perhaps also reworking the event class to carry locale information

antoinemzs (approved):
lgtm thanks!

PR #7744, opened 2026-09-03
feat(multi-tenancy): default-secure endpoint TxCtx compile rule, enforcing (#7726)

### LINKED ISSUE #7726: feat(Multi-tenancy): compilation check on TxCtx for all http call

laugiov:
## What

The default-secure tenant-scope rule for HTTP controllers, the compile-time follow-up named in
`TenantScopedEntrypointsTxCtxArchTest`:

- A new annotation processor `EndpointTxScopeRule` (twin of `EndpointTransactionalRule`): every REST
  endpoint (`@GetMapping`, `@PostMapping`, ...) must take a `io.openaev.context.TxCtx` parameter OR
  carry an exclusion marker, otherwise it fails compilation. So a new endpoint cannot ship without an
  explicit tenant-scope decision.
- Two exclusion markers `@NoTenantScope` / `@PlatformScoped` (in `io.openaev.annotation`, next to
  `@AllowRawJdbc`).

## Shipped disabled (why no grandfather list)

The controller sweep is not done, so most endpoints have neither yet. Rather than
grandfather hundreds of them, the rule ships **disabled by default** (`super(false)`, using the
family's `-A<FQCN>=true|false` flag):

- Now: the rule compiles, runs on every `openaev-api` build, and does nothing. The build stays green,
  no grandfather list.
- During the sweep: `-Aio.openaev.annotation.processor.EndpointTxScopeRule=true` turns it into a
  progress meter that lists the endpoints still missing a `TxCtx` or a marker.
- At sweep end: a one-line PR adds that arg to `openaev-api/pom.xml` and the rule becomes enforcing,
  compile-time, with zero exemptions.

The runtime fail-closed detector (#7723) is the net during the disabled window.

## Scope and safety

- The rule + markers only, shipped disabled. No controller changes, no `TenantScopedEntrypointsTxCtxArchTest`
  change, no pom enable.
- Matched entirely by fully-qualified name, so the processor keeps no dependency on the checked modules.
- Markers are `@Retention(SOURCE)`: no runtime effect. Full `openaev-api` build stays green with the
  rule registered and off (verified), so hundreds of not-yet-swept controllers compile fine.

## Tests

In-process compilation, JDK-only harness (no new dependency), 5 cases: an unmarked endpoint fails with
this rule's error; a `TxCtx` parameter passes; a method marker passes; a type-level marker covers its
methods; a non-endpoint method is ignored.

Closes #7726. Part of #6391.


corinnekrych (changes requested):
before merging let's make sure we marked the exception API first to not block sub-sequent PRs

laugiov:
Already done on the current head: @NoTenantScope is on UserApi.login, passwordReset, changePasswordReset, validatePasswordResetToken and StreamApi.streamFlux, and CI's Backend Compile is green with the rule enabled, so the build is proven. 
You are right about the second half though: from this merge on, any open branch that adds a @Transactional controller endpoint without a TxCtx or a marker will stop compiling, with an explicit error naming the fix, so it is worth a heads-up to the team.

corinnekrych (approved):

PR #7745, opened 2026-09-03
feat(engine): modular engine implementations (#7756)

### LINKED ISSUE #7754: 🛠️  chore(upgrade): Epic for Elasticsearch 9 migration major bump
(epic — no parent link needed)

## Goal

How can we enable incremental adoption of ES9 across our onprem customers
and our SaaS platform team? Aim for an approach similar to OCTI's, seeking
a common strategy where possible — see
[OCTI strategy](https://app.notion.com/p/filigran/Moving-OpenCTI-to-Elasticsearch-9-3c38fce17f2a8162bc16c0a5a6513e50).

> This topic is in active brainstorming — no final decisions yet. This
> epic tracks the topic at a high level; a detailed implementation plan
> and sub-issues will be split out once the brainstorming concludes.

> Pending PR from Romuald: #7665

## Independent of the Spring Boot 4.1 migration

This migration is **fully decoupled** from the Spring Boot 4.1 epic — see
that epic for the investigation proving Boot 4 does not require the ES9
client. Spring Boot major version and Elasticsearch client major version
can each move on their own schedule with no ordering constraint between
them.

## Key findings so far

- **The ES9 client cannot talk to an ES8 cluster** — confirmed empirically
  (`media_type_header_exception`, hard rejection at the wire level). No
  supported rolling/in-place upgrade path avoiding a real ES9 cluster
  exists. An unsupported jar-patch workaround was investigated and
  rejected (unsupported, no vendor guarantee, only basic CRUD/search
  verified).
- **Candidate approach under discussion (not yet decided)**: two
  compile-time ES-client stacks maintained on `main` simultaneously
  (`es8` / `es9` Maven profiles), both built and tested by CI on every
  commit, selected per-environment at deploy time via an environment
  variable — not a code branch choice. This is one option being weighed,
  not a locked-in decision.
- **Missing forced-reindex migration**: PR #7665 does not include a
  `Force_es_reindex_all`-style Flyway migration. Without resetting
  `indexing_status` cursors (which live in Postgres, not Elasticsearch) at
  cutover time, the sync job will silently under-index against the new ES9
  cluster.
- **Simplifying the index migration/cutover — in discussion with the SaaS
  platform team, not decided**: a multi-hour degraded-dashboard window
  during cutover to a cold ES9 cluster is undesirable, so we're exploring
  options to reduce it. One candidate under discussion is a read-replica-
  based pre-warm (stand up the new cluster ahead of time, warm it via a
  Postgres read replica, monitor convergence, then cut over with only a
  small delta left) — but this is one proposal among others being
  weighed, not a committed plan.

## Definition of Done (epic-level, to be refined)

- [ ] Detailed implementation plan finalized (brainstorming concluded)
- [ ] Sub-issues split out and tracked as sub-issues of this epic
- [ ] Forced-reindex migration added to PR #7665 (or equivalent)
- [ ] Migration/cutover strategy to minimize dashboard downtime decided
      (in discussion with the SaaS platform team) and documented for SaaS
      and onprem
- [ ] Both `es8`/`es9` profiles green in CI on every `main` commit

### LINKED ISSUE #7756: chore(upgrade): spike/PoC — modular engine implementation for ES8/ES9
**Epic**: [🛠️  chore(upgrade): Epic for Elasticsearch 9 migration major bump](https://github.com/OpenAEV-Platform/openaev/issues/7754)

## Summary

PR: https://github.com/OpenAEV-Platform/openaev/pull/7745 (open, not yet
merged)

`[P]` — spike/PoC, no hard dependency on the other Epic 2 sub-issue.

With very little refactoring, the concrete implementations of the "Engine
Sync" were extracted into Java modules, enabling support for both the ES8
and ES9 client versions side by side. The `EngineComponent` implementation
selector is promoted to the `openaev-api` module and, driven by
configuration, hands out either the OpenSearch impl, the ES8 impl, or a
possible new ES9 impl of the Elasticsearch-based engine sync.

This PoC is the underpinning for **Approach 1 (two compile-time stacks,
selected at deploy time)** discussed for the ES9 migration — one candidate
approach under discussion, not yet decided.

## Definition of Done

- [ ] Modular engine implementation reviewed and merged (or explicitly
      superseded by a different approach if the brainstorming lands
      elsewhere)
- [ ] Selector (`EngineComponent`) behavior validated for all three targets:
      OpenSearch, ES8, ES9
- [ ] No regression for existing OpenSearch/ES8 deployments

### LINKED ISSUE #7757: chore(upgrade): [P] Bump Elasticsearch client dependencies 8 to 9
**Epic**: [🛠️  chore(upgrade): Epic for Elasticsearch 9 migration major bump](https://github.com/OpenAEV-Platform/openaev/issues/7754)

## Summary

PR: https://github.com/OpenAEV-Platform/openaev/pull/7665 (open, not yet
merged)

`[P]` — dependency bump, no hard dependency on the other Epic 2 sub-issue.

- Bumps the Elasticsearch Java client 8.19 → 9.5.1 (dev and test images
  updated to match).
- `ElasticDriver` moves to `Rest5Client`: the 9.x client drops the
  HttpClient 4 transport, so auth, the trust-all SSL context, and
  preemptive basic auth are rebuilt on HttpClient 5.
- OpenSearch deployments are unaffected.
- **No data migration needed**: Elasticsearch holds nothing Postgres
  doesn't — all 13 `Handler` implementations read through a JPA repository,
  and the full rebuild path is already exercised by migrations like
  `V4_16__Force_es_reindex_injects`. An empty 9.x cluster is therefore a
  valid target: indices, templates, and the lifecycle policy are recreated
  at startup and reindexed.
- None of the Elasticsearch 9 breaking changes apply to us.

**After this change OpenAEV can no longer run against an Elasticsearch 8
cluster.** The 9.x Java client sends a content type the 8.x server
refuses, so every engine call fails and the platform does not start.
Existing deployments have to upgrade their cluster to Elasticsearch 9
first — the added migration page walks through it.

**Still missing** (flagged in the epic): a `Force_es_reindex_all`-style
Flyway migration to reset `indexing_status` cursors at cutover time —
without it, environments cutting over from an ES8 to an ES9 cluster will
silently under-index rather than doing a full rebuild.

## Definition of Done

- [ ] PR reviewed and merged
- [ ] Forced-reindex migration added (or explicit ops runbook step
      documented as a stopgap)
- [ ] Migration page for existing onprem/SaaS clusters reviewed

antoinemzs:
With very little refactoring, we are able to extract the concrete implementations of the "Engine Sync" into java modules, which enable us to support both the v8 and v9 versions of the elastic client.

Closes #7756 
Closes #7757 
Closes #7754


openaev-es9-client/pom.xml
```diff
@@ -0,0 +1,55 @@
...
+
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <elasticsearch.version>9.5.1</elasticsearch.version>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>co.elastic.clients</groupId>
+            <artifactId>elasticsearch-java</artifactId>
```
laugiov, line 39: elasticsearch-java:9.5.1 pulls commons-logging:1.3.5, which is not on main. It ships in BOOT-INF/lib next to spring-jcl, and both provide org.apache.commons.logging.Log, LogFactory, impl/NoOpLog and impl/SimpleLog. Spring Boot excludes commons-logging everywhere for this reason. The OpenSearch module already has the exclusion, this one needs the same.

openaev-es8-client/pom.xml
```diff
@@ -0,0 +1,65 @@
...
+                        <relocation>
+                            <pattern>co.elastic.clients</pattern>
+                            <shadedPattern>es8.co.elastic.clients</shadedPattern>
+                        </relocation>
+                        <relocation>
+                            <pattern>org.elasticsearch.client</pattern>
+                            <shadedPattern>es8.org.elasticsearch.client</shadedPattern>
+                        </relocation>
+                    </relocations>
+                    <createDependencyReducedPom>false</createDependencyReducedPom>
```
laugiov: With createDependencyReducedPom=false the module still declares its original dependencies, so the unshaded clients ship next to the shaded ones. In the built openaev-api.jar I get the three uber jars and elasticsearch-java-9.5.1.jar (14.2 MB), elasticsearch-rest5-client-9.5.1.jar, elasticsearch-rest-client-8.18.8.jar and opensearch-java-3.9.0.jar (10.6 MB). That is about 25 MB of dead weight. Total jar goes from 405.6 MB to 464.8 MB. Same setting in openaev-es9-client/pom.xml:42 and openaev-opensearch-client/pom.xml:47.

openaev-es8-client/pom.xml
```diff
@@ -0,0 +1,65 @@
...
+        </dependency>
+    </dependencies>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-shade-plugin</artifactId>
+                <version>3.6.2</version>
+                <configuration>
+                    <relocations>
```
laugiov: The relocations only cover the ES and OS packages, so all the other libraries are bundled unrelocated (Jackson 2.x and 3.x, httpclient 4.x and 5.x, slf4j-api, parsson, yasson). They are not the same builds: org/apache/http/impl/client/InternalHttpClient.class has CRC ebe9b31e in es8 and e4aeaa4f in es9 (4.5.14 vs 4.5.13), and httpclient-4.5.13.jar is also a normal entry in BOOT-INF/lib. Which one loads depends on jar order. <filters> or full relocation would fix it, and banDuplicateClasses in the enforcer would keep it fixed. Note also that each uber jar keeps only one META-INF/LICENSE and NOTICE, so the attribution of the bundled libraries is lost.

openaev-api/src/test/java/io/openaev/utilstest/DatabaseSnapshotManager.java
```diff
@@ -101,19 +104,16 @@ public void restoreToSnapshotState() {
...
     try {
-      esClient
-          .indices()
-          .delete(
-              d ->
-                  d.index(config.getIndexPrefix() + "_*")
-                      .ignoreUnavailable(true)
-                      .allowNoIndices(true));
+      for (EsModel<EsBase> model : models) {
+        service.cleanUpIndex(model.getName());
```
laugiov: cleanUpIndex also deletes the index template, and setupIndex only runs once per Spring context at driver init. 
This method runs afterTestClass, so from the second test class on, indices are auto-created without dynamic: strict and without the string_normalizer. 
The previous wildcard delete kept the templates. 
Tests pass today -> but the harness no longer matches production mappings.

openaev-es9-client/pom.xml
antoinemzs: excluded in next commits

openaev-es8-client/pom.xml
antoinemzs: > Note also that each uber jar keeps only one META-INF/LICENSE and NOTICE, so the attribution of the bundled libraries is lost.

addressed

openaev-es8-client/pom.xml
antoinemzs: removed; note that Spring Boot independently declares `elastisearch-java:8.18.8` but it's not used in our stack; we only use the shaded clients.

openaev-es8-client/pom.xml
antoinemzs: as discussed: we will go with partial shading of only the relevant engine client classes; transitive deps are provided via classpath and not included in the uberjar

openaev-api/src/test/java/io/openaev/utilstest/DatabaseSnapshotManager.java
antoinemzs: direct usage of the elastic client in here is problematic because it breaks the EngineService facade. I've split the method in two parts which will optionally delete the template, and use the non-deleting version in tests.

antoinemzs:
@laugiov at your convenience :pray:

.github/workflows/nightly-ci.yml
```diff
@@ -39,6 +39,14 @@ jobs:
           {"shard":"6","shard_name":"6","search_engine":"elasticsearch","engine_selector":"elk","artifact_suffix":"","includes":"shardfile","excludes":""},
           {"shard":"7","shard_name":"7","search_engine":"elasticsearch","engine_selector":"elk","artifact_suffix":"","includes":"shardfile","excludes":""},
           {"shard":"8","shard_name":"remaining","search_engine":"elasticsearch","engine_selector":"elk","artifact_suffix":"","includes":"catchall","excludes":""},
+          {"shard":"1","shard_name":"elk9-1","search_engine":"elasticsearch","engine_selector":"elk9","artifact_suffix":"","includes":"shardfile","excludes":""},
```
laugiov: These 8 elk9 shards have artifact_suffix: "", same as the 8 elk8 shards above, so both sets upload jacoco-exec-shard-1..8. upload-artifact fails on a duplicate name, so 8 nightly jobs go red. It can't show up here since nightly doesn't run on PRs.

"artifact_suffix":"-elk9" on these 8 lines fixes it, like -opensearch does above.

antoinemzs: done

laugiov (approved):

PR #7758, opened 2026-09-03
fix(rbac): set the tenant scope before the permission check runs

### LINKED ISSUE #7765: fix(scope): Role with assessment access cannot see execution results for scenarios and atomic tests
## Description

Description

We’ve encountered an issue where a role that has access to assessments doesn’t have access to see the execution results for assessments. This behavior is consistent for both scenarios and atomic tests.

Role capabilities:

Role/user can access assessments, but execution results UI is not visible.

Observed behavior:

User with this role’s view of a scenario result: the Results tab (right side) is not shown at all.

Admin/bypass permissions view: Results tab is visible.

Expected behavior:

Any role with access to view an assessment/scenario/atomic test execution should be able to see the execution results (Results tab) according to its permissions.

Insights

Why this specific due date?: This is a critical permission issue affecting user workflow.

Need

Roles with assessment access are unable to see execution results, which is critical for their workflow.

RomuDeuxfois:
## Facts

A non-admin caller with `ACCESS_ASSESSMENT` reads a simulation inject and gets **200 with `inject_type: null`**. The frontend gates the whole "Results by target" panel on that field, so the column renders empty — no results, no empty state, no error. Same user, same session:

| Endpoint | RBAC path | Result |
|---|---|---|
| `GET /injector_contracts/{id}` | pure mapping, no entity load | 200, `injector_contract_injector_type: "openaev_implant"` |
| `GET /atomic-testings/{injectId}` | `resolveTarget`, **loads the Inject** | 200, `inject_type: null` |

Admins are never affected.

## Cause

The permission check ran before the tenant scope was set.

`resolveTarget` loads the `Inject` to check permissions on it. At that point `app.current_tenants` was still empty, so the tenant filter denied the `injectors` rows (fail-closed). `@NotFound(IGNORE)` turned the missing association into `null`, that `null` stayed in the persistence context, and the controller served it.

The order was undefined: `TenantScopeTransactionAspect` and `AccessControlAspect` were both on `Ordered.LOWEST_PRECEDENCE` (a tie Spring breaks arbitrarily) and `HibernateFilterTransactionAspect` had no `@Order` at all. Admins short-circuit at `user.isAdmin()` before `resolveTarget`, which is why only non-admins hit it.

## Fix

An explicit order for every advice, no ties: lock `LP-5` → transaction `LP-4` → tenant filter `LP-3` → tenant scope `LP-2` → audit `LP-1` → rbac `LP`. `@Lock` still wraps the transaction, audit still wraps RBAC. No production logic changed.

`TenantScopeLockOrderingIntegrationTest` asserted the chain but omitted the two scoping aspects — the tied pair; it now covers all of them.

## Review note

- The transaction boundary moves two slots: anything ordered between `LP-4` and `LP-2` would now sit outside the transaction. Nothing else in the codebase uses that band.
- No functional regression test: I wrote one and it passed with the old `@Order` too. `AtomicTestingApiTenantIsolationTest` is `@Transactional`, so MockMvc requests share one transaction and the create request's scope is still set when the detail request runs. A real guard needs a non-transactional test (one transaction per request) — test-infrastructure work, out of scope here. The existing `detailExposesInjectTypeUnderV2Activation` covers this field but runs `isAdmin = true`, which is why the suite never saw it.

### DIFF
--- a/openaev-api/src/main/java/io/openaev/aop/AccessControlAspect.java
+++ b/openaev-api/src/main/java/io/openaev/aop/AccessControlAspect.java
@@ -33,7 +33,10 @@ import org.springframework.web.server.ResponseStatusException;
 
 @Aspect
 @Component
-@Order(Ordered.LOWEST_PRECEDENCE) // Inner to audit aspect (LOWEST_PRECEDENCE - 1)
+// Innermost of the chain: inner to the audit aspect (LP-1) and, crucially, to both tenant scoping
+// aspects (LP-3 / LP-2). The permission check below loads entities to resolve a resource's parent,
+// so it must never run before the tenant scope of its own transaction has been written.
+@Order(Ordered.LOWEST_PRECEDENCE)
 @RequiredArgsConstructor
 @Slf4j
 public class AccessControlAspect {
--- a/openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
+++ b/openaev-api/src/main/java/io/openaev/aop/audit_log/AccessControlAuditLogAspect.java
@@ -52,10 +52,11 @@ import org.springframework.web.server.ResponseStatusException;
  * "unauthorized" audit event before being re-thrown.
  *
  * <p>The aspect order is {@code LOWEST_PRECEDENCE - 1} so it runs <b>inside</b> the transaction
- * boundary ({@code LOWEST_PRECEDENCE - 2}) and <b>outside</b> the RBAC aspect ({@code
- * LOWEST_PRECEDENCE}). When halt-on-failure is active and the audit transport fails, the thrown
- * {@link AuditLogFailureException} propagates through the transaction interceptor, which rolls back
- * the mutation.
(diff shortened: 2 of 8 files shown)


laugiov:
Your diagnosis is right, I checked it end to end. Clean build at pr7758~1: inject_type comes back null. Clean build here: green. Non-admin, active-tables=injectors, no class-level @Transactional.

Worth adding that as a test: IntegrationTest is already non-transactional and createTenantWithCapabilities exists.

Dimfacion:
For the record, I tested the halt-on-failure of audit logs, and no issues detected

RomuDeuxfois:
> Your diagnosis is right, I checked it end to end. Clean build at pr7758~1: inject_type comes back null. Clean build here: green. Non-admin, active-tables=injectors, no class-level transactional.
> 
> Worth adding that as a test: IntegrationTest is already non-transactional and createTenantWithCapabilities exists.

Done here openaev-api/src/test/java/io/openaev/rest/atomic_testing/AtomicTestingApiNonAdminTenantScopeTest.java

laugiov (approved):

PR #7777, opened 2026-09-03
feat(audit-log): prevent renew of admin token (#7491)

Dimfacion:
### Proposed changes

* Prevent the user from renewing his token if he is admin
* Adding some checks on the backend to prevent the change of password, email and token for the admin user just like in the UI

### Testing Instructions

1. Try to renew the token as the admin user
2. Check it's not possible
3. Try to renew the token as a normal user
4. Check it's working
5. Try to change the token, the password or the email of the admin user
6. Check it's not possible
7. Try to change the token, the password or the email of a normal user
8. Check it's working

### Related issues

* Related #7491 
*

### Checklist

### DIFF
--- a/openaev-api/src/main/java/io/openaev/rest/user/MeApi.java
+++ b/openaev-api/src/main/java/io/openaev/rest/user/MeApi.java
@@ -17,6 +17,7 @@ import io.openaev.database.repository.OrganizationRepository;
 import io.openaev.database.repository.TokenRepository;
 import io.openaev.database.repository.UserRepository;
 import io.openaev.rest.exception.ElementNotFoundException;
+import io.openaev.rest.exception.ForbiddenException;
 import io.openaev.rest.exception.InputValidationException;
 import io.openaev.rest.helper.RestBehavior;
 import io.openaev.rest.user.form.me.UpdateMePasswordInput;
@@ -78,7 +79,7 @@ public class MeApi extends RestBehavior {
             .findById(currentUser().getId())
             .orElseThrow(() -> new ElementNotFoundException("Current user not found"));
     if (user.isExternal() && !input.getEmail().equals(user.getEmail())) {
-      throw new IllegalStateException(
+      throw new ForbiddenException(
           "The admin email cannot be changed using the API. Please change it in the configuration.");
     }
     user.setUpdateAttributes(input);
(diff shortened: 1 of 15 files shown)


openaev-api/src/main/java/io/openaev/rest/user/MeApi.java
```diff
@@ -77,6 +78,10 @@ public User updateProfile(@Valid @RequestBody UpdateProfileInput input) {
         userRepository
             .findById(currentUser().getId())
             .orElseThrow(() -> new ElementNotFoundException("Current user not found"));
+    if (user.isExternal() && !input.getEmail().equals(user.getEmail())) {
```
RomuDeuxfois, line 81: improvment: forgot password mechanism should be also blocked for admin

RomuDeuxfois, line 81: **todo**: please keep logic in service and in a single method to be sure every call go to the same defense mechanism

openaev-api/src/main/java/io/openaev/service/UserService.java
```diff
@@ -1,5 +1,6 @@
 package io.openaev.service;
 
+import static io.openaev.config.SessionHelper.currentUser;
```
RomuDeuxfois, line 3: **todo**: why we have this import ?

openaev-api/src/main/java/io/openaev/rest/user/MeApi.java
RomuDeuxfois, line 81: **todo**: call in the backend shoud used equalsIgnoreCase for email check.

openaev-api/src/main/java/io/openaev/rest/user/MeApi.java
```diff
@@ -17,6 +17,7 @@
 import io.openaev.database.repository.TokenRepository;
```
RomuDeuxfois, line 17: **todo**: pr name is not relevant for this work

openaev-api/src/main/java/io/openaev/rest/user/MeApi.java
RomuDeuxfois, line 81: **nitpick**: is external is misleading, can be good to change it

PR #7779, opened 2026-09-03
fix(communication): fix regex to retrieve injectId from email footer (#7202)

### LINKED ISSUE #7202: fix(mails): received mails doesn't appear on the mails inject menu
## Description

Response made by the players on the "Mails" menu on the "Execution" tab should be displayed. 

## Environment

1. OS: Windows 11
2. Version: 3.260805.0

## Reproducible steps

1. Run a simulation with mail sender injects ("Send multi-recipients mail" or "Send individual mail" injects)
2. Respond to the mail received on the player mailbox
3. Go to the "Execution" tab on the simulation, and select the "Mails" menu on the right
4. You should see here the injects who receive a response from the player
5. clic on one to display the details

## Expected output

The mail and reply button should appear :
Image

## Actual output

The mails section is empty
Image

## Additional information

On main-ff, some simulations display the mails (7d28c936-12ae-491b-976a-c1dc253039ba) and new ones does not (bb07d0a0-0cdf-42bd-941a-9fd40dc0c619)

EvaE-Filigran:
### Proposed changes

* Fix `ImapService.INJECT_ID_PATTERN` greedy regex: `(.*)` matched up to the last `]` in the message body instead of the first one, so when the footer contains `[inject_id=<uuid>]<br/>[base_url=<url>]`, the captured id included the trailing `<br/>[base_url=...` fragment, `injectRepository.findById` failed to resolve the inject, and the incoming/sent email was silently dropped (no `Communication` persisted).
* Replace `(.*)` with `([^\]]*)` so the capture stops at the first closing bracket.

### Testing Instructions

1. Create a time-based scenario with an email sender
2. Respond to this email; in the Execution > Email tab you should see the exchange

### Related issues

* Related #7202

### Checklist

### Further comments

Regression introduced by `chore(email-info): add more info on email footer`, which added a second bracketed token (`[base_url=...]`) right after `[inject_id=...]` in the footer. The pre-existing greedy regex had always been fragile, but only surfaced once a second `[...]` block was appended after it.


openaev-api/src/main/java/io/openaev/injectors/email/service/ImapService.java
```diff
@@ -34,7 +34,7 @@
 @RequiredArgsConstructor
 public class ImapService extends ExternalServiceBase {
 
-  private static final Pattern INJECT_ID_PATTERN = Pattern.compile("\\[inject_id=(.*)\\]");
+  private static final Pattern INJECT_ID_PATTERN = Pattern.compile("\\[inject_id=([^\\]]*)\\]");
```
antoinemzs: an inject ID is a UUIDv4 yes? Could this be a more reliable regex? No doubt this works as long as we put the footer between square brackets though.

EvaE-Filigran: fixed

antoinemzs (approved):

Issue #7389, opened 2026-08-13
fix(chaining): compatibility for Tabletop


mergenhan:
## Description

chaining should be compatible with tabletop actions

## Reproducible steps

Steps to create the smallest reproducible scenario:
1. Create a chaining simulation
2. Add scope as team or person
3. These team and person could not use during the simulation

## Expected output

If user add any item in the scope, it should be usable during the simulation if needed.

mergenhan:
https://github.com/OpenAEV-Platform/filigran-private/issues/308

mergenhan:
Lesson learned too
import for allow list capability.

PR #7793, opened 2026-09-04
feat(chaining): duplicate chaining simulation/scneario (#5775)

### LINKED ISSUE #5775: feat(chaining): duplicate scenario/simulation
Dod:

Chained scenario/simulation can be duplicated

camrrx:
### Proposed changes

* **Single duplicate endpoint.** `POST /api/scenarios/{id}` and `POST /api/exercises/{id}` now also
  copy the chaining logic map when the object is chained. No more dedicated `ChainingApi` duplicate
  routes (removed upstream by #7389). Implements ADR-007, added here.
* **Fixed the workflow graph copy** (`StepService.copyStepTemplate`, the single primitive shared by
  launch, autonomous plan mode, convert-to-manual and duplication). The copy is now three-phase, so
  `DEPEND_ON.value` and `Condition.stepFrom` are remapped to the copied steps instead of still
  pointing at the source workflow, and `step_data` is re-pointed at the new owner with its execution
  artefacts stripped.
* **A chained simulation no longer clones its runtime injects** — they are created at execution time,
  so duplicating a launched simulation was bleeding its run into the copy.
* **EE licence checked on the chained branch only** (the endpoint is shared with time-based objects),
  and it rejects rather than silently returning a copy without its logic map.
* **Frontend:** duplication is no longer disabled for chained objects, and
  `scenarios/scenario/Index.tsx` now refetches on `scenarioId` change — without it the duplicate
  opened on time-based tabs until a manual refresh.

### Testing Instructions

1. Duplicate a chained scenario from its detail page → opens on the chaining tabs, same logic map, no
   refresh needed.
2. Launch it, duplicate the simulation → authored steps only, no inject / status / expectation.
3. Source object unchanged in both cases.
4. Inactive EE licence → chained duplication rejected, time-based still works.

### Related issues

* Closes #5775


adr/ADR-007-duplicating-a-chained-scenario-simulation.md
```diff
@@ -0,0 +1,362 @@
...
+
+A **chained** object carries a second body of authored content those services know nothing about: its
+**workflow** (`workflows`, `steps`, `conditions`, `conditions_steps`, `workflow_scope_rules`,
+`scope_variables`). It also carries execution state the copy must never inherit: RUN workflows,
+runtime steps, `workflow_state` rows, `steps_delay_queue` rows, scope snapshots (ADR-006), in-flight
+RabbitMQ messages and runtime-generated injects.
+
+Two half-built pieces are the starting point of this decision:
+
+1. **The backend endpoints exist but are unreachable.** `ChainingApi.duplicateExercise`
```
heditar, line 26: I think it is no more relevant after your update, we should also probably remove duplicate existing methods

openaev-api/src/main/java/io/openaev/rest/scenario/ScenarioApi.java
```diff
@@ -190,8 +190,44 @@ public List<ScenarioSimple> updateScenariosWithInjectorContracts(
...
       actionPerformed = Action.DUPLICATE,
       resourceType = ResourceType.SCENARIO)
-  public Scenario duplicateScenario(@PathVariable @NotBlank final String scenarioId) {
-    return scenarioService.getDuplicateScenario(scenarioId);
+  @Operation(
+      summary = "Duplicate a scenario",
+      description =
+          "Duplicates a scenario's authored content. A chained scenario also gets a copy of its logic map (workflow, steps and conditions); execution artefacts are never copied. Requires an Enterprise Edition license when the scenario is chained.")
+  public Scenario duplicateScenario(@PathVariable @NotBlank final String scenarioId)
+      throws ChainingException {
```
heditar, line 198: Why do we throw that here but not in ExerciseApi ?

openaev-api/src/main/java/io/openaev/rest/exercise/ExerciseApi.java
```diff
@@ -529,8 +529,41 @@ public Exercise createExercise(@Valid @RequestBody CreateExerciseInput input) {
...
+   * <p>Chaining is an Enterprise Edition feature, but this endpoint also serves time-based
+   * simulations, so it cannot carry an unconditional {@code isEnterpriseEdition = true} on {@link
+   * AccessControl} - that flag is applied before anything else. The licence is therefore checked
+   * programmatically, and only on the chained branch: silently degrading to a metadata-only copy
+   * would make the user lose the logic map with no signal.
+   *
+   * @param exerciseId the source simulation
+   * @param duplicate the persisted duplicate to attach the copied workflow to
+   */
+  private void duplicateChainingWorkflowIfAny(final String exerciseId, final Exercise duplicate) {
```
heditar, line 558: Logic should be handle by the service not the api class

heditar:
tested ok

openaev-api/src/main/java/io/openaev/rest/scenario/ScenarioApi.java
camrrx: the scenario finder returns a List (so it throws a checked ChainingException if more than one TEMPLATE is found), while the simulation finder returns a single Workflow, that case simply cannot happen there !

PR #7836, opened 2026-09-07
feat(multi-tenancy): activate asset_groups on v2 isolation (#6435)

laugiov:
### Proposed changes

* Activate `asset_groups` on multi-tenancy v2: the v1 `@Filter` is removed and the table is added to `openaev.tenant.active-tables`, so reads are scoped by `TenantStatementInspector` instead of the Hibernate thread-local filter.
* Make write attribution explicit on every create path. `tenantId` became a required parameter of `AssetGroupService.createAssetGroup`, which turned each missed call site into a compile error: HTTP resolves it through `TenantWriteScopeResolver` (400 on an ambiguous multi-tenant scope), the two tenant-provisioning datapacks pass `tenant.getId()`, and the five executor connectors pass the `executor.getTenantId()` their own scoped transaction already opened.
* Fix an activation-caused break in the asset-group search endpoint. PostgreSQL allows selecting ungrouped columns when the `GROUP BY` covers a base table's primary key; the inspector rewrites the table into a derived table, where that no longer holds, and the query became invalid SQL (a 500 in production). `AssetGroupQueryHelper` now groups on every projected column.
* Scope the `asset_groups_total` telemetry gauge under `TxCtx.allTenants()`, and correct the collector's class javadoc, which documented a v1-specific rationale that v2 falsifies.
* Remove the three `tenant_id = :#{#tenantContext.currentTenant}` predicates from the native queries on this table. The one left on `injects` carries a `TODO v2` marker for that table's own activation.
* Add isolation coverage: `AssetGroupHttpIsolationTest` (admin), `AssetGroupNonAdminIsolationTest`, `AssetGroupBackgroundIsolationTest` (the ES indexing scope), plus seven inspector pins on the table's native queries and 32 entries in `TX_SCOPED_ENTRYPOINTS`.
* Extend the two go-live guards: `asset_groups` in `GUARDED_TABLES`, and the shared production-config guard so dropping the table from the allowlist fails the build.

**`TenantBaseListener` is removed, as the runbook requires at go-live.** It stamped `tenant_id` from the v1 thread-local on every insert, which is the wrong tenant as often as the right one on a background or provisioning path, and it made an unattributed write look successful instead of failing. With it gone, an unattributed write fails on the NOT NULL column rather than landing silently in the default tenant. Every create path already resolves the tenant explicitly, so nothing depended on it in production: `AssetGroupService.createAssetGroup` takes it as a required parameter, HTTP callers get it from `TenantWriteScopeResolver`, and background callers pass the tenant their own scope was opened for.

An earlier revision of this PR kept the listener, on the grounds that removing it broke 19 tests. That was true of the removal alone and it was the wrong conclusion: the dependency was concentrated in `AssetGroupFixture`, not spread across call sites. `AssetGroupFixture` now stamps the tenant itself, exactly as `SecurityCoverageFixture` already did for the other activated entity that carries no listener. That single line is the whole cost.

Eight of the nineteen tables in `active-tables` still carry a tenant listener (`import_mappers`, `mitigations`, `cwes`, `attackpath_execution`, `attackpath_finding`, `secret_references`, `secrets`, `connector_instances`). Removing those is a platform-wide cleanup of its own, tracked in #7844, and is not in this PR.

### Testing Instructions

1. Start the services from `openaev-dev/docker-compose.yml` (PostgreSQL, MinIO, Elasticsearch, RabbitMQ).
2. `mvn -ntp clean install -DskipTests` from the repository root.
3. Isolation and activation coverage: `mvn -o -pl openaev-api test -Dtest='AssetGroupHttpIsolationTest,AssetGroupNonAdminIsolationTest,AssetGroupBackgroundIsolationTest,AssetGroupApiTest,AssetGroupApiSearchTest,TenantStatementInspectorTest,ImportMapperActivationConfigTest'`
4. Architecture guards: `mvn -o -pl openaev-api test -Dtest='TenantActiveTableAccessArchTest,TenantNonOrmAccessArchTest,TenantScopedEntrypointsTxCtxArchTest,TenantBackgroundTransactionArchTest'`
5. Full regression: `mvn -ntp -pl openaev-api test`. Locally this gave 650 classes and 6061 tests with 0 failures; three errors appeared under load (RabbitMQ queue creation, a cascade-delete data race) in two classes that never touch asset groups, and both passed cleanly when re-run in isolation.
6. Fail-closed detector: `mvn -pl openaev-api test -Dopenaev.failclosed.detector=on`. Five of its failures pre-date this branch on `main`; the sixth is `AssetGroupBackgroundIsolationTest`, which reads the indexing handler unscoped on purpose and is now a traced SCOPED-OK entry in `failclosed-baseline.txt`.

**Client impact to be aware of when testing.** A create with no tenant selector now returns 400 instead of being silently attributed; callers on the `/api/tenants/{tenantId}/...` route are unaffected. And the asset-group options list in its atomic-testing mode now returns every in-scope tenant's groups under a multi-tenant scope, where it previously returned only the thread-local tenant's.

### Related issues

* Related #6435
* Related #6393

### Checklist

### Further comments

Two defects were found while doing this work and are worth flagging separately, because neither is caused by this change.

The functional-dependency break above is not specific to `asset_groups`. Five other call sites group on a primary key and select ungrouped columns, on tables this epic will activate: `ScenarioService:365`, `ExerciseService:880`, `TeamQueryHelper:68`, `UserQueryHelper:58` and `PlayerQueryHelper:43`. Each becomes invalid SQL the day its own table is activated, with an error message that says nothing about tenancy. This is raised on #7618, which generalises the same rewriter, with a proposal to fix it at the source rather than one call site at a time.

Separately, `import_mappers_total` in `ProductInventoryMetricCollector` has been reporting zero in production since the pilot activation: it is the same unscoped count that this PR fixes for asset groups, on a table that has been active for much longer. It is left untouched here because it is outside this issue's scope, but it needs its own fix, and the question of whether to scope that collector gauge by gauge or once and for all is worth deciding before `findings` activates.


openaev-model/src/main/java/io/openaev/database/model/AssetGroup.java
```diff
@@ -23,21 +23,42 @@
...
 @Table(name = "asset_groups")
 @EntityListeners({ModelBaseListener.class, TenantBaseListener.class})
-@Filter(name = "tenantFilter", condition = "tenant_id = :tenantId")
 @NamedEntityGraphs({
   @NamedEntityGraph(
       name = "AssetGroup.tags-assets",
       attributeNodes = {@NamedAttributeNode("tags"), @NamedAttributeNode("assets")})
 })
 public class AssetGroup implements TenantBase {
+
```
corinnekrych, line 38: I think we should remove the TenantBaseListener. i dont know why some activated table where not removed but they should it's part of the full clean-up

openaev-api/src/main/java/io/openaev/rest/asset_group/AssetGroupQueryHelper.java
```diff
@@ -38,14 +42,27 @@ public static void select(
...
-                .alias("asset_group_dynamic_filter"),
+            dynamicFilterAsJsonb.alias("asset_group_dynamic_filter"),
             assetIdsExpression.alias("asset_group_assets"),
             tagIdsExpression.alias("asset_group_tags"))
         .distinct(true);
 
-    // Group by
-    cq.groupBy(Collections.singletonList(assetGroupRoot.get("id")));
+    // Group by every non-aggregated column, not just the id.
+    //
```
corinnekrych, line 51: As discussed let's add those recommendation in SKILL so that AI has a pattern to apply for fuure Exercice etc... ativation

corinnekrych:
@laugiov Tested locally doing crud on assets / assets group. Runtime ok.

openaev-api/src/test/java/io/openaev/utils/fixtures/AssetGroupFixture.java
```diff
@@ -13,6 +15,12 @@ public static AssetGroup createDefaultAssetGroup(@NotNull final String name) {
     AssetGroup assetGroup = new AssetGroup();
     assetGroup.setName(name);
     assetGroup.setDescription("An asset group");
+    // asset_groups is tenant-active and its TenantBaseListener was removed at go-live: stamp the
+    // tenant explicitly here, matching what the listener used to do, instead of leaving every call
+    // site to remember it. TenantContext.getCurrentTenant() never throws (it defaults to
+    // Tenant.DEFAULT_TENANT_UUID), so this is safe even outside a request context. Same shape as
+    // SecurityCoverageFixture, the other activated entity that carries no listener.
+    assetGroup.setTenant(new Tenant(TenantContext.getCurrentTenant()));
```
corinnekrych, line 23: could we removed that occurence to TenantContext.getCurrentTenant()

corinnekrych (approved):

openaev-api/src/test/java/io/openaev/utils/fixtures/AssetGroupFixture.java
laugiov: Removing the TenantBaseListener, as you asked above, is exactly why the fixture has to stamp the tenant itself, the same way SecurityCoverageFixture already does on main

openaev-api/src/main/java/io/openaev/service/AssetGroupService.java
```diff
@@ -47,7 +47,24 @@ public class AssetGroupService {
...
 
-  public AssetGroup createAssetGroup(@NotNull final AssetGroup assetGroup) {
+  /**
+   * Creates an asset group owned by {@code tenantId}.
+   *
+   * <p>The tenant is an explicit parameter, never inferred. {@code asset_groups} is v2-active and
+   * the go-live commit removed {@code TenantBaseListener}, so nothing stamps the row implicitly any
+   * more: a create that does not carry its tenant would write {@code tenant_id = NULL}. HTTP
+   * callers resolve it through {@link io.openaev.config.TenantWriteScopeResolver}, which refuses an
+   * ambiguous multi-tenant scope; background callers pass the tenant they are already scoped to.
```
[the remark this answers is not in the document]
laugiov: Fixed in d2868baa3, on the lot B branch (feat/mtv2-activate-assets) since this PR is already merged. You were right: AssetGroup keeps TenantBaseListener, so the comment described a state that does not exist. It now says the real reason the tenant is passed explicitly, and names #7844 as what removes the listener.

openaev-api/src/test/java/io/openaev/rest/asset_group/AssetGroupHttpIsolationTest.java
```diff
@@ -0,0 +1,293 @@
...
+        rawSingle("SELECT tenant_id FROM asset_groups WHERE asset_group_id = ?", createdId);
+    assertEquals(tenantA, storedTenant, "the created asset group must belong to tenant A");
+  }
+
+  @Test
+  @DisplayName("a create with no tenant selector is refused, not silently attributed")
+  void createWithoutSelectorIsRejected() throws Exception {
+    // Today TenantBaseListener silently stamps whatever the v1 thread-local holds, so an ambiguous
+    // create succeeds and lands somewhere. The go-live commit removes that listener; attribution
+    // must be explicit and an ambiguous write must be refused loudly, per TenantWriteScopeResolver.
```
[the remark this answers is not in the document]
laugiov: Fixed in d2868baa3, on the lot B branch (feat/mtv2-activate-assets) since this PR is already merged. You were right: AssetGroup keeps TenantBaseListener, so the comment described a state that does not exist. It now says the real reason the tenant is passed explicitly, and names #7844 as what removes the listener.

openaev-api/src/main/java/io/openaev/rest/asset_group/AssetGroupQueryHelper.java
laugiov: Done, in .github/skills/activate-tenant-table/SKILL.md. The GROUP BY trap has its own section ("GROUP BY on a wrapped table: valid SQL before activation, a 500 after") with the grep to run over every GROUP BY that can reach the table, plus a short version in the phase checklist. Two more rules landed there from lots B and C: removing the v1 @Filter silently disarms every isolation suite that does not activate the table, and a sink pin must not be @Transactional or it proves nothing. Leaving this thread for you to close.

PR #7838, opened 2026-09-08
feat(multitenancy): activate assets on v2 isolation (#6438)

laugiov:
### Proposed changes

Activates the `assets` table on multi-tenancy v2. The table is `SINGLE_TABLE` with **three**
discriminators (`Endpoint`, `SecurityPlatform`, and the bare `Asset` an AI target is), so everything
below covers all three.

* **Go-live.** `Asset`'s v1 `@Filter` is removed and `assets` is appended to
  `openaev.tenant.active-tables`, so reads are scoped by `TenantStatementInspector` instead of the
  Hibernate thread-local filter. `GUARDED_TABLES` and the shared production-config guard are both
  extended, the latter so that dropping the table from the allowlist fails the build.
* **`LATERAL` on the endpoint MAC lookup.** `filterFromItem` accepts a `TableFunction` only with a
  `LATERAL` prefix, so `findByHostnameAndAtleastOneMacAddress` would have been refused fail-closed
  once the table was active, breaking endpoint lookup by hostname and MAC: the agent registration
  path. Pinned by `EndpointMacLookupTenantScopeTest` and by an inspector test that reads the query
  reflectively, so a future edit to the SQL is caught.
* **Explicit write attribution on every create path.** `tenantId` became a required parameter of
  both `EndpointService.createEndpoint` overloads, which turned each forgotten path into a compile
  error and surfaced seven call sites, two of which grep had missed. The security platform, AI
  target and `V1_DataImporter` paths resolve it through `TenantWriteScopeResolver`.
* **Read isolation coverage** for all three discriminators plus the generic `/api/assets` route that
  reads across them, as admin (`AssetHttpIsolationTest`) and as a non-admin
  (`AssetNonAdminIsolationTest`).
* **Three telemetry gauges scoped**, not one. `assets_total`, `endpoints_total` and
  `security_platforms_total` all count rows of this table and would have reported zero silently.
  `AssetMetricCollector`'s javadoc used to prescribe raw JDBC for this case, which contradicts the
  `TenantNonOrmAccessArchTest` guardrail; it now opens `TxCtx.allTenants()` like `countAssetGroups`.
* **`InjectsFinalizationJob.handlePendingInject` scoped per tenant.** It resolved each pending
  inject's agents with no scope, which walks the inject's asset groups; `AssetGroupService` throws
  `ElementNotFoundException` when the lookup finds nothing, and that escapes to `execute()`. One
  inject targeting an asset group stopped the whole finalization sweep, for every tenant. This was
  **live on main** since `asset_groups` was activated, not a consequence of this PR.

**Two traps worth knowing about for the remaining tables.**

A signature that takes a tenant is not evidence that the row gets one: `createEndpoint(input,
tenantId)` and `upsertEndpoint(input, tenantId)` both accepted a tenant, used it only to validate the
linked person, and never attributed it.

And an eager mandatory association onto an activated table does not degrade, it disappears.
`Agent.asset` is a `@ManyToOne` with no `fetch` attribute, so an unscoped agent read returns **no
agents at all**, not agents with a null asset. An empty result is indistinguishable from "this tenant
has no agents". `AgentAssetAssociationTenantScopeTest` pins both halves.

### Removing a v1 `@Filter` disarms isolation tests that do not activate the table

The test profile declares no `openaev.tenant.active-tables`, so the inspector never fires in a test
context that does not set it. Any suite that asserted this table's isolation **before** the
activation was relying on the v1 `@Filter`; removing it takes that isolation away while the
assertions keep running.

`EndpointApiTest`'s `TenantIsolation` class had four cross-tenant tests and no
`@TestPropertySource`. Removing the filter turned **one** of them red and left the other three green
**while isolating nothing**. `SecurityPlatformInjectorLifecycleTest` had the same shape one table
over: its foreign-tenant upsert assertion started resolving the foreign row. Both now activate the
table, and the trap is written into `.github/skills/activate-tenant-table/SKILL.md` so the next
activation does not pay for it again.

Note for reviewers: put that annotation on the **outermost** test class. On a `@Nested` one it builds
a second Spring context where the mock user provisioned into the parent's `TestUserHolder` is
missing, and every test dies before reaching its assertion.

### Deliberate deviations, both tracked

* **`TenantBaseListener` stays on `Asset`.** The runbook asks for its removal at go-live. On
  `asset_groups` that cost one fixture line; here **63 test sites** build an asset by hand and
  persist it straight through a repository, reaching neither fixture nor composer. Read isolation
  does not depend on it. #7844 owns the removal for the eight activated tables that still carry one,
  and finds the ground prepared since the fixtures and composers now stamp the tenant.
* **No per-table `ArchRule`.** `domains` has four allowlisted accessors; `assets` would have 31. A
  list that size is only worth something if every entry is genuinely reviewed, and a rubber stamp is
  worse than none because the next reader trusts it.

### Known limitation

`EndpointApiTest` declares `active-tables=assets`, and `@TestPropertySource` **replaces** the
allowlist, so `executors` is inactive there and its `AgentExecutorJoin` nested test exercises the
join without the rewrite. Adding `executors` turns four of that class's tests red on
`executors_pkey` duplicates (the fixture reuses a fixed executor id, the lookup that would find it
fails closed, the insert collides) — separate fixture work, noted in the annotation's comment.

### Testing Instructions

1. Start the services from `openaev-dev/docker-compose.yml` (PostgreSQL, MinIO, Elasticsearch, RabbitMQ).
2. `mvn -ntp clean install -DskipTests` from the repository root.
3. Isolation and activation coverage: `mvn -pl openaev-api -am test -Dtest='AssetHttpIsolationTest,AssetNonAdminIsolationTest,AssetWriteAttributionTest,AssetAssociationSinkTest,AgentAssetAssociationTenantScopeTest,EndpointSearchCountScopeTest,EndpointMacLookupTenantScopeTest,AssetMetricTenantScopeTest' -Dsurefire.failIfNoSpecifiedTests=false`
4. The suites the activation repaired: `mvn -pl openaev-api -am test -Dtest='EndpointApiTest,SecurityPlatformInjectorLifecycleTest,SecurityPlatformApiTest,AiTargetApiTest' -Dsurefire.failIfNoSpecifiedTests=false`
5. Architecture and config guards: `mvn -pl openaev-api -am test -Dtest='TenantActiveTableAccessArchTest,TenantScopedEntrypointsTxCtxArchTest,TenantNonOrmAccessArchTest,TenantBackgroundTransactionArchTest,ImportMapperActivationConfigTest,TenantStatementInspectorTest' -Dsurefire.failIfNoSpecifiedTests=false`
6. Full regression: `mvn -ntp -pl openaev-api -am test`. Locally 6311 tests with no assets-related failure; three errors appeared on `uk_parameters_key_platform` during the startup snapshot restore, a pre-existing flakiness on classes that never touch assets.

**Use `-am`.** `mvn -pl openaev-api test` resolves `openaev-model` from `~/.m2`, and that jar corrupted five times in a day on this branch, producing convincing false regressions. With `-am` the module is built in the same reactor and read from `target/classes`.

### Related issues

* Related #6438
* Related #6422
* Related #6393


openaev-api/src/test/java/io/openaev/utils/fixtures/AssetFixture.java
```diff
@@ -12,6 +14,11 @@ public static Asset createDefaultAsset(@NotNull final String name) {
     asset.setUpdatedAt(Instant.now());
     asset.setName(name);
     asset.setDescription("asset description");
+    // assets is tenant-active and Asset's TenantBaseListener was removed at go-live: stamp the
+    // tenant explicitly here, the way SecurityCoverageFixture and AssetGroupFixture already do,
+    // instead of leaving every call site to remember it. TenantContext.getCurrentTenant() never
+    // throws (it defaults to Tenant.DEFAULT_TENANT_UUID), so this is safe outside a request too.
+    asset.setTenant(new Tenant(TenantContext.getCurrentTenant()));
```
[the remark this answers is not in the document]
laugiov: Fixed in d2868baa3. You were right: Asset keeps TenantBaseListener (see the note on the entity itself, and #7844), so the comment was describing a state that does not exist. Corrected to say the real reason the fixture stamps the tenant: it is what lets #7844 be a change to production code alone, not a guard against a NOT NULL failure today.

openaev-api/src/test/java/io/openaev/utils/fixtures/EndpointFixture.java
```diff
@@ -66,6 +68,11 @@ private static Endpoint baseEndpoint(String name, Endpoint.PLATFORM_TYPE platfor
     endpoint.setIps(EndpointMapper.setIps(IPS));
     endpoint.setPlatform(platform);
     endpoint.setArch(Endpoint.PLATFORM_ARCH.x86_64);
+    // assets is tenant-active and Asset's TenantBaseListener was removed at go-live: stamp the
+    // tenant explicitly here, the way SecurityCoverageFixture and AssetGroupFixture already do,
+    // instead of leaving every call site to remember it. TenantContext.getCurrentTenant() never
+    // throws (it defaults to Tenant.DEFAULT_TENANT_UUID), so this is safe outside a request too.
+    endpoint.setTenant(new Tenant(TenantContext.getCurrentTenant()));
```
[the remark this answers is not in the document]
laugiov: Fixed in d2868baa3. You were right: Asset keeps TenantBaseListener (see the note on the entity itself, and #7844), so the comment was describing a state that does not exist. Corrected to say the real reason the fixture stamps the tenant: it is what lets #7844 be a change to production code alone, not a guard against a NOT NULL failure today.

openaev-api/src/test/java/io/openaev/utils/fixtures/SecurityPlatformFixture.java
```diff
@@ -1,13 +1,20 @@
...
   public static SecurityPlatform createDefault(String name, String type) {
     SecurityPlatform edr = new SecurityPlatform();
     edr.setSecurityPlatformType(SecurityPlatform.SECURITY_PLATFORM_TYPE.valueOf(type));
     edr.setName(name);
     edr.setDescription("I don't see anything, hear anything, say anything");
+    // assets is tenant-active and Asset's TenantBaseListener was removed at go-live: stamp the
+    // tenant explicitly here, the way SecurityCoverageFixture and AssetGroupFixture already do,
+    // instead of leaving every call site to remember it. TenantContext.getCurrentTenant() never
+    // throws (it defaults to Tenant.DEFAULT_TENANT_UUID), so this is safe outside a request too.
+    edr.setTenant(new Tenant(TenantContext.getCurrentTenant()));
```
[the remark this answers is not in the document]
laugiov: Fixed in d2868baa3. You were right: Asset keeps TenantBaseListener (see the note on the entity itself, and #7844), so the comment was describing a state that does not exist. Corrected to say the real reason the fixture stamps the tenant: it is what lets #7844 be a change to production code alone, not a guard against a NOT NULL failure today.

openaev-api/src/test/java/io/openaev/utils/fixtures/composers/EndpointComposer.java
```diff
@@ -43,6 +45,18 @@ public Composer withTag(TagComposer.Composer tagComposer) {
 
     @Override
     public Composer persist() {
+      // assets is tenant-active and Asset carries no TenantBaseListener any more, so an entity a
+      // test built by hand has no tenant and the insert fails on the NOT NULL column. Stamp the
+      // ambient tenant here, and only when the caller left it unset, so a test that attributes
+      // deliberately (isolation tests, cross-tenant fixtures) keeps full control.
+      //
```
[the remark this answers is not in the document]
laugiov: Fixed in d2868baa3. You were right: Asset keeps TenantBaseListener (see the note on the entity itself, and #7844), so the comment was describing a state that does not exist. Corrected to say the real reason the fixture stamps the tenant: it is what lets #7844 be a change to production code alone, not a guard against a NOT NULL failure today.

openaev-api/src/test/java/io/openaev/utils/fixtures/composers/SecurityPlatformComposer.java
```diff
@@ -18,6 +20,18 @@ public Composer(SecurityPlatform securityPlatform) {
 
     @Override
     public Composer persist() {
+      // assets is tenant-active and Asset carries no TenantBaseListener any more, so an entity a
+      // test built by hand has no tenant and the insert fails on the NOT NULL column. Stamp the
+      // ambient tenant here, and only when the caller left it unset, so a test that attributes
+      // deliberately (isolation tests, cross-tenant fixtures) keeps full control.
+      //
```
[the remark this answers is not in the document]
laugiov: Fixed in d2868baa3. You were right: Asset keeps TenantBaseListener (see the note on the entity itself, and #7844), so the comment was describing a state that does not exist. Corrected to say the real reason the fixture stamps the tenant: it is what lets #7844 be a change to production code alone, not a guard against a NOT NULL failure today.

laugiov:
Three things landed on this branch since the review request.

**The five Copilot comments are fixed and their threads resolved** (d2868baa3). They were all one real inconsistency: the fixture and composer comments said `TenantBaseListener` had been removed at go-live, while `Asset` deliberately keeps it (#7844). The comments now give the actual reason the tenant is stamped there, which is to let #7844 be a change to production code alone rather than a guard against a failure that cannot happen today. Comments only, no behaviour change. The same two comments on the merged #7836 are corrected here too, since that PR is closed.

**A defect that was live on this branch is now fixed** (d60faa4b8). `VulnerableEndpointRepository.findForIndexing` grouped on `a.asset_id` while projecting five more columns of `assets`, relying on PostgreSQL's functional-dependency rule. That rule holds for base tables only, and the inspector rewrites the `assets` join into a derived table once the table is active, so the query stops being valid SQL and the vulnerable-endpoint search index does not build at all. It was found while arming the findings suites and originally fixed there; it is cherry-picked here because it belongs to this activation, and because merging this branch on its own would otherwise put a broken index on `main`.

**One thing worth knowing before reviewing the test coverage.** `@WithMockUser` builds a user with no `users_tenants` row, so an HTTP request it makes resolves `TxCtx.missing()` and every read of an activated table is denied. Production never has that state (`V4_95__Migrate_users_to_default_tenant` attaches every user to the default tenant). The practical effect here is that `EndpointApiTest`'s create tests assert on an entity returned from the persistence context after an INSERT, which the inspector never rewrites, so they pass whatever the scope is. The remedy is one line per suite, `TenantIsolationTestHelper.attachCurrentUserToTenant(...)`, which already exists. It must not be made a global default: doing so broke 8 tests across 6 classes, because `DEFAULT_TENANT_UUID` has a special meaning in `TxCtxArgumentResolver.fallbackSelector`.

Full suite on this stack: 6318 tests, 0 failures, 1 error which is the known snapshot flakiness (#7854, surefire reports `Run 2: PASS`).

laugiov:
Six remediation commits since the review request, after an adversarial self-review of this branch. Each fix carries a mutation check: the fix removed, the test observed to fail, the fix restored.

**Five defects with production impact**

- `b7e90e844` — **filtered LIVE notification triggers silently stopped firing.** `NotificationMatchingService.matches` is `@Transactional` with no `TxCtx`, reached from an `@Async` `@TransactionalEventListener` — a pool thread, after commit, no ambient transaction. `NotificationEngineService` documents in its own comment that it depends on the v1 Hibernate filter, which this branch removes from `Asset`. The catalog maps ASSET and ASSET_GROUP to activated entities, so a trigger with any non-empty filter matched nothing. No log, no exception: `matches` catches and returns false. Only the re-check is now scoped — `recordEvents` writes a non-activated table and `dispatch` performs external delivery that must not sit inside a transaction.
- `9dfd4c70d` — `vulnerable_endpoints_total` and `mappers_total` reported zero. `VulnerableEndpointRepository` is a `JpaRepository<Endpoint, String>`, so it reads the activated `assets` table; both gauges sat unscoped beside three that were not.
- `7a68cd55a` + `246a0144b` — **`QueueChainingJob` ran its whole transaction with no scope.** Four levels below what the job imports, `createReadySteps` reaches `ScopeService.getValidAssets` → `assetService.assets(ids)`. An ASSET_ID allowlist resolved to nothing and the delayed inject fired with no per-asset target.
- `2c6930c5d` — **`writeAllowlistScopeIsolated` persisted `inject_assets: []`.** `REQUIRES_NEW` suspends the caller's transaction, so the caller's scope does not travel; the realignment then wrote the fail-closed empty list onto every asset-centric step template. It now declares its own `TxCtx`.
- `246a0144b` — `ScopeSnapshotService` scoped from `TenantContext.getCurrentTenant()`, which falls back to the default tenant. On the timeout job a run of another tenant froze its rules as `DELETED_DURING_EXECUTION` for assets that still exist.

**Guards and tests**

- `20b203d4f` — the activation guard was a name registry: `every_active_table_is_guarded` asserted set membership only, so adding a string was enough. It now asserts a rule **exists** per active table. Four tables owe one and are listed in a shrink-only debt register with the reason (#7874). I did not write the `assets` allowlist: 26 accessor classes, and a list written without tracing each one's scope mechanism looks like a guard without being one.
- The vacuous assertion in `AssetHttpIsolationTest` is replaced — `AssetService.searchAssets` excludes SECURITY_PLATFORM unconditionally, so asserting B's platform is absent held with isolation entirely removed. Two discriminators now have positive halves.
- A comment citing `FindingApi.withScopedAssociationsInitialized`, which lives on the stacked branch and not here, is corrected.

**One thing the frozen ArchUnit store taught me.** My first chaining fix added the primitive beside the raw `TransactionTemplate`. The store refused: it is a ratchet that only shrinks, so a new violation must be fixed rather than frozen. Removing the raw plumbing entirely was the right answer and took that baseline from 6 violations to 3.

Also filed, none of it belonging to this PR: #7872, #7873, #7874.

laugiov:
Correction to my earlier comment on this PR, about `@WithMockUser` and tenant membership.

I wrote that `@WithMockUser` builds a user with no `users_tenants` row, so an armed suite reads nothing and `EndpointApiTest`'s coverage is weaker than it looks. **That was too general, and the part about this PR's suite is not something I demonstrated.**

What I actually measured, with a probe inside a failing test, was `FindingApiTest` on the stacked branch: `memberships=[]`, the seeded row present and correctly attributed, and the response empty. That is real and it is fixed there.

What I did not check before generalising is that `IntegrationTest.tenantUri(...)` attaches the user to a tenant as a side effect, so any test that builds a tenant-prefixed URL gets a membership. The gap therefore bites suites that use the non-prefixed routes and never call that helper — not `@WithMockUser` as such.

And `EndpointApiTest` is green here with genuinely non-empty read assertions (`$.content.[0].asset_id`, `hasSize(1)`, `$[0].asset_agent_id`) while `assets` is active, which is not the behaviour of a suite reading nothing. So the claim that its create tests pass whatever the scope does not carry, and I should not have relayed it as established.

The rest of that comment stands: the five production defects and the guard change are measured and each carries a mutation check.

.github/skills/activate-tenant-table/SKILL.md
```diff
@@ -49,6 +49,16 @@ incident. Do not trade them away to make a test pass.
    for the expected reason before you write any production code. Never weaken,
    delete or `@Disabled` an existing test to get green, with one exception
    introduced in Phase 2 and resolved in Phase 6 (the documented go-live guard).
+
+   **Red for the intended reason is not enough. Ask: "name a change to the
```
corinnekrych: I really dont understand that new paragraph...

.github/skills/activate-tenant-table/SKILL.md
```diff
@@ -276,6 +277,95 @@ Classify every hit:
...
   converting it in this run, it is a blocker: stop and report (Phase 0)
 
+#### Removing the v1 `@Filter` silently disarms every isolation test that does not activate the table
+
+The test profile declares no `openaev.tenant.active-tables`, so the inspector never fires in a test
+context unless that context sets it. Every isolation test therefore needs
+`@TestPropertySource(properties = "openaev.tenant.active-tables={table}")`, which the RED phase
+above already tells you to write.
+
+The trap is the other direction, and it is about tests you did NOT write. A suite that asserted this
```
corinnekrych, line 319: Good catch, i wonder all those Nested V1 isolation tests, are they still usefull? should we removed them and used the generated seeded v2 test. Those "old" test also clash in the way they are written (siwtch tenant in one test transaction).
Should we remove them or adapt them?

openaev-model/src/main/java/io/openaev/database/repository/EndpointRepository.java
damgouj, line 107: question: I am not sure I have understood why we keep the V1 and all is linked as here with :#{#tenantContext.currentTenant}

openaev-api/src/main/java/io/openaev/notification/engine/NotificationEngineService.java
```diff
@@ -78,7 +81,21 @@ public void handleEventWithMessage(
...
+          // catalog now spans v2-activated tables (assets, asset_groups, findings). Those no longer
+          // carry the v1 @Filter the comment above relies on, and this runs on an @Async thread
+          // after commit, so there is no ambient transaction and no scope: without this the count
+          // is filtered by can_access_tenant against an empty app.current_tenants, comes back zero,
+          // and every filtered LIVE trigger on those resources silently stops firing.
+          //
+          // Only the re-check is wrapped. recordEvents writes notification_event_records, which is
+          // not activated, and dispatch performs external delivery that must not sit inside a
+          // transaction.
+          boolean matched =
```
corinnekrych: Good catch. But this is outside the scope of assets activation, ie: fixing an existing bug or is it linked to assets v2 activation?

.github/skills/activate-tenant-table/SKILL.md
laugiov: You are right, the paragraph is badly written.

What I meant is one sentence: a test that turns red when you remove the fix does not yet prove it tests the behaviour, it may only test that one fix.

The real example found by the AI : a telemetry gauge and its scoped counting method. The test called the method directly. Remove the scope from the method → red. But re-wire the gauge to the unscoped repository → still green, and that wiring was the part that was broken in production.

I will rewrite it, name the technique (mutation testing) and give the action: remove the fix a second time, from a different place.

.github/skills/activate-tenant-table/SKILL.md
laugiov: Good question, and I hit this twice yesterday.

The two styles cannot live in the same class, for a structural reason: the v1 style switches tenant inside one transaction, while a v2 scope is per transaction and is set by the request. When I made the helper move the v2 scope too, 10 HTTP isolation tests failed with "scope is already set … tried to change it", the aspect refuses to redefine what the request resolved.

So: a class that goes through HTTP should create real tenants and users and let the request resolve the scope, with no thread-local. A class that calls a service directly needs an explicit scope, and cannot host HTTP tests as well.

I would not delete them: for many endpoints they are the only isolation coverage we have. 26 suites have a nested TenantIsolation class and 6 declare active-tables; how many of the rest are actually inert needs a proper survey, not a grep, I do not have a number I trust yet.

The real fix is structural: a guard that fails any suite asserting isolation without declaring its table.

openaev-api/src/main/java/io/openaev/notification/engine/NotificationEngineService.java
laugiov: Fair question. Two parts.

It was in scope: the defect is caused by this PR, not pre-existing. NotificationEngineService says in its own comment that it relies on the v1 Hibernate filter, and this PR removes that filter from Asset. The catalog maps ASSET and ASSET_GROUP to activated entities, so a LIVE trigger with a filter stopped firing, with no log and no exception.

But it is moot now: main fixed the same defect in #6406, more broadly than I did. I merged main and kept their version, so the file left this PR's diff.

damgouj (approved):
Tests OK but 1 question remaining

openaev-model/src/main/java/io/openaev/database/repository/EndpointRepository.java
laugiov: This query crosses two tables with different status.

It selects a (assets, activated by this PR) through Inject (injects, not activated). The inspector rewrites the assets side with can_access_tenant, but it does nothing for injects, that table is still on v1. So i.tenant.id = :#{#tenantContext.currentTenant} is the only thing isolating the inject side. Remove it and the query would return endpoints reached through another tenant's injects.

That is why v1 is still here: the migration is table by table, and a query that joins an activated table to a non-activated one needs both mechanisms at once.

The line just below, in findAllEndpointsForAtomicTestingsSimulationsAndScenarios, is the opposite case: it filters e.tenant_id on assets, so it is now redundant with can_access_tenant. I left it, but it can go once the migration is finished, same family as #7844.

openaev-api/src/main/java/io/openaev/scheduler/jobs/WorkflowTimeoutJob.java
```diff
@@ -34,7 +36,21 @@ public void execute(JobExecutionContext jobExecutionContext) {
 
     for (Workflow workflow : expiredWorkflows) {
       try {
-        workflowEndService.forceCompleteWorkflowByTimeout(workflow);
+        // findAllExpiredRunWorkflows spans tenants and this job carries no scope of its own, so the
+        // end snapshot resolved its assets with none: every rule of a non-default tenant froze as
+        // DELETED_DURING_EXECUTION for assets that still exist. One scope per workflow, taken from
+        // the simulation that owns it (workflows carry no tenant column).
+        String tenantId = tenantOf(workflow);
+        if (tenantId == null) {
```
corinnekrych, line 44: I'm puzzled: why workflow table does not have a tenantId?

corinnekrych: With chk_workflow_simulation_or_scenario requiring:
```
(workflow_simulation_id IS NOT NULL AND workflow_scenario_id IS NULL)
OR
(workflow_simulation_id IS NULL AND workflow_scenario_id IS NOT NULL)
```
should we either take tenantId from simulation or sceanio?

openaev-api/src/main/java/io/openaev/scheduler/jobs/QueueChainingJob.java
```diff
@@ -69,6 +98,20 @@ public void execute(JobExecutionContext jobExecutionContext) throws JobExecution
...
           }
+          TenantContext.clearCurrentTenant();
         });
   }
+
+  /**
+   * The tenant a delayed step belongs to. {@code workflows} carries no tenant column of its own;
+   * the owning simulation does, through {@code TenantBase}.
+   */
+  private static String tenantOf(StepDelayQueue stepDelayQueue) {
```
corinnekrych, line 111: should we try to find the tenantId only from Simulation? could it be from scenario?

openaev-api/src/main/java/io/openaev/scheduler/jobs/WorkflowTimeoutJob.java
laugiov: Not a design decision, a gap: workflows was created in V4_72, three migrations before V4_75 which added tenant_id everywhere, and it was not in that migration's list. Same for steps, steps_delay_queue, conditions and workflow_scope_rules.

As long as these tables have no tenant column, we cannot activate them on v2 and we keep deriving the tenant by hand on every new path. Worth raising with Vertical 1.

laugiov: Good catch on With chk_workflow_simulation_or_scenario requiring, i will fix this

openaev-api/src/test/java/io/openaev/rest/AiTargetApiTest.java
```diff
@@ -32,6 +35,18 @@
 class AiTargetApiTest extends IntegrationTest {
 
   @Autowired private MockMvc mvc;
+  @Autowired private TenantIsolationTestHelper tenantIsolationHelper;
+
+  @BeforeEach
+  void attachMockUserToDefaultTenant() {
```
corinnekrych: on main, @WithMockUser does have a tenant-related property now: autoJoinDefaultTenant.
test helper could potentially be replaced with @WithMockUser(autoJoinDefaultTenant = true)

openaev-api/src/test/java/io/openaev/rest/ImportExportMapperApiTest.java
```diff
@@ -35,9 +38,20 @@
 public class ImportExportMapperApiTest extends IntegrationTest {
 
   @Autowired private MockMvc mvc;
+  @Autowired private TenantIsolationTestHelper tenantIsolationHelper;
   @Autowired private EndpointRepository endpointRepository;
   @Autowired private TagRepository tagRepository;
 
+  @BeforeEach
```
corinnekrych: could you use the new annotation?

openaev-model/src/main/java/io/openaev/database/model/Asset.java
```diff
@@ -54,9 +53,35 @@
 @DiscriminatorColumn(name = "asset_type", discriminatorType = STRING)
 @DiscriminatorValue(AssetType.Values.ASSET_TYPE)
 @EntityListeners({ModelBaseListener.class, TenantBaseListener.class})
```
corinnekrych: i'd removed TenantBaseListener

openaev-api/src/main/java/io/openaev/scheduler/jobs/QueueChainingJob.java
laugiov: You are right, and it was the same helper in both jobs. chk_workflow_simulation_or_scenario allows exactly one of a simulation or a scenario, and Scenario is TenantBase like Exercise, so a scenario-backed workflow does have a tenant, and my code dropped it and ran unscoped.

Fixed in d27c2be90: simulation first, then scenario, in QueueChainingJob and WorkflowTimeoutJob. The javadoc above was wrong too and is corrected.

One nuance for this file: steps_delay_queue_workflow_run_id is nullable, so an entry with no workflow run at all is still possible and keeps the null branch. When that happens the step is processed with a warning rather than skipped, dropping a queued step would be the same silent loss the scoping is meant to prevent.

Good catch. I checked "no simulation" and stopped there, without reading the constraint.

openaev-api/src/test/java/io/openaev/rest/AiTargetApiTest.java
laugiov: Good catch, thanks. Done in a25a3e4e6: the class now uses
@WithMockUser(isAdmin = true, autoJoinDefaultTenant = true) and the @BeforeEach is gone.

openaev-api/src/test/java/io/openaev/rest/ImportExportMapperApiTest.java
laugiov: Done.

openaev-model/src/main/java/io/openaev/database/model/Asset.java
laugiov: Done.

PR #7852, opened 2026-09-08
chore(deps): track every Spring Boot BOM override and check it in CI

laugiov:
### Proposed changes

| file | change |
| --- | --- |
| `renovate.json5` | +11 / -7. Adds the `kotlin.version` customManager, removes the `html-to-image` rule. |
| `pom.xml` | -1. Removes the dead `validation-api.version` property. |
| `.github/scripts/check-bom-overrides.py` | new, 367 lines. The check. |
| `.github/actions/bom-override-guard/action.yml` | new, 53 lines. Runs it. |
| `.github/workflows/_ci-pipeline.yml` | +23 / -3. New job, added to the gate as advisory. |
| `.gitignore` | +5. `__pycache__/` and `*.pyc`. |
| `.github/dependabot.yml` | deleted, 34 lines. |

A version pinned by hand for a security reason can stop being tracked without anyone seeing it. That
is what happened to Kotlin: it was pinned on 2026-08-19 and between then and today nothing has
proposed an update for it and nothing would have reported it going stale.

This PR fixes that one case, removes two settings that describe a repository we no longer have, and
adds a CI check so the next one is found by a machine instead of by reading the pom.

#### 1. `kotlin.version` had no tracking path

The root `pom.xml` redefines 17 properties that the `spring-boot-dependencies` BOM also defines. Each
one silently changes the version of every artifact the BOM manages through that property. Sixteen of
the 17 are tracked: eight by a `customManagers` entry, eight because a declared dependency uses the
property so the maven manager sees it. `kotlin.version` had neither.

Reproduce the list of overrides:

```
V=$(grep -oP '(?<=<version>)[^<]+' pom.xml | head -1)
curl -sfL "https://repo1.maven.org/maven2/org/springframework/boot/spring-boot-dependencies/$V/spring-boot-dependencies-$V.pom" -o /tmp/boot-bom.pom
for p in $(grep -oE "<[a-zA-Z0-9._-]+\.version>" pom.xml | tr -d '<>' | sort -u); do
  grep -q "<$p>" /tmp/boot-bom.pom && echo "$p boot=$(grep -oP "(?<=<$p>)[^<]*" /tmp/boot-bom.pom | head -1) repo=$(grep -oP "(?<=<$p>)[^<]*" pom.xml | head -1)"
done
```

Spring Boot 3.5.16 ships Kotlin 1.9.25, the repo pins 2.1.21. Nothing in any pom references
`${kotlin.version}`, which is why the maven manager cannot see it either:
`grep -rF --include=pom.xml '${kotlin.version}' .` returns nothing.

The property arrived in `0d0e99ead`, `fix(security): pin patched dependency versions`, on 2026-08-19,
in the same hunk as `logback.version`, `log4j2.version`, `httpclient5.version` and
`opentelemetry.version`. `git show 0d0e99ead -- pom.xml` shows all five. Nine days later
`b77a519be`, `chore(deps): let Renovate see the Spring Boot BOM overrides`, added managers for the
ones that needed one. Kotlin was missed.

Two things a reviewer should know before deciding what to do next.

`kotlin-stdlib 2.1.21` was released on 2025-05-12. The current release is `2.4.20`, published
2026-09-07. So the pin was already fifteen months old when it was set, and it is now three minor
lines behind:

```
curl -sfL https://repo1.maven.org/maven2/org/jetbrains/kotlin/kotlin-stdlib/maven-metadata.xml | grep -o '<release>[^<]*'
curl -sfI https://repo1.maven.org/maven2/org/jetbrains/kotlin/kotlin-stdlib/2.1.21/kotlin-stdlib-2.1.21.pom | grep -i last-modified
```

And the security reason for this specific line is not recorded anywhere I can find. The commit is
labelled `fix(security)` but bundles five properties, the `(#7497)` in its subject points at an
unrelated CI issue about a nightly Trivy scan, and OSV reports no advisory for `kotlin-stdlib` at
either the BOM version or the pinned one:

```
curl -sfL -X POST https://api.osv.dev/v1/query \
  -d '{"package":{"name":"org.jetbrains.kotlin:kotlin-stdlib","ecosystem":"Maven"},"version":"1.9.25"}'
# {}  and the same for 2.1.21
```

This PR does not bump Kotlin. It adds the manager and lets Renovate open that PR, for two reasons.
Kotlin is a runtime library that belongs to the class where a green pipeline is not evidence, so the
bump wants a named reviewer rather than being smuggled into a config fix. And the Renovate PR
appearing is the acceptance test for the fix: if no PR shows up on the next scheduled run, the
manager does not work.

The reviewer's second question is worth asking out loud: is the pin still wanted at all, and if so
why? If nobody can answer, the answer is probably to drop the override and let the BOM manage Kotlin.

#### 2. `html-to-image` was marked `[DANGER]` and carries no patch

`renovate.json5` marked it with `commitMessagePrefix: '[DANGER]'` and the suffix
`(html-to-image has a yarn patch)`. There is no such patch. `openaev-front/patches/` holds one file,
`react-apexcharts.patch`, and the `resolutions` block in `openaev-front/package.json` references only
that one. `html-to-image` is a plain dependency pinned at `1.11.13`.

```
ls openaev-front/patches/                       # react-apexcharts.patch
grep -n 'html-to-image\|patch:' openaev-front/package.json
```

A marker that fires on something harmless is worse than no marker, because it teaches the reviewer
that `[DANGER]` does not mean danger. Removing it leaves the marker on the one dependency that does
carry a patch.

#### 3. `validation-api.version` is dead

`pom.xml` set `<validation-api.version>2.0.1.Final</validation-api.version>`. Nothing references it,
and it is not a Spring Boot BOM property either, so it changed nothing at all. The validation API in
use is `jakarta.validation:jakarta.validation-api`, declared with no version in
`openaev-model/pom.xml` and `openaev-api/pom.xml`, so the BOM's `jakarta-validation.version` (3.0.2)
governs it.

```
grep -rn 'validation-api' --include=pom.xml .
grep -c 'validation-api.version' /tmp/boot-bom.pom     # 0
```

Dependency resolution is unchanged by the removal. `mvn -o -pl openaev-api -am dependency:list
-DincludeScope=runtime` returns 347 resolved entries before and after, and a `diff` of the two sorted
lists is empty.

#### 4. A CI check so the next one is found by a machine

`.github/scripts/check-bom-overrides.py` checks three things about every Spring Boot BOM override:

- that Renovate tracks it, through a declared dependency or through a `customManagers` entry;
- that it still wins, meaning Maven resolves the version the property gives and not the version the
  BOM would give;
- that it wins everywhere, meaning no artifact it governs resolves to two different versions across
  the modules. `dependency:list` is run with `appendOutput`, so the file holds one section per module,
  and an artifact that resolves to 10.1.59 in one and 10.1.55 in another is the drift this check
  exists to find rather than something to average away.

Nothing in it is a hand-maintained list, which is the whole point: a list is what failed on
2026-08-19. The set of overrides is the intersection of the pom properties with the BOM properties.
The property to artifact mapping comes from the BOM's own `dependencyManagement`, followed one level
into imported BOMs such as `jackson-bom` and `netty-bom`. Expected versions come from those BOMs
rather than being assumed equal to the property, because a BOM member can carry its own version:
`jackson-bom 2.22.2` ships `jackson-annotations 2.22`, and `netty-bom 4.2.17.Final` ships
`netty-tcnative-boringssl-static 2.0.81.Final`. Getting that wrong is what produced eight false
positives in the first draft.

Run it locally:

```
mvn -q -pl openaev-api -am dependency:list -DincludeScope=runtime \
    -DoutputFile="$PWD/target/deplist.txt" -DappendOutput=true
python3 .github/scripts/check-bom-overrides.py --dependency-list target/deplist.txt
```

On `main` at `19d90520a` it reports exactly the finding above and nothing else:

```
kotlin.version                   2.1.21                 1.9.25             NOTHING
...
::warning title=BOM override with no Renovate tracking::kotlin.version. Nothing will ever propose an
update, and nothing will report it going stale.
::notice title=Version decided outside the BOM::org.jetbrains.kotlin:kotlin-stdlib-common resolves to
1.9.10, neither 2.1.21 from ${kotlin.version} nor the Spring Boot version 1.9.25
FOUND 1 problem(s): 1 untracked, 0 not taking effect.
```

With the `renovate.json5` change in this PR it is green: 17 overrides, 74 artifact assertions passing,
0 drift, exit 0, and one notice left over.

That notice is a real observation and it is not fixed here. `kotlin-stdlib-common` resolves to 1.9.10
while `kotlin-stdlib` resolves to 2.1.21. Kotlin 2.x folded `-common` into `-stdlib`, so this may be
a duplicate-class situation rather than a version that is simply behind. I did not establish which
dependency pulls it in, and `dependency:tree -Dincludes=` did not answer it. It is reported as a
notice, not a failure, and it needs whoever owns the Kotlin pin.

**Advisory, not blocking.** The job is in the `pipeline-gate` needs list, and the gate's advisory
list, so a finding prints a warning and the gate still passes. Making it blocking means deleting the
`bom-override-guard)` branch from that list in `_ci-pipeline.yml`. That is deliberately a separate
change: the first draft of this check produced eight false positives from the BOM member versions and
six more from BOM imports, and none of them are things you want to discover while blocking a merge.
The next step is to watch it on real PRs for a few weeks and then flip it, and that step should also
add a unit test with fixture poms, which this PR does not have.

The job reuses the jars from Backend Compile rather than building again, the same way the API test
shards do. That is not cosmetic: `dependency:list` is a bare plugin goal, so it never builds the
reactor modules, and without the `io.openaev` jars in the local repository it fails on
`openaev-framework` with `openaev-model:jar ... has not been downloaded from it before`. That was
verified by running it against a local repository with `io/openaev` deleted.

#### 5. `.github/dependabot.yml` is deleted

**In one line: delete it, because it produces nothing and a second dependency bot declared in a
public repository is a trap for contributors.**

All four ecosystems were declared with `open-pull-requests-limit: 0`, which turns off version-update
PRs, and repository Dependabot security updates are disabled:

```
gh api repos/OpenAEV-Platform/openaev --jq .security_and_analysis.dependabot_security_updates
# {"status":"disabled"}
```

Deleting the file is behaviour-neutral. No `dependabot.yml` means no version updates, which is the
same outcome as a limit of 0. Dependabot **alerts** are a separate repository setting and stay on:
`gh api repos/OpenAEV-Platform/openaev/vulnerability-alerts` returns 204, which means enabled, and
this PR does not touch it.

---

### Testing Instructions

1. `renovate-config-validator --no-global --strict renovate.json5`. Output below.
2. Run the check locally with the two commands in section 4 and confirm it exits 0 and prints
   `all 17 Spring Boot BOM overrides are tracked and take effect`.
3. Break it on purpose to confirm it is not just printing OK. Three ways, all of which were run:
   - delete the `kotlin.version` `customManagers` entry and re-run: reports
     `kotlin.version ... NOTHING` and exits 1;
   - rewrite `tomcat-embed-core:jar:10.1.59` to `10.1.55` in `target/deplist.txt` and re-run: reports
     `DRIFT ... = 10.1.55, which is the Spring Boot version`;
   - append a second `tomcat-embed-core` line at `10.1.55` and re-run: reports
     `SPLIT ... = {10.1.55, 10.1.59}, more than one version on the classpath`.
4. `mvn -pl openaev-api -am dependency:list -DincludeScope=runtime` before and after the `pom.xml`
   change and diff the two: identical.
5. In CI, look at the `✅ Pipeline Gate` step output. `🔎 BOM Override Guard` should appear with a
   `✅`, and if it ever fails it appears with `⚠️` and the line
   `⚠️ 1 advisory job(s) reported problems — not blocking`.

The advisory wiring was checked by extracting the gate's shell block and running it with a synthetic
`needs` payload: guard red alone exits 0, guard red plus a real failure exits 1.

### Related issues

* Related #ISSUE-NUMBER

### Checklist

### Further comments

`renovate-config-validator` output, renovate 44.69.9:

```
$ renovate-config-validator --no-global --strict renovate.json5
 INFO: Validating renovate.json5 as repo config
 INFO: Config validated successfully against 1 file(s)
$ echo $?
0
```

Two notes on that. The validator checks option names, not option values: a deliberately invalid
`matchUpdateTypes: ['bogusType']` passes, while an invalid option name fails. And without
`--no-global` it validates against the global-config schema, which is the wrong one for this file.

The `kotlin.version` manager was also checked by running Renovate's own regex extractor against
`pom.xml`, which returns
`{"depName":"org.jetbrains.kotlin:kotlin-stdlib","currentValue":"2.1.21","datasource":"maven"}`.

This PR does not depend on PR 2 and PR 2 does not depend on this one. Either can merge first.

---

#### What a reviewer should check

Ordered by how much it matters, and each one is checkable without redoing the analysis.

1. **Is the Kotlin pin still wanted, and why?** This is the only question in the PR that I cannot
   answer from the repository. OSV has no advisory for `kotlin-stdlib` at 1.9.25 or 2.1.21, and the
   commit that added the pin bundles five properties under one `fix(security)` subject with an issue
   reference that points somewhere else. If the reason is gone, deleting the override is better than
   tracking it.
2. **Is `kotlin-stdlib-common` at 1.9.10 a problem?** The check reports it as a notice. Reproduce with
   `grep org.jetbrains.kotlin target/deplist.txt`.
3. **Does the guard job add wall time to the pipeline?** I have not run it in real CI. It waits for
   the Backend Compile artifacts, then resolves and checks, which took about four minutes locally on
   a warm cache. The E2E shards dominate the critical path at around thirteen minutes, so it should
   be free, but that is a projection and the first CI run settles it. `timeout-minutes: 15`.
4. **Is a new network dependency in CI acceptable?** The check downloads BOM poms from
   `repo1.maven.org` at run time. Maven already talks to Central in this pipeline, and a failure is
   advisory, so a Central outage prints a warning rather than blocking. Say so if you disagree.
5. **Is running a `.github/scripts/` file from CI a new pattern here?** It is. That directory holds
   three python helpers today, all run by hand. The composite action is the thin wrapper the other
   jobs use, so the job shape matches; only the script location is new.
6. **The advisory list in `_ci-pipeline.yml` changed shape**, from an `if` to a `case`, so a second
   job can be advisory. The existing `container-vulnerability-scan` behaviour is unchanged, and the
   comment explaining why it is decided in the gate rather than with `continue-on-error` is kept. I
   used an explicit `if` inside the branch rather than `[ test ] && var=1`, which survives `set -e`
   but only if you know the AND-OR-list rule.
7. **No unit test for the script.** Correctness was established by running it against this repository,
   17 overrides and 74 artifact assertions, plus a deliberate negative test. That is acceptable while
   the check is advisory and it is not acceptable when it goes blocking.


laugiov:
The open question in the description about wall time is settled by the first run.

`🔎 BOM Override Guard` took 3 minutes 39 seconds, 18:32:16 to 18:35:55, and the Pipeline Gate
finished at 18:54:26, nineteen minutes later. So it is not on the critical path, which is what the
description said needed confirming.

Output from the run:

```
Spring Boot parent      : 3.5.16
poms scanned            : 6
of which BOM overrides  : 17
artifacts on classpath  : 322
kotlin.version   2.1.21   1.9.25   renovate customManager
OK     org.jetbrains.kotlin:kotlin-stdlib = 2.1.21
NOTE   org.jetbrains.kotlin:kotlin-stdlib-common = 1.9.10, expected 2.1.21
OK: all 17 Spring Boot BOM overrides are tracked and take effect (1 notice(s), 0 not on the runtime classpath)
```

The `kotlin.version` line now reads `renovate customManager` rather than `NOTHING`, which is the
acceptance test for the fix in this PR.

The single notice is the `kotlin-stdlib-common` at 1.9.10 observation described above. It is reported
and does not fail the job, and it still needs whoever owns the Kotlin pin.

One thing worth flagging from the review of this PR rather than hiding it: Copilot reports it reviewed
6 of the 7 changed files and generated no comments. The file it most likely skipped is
`check-bom-overrides.py`, 367 lines, which is all of the new logic here. So the clean automated review
should not be read as coverage of the script.

guillaumejparis (approved):

Issue #7844, opened 2026-09-08
chore(multi-tenancy): remove TenantBaseListener from the 8 activated tables that still carry it


laugiov:
## Context

Raised by @corinnekrych while reviewing the `asset_groups` activation (#6435, PR #7836):

> I think we should remove the TenantBaseListener. i dont know why some activated table where not
> removed but they should it's part of the full clean-up

The observation is correct and it is about the platform, not that one table, so it needs its own
issue rather than disappearing with that PR.

`TenantBaseListener.@PrePersist` stamps `tenant_id` from `TenantContext.getCurrentTenant()` on every
insert. On a v2-active table that is the wrong mechanism in two ways:

- **It attributes from the v1 thread-local**, which falls back to `Tenant.DEFAULT_TENANT_UUID` when
  nothing set it. Off the request thread (a scheduled job, a queue consumer, a provisioning
  datapack) that default is the wrong tenant as often as the right one, and nothing says so.
- **It makes an unattributed write look successful.** With the listener gone, a create that forgot
  its tenant fails on the NOT NULL column, loudly, instead of silently landing in the default
  tenant. That is the whole point of explicit write attribution.

The activate-tenant-table skill already requires removing it at go-live. It has not been applied
consistently.

## Current state

Of the 19 tables in `openaev.tenant.active-tables`, **8 still carry a tenant listener**:

| still carrying it | already clean |
|---|---|
| `import_mappers` | `lessons_templates`, `collectors`, `executors`, `injectors` |
| `mitigations` | `autonomous_runs`, `autonomous_events`, `autonomous_directives` |
| `cwes` | `kill_chain_phases`, `security_coverages`, `domains`, `asset_groups` |
| `attackpath_execution`, `attackpath_finding` | |
| `secret_references`, `secrets` | |
| `connector_instances` | |

## What removing it takes

Smaller than it looks, on the evidence of `asset_groups`. The dependency is usually concentrated in
the test fixture, not spread across call sites: removing the listener there broke a batch of tests,
and a single line in `AssetGroupFixture` fixed all of them, mirroring what `SecurityCoverageFixture`
already did.

Per table:

1. Confirm every production create path sets the tenant explicitly (a required `tenantId` parameter
   on the service method is the pattern that makes a missed path a compile error).
2. Drop `TenantBaseListener` from the entity's `@EntityListeners`.
3. Stamp the tenant in the entity's test fixture, the way `SecurityCoverageFixture` and
[...]

PR #7856, opened 2026-09-09
feat(multitenancy): activate findings on v2 isolation (#6420)

laugiov:
### Proposed changes

Activates the `findings` table on multi-tenancy v2 (#6420, #6421, #6442), and fixes the defects the
activation exposed. It branches off `feat/mtv2-activate-assets` (PR #7838) rather than `main`, so
review it as a stack.

The go-live is the usual shape: `findings` appended to `openaev.tenant.active-tables`, the v1
`@Filter` removed from `Finding` (`TenantBaseListener` deliberately kept, as on `assets`; #7844
removes both), `GUARDED_TABLES` and the production-config guard extended, and the four suites that
touch findings armed with `@TestPropertySource`.

### Two production defects, both live before any activation

* **A finding created one at a time was attributed to the ambient tenant, not to its inject's.**
  `createFindings` (plural) sets `finding.setTenant(inject.getTenant())`; `createFinding` (singular)
  left it to `TenantBaseListener`, which stamps whatever `TenantContext` holds — ambient, and
  defaulting to `DEFAULT_TENANT_UUID` off the request path. `FindingWriteAttributionTest` sets the
  ambient tenant to a *different* tenant from the inject's on purpose, which is the only arrangement
  that tells the two apart, and reads ground truth back through raw JDBC because a scoped read could
  not see a row sitting in the wrong tenant.

* **`GET /api/findings/{id}` returned `finding_asset_groups: []` for every finding.**
  `Finding.getAssetGroups()` is a computed `@JsonProperty` walking `inject.getAssetGroups()` on the
  already-active `asset_groups`, and `Finding.assets` is a lazy `@ManyToMany` on `assets`. Jackson
  resolves both AFTER the handler returns, through open-in-view: the session is still open so nothing
  throws, but the transaction and its scope are gone and the inspector fail-closes. Both are now
  resolved inside the scoped transaction, on the three handlers that serialize a `Finding` — the same
  fix as `SecurityPlatformApi` (#7026).

### A third defect the activation itself would have caused

`FindingWriter.saveCompleteFinding` is `REQUIRES_NEW`. That suspends the caller's transaction, so the
caller's scope does not travel with it and the new transaction starts with `app.current_tenants`
unset. The upsert's `ON CONFLICT DO UPDATE … WHERE can_access_tenant(...)` is then false for every
row: **a re-detected finding could never be updated**, on every write path, once the table is active.
The null id was then inserted into `findings_assets`, so the failure surfaced as a NOT NULL violation
on a different table.

Both production entry points were checked before touching anything: the queued path scopes through
`BatchingInjectStatusService`'s `tenantTx.execute(TxCtx.forTenant(...))`, the direct path through
`InjectApi#injectExecutionCallback`'s `TxCtx`. Both carry a scope; neither survives `REQUIRES_NEW`.
`saveCompleteFinding` now scopes its own transaction from the `tenantId` it already receives, which
widens nothing: that same value is what the INSERT writes into `tenant_id`.

### The search index would not build

`FindingRepository.findForIndexing` grouped on `f.finding_id` while projecting seven more columns,
relying on PostgreSQL's functional-dependency rule. That rule holds for base tables only, and the
inspector rewrites `FROM findings f` into a derived table, so the query becomes invalid SQL rather
than degrading.

### What arming the suites actually took, and the false greens it removed

**Removing a v1 `@Filter` disarms every isolation suite that does not activate the table.** The test
profile declares no `active-tables`, so the inspector never fires in a context that does not set one.
`FindingApiTest`, `FindingSearchApiTest`, `FindingServiceTest` and `DashboardApiTest` carry dozens of
cross-tenant references between them and none declared `@TestPropertySource`.

Arming them exposed four distinct false greens:

* **`FindingApiTest` was comparing two empty sets.** Its nested classes asserted the HTTP response
  against `findingRepository.findAll()` read from the same unscoped test transaction. Both sides were
  empty, so eight tests passed on `[] == []`. `@WithMockUser` builds a user with no `users_tenants`
  row, so every scope its requests resolve is `TxCtx.missing()` — a state production never has, since
  `V4_95__Migrate_users_to_default_tenant` attaches every user to the default tenant. The suite now
  calls the existing `TenantIsolationTestHelper.attachCurrentUserToTenant(...)`.
* **`FindingServiceTest` switched tenants through `TenantContext` alone**, the v1 thread-local v2
  ignores. `notBeReadableFromTenantY` passed because **nothing** was readable, and
  `beReadableFromTenantX` failed. `switchToTenant` now moves both mechanisms.
* **`DashboardApiTest` indexed one layer below the scope.** It called `engineService.bulkProcessing`
  directly, where `EngineSyncExecutionJob` is what carries `TxCtx.allTenants()`. The index stayed
  empty, and the three `all shards failed` errors were sorts against an index with no mapping, not
  query errors.
* **The inspector rewrites an `UPDATE`, not only a `SELECT`.** `DashboardApiTest`'s test-only date
  setters silently updated zero rows without a scope, every endpoint kept `now()` as its creation
  date, and the date-range assertions counted all of them.

A sink pin must also not be `@Transactional` or it proves nothing: with the test holding the
transaction open, the handler joins it and Jackson serializes inside it, where the scope still lives.
The same flaw was in #7838's `AssetAssociationSinkTest`, and correcting it exposed a live defect on
`AssetGroupApi`. Both rules are in `.github/skills/activate-tenant-table/SKILL.md`.

### The fail-closed detector found one defect this branch had missed

Run with `-Dopenaev.failclosed.detector=on`, the gate named `MultiIdListSerializer.serialize` reading
`assets` with no scope. `AssetGroup.assets` is serialized as ids by that serializer, which walks the
lazy collection while Jackson runs, after the transaction and its scope are gone. The fix above
initialised `inject.getAssetGroups()` - the groups - but not each group's own assets, so
`GET /api/findings/{id}` returned its groups with `asset_group_assets: []`. The same defect, one
level down.

Worth knowing why the test did not catch it: it asserted `body.contains(assetId)` on the whole JSON
body, and that id appears twice in the response - once in `finding_assets`, which *was* initialised,
and once inside `asset_group_assets`, which was not. The first occurrence satisfied the assertion.
It now asserts on exact JSON paths, and the seed links the asset to the group so the nested array
means something. Confirmed by mutation.

The same run named two production reads that pre-date this branch, on `domains` and
`kill_chain_phases`. They are filed as #7872 and deliberately not waived here: adding a signature
nobody has traced to its production entry point is how a real defect gets silenced.

### Also verified, and safe

* The `cascade = ALL, orphanRemoval = true` on `Inject.findings` does not delete data.
  `FindingOrphanRemovalTest` initialises the collection unscoped on purpose (it comes back empty),
  saves the inject, and reads the rows back through raw JDBC: both survive.
* The public tracking callbacks are not the blocker the spec expected. `PhishingPublicApi` carries
  the tenant in its path; `HostedPublicApi` does not, but only ever writes findings, with the tenant
  taken from the inject, and reads no active table.
* `AttackPathExecutionIngestionService` and `AttackPathFindingIngestionService` reach the primitive
  with an explicit tenant. `QueueChainingJob` is raw plumbing already on the frozen ArchUnit
  baseline, and touches no active table.
* `findings_total` telemetry is scoped with `TxCtx.allTenants()`, like the asset gauges in #7838.

### Testing Instructions

1. Start the services from `openaev-dev/docker-compose.yml`.
2. `mvn -ntp clean install -DskipTests && mvn -ntp -pl openaev-api test -Dtest='FindingApiTest,FindingSearchApiTest,FindingServiceTest,DashboardApiTest,FindingWriteAttributionTest,FindingAssetGroupSinkTest,FindingOrphanRemovalTest,ProductInventoryTenantScopeTest,TenantActiveTableAccessArchTest,ImportMapperActivationConfigTest' -Dsurefire.failIfNoSpecifiedTests=false`

Two environment notes that cost real time here:

**Run the two goals in one command and do not use `-am`.** The VSCode Java language server
recompiles into the same `target/classes` Maven reads, with its own Lombok agent, so a build started
right after an edit can fail with `bad class file`, a Lombok-less cascade of `cannot find symbol`, or
`failed to discover tests`. It is intermittent and a rebuild clears it. This is the same phenomenon
recorded as a "corrupted `openaev-model` jar" during lots A and B.

**Check `Time to start RabbitMQ` before believing a broker failure.** The dev broker accumulates
per-tenant test queues (51 857 queues and 69 118 exchanges here, zero consumers) and took 348 seconds
to finish starting after a reboot. Runs launched inside that window fail with an AMQP `EOFException`
that looks exactly like a code defect.

### Related issues

* Related #6420
* Related #6421
* Related #6442
* Related #6393


openaev-api/src/main/resources/application.properties
```diff
@@ -628,7 +628,7 @@ openaev.enabled-dev-features=
 # table, filtered through the finding it is joined to).
 # autonomous_runs / autonomous_events / autonomous_directives are v2-native (TenantBase, no
 # @Filter; TenantBaseListener removed on activation) so they MUST stay here (#7396).
-openaev.tenant.active-tables=import_mappers,lessons_templates,mitigations,cwes,collectors,executors,injectors,attackpath_execution,attackpath_finding,secret_references,secrets,connector_instances,autonomous_runs,autonomous_events,autonomous_directives,kill_chain_phases,security_coverages,domains,asset_groups,assets
+openaev.tenant.active-tables=import_mappers,lessons_templates,mitigations,cwes,collectors,executors,injectors,attackpath_execution,attackpath_finding,secret_references,secrets,connector_instances,autonomous_runs,autonomous_events,autonomous_directives,kill_chain_phases,security_coverages,domains,asset_groups,assets,findings
```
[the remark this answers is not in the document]
laugiov: You read a description that was out of date, and it was mine to fix. This branch does activate `findings`, and the description now says so from its first paragraph. Rewritten today, along with the title.

openaev-api/src/test/java/io/openaev/rest/finding/FindingAssetGroupSinkTest.java
```diff
@@ -0,0 +1,164 @@
...
+
+  @org.junit.jupiter.api.AfterEach
+  void sweep() {
+    jdbc.update("DELETE FROM findings_assets WHERE asset_id = ?", assetId);
+    jdbc.update("DELETE FROM findings WHERE tenant_id = ?", tenantId);
+    jdbc.update("DELETE FROM assets WHERE tenant_id = ?", tenantId);
+    jdbc.update("DELETE FROM injects_asset_groups WHERE asset_group_id = ?", assetGroupId);
+    jdbc.update("DELETE FROM asset_groups WHERE tenant_id = ?", tenantId);
+    jdbc.update("DELETE FROM injects WHERE tenant_id = ?", tenantId);
+  }
```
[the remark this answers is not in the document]
laugiov: Right on both counts, fixed in 06c165527. The class is deliberately not `@Transactional`, so nothing rolled the committed tenant back: the sweep now deletes its own rows first and then calls `tenantHelper.deleteCommittedTenants(tenantId)`, the order that helper documents, and clears `TenantContext` because `switchToTenant` sets a thread-local that is not transactional either.

One thing your comment led me to that is worth stating: deleting the tenant row does not remove its RabbitMQ queues. `TenantQueueService.createDependencyForTenant` has no symmetric teardown, so the broker residue survives regardless. That is filed separately as #7873, with the measurements: left alone it reached 58 595 queues, took broker startup from 2 seconds to 349, and failed a regression run with `Failed to create RabbitMQ queues for tenant`.

openaev-api/src/test/java/io/openaev/rest/finding/FindingOrphanRemovalTest.java
```diff
@@ -0,0 +1,127 @@
...
+                    tenantId);
+              }
+            });
+  }
+
+  @AfterEach
+  void sweep() {
+    jdbc.update("DELETE FROM findings WHERE tenant_id = ?", tenantId);
+    jdbc.update("DELETE FROM injects WHERE tenant_id = ?", tenantId);
+  }
```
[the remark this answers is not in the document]
laugiov: Fixed in 06c165527, using `deleteCommittedTenants` as you suggested, after the class deletes its own findings and injects (the order that helper requires). Same caveat as the sink test: the tenant row goes, its broker queues do not, which is #7873.

openaev-api/src/test/java/io/openaev/rest/finding/FindingWriteAttributionTest.java
```diff
@@ -0,0 +1,106 @@
...
+  private String seedInject(String tenantId) {
+    // The fixture, not a hand-built entity: Inject carries required fields (dependsDuration among
+    // them) that a bare constructor leaves null.
+    Inject inject = io.openaev.utils.fixtures.InjectFixture.getDefaultInject();
+    inject.setTenant(new Tenant(tenantId));
+    entityManager.persist(inject);
+    entityManager.flush();
+    return inject.getId();
+  }
+}
```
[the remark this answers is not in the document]
laugiov: Fixed in 06c165527. This class is `@Transactional`, so its rows roll back, but the thread-local is not transactional and would have carried a deliberately-wrong ambient tenant into the next test on the same thread. Cleared in an `@AfterEach`.

laugiov:
Five remediation commits since the review request, after an adversarial self-review of this branch. Each fix carries a mutation check: the fix removed, the test observed to fail, the fix restored.

**A production defect this branch introduced**

`c7d3193a6` — **`POST /api/findings` could write into another tenant.** The change that attributes a finding from its inject (`finding.setTenant(inject.getTenant())`) takes that inject from `InjectService.inject(injectId)`, a primary-key load that no tenant predicate filters, and the endpoint's `@AccessControl` checks a capability with no `resourceId` — so `finding_inject_id` arrives straight from the request body, unauthorized. A member of tenant A posting an inject id owned by tenant B got the row stamped B. Before the change it landed in A, so the change flipped the direction from junk in your own tenant to a write into someone else's. The inject's tenant is now validated against the request scope through `tenantForWrite`, which refuses with a 400.

**Guards**

`ca014696a` — `findings` was added to `GUARDED_TABLES` with no rule behind it, and the meta-check only asserted set membership. There is now a real `findings_repository_access_is_reviewed` with all twelve accessors and the scope mechanism of each: the HTTP handlers carry a `TxCtx`, `FindingWriter` declares its own for its `REQUIRES_NEW` boundary, the attack-path and autonomous writers reach the primitive with the tenant of the inject or run, the two ES handlers run under `EngineSyncExecutionJob`'s `allTenants()`, and the gauge is explicitly `allTenants()`.

**Four tests that were green for the wrong reason**

- `c1265e3d3` — **the findings gauge compared zero to zero.** The class seeded no findings at all and read its baseline at assertion time, so `0 == 0` held with or without the scope. Two findings are seeded now, one per tenant, with the delta shape its asset-group sibling already used.
- `c1265e3d3` — **the GROUP BY test read past into the ORDER BY.** `sql.substring(sql.indexOf("GROUP BY"))` ran to the end of a query ending `ORDER BY f.finding_updated_at LIMIT`, so `contains("f.finding_updated_at")` was satisfied by the wrong occurrence and deleting that column from the real GROUP BY left it green. `i.inject_exercise`, projected and non-aggregated, was missing from the checked list entirely.
- `c1265e3d3` — the serialization sink test declared `active-tables=asset_groups,assets`, deactivating the table this PR activates, on the endpoint it modifies — the mirror image of the rationale three other files in this diff carry.
- `b9bf624cc` — **`FindingApiTest`'s expectations moved with the response.** The tenant aspect is `@Before`-only with `set_config(..., true)`, and the class is `@Transactional`, so the scope the request resolved is still set when the expectation is read through `findingRepository.findAll()` in the same transaction. If the search failed closed, both sides came back empty and the comparison held on `[] == []`. Eight assertion sites now guard that the expectation is non-empty; removing the `TxCtx` from the search makes them fail, where before it did not.

`ece9e1c3d` — `FindingWriter`, the file with the most new logic here, had no test. Both refusals are covered now, with a positive half proving the scope is what separates them rather than a blanket refusal.

Also filed, none of it belonging to this PR: #7872, #7873, #7874.

openaev-api/src/main/java/io/openaev/rest/finding/FindingService.java
```diff
@@ -80,9 +89,26 @@ public FindingSummaryOutput findingSummary(@NotNull final String id) {
         .build();
   }
 
-  public Finding createFinding(@NotNull final Finding finding, @NotBlank final String injectId) {
+  public Finding createFinding(
+      @NotNull final TxCtx ctx, @NotNull final Finding finding, @NotBlank final String injectId) {
     Inject inject = this.injectService.inject(injectId);
     finding.setInject(inject);
+    // The tenant comes from the inject that produced the finding, as it does in createFindings.
```
corinnekrych: let's remove this block of comment - AI is too verbose, the code reads well

laugiov: Done.

openaev-api/src/main/java/io/openaev/rest/finding/FindingService.java
```diff
@@ -184,7 +210,12 @@ public void processAgentFindings(
   public void saveAgentFinding(
       Inject inject, Asset asset, ContractOutputContext contractOutputContext, String value) {
 
+    // One value drives both the scope of the writer's own REQUIRES_NEW transaction and the tenant
+    // the row is attributed to, so the conflict branch can only ever match what the insert writes.
+    String tenantId = inject.getTenant() != null ? inject.getTenant().getId() : null;
+
     findingWriter.saveCompleteFinding(
+        tenantId == null ? TxCtx.missing() : TxCtx.forTenant(tenantId),
```
corinnekrych, line 208: why do we need both tenatId and TxCtx as param?

laugiov: Done

openaev-api/src/main/java/io/openaev/rest/finding/FindingWriter.java
```diff
@@ -30,9 +40,34 @@ public void saveCompleteFinding(
       String assetId,
       String[] tagIds,
       String tenantId) {
+    // findings.tenant_id is NOT NULL, so a blank tenant already failed here, as a constraint
```
corinnekrych: I'd pass only TxCtx and hre resolve it:
```
    String tenantId =
        switch (ctx) {
          case TxCtx.Restricted restricted when restricted.tenantIds().size() == 1 ->
              restricted.tenantIds().get(0);
          default -> null;
        };
```

laugiov: Adopted in 4bb16c5de, with the default arm throwing instead of returning null.

corinnekrych (approved):
Tested at runtime. LGTM

PR #7869, opened 2026-09-09
feat(finding): derive sensitive finding masking from contract output type (#7500)

heditar:
> Solves #7500 by deriving sensitivity from the finding type. #7587 was opened for the same issue with a stored `finding_is_sensitive` column; its branch now carries this same content, so only one of the two is needed.

### Proposed changes

Credentials discovered during simulations were displayed in cleartext on the Findings page. They are now masked on the way out of the API, while the database keeps the cleartext — deduplication, correlation and attack path computation all need it.

**Sensitivity is derived, not stored.** A value is secret because of the `PrimitiveType` it was extracted as — `Password`, `Hash` or `Key` — read from the `ChainingTypeRegistry` recipe of its `ContractOutputType`. No column, no migration, no `sensitive` flag threaded through the output processor chain. Adding a type that carries a hash makes it masked automatically.

**Masking is per segment.** `VALUE_COMPOSITIONS` declares how each value is built — the inverse of the processors' `toFindingValue` — so a credential comes back as `admin:mo******`: the account name stays readable, since it is what an analyst acts on and is not itself a secret. Masked parts keep a two character fragment, enough to recognise a secret you already know, never enough to disclose it. Anything unexpected — unknown composition, unexpected split — masks the whole value: an omission can only over-mask, never leak.

Main changes:

* `SensitiveValueMaskingUtils` (new) — the single masking implementation for the whole platform. It replaces `PrimitiveValueMaskingUtils`, whose policy disclosed the exact length of every secret, and the attack path's own private rule.
* `FindingOutput` (new) + `FindingApi` / `FindingMapper` / `FindingService` — the finding endpoints stop serializing the JPA entity, which is what made the cleartext reachable. All four finding representations mask in one place.
* `AttackPathGraphService` and friends — the graph masks server-side. It previously sent the cleartext in the payload and masked on render, which any network inspection defeated. Node and edge ids that encoded the raw value are hashed; `ConsumedFindingKeyDTO` keeps the cleartext in a non-serialized field used only by the matcher, so causal resolution stays exact.
* Frontend — the five finding views retyped to `FindingOutput`; client-side masking removed.
* Tests — masking policy, API behaviour, and an anti-drift test that walks each output processor and checks the declared composition takes its value apart into exactly the segments it was built from.
* `docs/docs/usage/evaluate/findings/findings.md`.

`PasswordPolicy` is the one explicit exception: its `key` holds a policy setting name (`MinimumPasswordLength`…), not a cryptographic key.

### Testing Instructions

1. Start the services: `cd openaev-dev && podman compose up -d openaev-dev-pgsql openaev-dev-minio openaev-dev-elasticsearch openaev-dev-rabbitmq`
2. Build with JDK 21: `mvn install -DskipTests -Dspotless.check.skip=true`
3. Backend: `mvn -pl openaev-api test -Dtest='FindingApiTest,FindingServiceTest,SensitiveValueMaskingUtilsTest,SensitiveValueCompositionConsistencyTest,ConsumedFindingKeyDTOTest,WorkflowServiceTest,WorkflowConfigurationMapperTest,AttackPath*Test' -Dsurefire.failIfNoSpecifiedTests=false`
4. Formatting: `mvn spotless:check`
5. Frontend: `cd openaev-front && yarn check-ts && yarn lint`
6. Manually: run a simulation producing a `Credentials` finding, open **Findings** — the value reads `admin:mo******` in the list, the detail page and the attack path drawer. `SELECT finding_value FROM findings WHERE finding_id = '…'` still returns the cleartext.
7. Chaining engine regression: open a workflow holding a `Password` scope variable, save without touching it, and check the raw value is preserved.

### Related issues

* Related #7500

### Checklist

### Further comments

The composition table is static and lives next to the masking, not on the processors: reaching them would need `OutputProcessorFactory`, which makes `FindingMapper` depend on every `OutputProcessor` bean — one of which depends back on `FindingService`. That is a Spring cycle, and `@Lazy` would hide it rather than remove it. The anti-drift test is what keeps the table honest instead, since it can reach the processors without creating a cycle at runtime.

**The secret is not deleted.** Masking happens on the way out of the API. Anyone with database or Elasticsearch access still sees the cleartext.


heditar:
Closing as a duplicate of #7587, which now carries this exact same content (both branches point at `33ffa7599`).

#7587 is the one to review: it holds the earlier review history on this issue, including the still-open point about `POST /api/findings/search` filtering on the cleartext column. Its title and description have been rewritten to describe the derived approach, since they still described the stored `finding_is_sensitive` column that this content no longer contains.

The branch is kept, not deleted, so nothing here is lost.

Issue #7698, opened 2026-09-01
fix(simulation): Simulation marked "finished" while Prevention/Detection expectations are still pending (Nuclei injects)


MarineLeM:
## Description

A simulation should only be `finished` once **all** inject expectations have a result. 

## Environment

1. Version of 01 september 2026

## Reproducible steps

In the affected simulation:
- Simulation status = **finished**
- Nuclei injects have a result for **Vulnerability** ✅
- Nuclei injects still have **Prevention** and **Detection** expectations at **pending** ❌

The simulation completes without waiting for the new Prevention/Detection results.

## Expected output

Simulation status should stay in progress until **all** expectation types (Prevention, Detection, Vulnerability) on every inject have a result.

## Impact
- False "finished" status misleads users on simulation completeness.
- OpenCTI expect to receive the stix bundle since the simulation is finished.. ( they didn't received it )

## Additional information
Prevention and Detection expectations were recently added to Nuclei injects, in addition to the existing Vulnerability expectation.

<img width="1569" height="790" alt="Image" src="https://github.com/user-attachments/assets/50ae2e1c-1a12-4960-9821-16260deb30f4" />
related inject details :

<img width="1920" height="2185" alt="Image" src="https://github.com/user-attachments/assets/c42702ea-9421-421a-b306-0b8f0f3a162a" />

camrrx:
This issue is fixed by @MarineLeM in the ticket #5384. 
I tried to reproduce the original bug and couldn't trigger it : when no Detection/Prevention collector is present, the simulation no longer waits on those expectations; it now correctly completes once the Vulnerability result comes back.

MarineLeM:
## Technical explanation on what happened :

1. At initialization, an expectation is created on the leaf with an empty `results` array (`results = []`), even when no SecurityPlatform of the requested type is registered.
2. The system then waits up to **6h** expecting a result to be collected on that leaf before computing the parent (AssetGroup/Asset) scores.
3. Since no collector exists, no result is ever produced.
4. `handleInjectExpectationCollectStatus()` computes `collectDone` as:
   ```java
   results == null || results.isEmpty() || hasValidResults(results) || expectation.isExpired()
   ```
   But it has the side effect of also matching a freshly-created, never-resolvable expectation for a non-existent SecurityPlatform.
5. → The inject is marked `collectExecutionStatus = COMPLETED` and saved.
[...]

PR #7897, opened 2026-09-11
fix(inject-expectation): set failed result when no collector found (#7698)

MarineLeM:
### Proposed changes

For expectations that require a collector and none is found :
- Still initialize the expectation on the leaf with an empty result (so it remains editable/manually fillable).
- Immediately set the leaf's score to 0, instead of leaving it pending indefinitely waiting for a collector that will never report.
- Surface this state clearly on the front-end with a message such as "No automated security platform detected".

### Related issues

* Related #7698


GaetanSantucci (approved):
Image

test ok

Issue #7990, opened 2026-09-15
feat(credentials): FT3.3 - part 2 -  Inject Authorisation emission during inject execution


MarineLeM:
## Description 

On inject execution with a secret attached, create/replace (retry = overwrite, no accumulation) the InjectAuthorisation row and enrich ExecutableInjectDTO to publish attachment in RabbitMq

PR #7997, opened 2026-09-16
feat(credentials): create injectAuthorisation if needed and send into rabbitMq (#7990)

MarineLeM:
### Proposed changes

* On inject execution with a secret attached, create/replace (retry = overwrite, no accumulation) the InjectAuthorisation row and enrich ExecutableInjectDTO to publish attachment in RabbitMq

### Related issues

* Related #7990


openaev-api/src/main/java/io/openaev/rest/inject/service/InjectService.java
```diff
@@ -1636,4 +1638,33 @@ private ObjectNode convertContent(String content) {
...
+    String rawCode = UUID.randomUUID().toString();
+    String hashedCode = AUTHORISATION_CODE_ENCODER.encode(rawCode);
+
+    injectAuthorisationRepository.deleteAllByInjectId(inject.getId());
+
+    InjectAuthorisation authorisation = new InjectAuthorisation();
+    authorisation.setInject(inject);
+    authorisation.setCode(hashedCode);
+    authorisation.setIssuedAt(Instant.now());
+    injectAuthorisationRepository.save(authorisation);
```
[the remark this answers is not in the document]
gabriel-peze: It's right, you should set the method as @ Transactional, or use saveAndFlush.

openaev-api/src/test/java/io/openaev/executors/ExecutorTest.java
```diff
@@ -1,19 +1,17 @@
...
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.mockito.Mockito.*;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import io.openaev.database.model.ExecutionStatus;
-import io.openaev.database.model.Inject;
-import io.openaev.database.model.InjectStatus;
-import io.openaev.database.model.Injection;
+import io.openaev.database.model.*;
```
gabriel-peze, line 9: This comment is not related to this line.

I was looking at the Inject.java object and especially the authorisation attribute :
```
@Getter
  @OneToOne(
      mappedBy = "inject",
      fetch = FetchType.LAZY,
      cascade = CascadeType.ALL,
      orphanRemoval = true)
  @JsonProperty("inject_authorisation")
  @JsonSerialize(using = MultiModelSerializer.class)
  @JsonDeserialize(using = MonoIdDeserializerHelper.class)
  private InjectAuthorisation authorisation;
```
I think there is a issue here with the JsonSerialize annotation, it's using a MultiModelSerializer, it need a list in entry, it should use a MonoIdSerializer instead.

And i question myself if this attribute shouldn't be in JsonIgnore ?
It's a sensitive data i think, is it mandatory to be exposed on the inject ?

openaev-api/src/main/java/io/openaev/rest/inject/service/InjectService.java
```diff
@@ -1636,4 +1638,33 @@ private ObjectNode convertContent(String content) {
     }
     return null;
   }
+
+  private static final Argon2PasswordEncoder AUTHORISATION_CODE_ENCODER =
```
gabriel-peze: Is it mandatory to use Argon2 here, it can be ressource consuming, if there is a lot of inject to handle it will take an unecessary time to compute.
I think you can use a SHA-256 here instead.

gabriel-peze (changes requested):

openaev-api/src/test/java/io/openaev/executors/ExecutorTest.java
MarineLeM: actually it has been removed

openaev-api/src/main/java/io/openaev/rest/inject/service/InjectService.java
MarineLeM: Good point

gabriel-peze (approved):

PR #8001, opened 2026-09-16
feat(chainnig): input validation (#5049)

camrrx:
### Proposed changes

* New format-validation core in openaev-model (FormatRuleKind, PrimitiveTypePolicy, PrimitiveFormatValidator): one source of truth for what each PrimitiveType supports (numeric comparison, case sensitivity) and what format its values must match. Rules are either a regex written in the Java ∩ ECMAScript intersection, shipped as-is to the frontend, or a named parser the frontend reimplements.
* New GET /api/chaining/primitive-types endpoint exposing those descriptors, so the UI stops restating backend rules. 
`{
  "primitive_type": "port",
  "primitive_type_capabilities": {
    "numeric_value": true,
    "case_sensitivity": false
  },
  "primitive_type_validation": {
    "applies_to": [
      "EQ",
      "NEQ"
    ],
    "rules": [
      {
        "kind": "PORT",
        "pattern": "^(?:6553[0-5]|655[0-2][0-9]|65[0-4][0-9]{2}|6[0-4][0-9]{3}|[1-5][0-9]{4}|[0-9]{1,4})$",
        "error_message_key": "Expected a valid port (0-65535)"
      }
    ]
  }
}` 
* Event condition form and scope variable dialog now driven by the descriptors.
* Scope variable values are now format-validated on create and update. 
* PrimitiveValueValidator split into format vs scope membership. isValidPort() now delegates to FormatRuleKind.PORT, so output processors cannot drift from the rule advertised to the frontend

### Testing Instructions

2. Go on logic tabs
3. Add component and choose event
4. Try to add a ipv4 condition and write wrong value
5. Try to add email/number/port ....
6. You should have a message if you write the wrong format

### Related issues

* Related #ISSUE-NUMBER


openaev-api/src/main/java/io/openaev/api/chaining/ChainingApi.java
```diff
@@ -54,6 +55,7 @@ public class ChainingApi extends RestBehavior {
   public static final String CHAINING_URI = "/chaining";
```
savacano28: These endpoints are never used then this class chainingApi was deleted in a previous pr.

openaev-api/src/main/java/io/openaev/api/chaining/ChainingApi.java
```diff
@@ -88,6 +90,22 @@ public ChainingOutput findAll(TxCtx ctx) {
...
+          "Returns, for each primitive type, the operators it supports and the format rules its"
+              + " values must satisfy. Lets the condition editor and the scope page offer the right"
+              + " operators and validate values without restating any backend rule.")
+  @ApiResponses({
+    @ApiResponse(responseCode = "200", description = "Primitive type descriptors retrieved")
+  })
+  @AccessControl(skipRBAC = true, isEnterpriseEdition = true)
+  @Transactional(readOnly = true)
+  @GetMapping(PRIMITIVE_TYPES_URI)
+  public List<PrimitiveTypeDescriptorOutput> primitiveTypeDescriptors(TxCtx ctx) {
```
savacano28: Maybe we could include this endpoint in ThreatArsenalApi like getArgumentTypes()?

openaev-api/src/main/java/io/openaev/api/chaining/dto/PrimitiveTypeCapabilitiesOutput.java
```diff
@@ -0,0 +1,27 @@
...
+@Builder
+@Schema(description = "Operator capabilities of a primitive chaining type.")
+public class PrimitiveTypeCapabilitiesOutput {
+
+  @Schema(
+      description =
+          "Values are numbers: the greater-than / less-than operators are meaningful and must be"
+              + " offered, and a value must be numeric whatever the operator.")
+  @JsonProperty("numeric_value")
+  private boolean numericValue;
```
savacano28: isNumericValue/isCaseSentive

openaev-api/src/main/java/io/openaev/service/chaining/PrimitiveValueValidator.java
```diff
@@ -39,37 +41,27 @@ public static boolean isAcceptedForPrimitiveType(
...
-      case IPv6 -> IpAddressUtils.isIpv6Address(value) && isIpAllowedByScope(value, context);
-      case Domain -> DOMAIN_VALIDATOR.isValid(value) && isDomainAllowedByScope(value, context);
-      case IpSubnet ->
-          (IpAddressUtils.isIpv4Subnet(value) || IpAddressUtils.isIpv6Subnet(value))
-              && isSubnetAllowedByScope(value, context);
+      case IPv4, IPv6 -> isIpAllowedByScope(value, context);
+      case Domain -> isDomainAllowedByScope(value, context);
+      case IpSubnet -> isSubnetAllowedByScope(value, context);
       case AssetId -> isAssetIdAllowedByScope(value, context);
       case AssetGroupId -> isAssetGroupIdAllowedByScope(value, context);
```
savacano28, line 52: Todo in other task :) : We should add also allow/deny list for teams and personas

openaev-model/src/main/java/io/openaev/validator/primitive/PrimitiveTypePolicy.java
```diff
@@ -0,0 +1,183 @@
...
+
+  /**
+   * Where a format rule is enforced.
+   *
+   * <p>The distinction exists because tightening the <i>runtime ingestion</i> path is not the same
+   * risk as tightening <i>user input</i>: a third-party injector emitting a non-standard value
+   * would silently lose data. Newly introduced rules therefore start as {@link #INPUT_ONLY} and may
+   * be promoted once observed in the field.
+   */
+  public enum FormatEnforcement {
```
savacano28, line 46: Why we need this distinction?

openaev-front/src/admin/components/chaining/logic/primitive-types.ts
```diff
@@ -0,0 +1,139 @@
...
+ *
+ * Rules that the backend can serialize arrive as a `pattern` and are applied as-is: the backend
+ * writes them in the Java / ECMAScript intersection precisely so no second source of truth exists
+ * here. Rules backed by a parser the backend cannot express as a portable regex arrive with no
+ * pattern and are implemented below, one function per `kind`. Both runtimes are pinned to the same
+ * behaviour by the shared vectors in `openaev-api/src/test/resources/primitive-format-vectors.json`
+ * — never change one side without running both test suites.
+ */
+
+export type PrimitiveTypeLabel = string;
```
savacano28, line 16: Maybe we could also add here the validation for size of subnet ?

openaev-front/src/actions/payloads/primitive-type-actions.ts
```diff
@@ -0,0 +1,17 @@
...
+
+const CHAINING_URI = '/api/chaining';
+
+/**
+ * Operator capabilities and value format rules of every primitive chaining type.
+ *
+ * Static, tenant-independent metadata: it only describes the shape of a value, so it is safe to
+ * fetch once and share across the whole session.
+ */
+export const fetchPrimitiveTypeDescriptors = async (): Promise<PrimitiveTypeDescriptorOutput[]> => {
```
savacano28, line 12: Maybe it could be also used in threatArsenal to validate when we are adding arguments in payloads?

savacano28:
Tests:

**Config**uration: CaseSensitive and numericalValue
  * Scope variables: check types: port, number, domains, ips... OK
    image
  * Action inputs: we should check also when we are creating an action? In this case, we should have an error? ❓ 
    image
  * Events: OK 
    image
  * ThreathArsenal: Im not sure if we need validation here ❓ 
    image

**Execution:**
[[Stephanya] Input validation_2026-09-17T11_07_20.187928200Z_(with_variable_values & with_scope_definition).zip](https://github.com/user-attachments/files/32333976/Stephanya.Input.validation_2026-09-17T11_07_20.187928200Z_.with_variable_values.with_scope_definition.zip)
* Generation/validation ips, ports, credentials, text... : OK
* Trigger actions based on these validated types: OK
* Validation generated Findings: OK

  image
  image
  image
image
image
image

savacano28 (approved):

openaev-model/src/main/java/io/openaev/validator/primitive/PrimitiveTypePolicy.java
camrrx: The two paths fail differently: on user input a rejected value returns a visible 400 the user can fix, whereas at runtime ingestion it is silently dropped from the workflow state. So new rules start as INPUT_ONLY, and only formats already enforced at ingestion before this PR are RUNTIME_AND_INPUT

openaev-front/src/actions/payloads/primitive-type-actions.ts
camrrx: Yes we could but is it out of scope ? We can open an other issue to do refacto ?

Issue #300, opened 2023-09-29
Support SAML 2 authentication


SamuelHassine:
## Use case

Support SAML 2 authentication

PR #8002, opened 2026-09-16
fix(multi-tenancy): fix organization access validation across tenants (#300)

GaetanSantucci:
### Proposed changes

* Improved cross-tenant security handling for organization-related access.
* Ensured organization data is resolved within the correct tenant/request scope.
* Added safeguards to prevent unauthorized access across tenant boundaries.

### Testing Instructions

1. Reproduce the organization access flow in a multi-tenant environment.
2. Verify that access is allowed for the correct tenant.
3. Verify that cross-tenant access is rejected as expected.
4. Confirm existing organization functionality still works normally.

### Related issues

* Related OpenAEV-Platform/filigran-private#300


openaev-api/src/main/java/io/openaev/service/organization/OrganizationService.java
```diff
@@ -1,34 +1,105 @@
...
 public class OrganizationService {
 
   private final OrganizationRepository organizationRepository;
-
   private final BulkDeleteExecutor bulkDeleteExecutor;
+  private final TagService tagService;
+  private final EntityManager entityManager;
+
+  /** Lists organizations in the legacy tenant scope, bounded by the caller's authorized scope. */
+  @Transactional(readOnly = true)
```
impolitepanda: Seeing as these functions are all called from the API layer where all methods are all already under a transactional, so it's not necessary to add it here. (same for following methods)

openaev-api/src/main/java/io/openaev/service/organization/OrganizationService.java
```diff
@@ -1,34 +1,105 @@
...
 @RequiredArgsConstructor
 public class OrganizationService {
 
   private final OrganizationRepository organizationRepository;
-
   private final BulkDeleteExecutor bulkDeleteExecutor;
+  private final TagService tagService;
+  private final EntityManager entityManager;
+
+  /** Lists organizations in the legacy tenant scope, bounded by the caller's authorized scope. */
```
impolitepanda, line 48: Javadoc is incorrect (missing parameters, etc.)

openaev-api/src/main/java/io/openaev/service/organization/OrganizationService.java
```diff
@@ -1,34 +1,105 @@
...
+        (specification, pageable) ->
+            organizationRepository.findAll(scope.and(specification), pageable),
+        searchPaginationInput,
+        Organization.class);
+  }
+
+  /** Creates an organization in the tenant explicitly resolved by the API write-scope resolver. */
+  @Transactional(rollbackFor = Exception.class)
+  public Organization createOrganization(
+      TxCtx ctx, OrganizationCreateInput input, String tenantId) {
```
impolitepanda: ctx isn't useful here as the check was already done on the API layer

openaev-api/src/main/java/io/openaev/service/organization/OrganizationService.java
```diff
@@ -93,25 +165,78 @@ public List<String> bulkDelete(
...
+    }
+  }
+
+  private Specification<Organization> inTenantScope(TxCtx ctx) {
+    Set<String> tenantIds = TxCtxScopeUtils.tenantIdsFromHTTPCtx(ctx);
+    return (root, query, cb) ->
+        tenantIds.isEmpty() ? cb.disjunction() : root.get("tenant").get("id").in(tenantIds);
+  }
+
+  private Set<Tag> resolveTags(List<String> tagIds, String tenantId) {
```
impolitepanda, line 220: This function should not exist here. Instead, it whould call the tagSet function from TagService. But tagSet doesn't seem to support the tenant filter. So a new function doing this should be created in the tagService class.

Also, this class still fetches all tags requested by their ID and then filter them in memory instead of filtering directly in the select in the database which would be more efficient.

openaev-api/src/main/java/io/openaev/service/organization/OrganizationService.java
```diff
@@ -93,25 +165,78 @@ public List<String> bulkDelete(
...
+        .stream()
+        .map(
+            organization -> new FilterUtilsJpa.Option(organization.getId(), organization.getName()))
+        .toList();
+  }
+
+  private Organization findAccessibleById(TxCtx ctx, String organizationId) {
+    Specification<Organization> specification =
+        inTenantScope(ctx).and((root, query, cb) -> cb.equal(root.get("id"), organizationId));
+    Session session = entityManager.unwrap(Session.class);
```
impolitepanda, line 198: Not sure why we need to disable the filtering here before making our query

Issue #7991, opened 2026-09-15
feat(credentials): FT3.3 - part 3 - Authorisation validation on secret resolution endpoint


MarineLeM:
## Description

Create POST /api/inject/{inject_id}/attachment/secret — pure read endpoint.
Validate credential resolution ability and resolve it.

The endpoint should return 200 or an error:
- 403
- CREDENTIAL_NOT_FOUND
- CREDENTIAL_ACCESS_DENIED
- CREDENTIAL_INACTIVE
- 200 + secret in plaintext ⇒ SecretReference.status = ACTIVE or UNSET

In this part, each resolution attempt produces one audit entry.

PR #8005, opened 2026-09-16
feat(credentials): resolve endpoint (#7991)

MarineLeM:
### Proposed changes

Create POST /api/inject/{inject_id}/attachment/secret — pure read endpoint.
Validate credential resolution ability and resolve it.

The endpoint should return 200 or an error:
- 403
- CREDENTIAL_NOT_FOUND
- CREDENTIAL_ACCESS_DENIED
- CREDENTIAL_INACTIVE
- 200 + secret in plaintext ⇒ SecretReference.status = ACTIVE or UNSET

In this part, each resolution attempt produces one audit entry.

### Related issues

* Related #7991


openaev-api/src/main/java/io/openaev/rest/inject/service/InjectService.java
```diff
@@ -1664,4 +1667,79 @@ public String getAuthorisationCodeIfNeeded(ExecutableInject executableInject) {
...
+    return status == SecretReference.SECRET_STATUS.AUTH_FAILED
+        || status == SecretReference.SECRET_STATUS.PERMISSION_DENIED;
+  }
+
+  private boolean isCredentialInactiveStatus(SecretReference.SECRET_STATUS status) {
+    return status == SecretReference.SECRET_STATUS.TIMEOUT
+        || status == SecretReference.SECRET_STATUS.NETWORK_ERROR
+        || status == SecretReference.SECRET_STATUS.UNSUPPORTED
+        || status == SecretReference.SECRET_STATUS.FORMAT_ERROR
+        || status == SecretReference.SECRET_STATUS.UNKNOWN;
```
gabriel-peze, line 1692: It should never happend, because status is checked at creation or modification, but i think we should still secure it and add the UNSET status here too :)

what do you think ?

openaev-api/src/main/java/io/openaev/rest/inject/service/InjectService.java
```diff
@@ -1664,4 +1667,79 @@ public String getAuthorisationCodeIfNeeded(ExecutableInject executableInject) {
 
     return rawCode;
   }
+
+  private boolean verifyAuthorisationCode(String injectId, String code) {
+    Optional<InjectAuthorisation> authorisationOpt =
+        injectAuthorisationRepository.findByInjectId(injectId);
+    if (authorisationOpt.isEmpty()) {
```
gabriel-peze, line 1674: should we check for a timeout here ?
There is the value of issuedAt that exist, maybe it can be used to set a timeout on the credential access, we don't want the credential authorization to be indefinitly accessible.

openaev-api/src/main/java/io/openaev/rest/inject/InjectApi.java
```diff
@@ -704,6 +707,24 @@ public List<RawDocument> getPayloadDocumentsByInjectIdAndPayloadId(
     return documentService.documentsForPayload(payloadId);
   }
 
+  @Operation(description = "Resolve an inject attachment secret")
+  @PostMapping({
+    "/api/inject/{injectId}/attachment/secret",
+    TENANT_PREFIX + "/inject/{injectId}/attachment/secret"
+  })
+  @Transactional(readOnly = true)
```
gabriel-peze, line 715: The transactionnal oppening should not be on the API but on the Service.

openaev-api/src/main/java/io/openaev/rest/inject/service/InjectService.java
```diff
@@ -1664,4 +1667,79 @@ public String getAuthorisationCodeIfNeeded(ExecutableInject executableInject) {
...
+      return false;
+    }
+
+    InjectAuthorisation authorisation = authorisationOpt.get();
+    return hashWithSHA256(code).equals(authorisation.getCode());
```
gabriel-peze: hashWithSHA256 is not null safe, you should check for null before using the variables.

nitpick : reverse the test authorisation.getCode().equals(hashWithSHA256(code))

gabriel-peze (changes requested):

openaev-api/src/main/java/io/openaev/rest/inject/service/InjectService.java
MarineLeM: There is no timeout since we consider that if the injectAuthorisation exists, it's valid. When the inject passes to a terminated state, we will remove the injectAuthorisation. Actually, what is missing here is a check that the inject status is not in a terminated state — I will add this verification.

openaev-api/src/main/java/io/openaev/rest/inject/service/InjectService.java
MarineLeM: The UNKNOWN status exists, as you can see in the SECRET_STATUS enum.

The UNSET status is considered OK since it's unset because it's for credentials for which we cannot do a connectivity test, as for example hash and user/password secrets. So what will happen for them is that the injector itself, by doing its execution, will check if the creds are working or not and add a trace in the inject execution

gabriel-peze (approved):
LGTM

PR #8045, opened 2026-09-20
chore(ci): speed up PR pipelines and stop PRs writing caches (#1)

guillaumejparis:
Cache writes (yarn, Playwright, Maven, Docker layers) are gated to refs/heads/main. A PR-scoped cache entry is readable only by that same PR, so it consumed the shared 10 GB quota without ever warming another run, evicting the entries PRs do read.

Caching

shared maven-v1 key restored by compile/spotless/package/codecov, written only by backend-compile, replacing setup-java's default key whose contents differed per job
yarn cache key gains runner.arch so amd64 and arm64 stop sharing one entry
ubi9 docker layers are no longer cached; no PR job consumes that image
Nightly opts out of caching entirely, freeing browsers and image variants Core CI never reads
Sharding & matrix

API shards rebalanced to 9 + catch-all; PR E2E partitions every suite once across amd64 Chromium jobs
Failure handling

retries are handled once, at the workflow level: ci-retry.yml now re-runs failed jobs up to twice instead of once, covering back-to-back infra flakes. No per-step retry loops — including the one backend-compile previously had, which fired 2.1 s after the first failure and so never outlasted a registry blip
Playwright retries: 2 → 0, since every suite already polls internally via expect/toPass; trace moves to retain-on-failure because on-first-retry can no longer trigger
action/assertion timeouts tightened 60s/30s → 15s so a bad locator fails fast; infra suites get a 420 s per-project budget instead, as they poll for a real agent
Robustness

the four frontend jobs share a new yarn-install composite action instead of repeating the same Node/Corepack/cache/install block
that action no longer masks a Corepack failure: echo "dir=$(yarn config get cacheFolder)" exited 0 with an empty path, surfacing minutes later as actions/cache's opaque Input required and not supplied: path


.github/actions/backend-compile/action.yml
```diff
@@ -9,32 +15,41 @@ runs:
...
       shell: bash
+
+    # Exclude modules built in this run to prevent restoring stale OpenAEV artifacts.
+    - name: Save Maven dependencies
+      if: ${{ inputs.no-cache != 'true' && github.ref == 'refs/heads/main' && steps.maven-cache.outputs.cache-hit != 'true' }}
+      uses: actions/cache/save@v6
+      with:
+        path: |
+          ~/.m2/repository
+          !~/.m2/repository/io/openaev
```
laugiov, line 44: This exclusion is a no-op.

actions/cache globs paths with implicitDescendants: false, so `~/.m2/repository`
resolves to a single directory entry and tar recurses into it. I ran it against the
real lib: same resolved list with and without the `!` line. Negations only bite when
the include pattern enumerates below them.

So maven-v1 carries the whole io/openaev tree. That's 509 MB (the size of the
backend-compiled artifact), into a cache already sitting at 10.3 GB.

One catch on the fix: the upload on line 47 reads io/openaev, so you can't just rm
before saving. Moving the save after the upload and rm-ing there is the easy way out.

.github/actions/api-tests/action.yml
```diff
@@ -272,16 +280,22 @@ runs:
...
+          ${{ steps.suite.outputs.surefire_arg }} \
+          ${{ steps.tenant.outputs.args }}
+
+    # A single main-branch shard writes the shared dependency cache.
+    # Resolved dependencies are still saved after test failures.
+    - name: Save API test Maven dependencies
+      if: ${{ !cancelled() && inputs.no-cache != 'true' && github.ref == 'refs/heads/main' && inputs.shard == '1' && steps.maven-cache.outputs.cache-hit != 'true' }}
+      uses: actions/cache/save@v6
+      with:
+        path: ~/.m2/repository
```
laugiov, line 296: Same thing here, minus the attempt: this saves all of ~/.m2/repository, and it runs
after the backend-compiled download, so the 509 MB of io/openaev lands in it too.

Issue #8049, opened 2026-09-21
feat(inject): make time-based staleness threshold a config


antoinemzs:
## Use case
Make it possible to configure a custom staleness threshold

## Current workaround
None -- hardcoded

## Proposed solution
New config; default 4 mins (current hardcoded value)
Time unit: minutes (the OAEV scheduler and general workflow reason in minutes)

PR #8053, opened 2026-09-21
feat(inject): configure staleness threshold (#8049)

antoinemzs:
### Proposed changes

* Staleness threshold configuration key added

Note there is a drive-by relocation of a test file that got stranded in a previous merge

### Testing Instructions

1. Step-by-step how to test

### Related issues

* Related #8049
*

### Checklist

### Further comments


openaev-api/src/main/java/io/openaev/executors/Executor.java
```diff
@@ -48,6 +49,7 @@ public class Executor {
   private final ExecutableInjectDTOMapper executableInjectDTOMapper;
   private final ConnectorInstanceService connectorInstanceService;
   private final InjectExpectationService injectExpectationService;
+  private final OpenAEVConfig openAEVConfig;
```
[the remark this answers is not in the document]
antoinemzs: that's why you don't write test classes with a thousand mocks which break at every turn

on another note the tests aren't broken so here we are

openaev-api/src/main/java/io/openaev/utils/InjectionUtils.java
```diff
@@ -26,12 +26,14 @@ private InjectionUtils() {}
...
+   * @param stalenessThreshold number of minutes after the desired inject start after which the
+   *     inject is deemed too old
+   * @return {@code true} if the injection date is within the threshold before now
    * @throws java.util.NoSuchElementException if the injection has no scheduled date
    */
-  public static boolean isInInjectableRange(Injection injection) {
+  public static boolean isInInjectableRange(Injection injection, Integer stalenessThreshold) {
     Instant now = Instant.now();
-    Instant start = now.minus(Duration.parse("PT4M"));
+    Instant start = now.minus(stalenessThreshold, ChronoUnit.MINUTES);
```
laugiov, line 36: One suggestion : with 0 or a negative value the window start lands at or after now, so no scheduled inject ever passes again, and nothing warns at startup. Since the whole point is that we set this by hand on the armasuisse instance before their exercise, a typo would be expensive. 
Could we guard it? Min(1) or just a check at startup, whatever's cleanes

laugiov (approved):

openaev-api/src/main/java/io/openaev/utils/InjectionUtils.java
antoinemzs: yes thats the commit above

